PR #1, opened 2025-03-05
Basic player and map

manuq:
Fix https://github.com/endlessm/threadbare/issues/2


wjt (approved):
LGTM (though I confess I didn't read every line). I like the structure of interactions.

PR #11, opened 2025-03-13
Change "Talk" label based on interaction area

wjt:


scenes/interact_area.gd
```diff
@@ -4,6 +4,9 @@ extends Area2D
 signal interaction_started
 signal interaction_ended
 
+@export var action := "Talk"
```
wjt, line 7: One thing I'm not sure about here is localisation. If I mark this default for translation:

```suggestion
@export var action := tr("Talk")
```

and add this file to the list of files to extract strings from then of course this string is extracted for translation. But if I override this property in some scene and add that scene to the files to extract translations from - nothing is extracted to translate.

https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#converting-keys-to-text mentions that labels automatically fetch translations for their text if a translation exists so if e.g. "Kick" is translated somewhere else then using "Kick" as the action on an InteractArea somewhere in the project would end up using that translation...

manuq: Good point, I don't know how can we do the localization of this. Maybe a separate file somewhere in the project with the keywords like Kick, Talk, etc?

What you mention about labels automatically fetching translations means that the `tr()` I added here https://github.com/endlessm/threadbare/blob/d54fd60dde8bb825accceb285b8cc7149da6e250/scenes/interact_label.gd#L12 is probably unneeded or even wrong.

Issue #5, opened 2025-03-05
Initial mini game


manuq:
The quest for the MVP will have a mini game challenge.

Implement an initial mini game. This could be a jigsaw puzzle that fits thematically.

wjt:
I was thinking about how to do a music-based puzzle, inspired in part by the protagonist of LOOM (where playing melodies is one of the main actions you take to progress in the game) being named Bobbin Threadbare. As an initial idea, you could imagine that you talk to a bard who plays you a fragment of a traditional folk song; later in the map you come across a musical instrument or some particularly tuneful stones you can kick, and if you play the folk melody on them, a door opens. Or you could imagine an encounter where if you can accurately repeat a series of increasingly difficult melodies back to a musician, they give you some item.

I discussed with @jbourqueendless and he showed me [a video of a puzzle in Stardew Valley](https://youtu.be/0pmyvK-lzK4) where you have to play a melody back – similar idea.

Considerations here – not everyone can hear pitch (whether due to hearing loss/deafness, or just not having musical experience) so there needs to be some visual element too. (Justin's idea was that the bard gives you the sheet music.)

wjt:
I can see three ways to do the "play a tune" mechanic:

1. Copy the Stardew Valley tuneful-rocks
2. Interact with a single instrument which brings up some kind of keyboard
3. An item in the inventory (oops, don't have that yet) which when used brings up a keyboard 

Further idea for the puzzle. You learn a tune from the bard, picking out a few notes from a major scale: 1 2 3 5. But then next to the door are a series of markings:

- 0
- ↑
- ←→
- ↓

Indicating that you should play the melody four ways in turn:

- base melody 1 2 3 5
- shifted one note right 2 3 4 6
- flipped 6 4 3 2
- then shifted back to the left 5 3 2 1

And 4 lights by the markings light up in turn when you get each pattern right, or all switch off when you go wrong.

wjt:
I added a 6-note musical rock xylophone in https://endlessm.github.io/threadbare/branches/endlessm/musical-rocks/ where you interact with (kick) each rock to play its note. Colour is modulated based on the note number.

Next I'll add something that recognises patterns being played.

wjt:
https://endlessm.github.io/threadbare/branches/endlessm/musical-rocks/ in about 5 minutes once commit f1fd9f0b has finished building has a puzzle to solve. The NPC just directly tells you what to do but I would imagine that the "shift up, flip, shift down" would be in the environment.

[...]

PR #19, opened 2025-03-17
Add musical rocks puzzle

wjt:
This adds a series of rocks which the player can interact with to play musical notes. These are used for a puzzle where, if the player kicks the rocks in a certain series of sequences, they complete a puzzle and gain a reward. The player can interact with an NPC who tells some backstory, demonstrates the first part of the melody, and directs the player to some environmental clues for how to form the remaining melodies.

The concept is inspired by puzzles in Stardew Valley and Return to Monkey Island, with the environmental clues lovingly stolen from the Crystal Heart puzzle in chapter 6 of Celeste.

The components of the puzzle are structured with the general intent that they could be reused by other puzzles:

- The note played by each rock (both its audio sample and its note name (A-G)) are properties of the rock, with the colour of the rock being derived from the note name.

- The xylophone iterates its children to find all available notes, so notes can be added and removed by adjusting the rock child nodes.

- The tunes to be played are a property of the MusicPuzzle scene, which tracks the player's input, ignites fires (themselves another property) corresponding to each step of the puzzle, and emits a signal when the puzzle as a whole is solved.

The NPC -> Xylophone interaction is a little ad-hoc (using mood tags) but does work as one approach to having dialogue which interacts with the outside world.

The reward for completing the puzzle is that a bouncing sheep appears. (Maybe the player can shear it to make wool.) Of course, like the dialogue, assets, etc., this is a placeholder!

https://github.com/endlessm/threadbare/issues/5


scenes/main.tscn
```diff
@@ -1,9 +1,19 @@
...
 [ext_resource type="SpriteFrames" uid="uid://clx8sssi4tw8q" path="res://scenes/npcs/sprite_frames/sprite_frame_02.tres" id="5_tbgi4"]
+[ext_resource type="PackedScene" uid="uid://dv30a1p74xgs1" path="res://scenes/music_puzzle/musical_rock_xylophone.tscn" id="5_tefeu"]
+[ext_resource type="PackedScene" uid="uid://0ho4pvu8mn22" path="res://scenes/music_puzzle/bonfire.tscn" id="6_o6xl0"]
+[ext_resource type="Script" uid="uid://c68oh8dtr21ti" path="res://scenes/music_puzzle/music_puzzle.gd" id="8_tipki"]
+[ext_resource type="Texture2D" uid="uid://b6pfo5hc7quec" path="res://scenes/music_puzzle/clues/1.png" id="10_trceg"]
+[ext_resource type="Texture2D" uid="uid://3ieahlwwqb0t" path="res://scenes/music_puzzle/clues/2.png" id="11_a8y0u"]
+[ext_resource type="Texture2D" uid="uid://dgqq26xsu3hrf" path="res://scenes/music_puzzle/clues/3.png" id="12_jkv2x"]
+[ext_resource type="Texture2D" uid="uid://dybrvl42chyoi" path="res://scenes/music_puzzle/clues/4.png" id="13_jbj1t"]
+[ext_resource type="PackedScene" uid="uid://ckbqsrxnccwd0" path="res://scenes/music_puzzle/sheep.tscn" id="14_a8y0u"]
```
wjt, line 16: It's kind of sad to dump all this stuff into `main.tscn` – conflicts with other branches ahoy… – but I am not sure it is worth factoring the puzzle out into yet more scenes.

wjt:
BTW @manuq I rolled your hitbox & y-sorting feedback into this PR. I think it works a bit better now.

manuq (approved):
This looks great to me. Leaving it open because maybe @JuanFdS or someone else can give another review with their knowledge. Otherwise let's merge.

manuq:
> BTW @manuq I rolled your hitbox & y-sorting feedback into this PR. I think it works a bit better now.

Yes, I saw that immediately while testing the PR. Thanks!

scenes/music_puzzle/bonfire.gd
```diff
@@ -0,0 +1,8 @@
+class_name Bonfire
+extends StaticBody2D
+
+@onready var fire: AnimatedSprite2D = %Fire
+
+func ignite() -> void:
+	fire.animation = &"burning"
+	fire.play()
```
JuanFdS, line 8: The animation can be passed as a parameter to play as well! (unless there's something I'm missing I think this should be equivalent)
```suggestion
	fire.play(&"burning")
```

scenes/music_puzzle/musician.dialogue
```diff
@@ -0,0 +1,16 @@
+~ start
+Musician: Hello there, Storyweaver.
+Musician: Have you seen those stones all over this island?
+- What about them? They're just inert lumps of rock.
+	Musician: Oh, I wouldn't be so sure about that!
+- There's something unusual about them… it's like they have a certain energy.
+Musician: They're tuned to a major scale!
+	Legend has it that playing the right melody will open a door.
+=> demo_phrase_1
```
pablitar, line 9: Maybe we could rename this label to be related to the actual dialogue?

scenes/music_puzzle/music_puzzle.gd
```diff
@@ -0,0 +1,61 @@
...
+var _current_melody := 0
+var _position := 0
+
+
+func _ready() -> void:
+	xylophone.note_played.connect(_on_note_played)
+
+
+func _debug(fmt: String, args: Array = []) -> void:
+	if false:
```
JuanFdS, line 22: I think this false could be turned into an @export variable, so we can flip that switch while the game is running in the editor in case we want to check the debug lines.
```suggestion
	if false:
```

scenes/music_puzzle/music_puzzle.gd
```diff
@@ -0,0 +1,61 @@
+extends Node2D
+
+@export var xylophone: MusicalRockXylophone
+
+## Melodies expressed with the letters ABCDEFG.
+@export var melodies: Array[String]
+
+## A fire corresponding to each melody, ignited when the correct melody is played (in order).
+@export var fires: Array[Bonfire]
```
JuanFdS, line 9: This is totally not necessary!, but, a cool thing godot lets us do is add warnings to let devs know if a node isn't set up correctly. If we want this node to be modified by learners, we could also add an in-editor warning in case the melodies and fires don't add up:
```gdscript
	if melodies.size() != fires.size():
		return ["There should be one fire for each melody, but now there are %d melodies and %d fires" % [melodies.size(), fires.size()]]
	return []
```

Note: the script needs to be a `@tool` for this to work!

image

scenes/music_puzzle/musical_rock.gd
```diff
@@ -0,0 +1,42 @@
...
+
+func _modulate_rock():
+	if sprite_2d:
+		var i: int = NOTES.find(note)
+		sprite_2d.modulate = Color.from_hsv(i * 100.0 / NOTES.length(), 0.67, 0.89)
+
+
+func _on_interaction_started() -> void:
+	await play()
+	interact_area.interaction_ended.emit.call_deferred()
```
JuanFdS, line 34: Do you remember why this signal had to be emitted deferred?, can we add a comment about it?

JuanFdS (approved):
Looks great!

scenes/music_puzzle/music_puzzle.gd
wjt: Good idea! I applied this suggestion with a tweak to the function to work around an edge case of `gdformat` that I don't like...

scenes/music_puzzle/musician.dialogue
wjt: I think I'll not make any further tweaks here because this is placeholder dialogue but I'm interested in your advice on how to structure dialogue in general :)

wjt:
I'll go ahead and merge this. Thanks for all the feedback!

PR #21, opened 2025-03-18
Add linting with gdlint and gdformat, run via pre-commit and CI

wjt:
This PR reformats the GDScript source files with `gdformat`, and adjusts some declarations to pass `gdlint` and match the [GDScript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html). The diff is pleasingly small, perhaps because there's not much code yet.

It then adds a pre-commit config file to run these checks locally, and sets up a GitHub Actions workflow to run the pre-commit checks on pull requests.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,0 +1,20 @@
+name: Checks
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+# Cancel any ongoing previous run if a PR is updated
+concurrency:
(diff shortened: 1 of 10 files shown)


wjt:
I'm open to being persuaded to the point of view that having coding style checks in an early-stage project slows down development too much. But in my opinion, it's better to add them sooner, when the codebase is smaller, rather than to try to painfully reformat all the code later. Plus, we can see from the small diff here that it's actually not too hard to comply with these style guidelines so I don't think the overhead will be too great.

PR #25, opened 2025-03-19
editorconfig: Configure *.gd files to use tabs for indentation

wjt:
This is the Godot editor's default, but Zed doesn't know that.


manuq:
Oh are you using Zed for GDScript? Interesting.

PR #27, opened 2025-03-19
Create 2 basic audio buses: Music & SFX

JuanFdS:
Having separate audio buses allows us to change the volume of all the audio that plays in that bus independently. It also allows adding effects to the buses, but we probably won't even use that. However, I think having at least 2 basic buses for Music and Audio, and have all AudioStreamPlayers use one of them (or a new bus inherited from one of them), would be a good start since that lets us have 2 sliders in the config menu: one for music and one for sound effects.

### DIFF
--- a/default_bus_layout.tres
+++ b/default_bus_layout.tres
@@ -1,0 +1,15 @@
+[gd_resource type="AudioBusLayout" format=3 uid="uid://c7thbop54thnf"]
+
+[resource]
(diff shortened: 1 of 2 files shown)


manuq (approved):
Sounds good to me!

wjt (approved):
Great idea to set this up before we start adding sounds & music in earnest.

Issue #20, opened 2025-03-18
Splash screen


wjt:
The high-level outline of the MVP includes a splash screen:

![Image](https://github.com/user-attachments/assets/2cfdec3d-ad16-4efa-954f-629a4cef1a22)

While we do not have the animation assets etc. yet, we can still add a placeholder splash screen, where pressing any button starts the game. In due course we can develop this to show an animation, and potentially a "continue game / new game / options" menu.

wjt:
Further breakdown in this flow by @jofilizola:

![Image](https://github.com/user-attachments/assets/6fca9d25-ec8c-4f6c-9606-22f24a0196ce)

manuq:
@wjt yes, even a screen with a label "A fabric worlds getting unravelled and torn ..." would be a good placeholder.

PR #29, opened 2025-03-20
Placeholder splash and intro

wjt:
This adds a placeholder splash screen scene, followed by a placeholder scene introducing the world.

#20


wjt:
https://endlessm.github.io/threadbare/branches/endlessm/placeholder-splash-and-intro/

I don't know whether it's worth adding the animation from https://endlessm.github.io/threadbare/branches/endlessm/logo-stitch-experiment/ here; or making the placeholders a little less placeholder (e.g. "Press Space to continue" prompts).

assets/unsplash/README.md
```diff
@@ -0,0 +1,17 @@
+Files in this directory are covered by the [Unsplash license](https://unsplash.com/license).
+
+assorted-color-textiles.webp
```
wjt, line 3: TODO: Replace this image with the midjourney-generated embroidered world map. Keep the fade-out effect.

scenes/intro/intro.tscn
```diff
@@ -0,0 +1,123 @@
...
+grow_vertical = 2
+size_flags_horizontal = 4
+size_flags_vertical = 4
+theme_override_colors/font_outline_color = Color(0, 0, 0, 1)
+theme_override_constants/outline_size = 8
+theme_override_font_sizes/normal_font_size = 48
+text = "A fabric world is unravelled and torn..."
+autowrap_mode = 0
+visible_characters = 0
+visible_ratio = 0.0
```
wjt, line 118: Increase font size. Add larger black border/shadow to improve legibility. Centre the text.

scenes/splash/splash.gd
```diff
@@ -0,0 +1,15 @@
...
+		switch_to_intro()
+
+
+func switch_to_intro() -> void:
+	get_tree().change_scene_to_packed(preload("res://scenes/intro/intro.tscn"))
+
+
+func _on_logo_stitcher_finished() -> void:
+	await get_tree().create_timer(5.0).timeout
+	switch_to_intro()
```
JuanFdS, line 15: I've seeing an error on the `Output` tab because the `LogoStitcher` (which runs in the editor because it's a `@tool`) tries to call `_on_logo_stitcher_finished` through the signal `finished`. We could avoid this error by adding:

```gdscript
func _ready() -> void:
   $LogoStitcher.finished.connect(self._on_logo_stitcher_finished)
```

so the connection to the signal doesn't exist in the editor.
Another option is adding:
```gdscript
if Engine.is_editor_hint(): return
```
in the body of `_on_logo_stitcher_finished`, but I think we can just keep this node as it is and get the signal connected through code.

image

scenes/splash/splash.gd
```diff
@@ -0,0 +1,15 @@
+extends Node2D
+
+
+func _process(_delta: float) -> void:
+	if Input.is_action_just_pressed("ui_accept"):
```
JuanFdS, line 5: just to keep the same as in the rest of the code
```suggestion
	if Input.is_action_just_pressed(&"ui_accept"):
```

scenes/splash/logo_stitcher.gd
```diff
@@ -0,0 +1,128 @@
...
+
+	var points: PackedVector2Array = _stitch_points.slice(0, j * 2)
+	if not points:
+		return
+
+	draw_multiline(points, color, stitch_width, true)
+
+	# Round the ends of each stitch
+	for point: Vector2 in points:
+		draw_circle(point, stitch_width / 2, color, true, -1.0, true)
```
JuanFdS, line 128: wow, this is really cool!

JuanFdS (approved):
Great!, there's a thing with a `@tool` script giving an error in the editor that we can fix but apart from that I like it!

scenes/splash/splash.gd
wjt: I decided the tool script is still useful so I moved the signal connection to code as you suggest.

PR #30, opened 2025-03-20
Add explicit type annotations

wjt:
This squashes warnings that are shown in the debugger tab when running the game from the editor.

I have tried to find a way to check for these warnings in CI but it seems hard to do. If I adjust project.godot to make these warnings fatal rather than warnings, it blocks running the project (really annoying in development) or exporting it.

The alternative seems to be to write a CLI tool that feeds the source files into Godot's LSP server and extracts the diagnostics. Overkill.

### DIFF
--- a/scenes/music_puzzle/music_puzzle.gd
+++ b/scenes/music_puzzle/music_puzzle.gd
@@ -14,8 +14,8 @@ signal solved
 ## If enabled, show messages in the console describing the player's progress (or not) in the puzzle
 @export var debug: bool = false
 
-var _current_melody := 0
-var _position := 0
+var _current_melody: int = 0
+var _position: int = 0
 
 
 func _ready() -> void:
@@ -32,7 +32,7 @@ func _on_note_played(note: String) -> void:
 	if _current_melody >= melodies.size():
 		return
 
-	var melody := melodies[_current_melody]
+	var melody: String = melodies[_current_melody]
 	_debug(
 		"Current melody %s position %d expecting %s, received %s",
 		[melody, _position, melody[_position], note],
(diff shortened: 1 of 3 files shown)


scenes/music_puzzle/music_puzzle.gd
```diff
@@ -68,7 +68,10 @@ func _on_note_played(note: String) -> void:
 
 func _get_configuration_warnings() -> PackedStringArray:
 	if melodies.size() != fires.size():
-		var fmt = "There should be one fire for each melody, but there are %d melodies and %d fires"
-		return [fmt % [melodies.size(), fires.size()]]
+		var fmt: String = """
+			There should be one fire for each melody, \
+			but currently there are %d melodies and %d fires.
+		"""
+		return [fmt.strip_edges() % [melodies.size(), fires.size()]]
```
wjt, line 75: Adding the `: String` pushed this over gdlint's 100-column limit. The kind of eccentric use of a multiline string plus `strip_edges()` is to avoid some of gdformat's grosser formatting choices. But I wonder if I can find something neater…

wjt: I think this is the best I can do.

manuq (approved):

manuq:
> I have tried to find a way to check for these warnings in CI but it seems hard to do. If I adjust project.godot to make these warnings fatal rather than warnings, it blocks running the project (really annoying in development) or exporting it.

Yes, we don't want to make these warnings fatal to learners or to ourselves while doing explorations.

> The alternative seems to be to write a CLI tool that feeds the source files into Godot's LSP server and extracts the diagnostics. Overkill.

Yeah, too overkill for now.

Issue #22, opened 2025-03-18
World map: Create world map system and add Frays End area


pablitar:
The "World Map" will be the area in which the player is dropped by default. Here it can talk to NPCs and get quests.

The World Map probably will be separated in areas. These areas can be connected by "portals": When the player reaches the edge of an area, a door, or something that could lead to another area, it is transported to that area. We can keep it simple by fading to black, loading the scene, and then fading from black.

Once the player starts a quest, it gets transported to the quest scene. Once the quest is completed, the player is returned to the world map.

PR #31, opened 2025-03-21
Add an initial world map

pablitar:
Add two prototype world-map scenes, containing some props and buildings.

The two scenes are linked, switching to the other when the player enters a
particular region.

Solves #22


scenes/world_map/teleporter.gd
```diff
@@ -0,0 +1,16 @@
...
+func _ready() -> void:
+	self.body_entered.connect(_on_body_entered)
+	collision_layer = 0
+	collision_mask = 0
+	set_collision_layer_value(3, true)
+	set_collision_mask_value(1, true)
+
+func _on_body_entered(_body: PhysicsBody2D) -> void:
+	if scene_to_go_to:
+		get_tree().change_scene_to_file(scene_to_go_to)
```
wjt, line 15: I'm not quite sure how to make this work cleanly but it would be nice to link the two teleporters so that when you leave the song sanctuary you end up at the east edge of frays end, rather than some random starting place.

wjt: I wonder if something like this could work... Give each Teleporter 4 properties:

1. A reference to the Player in the current scene
2. A reference to the target scene
3. The NodePath of the Teleporter in the target scene
4. A node representing a spawn point

Then when changing to the target scene, call some method on the target teleporter, so that it moves the player to its spawn point.

scenes/world_map/teleporter.gd
```diff
@@ -0,0 +1,16 @@
+class_name Teleporter
+extends Area2D
+
+@export_file("*.tscn") var scene_to_go_to: String
```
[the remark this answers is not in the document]
pablitar: Exactly!

scenes/world_map/teleporter.gd
pablitar: Do we move the Player node to the other scene? Or do we add the player to each scene? What would be the pros and cons of each approach?

wjt: Good question. I assumed each scene would have its own instance of the Player scene, and its position would be adjusted in the case of having teleported into the scene.

project.godot
```diff
@@ -97,10 +97,13 @@ locale/translations_pot_files=PackedStringArray("res://scenes/npcs/talker/defaul
...
+2d_physics/layer_3="player_detectors"
 2d_physics/layer_5="walls"
 2d_physics/layer_6="interactable"
 
 [rendering]
 
 renderer/rendering_method="gl_compatibility"
 renderer/rendering_method.mobile="gl_compatibility"
+2d/snap/snap_2d_transforms_to_pixel=true
+2d/snap/snap_2d_vertices_to_pixel=true
```
wjt, line 109: What does this part of the change do? To me it seems worth explaining separately, perhaps in its own commit.

.gitignore
```diff
@@ -5,3 +5,4 @@
 # From Tiny Swords pack
 .DS_Store
 .aseprite
+.vscode/settings.json
```
wjt, line 8: This is unrelated to the rest of the change and should be its own commit.

wjt (commented):
I think it's fine to leave the player spawning at its normal location for now – we have #35 to track improving that.

I suggest dropping the "change main scene" patch & fixing the linter errors in the other patch.

scenes/world_map/teleporter.gd
JuanFdS: I like each scene having its own player scene so each level contains everything it needs to work and can be ran from Run Current Scene. The other option would be having the player (or a player spawner?) as a Global. But I prefer having it in the level itself to make it easier for learners to place it and modify it if needed🤔

I agree on most of what @wjt commented about the teleporter properties except:
- I don't think we should need to specify the player's NodePath, if the teleporter is interactable, we can get a reference to the player when it interacts with the teleporter.

The teleporters could be identified by either the NodePath or an id (pro: teleporter from level A doesn't break if level B changes the node hierarchy, cons: we need to add code to avoid duplicate ids and stuff like that). I think the NodePath is a pretty good solution for now.

I think the optimal experience for the people modifying the scene would be a dropdown with all the teleporters in the other scene, so they can pick which one. Now... the tricky part is, in order to do this, we should be able to extract from info from the other scene. This either means instantiating it in the editor or using the SceneState API:
https://docs.godotengine.org/en/stable/classes/class_scenestate.html

Mmm... I've just checked the API and I think it isn't hard to get the info we could need for this 😮 .
If you add all the teleporters to a "teleporter" group, you could get the NodePaths of a teleporter in a different level like this:

```gdscript
func get_teleporter_paths(packed_scene: PackedScene):
	var teleporter_node_paths = []
	var scene_state := packed_scene.get_state()
	for i in scene_state.get_node_count():
		if "teleporter" in scene_state.get_node_groups(i):
			teleporter_node_paths.push_back(scene_state.get_node_path(i))
	return teleporter_node_paths
```

project.godot
pablitar: Totally! It changes the sprites transforms so that pixels get translated as wholes and there is no subpixel movement. It reduces some jitter and wobblyness that you may see otherwise. I'll get it in another commit!

.gitignore
pablitar: Sure. I can also see if I can exclude that file locally since maybe not everyone is using VS Code

scenes/world_map/teleporter.gd
pablitar: Nicee! I'll see if I can add this. Thanks!

I agree with Juan that we don't need a reference to the player.

Regarding the player being moved from one scene to the other, I agree that's better to have the player in the scene so it can be easily played. If a reason comes up for the player to be moved from one scene to the other, maybe we can make it so the game detects when there's no player, and adds one. But that may come with unexpected consequences so let's go with having the player in the scene for now.

pablitar:
hey @JuanFdS @wjt, since I'll follow this up in #35, I think is ready to merge! Should I go for it?

.gitignore
wjt: I think it's fine to add it, but as a Git pedant I prefer it as a separate commit :)

PR #39, opened 2025-03-24
world_map: Add ScreenOverlay CanvasLayer nodes

wjt:
The Player scene has a label that is shown when an interaction is available. This label is reparented to the current scene's "ScreenOverlay" node. However, the Frays End and Song Sanctuaries scenes did not have a ScreenOverlay node, causing the label to be displayed at an incorrect position (or not visibly at all).

Add a ScreenOverlay node to each scene.


pablitar (approved):
Thanks for fixing this!

PR #41, opened 2025-03-24
Assemble playable game from disparate scenes

### LINKED ISSUE #32: Stitch the game together from beginning to end
Link together the elements we have prototyped into a playable game, to provide as much as currently possible of the MVP player flow:

Image

Acceptance criteria: when a player plays the game, they should first see the splash screen, then, they would be transported to the world map. From the world map, they can play the different challenges.

Notes/questions: how is the world map connected to the challenges?
Note: the 3 challenges are part of 1 scene.

wjt:
https://endlessm.github.io/threadbare/branches/endlessm/stitch-scenes-into-playable-game/

(Go to Song Sanctuary and talk to the new NPC there.)

Fixes #32

### DIFF
--- a/scenes/main.tscn
+++ b/scenes/main.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=15 format=4 uid="uid://7hoy2p14t6kc"]
+[gd_scene load_steps=17 format=4 uid="uid://7hoy2p14t6kc"]
 
 [ext_resource type="TileSet" uid="uid://07fq3rspk8ia" path="res://scenes/tileset.tres" id="1_0wfyh"]
 [ext_resource type="PackedScene" uid="uid://iu2q66clupc6" path="res://scenes/player/player.tscn" id="3_sugp2"]
@@ -14,6 +14,10 @@
 [ext_resource type="Texture2D" uid="uid://dgqq26xsu3hrf" path="res://scenes/music_puzzle/clues/3.png" id="12_jkv2x"]
 [ext_resource type="Texture2D" uid="uid://dybrvl42chyoi" path="res://scenes/music_puzzle/clues/4.png" id="13_jbj1t"]
 [ext_resource type="PackedScene" uid="uid://ckbqsrxnccwd0" path="res://scenes/music_puzzle/sheep.tscn" id="14_a8y0u"]
+[ext_resource type="Script" uid="uid://hqdquinbimce" path="res://scenes/world_map/teleporter.gd" id="15_jkv2x"]
+
+[sub_resource type="RectangleShape2D" id="RectangleShape2D_jbj1t"]
+size = Vector2(52, 127)
 
 [node name="Main" type="Node2D"]
 texture_filter = 1
@@ -118,7 +122,19 @@ position = Vector2(288, -84)
 texture = ExtResource("13_jbj1t")
 
 [node name="Sheep" parent="OnTheGround/MusicPuzzle" instance=ExtResource("14_a8y0u")]
-position = Vector2(304, 228)
+position = Vector2(672, 252)
+
+[node name="Teleporter" type="Area2D" parent="."]
+position = Vector2(2117, 1023)
+collision_layer = 4
+script = ExtResource("15_jkv2x")
+scene_to_go_to = "uid://db6vhi7s2f37c"
+spawn_point_path = NodePath("")
+metadata/_custom_type_script = "uid://hqdquinbimce"
+
+[node name="CollisionShape2D" type="CollisionShape2D" parent="Teleporter"]
+position = Vector2(-229, -487)
+shape = SubResource("RectangleShape2D_jbj1t")
 
 [node name="ScreenOverlay" type="CanvasLayer" parent="."]
 
--- a/scenes/stealth_level/stealth_level.tscn
+++ b/scenes/stealth_level/stealth_level.tscn
@@ -1,11 +1,13 @@
-[gd_scene load_steps=21 format=4 uid="uid://db6vhi7s2f37c"]
+[gd_scene load_steps=23 format=4 uid="uid://db6vhi7s2f37c"]
 
 [ext_resource type="Script" uid="uid://dnp0tjloec2d7" path="res://scenes/stealth_level/game_logic.gd" id="1_f6lvl"]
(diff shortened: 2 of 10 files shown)


scenes/stealth_level/stealth_level.tscn
```diff
@@ -152,6 +154,9 @@ dialogue = ExtResource("6_oyx4c")
 
 [node name="Talker2" parent="NPCs" instance=ExtResource("5_xtu6t")]
 position = Vector2(7249, 934)
+script = ExtResource("6_1b8xw")
+quest_scene = ExtResource("7_idy4y")
+quest_description = ""
```
wjt, line 159: This is kind of weird because the NPC here is not a quest *starter* but an ender. I guess I should make this a teleporter like in the music map.

scenes/main.tscn
```diff
@@ -118,7 +122,17 @@ position = Vector2(288, -84)
...
 
 [node name="Sheep" parent="OnTheGround/MusicPuzzle" instance=ExtResource("14_a8y0u")]
-position = Vector2(304, 228)
+position = Vector2(672, 252)
+
+[node name="Teleporter" type="Area2D" parent="."]
+position = Vector2(2117, 1023)
+script = ExtResource("15_jkv2x")
+scene_to_go_to = "uid://db6vhi7s2f37c"
+metadata/_custom_type_script = "uid://hqdquinbimce"
```
wjt, line 131: I imagine this would be replaced by a collectible which, when collected, triggers some victory animation and then a scene transition.

scenes/world_map/story_quest_starter.gd
```diff
@@ -0,0 +1,38 @@
...
+
+
+## At the end of the current interaction, enter [member quest_scene]. This is intended to be called
+## from dialogue.
+func enter_quest() -> void:
+	interact_area.interaction_ended.connect(_do_enter_quest)
+
+
+func _do_enter_quest() -> void:
+	SceneSwitcher.change_to_packed(quest_scene)
```
wjt, line 38: The reason it works like this is that, if you switch scenes while dialogue is running, the dialogue script will subsequently crash because its nodes have been removed from under it while it was `await`ing. I'm not wild about the API here and would welcome some other suggestions...

wjt (commented):
This does work but I'm not super happy with it, I have to say...

wjt:
Weirdly this seems to break the enemies in the stealth quest. I have no idea why!

wjt:
Ah, the stealth quest is already broken in the exported game.

wjt:
Filed https://github.com/endlessm/threadbare/issues/79 for that

scenes/stealth_level/stealth_level.tscn
```diff
@@ -228,7 +233,7 @@ curve = SubResource("Curve2D_46hcm")
 
 [node name="Guard7-GoingBackAndForth" parent="StealthQuest/EnemyGuards" node_paths=PackedStringArray("patrol_path") instance=ExtResource("8_lbv6i")]
 position = Vector2(4607, 990)
-velocity = Vector2(-71.1522, 70.2664)
+velocity = Vector2(-70.9997, 70.4205)
```
JuanFdS, line 236: I just noticed that if a guard has `move_while_in_editor` set to true, when saving the velocity it had at that moment will be stored in the scene. My bad 🤦🏾 , I will make it so it set to zero on save. Created an issue to handle it: https://github.com/endlessm/threadbare/issues/81

scenes/stealth_level/stealth_level.tscn
JuanFdS: I agree, but I think I can fix that in the collectibles branch since this npc acting as a teleporter is temporary

scenes/world_map/story_quest_starter.gd
JuanFdS: I feel like it'd be clearer to follow the intention if the scene switch happened as part of `_on_dialogue_ended` 🤔 , what do you think?

JuanFdS (approved):
Added some comments, but great work 🚀

scenes/stealth_level/stealth_level.tscn
wjt: Ah great, I had been manually removing this hunk from each patch but forgot here.

scenes/world_map/story_quest_starter.gd
wjt: I'm trying to remember why I didn't do that.... I think it's because it means you need extra flag in the class for "should switch scene when the dialogue ends", and a super() call, so it ends up being more code, whereas the signal connection itself serves as that flag in the current implementation. (It should probably be a ONESHOT connection I guess.) I'm on the fence, I think you're probably right that the intention is clearer with what you propose.

wjt: I pushed a patch that makes this change, adding a flag and overriding `_on_dialogue_ended()`. I'm fine with either implementation.

scenes/world_map/story_quest_starter.gd
```diff
@@ -31,8 +34,12 @@ func _get_configuration_warnings() -> PackedStringArray:
...
+
 
+func _on_dialogue_ended(dialogue_resource: DialogueResource) -> void:
+	await super(dialogue_resource)
 
-func _do_enter_quest() -> void:
-	SceneSwitcher.change_to_packed(quest_scene)
+	if _enter_quest_on_dialogue_ended:
+		SceneSwitcher.change_to_packed(quest_scene)
+		_enter_quest_on_dialogue_ended = false
```
JuanFdS, line 45: I'm curious about why we need this boolean, is `_on_dialogue_ended` triggered more than once? Because if not, and starting a quest changes to a different scene, this boolean would be back to true when we come back to the original scene

wjt: The Boolean is set by a particular branch of the dialogue tree calling the method above. It needs to be possible to have a branch of the dialogue that doesn't enter the quest.

wjt: I guess the other question is, why do we need to set it back to `false` when the scene is going away anyway? Well, I suppose we don't, but IMO it captures the intent better, and if `change_to_packed` fails it might be meaningful.

wjt:
Merging this based on the positive review

Issue #9, opened 2025-03-07
Initial game state system


manuq:
Add a way to persist the game state. Initially this could be as simple as a Dictionary for the inventory (mapping item name to quantity). A GameState autoload will provide the methods to change the game state and automatically save the state to disk when changed. For example, an NPC can give the player a gift. For this Dialogue Manager supports lines like `do Global.function()` or `set Global.state = "something"`.

Godot ConfigFile could be used initially. Another option is using JSON, but it might be too complex serializing/deserializing for our case, unless our data structures become complex, I'd avoid it.

This prototype can be used for reference:
- https://github.com/manuq/threadbare-dialogue-dm/blob/main/game_state/game_state.gd

wjt:
May not be necessary for the MVP:

1. We don't need to support keeping collectables from quests so we don't need global persistent state;
2. If we don't support saving within a quest, then we don't need persistent per-quest state.

manuq:
Then what we'll need for sure is a visual inventory in the UI, for keeping at least:
> Memory, Imagination, or Spirit  (M,I,S), each held on a spool in their Inventory and shown on the main screen as a sort of visual scoreboard.

pablitar:
For the time being I'll implement a non-persistent state. We can follow up with persistence if deemed necessary for the MVP

manuq:
Now that https://github.com/endlessm/threadbare/issues/432 is done by letting the player exit the quest, it becomes important to persist the state. So the Title screen can provide a Continue / New game options if there is progress, in place of the current Start.

At the same time we should be able to review/simplify the collectible system: https://github.com/endlessm/threadbare/issues/670#issuecomment-2893741349

So citing the comment, we would like to persist:

- Which quest are we playing (or no quest)
- How many threads have been collected (0, 1, 2, 3)
- Has the player activated the loom?
- What's the current scene?

PR #42, opened 2025-03-25
Improve music puzzle dialogue using state from the game

wjt:
Previously the NPC in the music puzzle had a static dialogue tree that was unaware of the player's progress through the puzzle and whether the player has interacted with the NPC previously.

Expose properties and methods of Talkers into their dialogue, and use this (plus an additional boolean on the music puzzle NPC) to adjust the bard's dialogue based on the state of the puzzle and whether the player has spoken to them before.

These changes fell out of my ongoing work to string together all the scenes we have into a linear game (#32).


scenes/music_puzzle/bard.gd
```diff
@@ -3,22 +3,10 @@
 class_name Bard
 extends Talker
 
-@export var xylophone: MusicalRockXylophone
+@export var puzzle: MusicPuzzle
 
+var first_conversation: bool = true
```
wjt, line 8: @pablitar since you are working on game state (#9) – I wonder if this the kind of thing you would expect to live there, rather than as an ad-hoc boolean on this NPC?

pablitar: It kinda depends. If the state should outlive the node, then yes. If not, then no. If the quest is replayable, maybe we don't want this stuff to be in the global state.

pablitar: Eventually there may be a need for "scoped" state. For instance, state that survives during the lifecycle of a StoryQuest. But we should follow up on that once we have a clearer picture of what state do we need.

manuq: The other option is to look at the game state from the player point of view. "Have I already talked to the bard?" Which suggests that the state should be global.

wjt: Or a property of the player...!

manuq (approved):
The branching narrative is an improvement to the puzzle.

scenes/music_puzzle/musician.dialogue
```diff
@@ -1,16 +1,39 @@
...
 Musician: Hello there, Storyweaver.
-Musician: Have you seen those stones all over this island?
-- What about them? They're just inert lumps of rock.
-	Musician: Oh, I wouldn't be so sure about that!
-- There's something unusual about them… it's like they have a certain energy.
-Musician: They're tuned to a major scale!
-	Legend has it that playing the right melody will open a door.
+if puzzle.get_progress() >= 1:
+	Musician: Oh! Have you been playing with the stones over there?
+	Player: I kicked them and something strange happened...
```
wjt, line 13: I don't think it's necessary here but you could imagine the player having a name, and it turns out that you can use `{{ variable access }}` in the "character name" position, i.e. before the colon.

https://github.com/nathanhoad/godot_dialogue_manager/blob/main/docs/Basic_Dialogue.md#variables-in-dialogue

So the way to do this (in future) would be to add the player who is doing the interaction that triggered the dialogue to the extra_game_states array too.

PR #44, opened 2025-03-26
Project structure: Group interact files

manuq:
Create a new folder scenes/interact/. And move interact ray, label and area files (scenes and scripts) into it. This way they are grouped and not spread in the scenes/ folder.


JuanFdS (approved):
👍🏾

Issue #6, opened 2025-03-05
Script the initial narrative


manuq:
The narrative moments for the MVP need to be written before it can be implemented:

> As a player, I begin by opening the Threadbare game. I see a Splash Screen that shows an animation of a fabric world getting unraveled and torn, turning a dull gray. I press start.
I approach an Elder (NPC) who provides a narrative (text) of why I have been summoned to Threadbare. The elder sends me on a quest to find a story and collect Memory, Imagination, and Spirit (just collectively) and bring it back to the Eternal Loom.
I then encounter a narrative screen telling me the beginning of a story. Next, I am faced with a challenge (this could be a tilemap maze, a combat experience, or an interactive puzzle). I earn a star or similar (one of three stars total).
I encounter another narrative screen, this one telling me the middle of the story. I do another challenge and earn another star.
I encounter a third and final narrative screen, the end of the story. I do my final challenge and earn my final star.
I visit the Elder again and celebrate my three earned stars. The collective Memory, Imagination, and Spirit I collected are woven into the Loom.
I see a Splash Screen that shows the Threadbare world from Splash Screen 1, that depicts color returning to Threadbare and a new section of the world added. (editado) 

For example, write in detail the exposition happening here: "I encounter another narrative screen, this one telling me the middle of the story".

pablitar:
In order to do this, I suggest we get somebody to write a small story that we can then adapt. Character dialog is specially useful in this regard!

pablitar:
> > In order to do this, I suggest we get somebody to write a small story that we can then adapt. Character dialog is specially useful in this regard!
> 
> Yes, Stephen, Andrea and Heather are supposed to work on this.

Should we then assign this to them somehow? Can we get updates on what they're doing so we can give feedback from an implementation standpoint?

manuq:
> I'll ask on Discord. Sorry about the mess, but I want to push so that we can start working on some of these uncertainties

No, you are right, ideally we would be working this project in the open. Let's ask for their GitHub handles.

Stregoica777:
@manuq @wjt **What's the deadline for this?**
I can start drafting these ASAP (but a deadline would be nice):
1) Dialogue with the Elder;
2) Narrative screen: beginning of the story;


> The narrative moments for the MVP need to be written before it can be implemented:
> 
> > As a player, I begin by opening the Threadbare game. I see a Splash Screen that shows an animation of a fabric world getting unraveled and torn, turning a dull gray. I press start.
[...]

PR #46, opened 2025-03-26
intro: Replace placeholder text with dialogue

wjt:
Previously, a one-line placeholder introduction to the world was displayed by animating the visible_ratio property of a RichTextLabel.

Instead, define the on-screen introduction text as a dialogue file. Trigger the fade-out animation with a mutation (a.k.a. method call) in the dialogue file. Replace the RichTextLabel with an invocation of that dialogue.

https://github.com/endlessm/threadbare/issues/6

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -91,7 +91,7 @@ running={
 
 [internationalization]
 
-locale/translations_pot_files=PackedStringArray("res://scenes/npcs/talker/default.dialogue", "res://scenes/music_puzzle/musician.dialogue")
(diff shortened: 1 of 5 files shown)


wjt:
Feedback from @Stregoica777 –

> THB I would like the text to be bigger (the font size I mean) because the game is about storytelling, so words have much weight. 

I have been thinking that we may want to adjust the dialogue balloons for the rest of the game anyway, so I am tempted to say, let's do that as a separate change and improve the font size (and legibility when scaled) while doing that.

manuq (approved):

wjt:
Ugh, this seems to crash in the web export...

Issue #37, opened 2025-03-21
Checkpoints


LauraFoglia:
- Saving points so is not really punishing if you loose: automatic

wjt:
As discussed on https://github.com/endlessm/threadbare/issues/32#issuecomment-2752033967 we will have one scene per challenge (so 3 scenes per storyquest). In the case where failing a challenge just means starting again from the beginning, there may not need to be a checkpoint – just restart the current scene. But we may want checkpoints within a challenge, or to save some state across attempts.

wjt:
> But we may want checkpoints within a challenge, or to save some state across attempts.

Happy to see (the videos on) #101. From #82, after the third time I acted as a human checkpoint to get the playtester back to the point they had reached, I changed my mind: the stealth challenge does need checkpoints.

PR #47, opened 2025-03-26
Stealth Level

JuanFdS:
This adds a new story quest challenge that's focused in stealth.

- It adds a new type of enemy, the Guard, which patrols along a path and tries to detect the player, if it does, it "alerts" about that.

- The way the stealth level works, if any Guard is alerted, the level is lost and has to be started again. Checkpoints will be added after completing #37 .

- The tileset is a bit more complex now, there's an occlusion layer and we have 2 terrains for stone because I wanted to be able to tell apart different parts of the stone: tiles that look the same have different occlusion shapes (c40ab7fe15e53427032e99ff96e633964e8038c5)
Autotiling setup displaying 2 different terrains using the same tilesheet. The tiles in the top belong to the stone terrain, and the ones in the botom to the stone_wall terrain. The stone terrain is used for stone floors, and the stone_wall terrain for the edges of the stone terrain Occlusion configuration for that same tilesheet. All tiles occlude almost completely, except the stone walls and the lower bit of the stone tiles that are connected to those walls, which is, the stone_wall terrain

- The lighting of the level was achieved by using a `CanvasModulate` node to darken the scene, and `LightPoint2D`s in the Guards.

- Added a new physics layer: `sight_occluders`, those are used to check if there is any wall or block between the Guard and the Player, in that case, the Player is not seen.

### GameLogic

A GameLogic node is on the scene to let learners modify some behavior of the level:
- `player_instantly_loses_on_sight` (boolean)
- `zoom` (float)
Screenshot of the godot editor showing the GameLogic node selected in the SceneTree as child of the root node. In the inspector, the configurable variables are shown.

Initially, I thought about adding more configuration in that node (guards speed, detection area scaling, etc.) but ended up deciding to let those things be configurable per Guard because that lets us have some different kinds of Guards just by tweaking the parameters and paths they follow.

### Guards and patrol paths

Each guard knows a patrol path that it follows while it is patroling. Added a button in the inspector to modify/create a patrol path.
The path itself is just a Godot's builtin Path2D.

Ideally, I'd have wanted to have both the Guard and the Path2D not as siblings but as parent and child to make the structure clearer in the scene tree, but if I did that, in runtime the position of the points would move relative to the Guard.
If you have any opinions or ideas about it, go ahead 😅 

Screenshot showing the stealth level. A guard and its patrol path can be seen. The guard is selected, in the inspector the following things can be seen: a tool button with the text "Edit Patrol Path", a patrol_path variable that has a Path2D assigned, wait_time, move_speed, and some more options that can be found in guard.gd


addons/shaker/plugin.gd
```diff
@@ -0,0 +1,12 @@
...
+
+
+func _enter_tree() -> void:
+	# Initialization of the plugin goes here.
+	pass;
+
+
+func _exit_tree() -> void:
+	# Clean-up of the plugin goes here.
+	pass;
```
manuq, line 12: This is very empty. I really dislike this part of the shaker plugin :)

addons/shaker/plugin.cfg
```diff
@@ -0,0 +1,7 @@
+[plugin]
+
+name="Shaker"
+description="Shake any properties of a node. You could use it as a camera shake effect."
+author="Lenny"
+version="1.0"
+script="plugin.gd"
```
manuq, line 7: Is this one? https://github.com/Eneskp3441/Shaker . The format of `plugin.cfg` doesn't have a field for the plugin source, which doesn't help.

addons/shaker/shaker.gd
```diff
@@ -0,0 +1,108 @@
+class_name Shaker
+extends Node
```
manuq, line 2: All the plugin provides is this `addons/shaker/shaker.gd` script. I think that we should be careful with the dependencies we add and weight the pros and cons. I can't forget that the well-stablished Dialogue Manager bited us today :) . A shake effect can be done with a simple AnimationPlayer or Tween. So I would add our custom implementation rather than adding a dependency for this case.

project.godot
```diff
@@ -99,6 +99,7 @@ locale/translations_pot_files=PackedStringArray("res://scenes/npcs/talker/defaul
 2d_physics/layer_1="player"
 2d_physics/layer_2="npcs"
 2d_physics/layer_3="player_detectors"
+2d_physics/layer_4="sight_occluders"
```
manuq, line 102: :+1:

manuq (changes requested):
This is working really well. Partial review for now, as I'm running out of time today but I wanted to give you some feedback.

I think that the lights and light occlussion add some complexity but is good to have.

If you think that the shaker plugin can be replaced by your own implementation or a simple AnimationPlayer, I would prefer that for now.

What is the source and license of the PNG assets added? If you did them yourself you can add that to the commit message.

Could you add a description to the last 3 commits? And you have added a very good explanation to the PR message, can you move that to the commit descriptions?

addons/shaker/shaker.gd
JuanFdS: Yeah, you make a good point 🤔 . I think I got a bad habit of adding that plugin to most projects without thinking too much about it 😅 .
In this case, I think I'd just remove the shaking effect, iirc it was used for a character animation and for a camera shake, and none of those things are just details and aren't really necessary for this PR.

I think it would be a good idea to have a bit of camera shake for juiciness, but that can be done later in a separate PR and we can implement it ourselves. So, I think I'll just remove the plugin and its uses for this PR. What do you think? @manuq

manuq: @JuanFdS yes, removing the effect for now sounds good!

addons/shaker/plugin.cfg
JuanFdS: It's this one: https://github.com/Rofle44-git/shaker-plugin-gd4

scenes/enemies/guard/guard.gd
```diff
@@ -0,0 +1,395 @@
...
+			guard_movement.stop_moving()
+
+
+## Updates where the Guard is looking at.
+func _update_direction(delta: float) -> void:
+	if velocity.is_zero_approx():
+		return
+
+	var target_angle: float = velocity.angle()
+	var turn_speed: float = 10.0
```
manuq, line 152: This `turn_speed` doesn't change, so it could be a `const` at the top of the script.

scenes/enemies/guard/guard.gd
```diff
@@ -0,0 +1,395 @@
+@tool
+class_name Guard extends CharacterBody2D
+## Enemy type that patrols along a path and raises an alert if the player is detected.
+
+## Emitted when the player is detected.
+signal player_detected(player: Node2D)
+
+enum State {
+	## Going along the path.
+	PATROLING,
```
manuq, line 10: The correct grammar would be:
```suggestion
	PATROLLING,
```

scenes/enemies/guard/guard.gd
```diff
@@ -0,0 +1,395 @@
...
+	## Player is in sight, it takes some time until the player is detected.
+	DETECTING,
+	## Player was detected.
+	ALERTED,
+	## Player was in sight, going to the last point where the player was seen.
+	INVESTIGATING,
+	## Lost track of player, walking back to the patrol path.
+	RETURNING
```
manuq, line 18: Although the linter doesn't detect it, it is better style to have a trailing comma here. See https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#trailing-comma
```suggestion
	RETURNING,
```

scenes/enemies/guard/guard.gd
```diff
@@ -0,0 +1,395 @@
...
+}
+
+@export_category("Patrol")
+@warning_ignore("unused_private_class_variable")
```
manuq, line 22: Can you explain why this warning ignore is needed?

scenes/enemies/guard/light.gd
```diff
@@ -0,0 +1,52 @@
+@tool
+extends PointLight2D
+
+@warning_ignore("unused_private_class_variable")
```
manuq, line 4: Can you explain why this warning ignore is needed?

scenes/stealth_level/game_logic.gd
```diff
@@ -0,0 +1,25 @@
...
+@export var player_instantly_loses_on_sight: bool = false:
+	set = set_player_instantly_loses_on_sight
+@export_range(0.5, 3.0, 0.1, "or_greater", "or_less") var zoom: float = 1.0:
+	set = set_zoom
+
+@onready var enemy_guards: Node2D = %EnemyGuards
+@onready var camera_2d: Camera2D = %Camera2D
+
+
+func set_player_instantly_loses_on_sight(new_value):
```
manuq, line 13: This is probably raising a warning because the return value is missing:
```suggestion
func set_player_instantly_loses_on_sight(new_value) -> void:
```

scenes/stealth_level/game_logic.gd
```diff
@@ -0,0 +1,25 @@
...
+	player_instantly_loses_on_sight = new_value
+	if enemy_guards:
+		for child: Node in enemy_guards.get_children():
+			var guard := child as Guard
+			if guard:
+				guard.player_instantly_detected_on_sight = player_instantly_loses_on_sight
```
manuq, line 19: This could be rewritten to avoid so many nesting, by returning early:
```suggestion
	if not enemy_guards:
		return
	for child: Node in enemy_guards.get_children():
		var guard := child as Guard
		if not guard:
			continue
		guard.player_instantly_detected_on_sight = player_instantly_loses_on_sight
```

scenes/enemies/guard/guard_movement.gd
```diff
@@ -0,0 +1,89 @@
...
+extends Node2D
+
+## Emitted when [member still_time_left] reaches 0
+signal still_time_finished
+## Emitted when [member guard] reached [member destination]
+signal destination_reached
+## Emitted when [member guard] got stuck trying to reach [member destination]
+signal path_blocked
+
+@export var guard: Guard
```
manuq, line 12: If we can safely assume that the owner of this node is a Guard, then this can be simply:
```suggestion
@onready var guard: Guard = owner
```
And the `_ready()` below can be removed. I'm switching to this in some nodes of Player in my branch.

scenes/stealth_level/stealth_quest.gd
```diff
@@ -0,0 +1,18 @@
...
+		var guard := node as Guard
+		if guard:
+			guard.player_detected.connect(self.on_player_detected)
+
+
+func on_player_detected(player: Node2D) -> void:
+	player.process_mode = ProcessMode.PROCESS_MODE_DISABLED
+	await get_tree().create_timer(2.0).timeout
+	await create_tween().tween_property(transition_fade, "color:a", 1.0, 0.5).finished
+	get_tree().reload_current_scene()
```
manuq, line 18: :+1: this is the simple way to restart the minigame, I'm doing exactly the same thing in my branch.

manuq (approved):
Looks good! I left a few suggestions and some questions.

scenes/enemies/guard/guard.gd
JuanFdS: do you know if that could be added to the linter? 🤔

scenes/enemies/guard/guard.gd
JuanFdS: Yes, that variable is not used in the rest of the code, its only use is being shown as a button in the inspector thanks to `@export_tool_button`

scenes/enemies/guard/light.gd
JuanFdS: Same as https://github.com/endlessm/threadbare/pull/47#discussion_r2016787760

scenes/enemies/guard/guard_movement.gd
JuanFdS: yeah, I think that for a node as tighly coupled as this one that would be fine 👍🏾

scenes/enemies/guard/guard.gd
wjt: The linter is https://github.com/Scony/godot-gdscript-toolkit/wiki/3.-Linter + https://github.com/Scony/godot-gdscript-toolkit/wiki/4.-Formatter . 

I believe that, like Python's `black`, the presence or absence of a trailing comma is something that's treated specially by the formatter. If you have:

```gdscript
enum Foo { A, B }
```

it will stay on one line if it can. If you have a trailing comma:

```gdscript
enum Foo { A, B, }
```

It will always be wrapped with one element per line.

wjt: I checked the documentation and there's no way to force one style or other.

JuanFdS: Ohhh I see, and also from what I've checked the formatter isn't configurable (well, only max-length)
https://github.com/Scony/godot-gdscript-toolkit/wiki/4.-Formatter

Issue #48, opened 2025-03-26
Web export crashes during introduction


wjt:
Midway through the introductory dialogue, the web export crashes:

![Image](https://github.com/user-attachments/assets/317ae75f-7983-4f30-86d3-0f7fc99d2ed2)

I don't know if this is due to #46 (which added dialogue to the intro screen) or #45.

wjt:
When I run the web export of commit 35518660 locally, I get the first 2 errors but not the RuntimeError (perhaps because it's using a different export template, the debug one not the release one).

The translation is not present due to #45 – one of the changes in GDM 3.5.0 is that it excludes its own `l10n` directory from the exported game, on the basis that the strings there are not needed at runtime I guess.

But actually that's wrong. The crash is at this block:

https://github.com/endlessm/threadbare/blob/3551866096447e851fb806f38a675e01d289f954/addons/dialogue_manager/dialogue_manager.gd#L980-L983

when resolving this line of dialogue:

https://github.com/endlessm/threadbare/blob/3551866096447e851fb806f38a675e01d289f954/scenes/intro/intro.dialogue#L4

The translation for the `runtime.method_not_found` message is looked up even if `found` is `true` (which indicates that the method was found and hence `show_error_for_missing_state_value` will early-return:

https://github.com/endlessm/threadbare/blob/main/addons/dialogue_manager/dialogue_manager.gd#L514-L515

I think this is a regression in GDM. I'll file a bug upstream and revert the upgrade for now.

wjt:
Reported upstream: https://github.com/nathanhoad/godot_dialogue_manager/issues/835

PR #49, opened 2025-03-26
Back out "Update Dialogue Manager addon to v3.5.0"

wjt:
This backs out commit 380f118a1c444d852a7e5810d4d8d52786fa4fd4.

There is a regression in this release: Dialogue Manager's export plugin strips out its own translations during export, presumably on the basis that they are not used at runtime, but in fact they are.

Revert this change for now, pending a fix in Dialogue Manager.

Fixes https://github.com/endlessm/threadbare/issues/48


manuq:
I see, let's revert for now then.

PR #50, opened 2025-03-26
Restore and save current scene to URL hash in web builds

### LINKED ISSUE #40: Allow "deep linking" to specific scenes (for development/testing)
In other projects we have found it useful during development to be able to link to a web build of a proposed branch, running a specific scene. We have per-branch builds at https://endlessm.github.io/threadbare/branches/ but currently there is no way to play a specific scene other than to temporarily change the main scene in `project.godot` in the branch.

For Candy Collective, I implemented a feature where you can use the URL hash to link to a particular "world". For example, <https://endlessm.github.io/candy-collective/#Worlds/World%20One> jumps straight to "World One", one of the "Community Worlds" that you can select from the main menu at <https://endlessm.github.io/candy-collective/>.

Another approach would be to have a "cheat code" to select a particular quest/scene from an in-game menu. This might be useful as the number of quests increases & for non-web use, but I think for development it might be nice to do something along the lines of the URL hash approach.

wjt:
During development, it is very useful to be able to link directly to a scene, so that a pull request can be easily playtested.

To enable this, use `SceneSwitcher` whenever we switch scenes.

Then, extend it to update the URL hash when switching scenes, if running on the web platform. On startup, check the URL hash, and if it matches a scene, switch to it. Later changes to the URL hash are ignored.

Fixes #40


wjt:
Try going to:

https://endlessm.github.io/threadbare/branches/endlessm/scene-in-url-hash/#uid://cufkthb25mpxy

Any resource path that points to a scene is accepted as input, but it will be resolved to its 
more human-readable `res://` path. Then try walking between the two scenes of the world map.

manuq:
Works very well! @pablitar is also adding a SceneSwitcher singleton: https://github.com/endlessm/threadbare/pull/51/commits

wjt:
I saw that – and actually this branch will also conflict with a local branch I have to string together the game. I'll leave this in draft until #51 lands then adapt it

wjt:
@JuanFdS tagging you for review since you liked the idea on #40 :)

pablitar:
This is great! I'd love to see a shorthand for loading scenes that doesn't require writing the whole scene path in the URL. Maybe by default complete the path using "scenes" as the root? Or any other convention

wjt:
That could work – if the path is relative, resolve it against `res://scenes/`. I'll try that tomorrow.

In Candy Collective I stripped the `res://` prefix and the `.tscn` suffix, which does make the URLs less ugly, but I switched here to using the full absolute path (with scheme) when I realised that it makes `uid://` paths work. Not strictly necessary but a nice easter egg :)

wjt:
> Try going to:
> 
> https://endlessm.github.io/threadbare/branches/endlessm/scene-in-url-hash/#uid://cufkthb25mpxy

Absolute paths (`uid://` or `res://`) still work, but if they match the standard path (starts with `res://scenes/`, ends with `.tscn`) these are stripped.

So this link gets normalised to https://endlessm.github.io/threadbare/branches/endlessm/scene-in-url-hash/#world_map/frays_end

pablitar (approved):
Love this.

PR #51, opened 2025-03-27
Adding a spawn point for teleporters

### LINKED ISSUE #35: Teleporters should have an spawn point

pablitar:
Since areas in the world map could have different entrances, it's important that a teleporter has information not only to which scene it should go to, but also to which point on that scene the player should spawn.

This PR adds a property called "spawn_point" to each teleporter, which can contain the node path of the spawn point in which the player should be moved at the start of the scene. Some considerations:
- Since the node reference of the spawn point is to another scene, we cannot use the default behaviour of the @export annotation with a NodePath. Instead, we can read the target scene and find all those nodes that belong to the "spawn_point" group, and list those as options. This approach work quite well, with some limitations. For instance, it doesn't work if the spawn point is defined in the target scene, but instead in a instanced scene (scene within a scene)
- To identify the player, a group called "player" was added
- To encapsulate the logic of changing scenes, a SceneSwitcher singleton was added. For the time being it is quite simple, but eventually could have logic regarding transitions, scene initialization, caching, and may other goodies.
- To communicate the spawn point to the target scene, we are using the GameState singleton. This is encapsulated in the SceneSwitcher singleton in case we want to change this implementation eventually.
- To prevent camera smoothing from misbehaving when the player spawns somewhere, smoothing is temporarily disabled during the process of spawn. This could be followed up with an issue of implementing a proper camera with its own behavior and script, and maybe make the camera "teleport aware"

Fixes #35


scenes/world_map/teleporter.gd
```diff
@@ -1,17 +1,72 @@
...
+
+@export_file("*.tscn") var scene_to_go_to: String:
+	set(new_value):
+		scene_to_go_to = new_value
+		_update_available_spawn_points()
+		notify_property_list_changed()
+
+var spawn_point_path: NodePath:
+	set(new_val):
+		if new_val == ^"NONE":
```
manuq, line 15: I see below that `^"NONE"` is something you are building with the hint string. Is it because there is no way to pass the empty path `^""` ?

manuq (approved):
All is very well explained. I only left a question I have about the NodePaths.

scenes/world_map/scene_switcher.gd
```diff
@@ -0,0 +1,6 @@
+extends Node
```
wjt, line 1: My only other nitpick would be that I don't think this should live in the `world_map` directory. I would expect this to be used to switch between the scenes that make up a quest, too, even though the `spawn_point` functionality would probably not be used, just so that we can have consistent transitions between scenes and (as in #50) do fancy stuff like putting the current scene into the URL hash in web builds.

In that branch I placed it at `res://scenes/scene_switcher/scene_switcher.gd` – there's no actual *scene* there but it seemed as good a home as any and it matches the `game_state.gd` script being in a `res://scenes/game_state` directory.

scenes/world_map/teleporter.gd
pablitar: Yes! Since the hint_string expects a comma separated list of values, if you have an empty string, it gets ignored. Or at least it did that when I was testing it. Maybe I did something wrong.

In any case, I also believe that having a NONE element is clearer than having an "" element.

JuanFdS (dismissed):
Great work!, but I think we may have an issue with portal loops (https://github.com/endlessm/threadbare/pull/51#discussion_r2016746689)

We can settle in having the spawn points not collide with the teleporters for now and handle that later (if you think that's the way to go, tell me and I'll unblock the PR), but I think we should let spawn points and teleporter be at the same position without teleporting the player back.

scenes/world_map/teleporter.gd
```diff
@@ -1,17 +1,72 @@
...
 	collision_mask = 0
 	set_collision_layer_value(3, true)
 	set_collision_mask_value(1, true)
 
+	if Engine.is_editor_hint():
+		return
+
+	# This is to stop a possible infinite loop if the player spawns on a
+	# teleporter that returns them to the teleporter that sent them
+	await get_tree().physics_frame
```
JuanFdS, line 34: Oh sorry, this comment was supposed to be part of my review!:

I'm not sure what case is being handled by this, but if teleporters and spawn points overlap, we can get a infinite portal loop (!)

https://github.com/user-attachments/assets/9fe40023-5c26-4b54-ab9d-9fc960f8eef9

scenes/world_map/teleporter.gd
```diff
@@ -1,17 +1,72 @@
...
+		_available_spawn_points = paths
+
+
+func _get_property_list() -> Array[Dictionary]:
+	var property_list: Array[Dictionary] = []
+
+	property_list.push_back(
+		{
+			"name": "spawn_point_path",
+			"type": TYPE_STRING,
```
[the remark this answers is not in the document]
JuanFdS: That would create a widget in the inspector to let us choose a nodepath... but in the current scene. I think there isn't a built-in way to pick a NodePath from a different scene.

image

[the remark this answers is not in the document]
manuq: > That would create a widget in the inspector to let us choose a nodepath... but in the current scene. I think there isn't a built-in way to pick a NodePath from a different scene.
Ah I see, thanks.

scenes/world_map/teleporter.gd
pablitar: Heh. Honestly, I didn't test this. I assumed that waiting for a physics frame before connecting the signal should prevent the loops. I'll see what I can do

scenes/world_map/teleporter.gd
```diff
@@ -1,17 +1,72 @@
...
+	# teleporter that returns them to the teleporter that sent them
+	await get_tree().physics_frame
+
+	self.body_entered.connect(_on_body_entered)
+
 
 func _on_body_entered(_body: PhysicsBody2D) -> void:
 	if scene_to_go_to:
-		get_tree().change_scene_to_file(scene_to_go_to)
+		SceneSwitcher.change_scene_to(scene_to_go_to, spawn_point_path)
```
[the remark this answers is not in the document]
pablitar: Fixed!

JuanFdS (approved):
👍🏾

scenes/world_map/teleporter.gd
pablitar: It's a bit more complicated than it should. Created #56 to follow up on this

PR #57, opened 2025-03-27
Combat 01 merge

manuq:


manuq:
This is now ready for review. Some comments:

1. I'm not happy that I ended up duplicating the player for this scene. This is because after the discussion on Friday it was decided to make player fighting a special case of the combat minigame. I think it's easier this way, but also I would like to make fighting a component of the player in the future. I tried but coudn't yet, and I don't want to delay merging this anymore. Still I left some commits that I did to the original player. I think they are good to have in any case.

2. This is not yet hooked with the game, no teleports yet. And there is no collectible at the end. I think this can happen after merging. Also we discussed making "fill all inkwells" the win condition on Friday.

3. There is no narrative in this scene yet.

scenes/ink_combat/ink_drinker/ink_drinker.gd
```diff
@@ -0,0 +1,65 @@
...
+func _on_timeout() -> void:
+	var player: FightingPlayer = get_tree().get_first_node_in_group("player")
+	if not is_instance_valid(player):
+		# or player.is_queued_for_deletion():
+		return
+	animated_sprite_2d.play(&"attack anticipation")
+	await animated_sprite_2d.animation_looped
+	animated_sprite_2d.play(&"attack")
+	var ink_blob: InkBlob = INK_BLOB.instantiate()
+	ink_blob.direction = player.global_position - ink_blob_marker.global_position
```
JuanFdS, line 36: Just so we don't have to normalize it later:
```suggestion
	ink_blob.direction = player.global_position.direction_to(ink_blob_marker.global_position)
```

scenes/ink_combat/ink_drinker/ink_drinker.gd
```diff
@@ -0,0 +1,65 @@
...
+	)
+	Engine.get_main_loop().current_scene.add_child(ink_blob)
+	await animated_sprite_2d.animation_looped
+	animated_sprite_2d.play(&"idle")
+
+
+func _on_got_hit(body: Node2D) -> void:
+	if body is InkBlob and not body.can_hit_enemy:
+		return
+	body.queue_free()
```
JuanFdS, line 49: I feel like we are handling some of the collisions from the InkBlob and some from the things that the InkBlob collides with 🤔 . I think it'd be easier to follow if InkBlob handled what it happens to itself depending on what it hits (just passing through, being freed, changing its direction) and if it made contact with other thing, it notified them with a `got_hit()` message to let them handle what they have to do.

scenes/ink_combat/ink_drinker/ink_drinker.gd
```diff
@@ -0,0 +1,65 @@
...
+	animation_player.play(&"got hit")
+	health -= 0.05
+	health_bar.value = clamp(health, 0., 1.)
```
JuanFdS, line 52: It might be just a style preference, but I usually use move_toward for these cases. It makes sure the final value doesn't go past the destination (0.)
```suggestion
	health = move_toward(health, 0., 0.05)
```

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
+class_name Inkwell
+extends StaticBody2D
+
+const SPLASH: PackedScene = preload("res://scenes/ink_combat/splash/splash.tscn")
+
+@export_enum("Cyan", "Magenta", "Yellow", "Black") var ink_color_name: int = 0
```
JuanFdS, line 6:
```suggestion
@export var ink_color_name: InkBlob.InkColorNames
```

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
...
+func got_hit(
+	ink_blob_global_position: Vector2, _hit_vector: Vector2, ink_blob_color_name: int
+) -> void:
+	var splash: Node2D = SPLASH.instantiate()
+	splash.ink_color_name = ink_blob_color_name
+	Engine.get_main_loop().current_scene.add_child(splash)
+	splash.global_position = ink_blob_global_position
+
+
+func _get_ink_drinkers() -> Array[Node]:  #Array[InkDrinker]:
```
JuanFdS, line 34: I wish Godot could type this as `Array[InkDrinker]` 🙏🏾 . 

I'm not sure if it'd work, but maybe you can cast it when returning it?
Otherwise, I would remove the comment and add godot docs comments to explain this is expected to return `InkDrinker`s
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_documentation_comments.html

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
...
+	var is_ink_drinker: Callable = func(enemy: Node2D) -> bool: return enemy is InkDrinker
+	var enemies: Array[Node] = get_tree().get_nodes_in_group(&"enemies")
+	return enemies.filter(is_ink_drinker)
```
JuanFdS, line 37: why not have a group for the `ink_drinkers` type of enemy?

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
...
+
+
+func fill() -> void:
+	animation_player.play(&"fill")
+	ink_amount += 0.334
```
JuanFdS, line 42: Mmmm, I know this is supposed to be 1/3 of the inkwell, but I'm not sure how clear it'd be for learners to realise what this number means.

We could also have an amount of `ink_blobs` needed to fill the inkwell, and then use integers for this (`ink_amount += 1`, or, `ink_blobs_amount += 1`).

JuanFdS (changes requested):
Awesome!, left some comments, I'd like us to at least address the repetition of the colors before merging.

Then, can we create a separate issue for making the fighting a separate component of the player so we don't need to duplicate the entire player? 🤔 . I'd also be fine with the player always having the "fighting" component as a child and being able to toggle it off or on through an exported variable.
We can also discuss approaches to how we'd like the player to be extended (adding child nodes that extend its behavior without it needing to know anything about it?, having it be one big thing with configurations?, etc)

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
...
+	interact_label.reparent(screen_overlay)
+	interact_label.label_text = InkBlob.InkColorNames.keys()[ink_color_name]
+
+
+func got_hit(
+	ink_blob_global_position: Vector2, _hit_vector: Vector2, ink_blob_color_name: int
+) -> void:
+	var splash: Node2D = SPLASH.instantiate()
+	splash.ink_color_name = ink_blob_color_name
+	Engine.get_main_loop().current_scene.add_child(splash)
```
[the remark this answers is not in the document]
manuq: I saw this in Dialogue Manager and started using it. It is in `scenes/player/player_interaction.gd` already as well. But you are right, `get_tree().current_scene` should be enough.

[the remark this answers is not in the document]
manuq: I fixed the new ones in this PR and also the existing one in a separate commit.

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
...
+@onready var interact_label: Control = %InteractLabel
+
+
+func _ready() -> void:
+	var color: Color = InkBlob.INK_COLORS[ink_color_name]
+	animated_sprite_2d.modulate = color
+
+	var current_scene: Node = Engine.get_main_loop().current_scene
+	var screen_overlay: CanvasLayer = current_scene.get_node("ScreenOverlay")
+	interact_label.reparent(screen_overlay)
```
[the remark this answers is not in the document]
manuq: The whole point of the ScreenOverlay is to keep the labels the same size, regardless of the camera zoom. See 2c085ccc9b13b434f822e9c34755f23cc6f1d141

However I see that's not working as expected anymore, or at least not in this PR. I will switch these labels to normal ones for now.

scenes/ink_combat/inkwell/inkwell.gd
manuq: Ouch I didn't meant to leave this comment. Initially I had `Array[InkDrinker]` but Godot complained with a runtime error: `Trying to return an array of type "Array[Node]" where expected return type is "Array[InkDrinker]".`

I now tried casting `return enemies.filter(is_ink_drinker) as Array[InkDrinker]` but the same error happens:
Captura desde 2025-03-31 14-46-51

So I removed the comment for now.

scenes/ink_combat/inkwell/inkwell.gd
manuq: I agree! The problem is that right now I only have 3 frames of fill animation :) . So I will do as you say, but hardcode the amount needed to 3.

scenes/ink_combat/inkwell/inkwell.gd
```diff
@@ -0,0 +1,49 @@
...
+func fill() -> void:
+	animation_player.play(&"fill")
+	ink_amount += 0.334
+	animated_sprite_2d.frame += 1
+	if ink_amount >= 1.:
+		var ink_drinkers: Array[Node] = _get_ink_drinkers()
+		var ink_drinker: InkDrinker = ink_drinkers.pick_random() as InkDrinker
+		ink_drinker.explode(ink_color_name)
+		queue_free()
+		interact_label.queue_free()
```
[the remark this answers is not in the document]
manuq: I will check after merge if it's worth keeping the "labels stay the same size when camera zoom in / out" feature.

scenes/ink_combat/big_splash/big_splash.gd
```diff
@@ -0,0 +1,13 @@
+class_name BigSplash
+extends Node2D
+@export_enum("Cyan", "Magenta", "Yellow", "Black") var ink_color_name: int = 0
```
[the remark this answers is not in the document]
manuq: This is much better! Thanks.

manuq:
> Awesome!, left some comments, I'd like us to at least address the repetition of the colors before merging.

Thanks! I addressed all your comments except the one for the labels in a separate CanvasLayer. I will iterate on it later. Maybe I'm overthinking that these labels (and the ones to interact with the NPCs) need to be unscaled with the camera.

> Then, can we create a separate issue for making the fighting a separate component of the player so we don't need to duplicate the entire player? 🤔 . I'd also be fine with the player always having the "fighting" component as a child and being able to toggle it off or on through an exported variable. We can also discuss approaches to how we'd like the player to be extended (adding child nodes that extend its behavior without it needing to know anything about it?, having it be one big thing with configurations?, etc)

Yes I will do that after merge. I like the idea of having the component as a child that can be toggled on/off, to keep things simple.

> EDIT: OH!, also, where are we using the ink_wall collision layer? I see that it's on the `AirStream` attack, but that node already collides with the ink because of its collision_mask. What thing should the ink_wall layer be applied to?

Sorry, that was for any other object that we want the pinball effect in the ink. But since none is part of this PR, I removed it.

PR #58, opened 2025-03-27
player: Use unit-length input vector

wjt:
Previously, the horizontal and vertical axes were fetched separately as a float between -1 and 1, and then combined into a Vector2. In the case where the player is pressing right and down, the resulting vector will be (1, 1), which has length sqrt(2) ≈ 1.4. This means that the player can move faster on the diagonal than horizontally or vertically.

Instead, use Input.get_vector(), which returns a unit vector based on the four actions provided. This means the player moves the same speed regardless of which direction they're walking/running in.

### DIFF
--- a/scenes/player/player.gd
+++ b/scenes/player/player.gd
@@ -13,10 +13,7 @@ func _process(delta: float) -> void:
 		velocity = Vector2.ZERO
 		return
 
-	var axis: Vector2 = Vector2(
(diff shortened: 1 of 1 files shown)


pablitar (approved):
Thank you for fixing this! I remember seeing it in the demo, and then completely forgot about it hahaha

PR #59, opened 2025-03-27
Add joypad mappings for running and skipping dialogue

wjt:
This binds B to ui_cancel (which skips to the end of the current dialogue line, for the impatient) and LB to run.

It also adjusts the ui_accept binding to only match joypad device 0, in common with all the other joypad mappings. Arguably for a single-player game we could allow input from any connected joypad, but we should be at least be consistent.


pablitar (approved):
LGTM!

PR #60, opened 2025-03-27
pre-commit: Run formatter before linter

wjt:
In my experience, gdlint often complains about problems (such as line length) that gdformat subsequently fixes.

Run the two commands in the opposite order to reduce noise in the output.


pablitar (commented):
LGTM!

pablitar (approved):

PR #63, opened 2025-03-27
Adding transitions singleton to handle transitions between scenes and a method to use transitions in Scene Switcher

### LINKED ISSUE #34: Add transitions when the player changes scenes in the world map
- Fade to black / Fade from black
- Nice to have: The player walking from the entrance

pablitar:
This is a draft until #50 is merged

Added a singleton that can be used to show transitions when changing scenes. There are currently 3 different effects: Fade, Wipe, and Radial. More effects could be added with different masks. See examples here: https://github.com/ninetailsrabbit/indie-blueprint-scene-transition/tree/main/addons/ninetailsrabbit.indie_blueprint_scene_transition/src/transitions/dissolve/textures

Also, made it so a teleporter is "one shot". This is because before, if the player moved during the transition, it could trigger the teleport again and cause issues.

Fixes #34


wjt:
This looks great – a really nice improvement.

But I think we should stop processing player input during the transition. Otherwise, if you switch direction as soon as the transition starts, the player character ends up in an unexpected location on the other side of the transition.

image

I guess we could pause the game during the transition – though this would mean the trees also stop blowing in the wind.

pablitar:
> This looks great – a really nice improvement.
> 
> But I think we should stop processing player input during the transition. Otherwise, if you switch direction as soon as the transition starts, the player character ends up in an unexpected location on the other side of the transition.
> 
>  Screencast.from.2025-03-28.10-58-19.webm 
> I guess we could pause the game during the transition – though this would mean the trees also stop blowing in the wind.

I agree! But I think we should follow up in another issue since pausing is something that can affect a lot of the game: https://github.com/endlessm/threadbare/issues/70

If you are OK, I'd like to merge this and then we can follow up with that issue

wjt:
Sure, I think that's fine. Sorry, I introduced conflicts in #50 :(

pablitar:
> Sure, I think that's fine. Sorry, I introduced conflicts in #50 :(

No worries! That was expected :D

pablitar:
@wjt ready for review and merge!

Issue #10, opened 2025-03-11
Visual style research


manuq:
Let's try to find the visual style. From design documents:

> The game world is shaped loosely by the three Sacred Elements – Memory, Imagination, and Spirit. 
> 
> The world of Threadbare, a woven fabric of life and wonder, made real by stories, is split into regions, loosely sewn together like a patchwork quilt. These regions lean toward one of these elements, but their borders are fluid, shifting and blending naturally rather than forming rigid divisions. 
> 
> Stories emerge, evolve, and intertwine as contributors add to the world, making it a living, ever-changing ecosystem.

It has also been mentioned that the world should look "tactile".

manuq:
For "tactile" I tried a soft floor effect, using Yoshi's Whooly World as reference. I think it can look good if the tiles are fabric patterns. Although it doesn't respect pixel art:

https://github.com/user-attachments/assets/c63ab8f4-cd2a-4691-998d-68290c06fb92

It lives in this branch: https://github.com/endlessm/threadbare/tree/soft-floor . The shader code is very bad as it is exploratory, I tried 2 different shaders and left the 2 active and mixed. It should be doable with just one (the one with a small ColorRect, not the one occupying the whole canvas!).

manuq:
I realized that the previous exploration had a glitch when running in web browser. A dark circle appears where the shader deforms the floor pixels. Then thanks to folks from the Godot Meetup Argentina I was able to constrain the bug. It was already reported and with a PR opened. So I confirmed that the PR fixes the glitch:
https://github.com/godotengine/godot/pull/103878#issuecomment-2715164989

It was very useful to learn how to download the web export template artifact from the PR, and export the game with a custom export template.

PR #64, opened 2025-03-28
ci: Export with Godot 4.4.1

wjt:
This maintenance release was made a couple of days ago. In particular this should fix a shader bug in the web export that manuq ran into when experimenting with deforming the floor as if made of fabric.

https://godotengine.org/article/maintenance-release-godot-4-4-1/ https://github.com/godotengine/godot/pull/103878#issuecomment-2715164989 https://github.com/endlessm/threadbare/issues/10#issuecomment-2715238382

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GODOT_VERSION: 4.4
+  GODOT_VERSION: 4.4.1
 
 jobs:
   build:


wjt:
Of course we can't easily test the shader bug fix since it's on a separate branch, but https://endlessm.github.io/threadbare/branches/endlessm/ci-export-with-4.4.1/ works correctly.

JuanFdS (approved):
👍🏾

manuq (approved):
Excellent! Thanks.

PR #65, opened 2025-03-28
Apply MPL-2.0 license to original code

wjt:
This is a soft copyleft license, which allows combining this project with closed-source parts (such as, hypothetically, console ports of Godot) while still requiring the source to this project itself to be made available to those who receive an executable build of the game.

Add the license as a COPYING file. Explain in the README that this applies to original code, and not to third-party assets and addons. (The only addon at present is under the MIT license.)

Add a license header to all source files.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,7 @@
+# This is the list of Threadbare's significant contributors.
+#
+# This does not necessarily list everyone who has contributed code,
+# especially since many employees of one corporation may be contributing.
+# To see the full list of contributors, see the revision history in
+# source control.
+Endless OS Foundation LLC
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,373 @@
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
+1.1. "Contributor"
+    means each individual or legal entity that creates, contributes to
+    the creation of, or owns Covered Software.
+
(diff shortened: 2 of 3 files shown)


wjt:
I believe it is good practice to include a license header (and copyright information!) in each source file. However this is the kind of thing which will tend to be forgotten over time.

https://reuse.software/dev/#pre-commit-hook can be used to enforce this. It would need some configuration though – by default it expects `.uid` files to have a copyright header! And `.tscn` files can't realistically have these headers, but they make up a really significant portion of the game.

So perhaps the path of least resistance is to drop that patch.

manuq:
> So perhaps the path of least resistance is to drop that patch.

With this you mean, do not add the header text to all the source files?

wjt:
That's what I mean – I'm not sure how much value it has to put a license header in all the `.gd` files when so much of the project is in other source files that can't have these headers. (`.tscn` files can have comments, but the editor deletes them when saving changes back to the file.) So as a first step, just add the license to the project.

I think we also need an `AUTHORS` file to list major copyright owners (currently just us).

wjt:
I'm actually trying to get the REUSE compliance stuff in place because I think it may help with the credits screen to have a machine-readable description of the author and license of every file/asset pack/etc. But that's a separate task to simply applying a license.

manuq (approved):

Issue #54, opened 2025-03-27
Make dialogue bubbles match the game style


LauraFoglia:
Currently we are using the default dialogue bubbles from Godot Dialogue Manager with Godot's fallback font.

We will want to adjust this in a few ways:

- Use a font that matches the art style of the project
- Adjust the font size, too!
- Adjust the appearance of the bubbles

wjt:
@PlayMatters says:

> can we use the Tiny Swords banner assets for text boxes, etc?

This would avoid needing to design new banners! The assets are [here](./assets/tiny-swords/UI/Banners).

wjt:
I think we should split the input hints to a separate task.

I've started looking at this because I realised that styled text & controls would also be needed for the credits scene (#53)

wjt:
#106 addresses the first half of this, setting a more appropriate font, as well as adding a custom balloon scene that we can modify as needed.

wjt:
The remaining piece here is to style the dialogue-choice buttons.

![Image](https://github.com/user-attachments/assets/b07fb585-8f8b-4eed-b521-9245b89c5761)

In the theme, we need to cover four button states:

- Normal (unfocused dialogue choice, of which there may be several)
- Focused (the dialogue choice that will be activated if you press Space/Enter/A)
- Pressed
- Disabled (we don't currently have any unavailable dialogue choices but GDM supports it)

We do not need to style the "Hover" state, which is used when the mouse is pointing at the element, because the dialogue_responses_menu.gd script has code to focus a button when the mouse enters it.

I learnt from the Button theme property documentation https://docs.godotengine.org/en/stable/classes/class_button.html#class-button-theme-style-focus that the focus state is handled differently: its stylebox is displayed *over* the base stylebox.

Looking at the assets provided by Tiny Swords, it includes various flavours of button assets: square 64×64 images, `3Slides` images which are 192×64 (i.e. 3 tiles wide, 1 tile tall), and `9Slides` 192×192 images (3 tiles wide, 3 tiles tall). 192×192 is way too big for dialogue choices, and I couldn't get the 64×64 images to tile nicely horizontally, so `3Slides` it is.

However I also struggled to set the texture margins & expand margins correctly for the `3Slides` images. I think I must be just misunderstanding how these properties relate to each other.

Also there are not quite the same set of button states as we need. There's a "hover" asset in a neutral yellow colour, then red/blue assets in normal & pressed states, and a disabled state.

[...]

PR #66, opened 2025-03-28
intro: Remove [next=auto] directives

wjt:
I had included [next=auto] directives at the end of each line of the introduction so that the intro text would advance automatically after a few moments (proportional to the length of the text).

What I had not appreciated is that if you configure the text to advance automatically, the player can't press ui_accept (space/A) to move forward more quickly. This is annoying if you want to skip the introduction.

Remove these directives.


JuanFdS (approved):
I'd like to check later if there's any way to configure it to have both it automatically move to the next text while also allowing a skip, but for now I prefer skipping manually 👍🏾

wjt:
Good point. The code that applies this rule actually lives in the balloon:

https://github.com/endlessm/threadbare/blob/267c0e325196f2f7cc1e1074e6c090ecc60beda1/addons/dialogue_manager/example_balloon/example_balloon.gd#L116-L127

So this is something we can easily adapt once we have custom balloons (#54) but, indeed, I think that for now this is less annoying.

PR #80, opened 2025-03-28
Fix #79: stealth level is broken on release builds

### LINKED ISSUE #79: Stealth quest doesn't work in exported game
https://endlessm.github.io/threadbare/#stealth_level/stealth_level

```
index.js:474 SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
index.js:474    at: GDScript::reload (res://scenes/enemies/guard/guard.gd:387)
```

JuanFdS:
It can be played here: https://endlessm.github.io/threadbare/#stealth_level/stealth_level

Having a reference to an editor's only type causes errors on release build. Changed the code so it dynamically finds the node.

Directly referencing a node like EditorInterface can only be done in plugins and scripts that only runs in the editor.

### DIFF
--- a/scenes/enemies/guard/guard.gd
+++ b/scenes/enemies/guard/guard.gd
@@ -383,8 +383,17 @@ func _notification(what: int) -> void:
 ## Function used for a tool button that either selects the current patrol_path
 ## in the editor, or creates a new one
 func edit_patrol_path() -> void:
+	if not Engine.is_editor_hint():
+		return
+
+	# Cannot directly reference [class EditorInterface] in code that isn't
+	# part of a script that runs only in the editor (like plugins).
+	# This function should only be called in the editor, but having a direct
+	# reference to the [class EditorInterface] causes errors on runtime builds.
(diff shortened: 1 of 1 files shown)


scenes/enemies/guard/guard.gd
```diff
@@ -383,8 +383,17 @@ func _notification(what: int) -> void:
...
 ## in the editor, or creates a new one
 func edit_patrol_path() -> void:
+	if not Engine.is_editor_hint():
+		return
+
+	# Cannot directly reference [class EditorInterface] in code that isn't
+	# part of a script that runs only in the editor (like plugins).
+	# This function should only be called in the editor, but having a direct
+	# reference to the [class EditorInterface] causes errors on runtime builds.
+	var editor_interface = Engine.get_singleton("EditorInterface")
```
manuq, line 393: An alternative to the early return `if not Engine.is_editor_hint()` would be:
```suggestion
	var editor_interface = Engine.get_singleton("EditorInterface")
	if not editor_interface:
		return
```
I guess both work.

Issue #53, opened 2025-03-27
Credits scene


LauraFoglia:
We need a scene that displays credits for contributions made to the game, and in particular attribution for third-party assets used in the game whose license requires attribution.

wjt:
I was wondering how best to keep track of the attribution required for different assets. My best idea so far is to define a new resource type with fields describing the author, license, and homepage; having an instance of this resource at the root of each directory of assets; and having the credits scene walk the `res://assets/` directory structure to find all such resources and list them.

wjt:
Removing this for now – I'm not actively working on it and I think other things have higher priority.

PR #88, opened 2025-04-01
Add machine-readable license info for all files

wjt:
This adds machine-readable licensing information for every file in the repo, following the [REUSE](https://reuse.software/) specification, along with a pre-commit hook to validate this for future changes.

Doing this forced me to be more specific about the licenses in the project, including the fact that we have some original assets, which I have placed under CC-BY-SA-4.0, along with some derived assets. I also removed a stray `.zip` file from the repo.

I hope to find a way to use this data to automatically generate information for the future credits screen –  #53 – perhaps by writing an addon that hooks the export process to generate an extra resource.


.pre-commit-config.yaml
```diff
@@ -13,6 +16,10 @@ repos:
   hooks:
   - id: gdformat
   - id: gdlint
+- repo: https://github.com/fsfe/reuse-tool
+  rev: v5.0.2
+  hooks:
+  - id: reuse
```
wjt, line 22: My spider sense tells me that this might be incredibly annoying during development...

wjt: Another option would be to have a separate CI job that checks this – less annoying during development but would still catch errors in CI, and also makes the "Linting and Formatting" check name not be inaccurate – it would be clearer if a "License information" check failed in the case where something is broken.

wjt: I think that's better. We even get a nice summary on the workflow page almost for free:

https://github.com/endlessm/threadbare/actions/runs/14201291432?pr=88

image

assets/midjourney/embroidered-world-map.webp.license
```diff
@@ -0,0 +1,9 @@
+SPDX-FileCopyrightText: 2025 The Threadbare Authors
+SPDX-License-Identifier: CC-BY-SA-4.0
+
+This image was created by Midjourney. Per the Midjourney Terms of Service:
+
+> You own all Assets You create with the Services to the fullest extent possible
+> under applicable law.
+
+https://docs.midjourney.com/hc/en-us/articles/32083055291277-Terms-of-Service#h_01JD8G8KYZ5KTCT2R7W0KM79PZ
```
manuq, line 9: :+1:

assets/unsplash/grey-fabric.jpg.license
```diff
@@ -0,0 +1,7 @@
+https://unsplash.com/photos/a-black-and-white-photo-of-a-blackboard-h-LcVG8W1XY
+
+Copyright Annie Spratt (https://unsplash.com/@anniespratt)
+
+Edited to rotate to landscape orientation, desaturate, and increase compression.
+
+SPDX-License-Identifier: LicenseRef-Unsplash
```
manuq, line 7: :+1:

manuq (approved):
Wow great job. I like that this is so specific and that CC-BY-SA for the original assets makes sense. I went through it and I have no other comments.

PR #89, opened 2025-04-01
Added Pause System

### LINKED ISSUE #70: Add a Pause System
It should allow pausing different elements of the game:
- The whole game
- The player input
- Player, npcs, everything that moves

Don't forget to pause the player during transitions.

pablitar:
Previously, if the player continues to hold a direction during the transition, the character would be moved in the new scene. This could lead to the character ending up in an unexpected location on the other side of the transition – perhaps even off the edge of the scene.

To solve this issue, and to move forward in adding all the systems that the game needs to function, a Pause System has been added, and it is used during scene transitions.

The Pause System allows pausing the whole game, or just the player input, and is flexible enough to allow adding other kinds of pauses. It also supports being requested pauses from a multitude of different origins, and handling it gracefully. This pattern of requests prevents some bugs that could appear if the pause logic becomes more complex. For instance let's say we have an inventory and a pause menu. And both pause the game. The pattern ensures that as long as one is opened, the game is paused. Another safety measure: Each pause request is attached to a node. If that node leaves the tree, then it is assumed that the request should be cleared.

This PR also adds a global singleton called Pause, which handles the pause requests and pausing / unpausing. This has to be global because pauses can affect scenes right after they load (for instance, during a transition), and also, because the mechanism for pausing the whole tree in Godot is global and transcends the current scene.

Fixes #70

### DIFF
--- a/scenes/globals/pause.gd
+++ b/scenes/globals/pause.gd
@@ -1,0 +1,78 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends Node
+
+signal pause_changed(system: System, paused: bool)
+
+enum System {
+	PLAYER_INPUT,
+	GAME,
+}
+
+var _pause_requests: Dictionary[System, Array] = {
+	System.PLAYER_INPUT: [],
+	System.GAME: [],
+}
+
+
+func _ready() -> void:
+	pause_changed.connect(_on_pause_changed)
+
+
+func _on_pause_changed(system: System, paused: bool) -> void:
+	if not is_inside_tree():
+		# This can happen when the game is quitting.
+		return
+
+	match system:
+		System.GAME:
+			get_tree().paused = paused
+
+
+func pause_system(system: System, node: Node) -> void:
+	var nodes_pausing: Array = _pause_requests[system]
+
+	if nodes_pausing.has(node):
+		print(
+			(
+				"Node %s requested pause for system %s but already exists"
+				% [node.name, System.find_key(system)]
+			)
+		)
+
+		return
+
+	var was_paused: bool = is_paused(system)
+
+	nodes_pausing.push_back(node)
+	node.tree_exited.connect(self._remove_pauses_of_node.bind(node), CONNECT_ONE_SHOT)
+
+	if not was_paused:
+		pause_changed.emit(system, true)
+
+
+func unpause_system(system: System, node: Node) -> void:
+	var nodes_pausing: Array = _pause_requests[system]
+
+	if nodes_pausing.has(node):
+		node.tree_exited.disconnect(self._remove_pauses_of_node.bind(system, node))
+
+		nodes_pausing.erase(node)
+
+		if nodes_pausing.is_empty():
+			pause_changed.emit(system, false)
+
+
+func is_paused(system: System) -> bool:
+	return not _pause_requests[system].is_empty()
+
+
+func _remove_pauses_of_node(removed_node: Node) -> void:
+	for system: System in _pause_requests.keys():
+		var nodes_pausing: Array = _pause_requests[system]
+
+		if nodes_pausing.has(removed_node):
+			nodes_pausing.erase(removed_node)
+
+			if nodes_pausing.is_empty():
+				pause_changed.emit(system, false)
--- a/project.godot
+++ b/project.godot
@@ -21,6 +21,7 @@ DialogueManager="*res://addons/dialogue_manager/dialogue_manager.gd"
 GameState="*res://scenes/game_state/game_state.tscn"
 SceneSwitcher="*res://scenes/scene_switcher/scene_switcher.gd"
 Transitions="*res://scenes/transitions/transitions.tscn"
+Pause="*res://scenes/globals/pause.gd"
 
 [debug]
 
--- a/scenes/globals/pause.gd.uid
+++ b/scenes/globals/pause.gd.uid
@@ -1,0 +1,1 @@
+uid://b4wpfgte45uhr
--- a/scenes/player/player.gd
+++ b/scenes/player/player.gd
@@ -48,12 +48,17 @@ func _process(delta: float) -> void:
 		velocity = Vector2.ZERO
 		return
 
-	var axis: Vector2 = Input.get_vector(&"ui_left", &"ui_right", &"ui_up", &"ui_down")
+	var axis: Vector2 = (
+		Input.get_vector(&"ui_left", &"ui_right", &"ui_up", &"ui_down")
+		if not Pause.is_paused(Pause.System.PLAYER_INPUT)
+		else Vector2.ZERO
+	)
+
 	if not axis.is_zero_approx():
 		last_nonzero_axis = axis
 
 	var speed: float
-	if Input.is_action_pressed(&"running"):
+	if Input.is_action_pressed(&"running") and not Pause.is_paused(Pause.System.PLAYER_INPUT):
 		speed = run_speed
 	else:
 		speed = walk_speed
--- a/scenes/player/player_fighting.gd
+++ b/scenes/player/player_fighting.gd
@@ -18,7 +18,9 @@ func _ready() -> void:
 
 
 func _process(_delta: float) -> void:
-	if Input.is_action_just_pressed(&"ui_accept"):
+	if Pause.is_paused(Pause.System.PLAYER_INPUT):
+		is_fighting = false
+	elif Input.is_action_just_pressed(&"ui_accept"):
 		is_fighting = true
 	elif Input.is_action_just_released(&"ui_accept"):
 		is_fighting = false
--- a/scenes/player/player_interaction.gd
+++ b/scenes/player/player_interaction.gd
@@ -37,7 +37,10 @@ func _process(_delta: float) -> void:
 	if not interact_area:
 		interact_label.visible = false
 		return
-	if Input.is_action_just_released(&"ui_accept"):
+	if (
+		Input.is_action_just_released(&"ui_accept")
+		and not Pause.is_paused(Pause.System.PLAYER_INPUT)
+	):
 		interact_area.interaction_ended.connect(_on_interaction_ended)
 		interact_area.start_interaction(interact_ray.target_position.x < 0)
 		interact_ray.enabled = false
--- a/scenes/scene_switcher/scene_switcher.gd
+++ b/scenes/scene_switcher/scene_switcher.gd
@@ -60,14 +60,12 @@ func _set_hash(resource_path: String) -> void:
 		_window.location.replace(url.href)
 
 
-func change_to_file_with_transition(scene_path: String, spawn_point: NodePath = ^""):
-	# We should probably pause the game during transitions. But pause is
-	# not trivial, so I suggest first tackling it in another issue, then we can
-	# add it here.
-
+func change_to_file_with_transition(scene_path: String, spawn_point: NodePath = ^"") -> void:
+	Pause.pause_system(Pause.System.PLAYER_INPUT, self)
 	await Transitions.leave_scene(Transition.Effect.RIGHT_TO_LEFT_WIPE)
 	change_to_file(scene_path, spawn_point)
 	await Transitions.introduce_scene(Transition.Effect.LEFT_TO_RIGHT_WIPE)
+	Pause.unpause_system(Pause.System.PLAYER_INPUT, self)
 
 
 func change_to_file(scene_path: String, spawn_point: NodePath = ^"") -> void:


scenes/globals/pause.gd
```diff
@@ -0,0 +1,114 @@
...
+
+func _remove_pauses_of_node(removed_node: Node) -> void:
+	for req: PauseRequest in _pause_requests.duplicate():
+		if req.node == removed_node:
+			_pause_requests.erase(req)
+
+	_refresh_systems()
+
+
+class PauseRequest:
```
manuq, line 108: It took me some time to find what was `PauseRequest` until I found it at the bottom of the script. Is the linter forcing this? Otherwise can it be moved above all the `func` declarations?

scenes/globals/pause.gd
```diff
@@ -0,0 +1,114 @@
...
+	var system: System
+	var node: Node
+
+	func _init(a_system: System, a_node: Node) -> void:
+		system = a_system
+		node = a_node
```
manuq, line 114: This class is only for state. I guess it means "this system is paused by this node". I wonder if it can be replaced by a Dictionary: `_systems_paused_by_node: Dictionary[System, Node] = {}`

scenes/globals/pause.gd
```diff
@@ -0,0 +1,114 @@
...
+
+
+func unpause_system(system: System, node: Node) -> void:
+	_clear_request(system, node)
+
+	_refresh_systems()
+
+
+func is_paused(system: System) -> bool:
+	return system in paused_systems()
```
manuq, line 29: Why not use the `_paused_systems` Array directly?

scenes/globals/pause.gd
```diff
@@ -0,0 +1,114 @@
...
+
+
+func paused_systems() -> Array:
+	return _paused_systems
+
+
+func unpaused_systems() -> Array:
+	return System.values().filter(func(system: System) -> bool: return !is_paused(system))
```
manuq, line 37: This function is not used, let's remove it.

scenes/globals/pause.gd
```diff
@@ -0,0 +1,114 @@
...
+
+
+# Private methods
```
manuq, line 40: I don't think this comment is needed, as by convention all private methods start like `_this`. Note to myself to start differentating private and public methods in my own PRs.

scenes/globals/pause.gd
```diff
@@ -0,0 +1,114 @@
...
+func _find_request(a_system: System, a_node: Node) -> PauseRequest:
```
manuq, line 41: If `PauseRequest` is replaced by a Dictionary, you could just use `Dictionary.get()` or `Dictionary.has()` instead of this private method.

manuq (changes requested):
I have left some comments. My fear is that this looks like a generic solution for problems we may not have. But I would like to know @wjt and @JuanFdS thoughts too.

I understand that a global is needed, because the player instance in the next scene is not the same player instance as the current scene. That is the explanation I'm missing in the commit message.

Ideally this could be split in 2 or 3 commits. In particular the player controller can be a separate commit.

scenes/globals/pause.gd
pablitar: The linter forces me to have it below :(

scenes/globals/pause.gd
pablitar: Hmm. Could be! But it's not 1:1. Systems can be paused by multiple nodes at once. Maybe it could be replaced by a Dictionary of `Dictionary[System, Array[Node]]` or something similar.

scenes/globals/pause.gd
pablitar: Because the _paused_systems array is an implementation detail. The contents of the array can actually be derived from the _pause_requests Array, so in a way, it's a cache. Having the method allows to change the implementation in the future if that's necessary.

That said, being the method is_paused defined inside the class, I guess it wouldn't hurt much to reference _paused_systems directly. I'm OK either way. Just say the word and I'll change it.

scenes/globals/pause.gd
pablitar: I thought that since this class in particular has many different methods, it was necessary to clearly differentiate (and emphasize) public vs private. But can remove the comment if you want!

pablitar:
> I have left some comments. My fear is that this looks like a generic solution for problems we may not have. But I would like to know @wjt and @JuanFdS thoughts too.

It kinda is. But my experience tells me the problems will come hahaha. Even simple games can have complex pause logics. I know it seems like unnecessary complexity, but it can save us several bugs in the future of things pausing and not pausing when they should.

> 
> I understand that a global is needed, because the player instance in the next scene is not the same player instance as the current scene. That is the explanation I'm missing in the commit message.

I'll add it! Another reason the global is needed is because pausing the game in Godot is a global state, independent of the current scene (`get_tree().paused`). If we handle pause in a local node, we could end up having a paused game without a node responsible from unpausing it.

> Ideally this could be split in 2 or 3 commits. In particular the player controller can be a separate commit.
OK! I'll split the PR in three commits:
- Adding the Player Controller
- Adding the pause system
- Using the pause system in the transitions

scenes/globals/pause.gd
pablitar: Yes! I'll see if I can implement the Dictionary of Arrays

pablitar:
Feedback addressed @manuq !

wjt:
Another commit message suggestion:

> Added Pausing of Player Inputs during scene transitions 
>
> In order to resolve the issue brought forward by @wjt in PR https://github.com/endlessm/threadbare/pull/63

Can you describe the problem instead of linking to it? I suggest:

> Pause player input during scene transitions
>
> Previously, if the player continues to hold a direction during the transition, the character would be moved in the new scene. This could lead to the character ending up in an unexpected location on the other side of the transition – perhaps even off the edge of the scene.
>
> Instead, stop processing player input when the transition begins, resuming when the transition is complete.

(Tagging me in the commit message also means I get an email each time a version of the patch is pushed!)

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
+extends Node
+
+enum System {
+	PLAYER_INPUT,
+	GAME
+	# For now is just these two but in the future we could have: Physics, Enemies, AI, etc
```
manuq, line 6:
```suggestion
	GAME,
```

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+}
+
+# Private vars
```
manuq, line 9:
```suggestion
```

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+var _pause_requests: Dictionary[System, Array] = {}
+
+# This should be an array of systems but it throws
+# a runtime error when using the filter function, so for now just "Array"
+var _paused_systems: Array = []
+
+
+# Public interface
```
manuq, line 17:
```suggestion
```

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+
+func is_paused(system: System) -> bool:
+	return system in paused_systems()
+
+
+func paused_systems() -> Array:
+	return _paused_systems
+
+
+# Private methods
```
manuq, line 36:
```suggestion
```

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+		a_node.tree_exited.disconnect(self._remove_pauses_of_node.bind(system, a_node))
+
+		_refresh_systems()
+
+
+func _refresh_systems() -> void:
+	_refresh_paused_systems()
+
+	if not is_inside_tree():
+		# This can happen when the game is quitting
```
manuq, line 72:
```suggestion
		# This can happen when the game is quitting.
```

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+	return system in paused_systems()
+
+
+func paused_systems() -> Array:
+	return _paused_systems
+
+
+# Private methods
+func _exists_request_for_system(a_system: System) -> bool:
+	return !_pause_requests.get_or_add(a_system, []).is_empty()
```
manuq, line 38: The `!` negation is deprecated in favor of `not`. Also why not just check if the key is in the Dictionary? Which also makes the function so trivial that `a_system not in _pause_requests` can be used directly when this function is needed.
```suggestion
	return a_system not in _pause_requests
```

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+
+
+func _add_request(system: System, a_node: Node) -> void:
+	var nodes_pausing: Array = _pause_requests.get_or_add(system, [])
+
+	if !nodes_pausing.has(a_node):
```
manuq, line 44:
```suggestion
	if not nodes_pausing.has(a_node):
```

scenes/player/player_controller.gd
```diff
@@ -0,0 +1,23 @@
...
+
+func _process(_delta: float) -> void:
+	reset()
+
+	if Pause.is_paused(Pause.System.PLAYER_INPUT):
+		return
+
+	axis = Input.get_vector(&"ui_left", &"ui_right", &"ui_up", &"ui_down")
+	running = Input.is_action_pressed(&"running")
+	interacting = Input.is_action_just_released(&"ui_accept")
```
manuq, line 17: I think that is better to keep the interacting in the `player_interaction.gd` script. Because a fighting player may not need NPC talking interaction. So adding a player_controller goes a bit against the modularity we want for the player. Please leave it out of this PR.

manuq (changes requested):
A couple more suggestions from my side.

scenes/globals/pause.gd
manuq: Actually not, since the key is not removed when the value array is emptied.

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+var _pause_requests: Dictionary[System, Array] = {}
+
+# This should be an array of systems but it throws
+# a runtime error when using the filter function, so for now just "Array"
+var _paused_systems: Array = []
+
+
+# Public interface
+func pause_system(system: System, node: Node) -> void:
+	_add_request(system, node)
```
manuq, line 19: The only thing that this function is doing is call `_add_request()`. So I would move the content of that function here.

scenes/globals/pause.gd
```diff
@@ -0,0 +1,91 @@
...
+
+
+func unpause_system(system: System, node: Node) -> void:
+	_clear_request(system, node)
+
+	_refresh_systems()
```
manuq, line 25: This function is calling `_clear_request` which already calls `_refresh_systems()` at the end. So `_refresh_systems()` can be called twice. I would just move the content of `_clear_request` here and get rid of that function.

manuq (changes requested):
And 2 more suggestions!

scenes/globals/pause.gd
```diff
@@ -0,0 +1,88 @@
...
+enum System {
+	PLAYER_INPUT,
+	GAME,
+}
+
+var _pause_requests: Dictionary[System, Array] = {}
+
+# This should be an array of systems but it throws
+# a runtime error when using the filter function, so for now just "Array"
+var _paused_systems: Array = []
```
manuq, line 12: The `_paused_systems` is a variable derived from `_pause_requests`, it is an Array built from the keys of the Dictionary. This is prone to errors and has to be constantly sync'ed. I propose that we get rid of it and use the Dictionary keys directly.

scenes/globals/pause.gd
```diff
@@ -0,0 +1,88 @@
...
+	_refresh_paused_systems()
+
+	if not is_inside_tree():
+		# This can happen when the game is quitting.
+		return
+
+	if is_paused(System.GAME):
+		get_tree().paused = true
+	else:
+		get_tree().paused = false
```
manuq, line 75:
```suggestion
	get_tree().paused = is_paused(System.GAME)
```

By the way, it would be better to use signals and callbacks, to avoid constantly checking if a system needs to be paused or not:

```
signal pause_changed(system: System, paused: bool)

func _ready() -> void:
        pause_changed.connect(_on_pause_changed)

func _on_pause_changed(system: System, paused: bool) -> void:
        if not is_inside_tree():
                # This can happen when the game is quitting.
                return

        match system:
                System.GAME:
                        get_tree().paused = paused

```

manuq (changes requested):
@pablitar since this review is becoming long, I ended up giving a try myself. See the changes for `pause.gd` here: https://github.com/endlessm/threadbare/pull/99/commits/9789001a9942d7c2af5c96e3c91afc66a559d8d2

This commit has all the suggestions I gave.

scenes/globals/pause.gd
manuq: Then in that case I prefer that we define the variable with empty arrays, rather than using `get_or_add(system, [])` everywhere:

```
var _pause_requests: Dictionary[System, Array] = {
        System.PLAYER_INPUT: [],
        System.GAME: [],
}
```

manuq (approved):
@pablitar this looks good! Before merging, please add to the commit message the actual reason why we are doing this. Something like: "a global player controll pause system is needed because when switching scenes, the player instance in the next scene is not the same player instance as in current scene".

PR #95, opened 2025-04-02
ci: Report errors from `reuse lint`

wjt:
When I added a job that runs `reuse lint`, I piped its output to `tee $GITHUB_STEP_SUMMARY` so that the output shows up in a nicely-formatted way in the workflow run summary.

Unfortunately, by default the exit status of a shell pipeline is the exit status of the rightmost command. This means that even if `reuse lint` fails, the pipeline will succeed, erroneously claiming that the licensing information is correctly specified.

Use bash's pipefail option to cause the pipeline to fail in this case.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -31,5 +31,7 @@ jobs:
       with:
         cache-dependency-glob: ".github/workflows/checks.yaml"
     - name: Check REUSE compliance
+      shell: bash
       run: |
+        set -o pipefail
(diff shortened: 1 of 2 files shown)


wjt:
That's more like it:

image

https://github.com/endlessm/threadbare/actions/runs/14217703789?pr=95

manuq (approved):
Oh sorry for introducing a script without the license header. The Bash pipefail would do it for the check.

Issue #84, opened 2025-03-31
Combat mechanic: Make it a component


manuq:
In https://github.com/endlessm/threadbare/pull/57 the player scene was duplicated for the ink combat. But we want a single Player class with the combat feature that can be turned on/off depending on the minigame.

PR #97, opened 2025-04-02
Player Fighting Component

manuq:
Fixes https://github.com/endlessm/threadbare/issues/84


scenes/player/player.gd
```diff
@@ -3,12 +3,37 @@
 class_name Player
 extends CharacterBody2D
 
+enum State {
+	COZY,
+	FIGHTING,
+}
+
+@export var state: State = State.COZY:
+	set = _set_state
```
wjt, line 12: Perhaps you can add some documentation comments here:

```suggestion
enum State {
	## Player can explore the world, interact with items and NPCs, but is not fighting. Combat actions are not available in this mode.
	COZY,
	
	## Player is engaged in combat. Player can use combat actions.
	FIGHTING,
}

## Controls how the player can interact with the world around them.
@export var state: State = State.COZY:
	set = _set_state
```

scenes/player/animation_player.gd
```diff
@@ -2,11 +2,52 @@
...
+
+func _process_walk_idle(_delta: float) -> void:
 	if player.velocity.is_zero_approx():
 		play(&"idle")
 	else:
 		play(&"walk")
+
+
+func _process_fighting(_delta: float) -> void:
+	if not player_fighting.is_fighting and current_animation not in FIGHT_ANIMATIONS:
```
wjt, line 28: This is not new in this branch but I think the name of `is_fighting` is a bit misleading – I expected it to mean the same as `player.state == COMBAT` but actually it means "the player is actively using the attack action".

scenes/player/animation_player.gd
```diff
@@ -2,11 +2,52 @@
...
 @onready var player: Player = owner
+@onready var player_fighting: Node2D = %PlayerFighting
 
 
 func _process(_delta: float) -> void:
+	match player.state:
+		player.State.COZY:
+			_process_walk_idle(_delta)
+		player.State.FIGHTING:
+			_process_fighting(_delta)
```
wjt, line 17: It does feel a bit weird to have both of these here… You could move the fighting bit to `player_fighting.gd`, give it a reference to this player and have some methods to control the animations, but then where would the walk-handling go...

wjt: Actually I don't think that you need the branch here. In the case where the player is in cozy mode, `player_fighting.is_fighting` will always be false, so it will fall back to calling `_process_walk_idle`. Maybe you could simplify that way.

wjt (commented):
This works. I feel like there must be a more modular way to handle the different animations depending on which mode (maybe mode would be better than state for the cozy vs fighter enum & property?) but TBH I can't immediately see it.

scenes/player/player.gd
```diff
@@ -11,6 +18,22 @@ extends CharacterBody2D
...
+
+func _set_state(new_state: State) -> void:
+	state = new_state
+	if not is_node_ready():
+		return
+	match state:
+		State.COZY:
+			player_fighting.process_mode = ProcessMode.PROCESS_MODE_DISABLED
+		State.FIGHTING:
+			player_fighting.process_mode = ProcessMode.PROCESS_MODE_INHERIT
```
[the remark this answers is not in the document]
manuq: Yeah, I was unsure if the narrative was going to be part of this scene or not. In any case, we could add a collision area in the bridge that changes the mode from cozy to combat.

manuq:
> This works. I feel like there must be a more modular way to handle the different animations depending on which mode (maybe mode would be better than state for the cozy vs fighter enum & property?) but TBH I can't immediately see it.

:+1: let's change from "state" to "mode".

manuq:
@wjt I addressed your comments in the fixup commits.

scenes/player/player.gd
wjt: I pushed another fixup commit that adds these comments.

wjt (approved):
LGTM, please squash the fixups before mergin!

scenes/player/player.gd
manuq: Ouch I knew I was missing something. Thanks.

manuq:
> LGTM, please squash the fixups before mergin!

I have adjusted the commit message too, because now the PlayerInteraction is also being disable/enable its process mode.

PR #100, opened 2025-04-03
FIX #81 - Reset guards velocity on save

### LINKED ISSUE #81: Saving the stealth level while any of the guards is moving updates and stores its current velocity
## Steps to reproduce

- Open `res://scenes/stealth_level/stealth_level.tscn`
- Select node `StealthQuest/EnemyGuards/Guard-GoingBackAndForth`
- Set `move_while_in_editor` to true
- Wait until the guard is moving
- Save the scene
- Check `res://scenes/stealth_level/stealth_level.tscn`, it will have a modified line where the velocity of that node has been changed.

## How it should work

On save, the velocity of the guards should be reset to Vector2.ZERO.

JuanFdS:
Fix #81

### DIFF
--- a/scenes/enemies/guard/guard.gd
+++ b/scenes/enemies/guard/guard.gd
@@ -371,15 +371,22 @@ func _is_patrol_path_closed() -> bool:
 	return first_point_position.is_equal_approx(last_point_position)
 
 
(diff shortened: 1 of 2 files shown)


manuq (approved):
Excellent!

PR #102, opened 2025-04-03
Adding a Player Controller

pablitar:
Since we need to be able to pause the inputs of a player, a Player Controller has been added as an intermediary between the Input singleton and the player scripts. This controller checks if the input has been paused, and if so, ignores them.

Player Controllers are also a good practice in general, since they allow to easily have player characters that are controlled by entities that are not input, like code for cinematics, or another player through multiplayer.

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -107,7 +107,7 @@ running={
 
 [internationalization]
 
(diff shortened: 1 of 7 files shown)


manuq (approved):
The wrapper looks good!

manuq (approved):
@pablitar feel free to resolve the conflict and then merge.

PR #106, opened 2025-04-04
Use m6x11plus pixel font

wjt:
This adds the [m6x11](https://managore.itch.io/m6x11) pixel font to the project, and adjusts dialogue balloons, interact labels, and signpost labels to use it.

Doing so required adding a custom dialogue balloon scene (a copy of Godot Dialogue Manager's built-in sample balloon) in order to adjust its theme. This will be useful in subsequent changes to customise the balloon further.

image
image
image

https://endlessm.github.io/threadbare/branches/endlessm/use-m6x11plus-pixel-font/

Helps #54


wjt:
Draft because the game crashed when I tried to interact with the bard. I will check if this is a regression in this branch or an existing bug.

wjt:
It's an existing bug on main, fixed in #109

manuq (approved):
This font is great.

For the last commit instead of working around the fact that some labels are not part of the ScreenOverlay node (so they are scaled with the camera), we could make them part of that.

But I also wonder what people thinks about that ScreenOverlay nd if should use it everywhere or drop it. Having labels that don't scale with the camera is a feature I added early in the project "by intuition", it wasn't a real request. Currently the only minigame that changes the camera zoom is the Stealth game.

wjt:
I saw that #113 is tracking a related issue with scaling labels… I suggest we merge this, considering it a preexisting issue – it was already the case that the signpost had a different (implicit) font size to the interact label – to be fixed separately.

Issue #107, opened 2025-04-04
Musician dialogue broken


wjt:
Since https://github.com/endlessm/threadbare/pull/93 was merged, talking to the musician crashes as follows:

```
SCRIPT ERROR: Invalid call. Nonexistent function 'has_method' in base 'Nil'.
          at: _thing_has_method (res://addons/dialogue_manager/dialogue_manager.gd:1344)
SCRIPT ERROR: Assertion failed: "is_solved" is not a callable method on "<null>"
          at: show_error_for_missing_state_value (res://addons/dialogue_manager/dialogue_manager.gd:522)
SCRIPT ERROR: Invalid call. Nonexistent 'bool' constructor.
          at: _check_condition (res://addons/dialogue_manager/dialogue_manager.gd:646)
```

I bisected it to commit bb68c70b8c84293de0560cb882c7cc9fc6ec7a8f. I think the problematic hunk is:

```
 [node name="Bard" parent="OnTheGround" node_paths=PackedStringArray("puzzle") instance=ExtResource("4_jyhfs")]
 position = Vector2(1344, 536)
 script = ExtResource("5_choun")
-puzzle = NodePath("../MusicPuzzle")
+puzzle = NodePath("InteractArea")
 dialogue = ExtResource("5_85g3d")
 look_at_side = 1
 sprite_frames = ExtResource("5_tbgi4")
```

PR #109, opened 2025-04-04
Music challenge: Fix assigning puzzle node to bard character

wjt:
In commit bb68c70b8c84293de0560cb882c7cc9fc6ec7a8f, the puzzle property of the bard was incorrectly changed to be an InteractArea rather than the MusicPuzzle that is needed. This caused the following warnings when talking to the bard:

    SCRIPT ERROR: Invalid call. Nonexistent function 'has_method' in base 'Nil'.
              at: _thing_has_method (res://addons/dialogue_manager/dialogue_manager.gd:1344)
    SCRIPT ERROR: Assertion failed: "is_solved" is not a callable method on "<null>"
              at: show_error_for_missing_state_value (res://addons/dialogue_manager/dialogue_manager.gd:522)
    SCRIPT ERROR: Invalid call. Nonexistent 'bool' constructor.
              at: _check_condition (res://addons/dialogue_manager/dialogue_manager.gd:646)

and prevented the bard playing the xylophone to demo the melody or reacting to the player having completed the puzzle.

Restore the correct value.

Fixes https://github.com/endlessm/threadbare/issues/107


JuanFdS (approved):
sorry! 👍🏾

PR #114, opened 2025-04-04
Make interact label position work even if camera's zoomed

JuanFdS:
ScreenOverlay had follow_viewport enabled set to true, so it is already passing a transform to its children. Changed the player interaction code so it doesn't try to compensate for the canvas layer, since ScreenOverlay is already handling that with the follow_viewport setting.

Also, replaced every instance of the ScreenOverlay canvas layer by a scene, so if we need to change any setting is done just once.

## Before (85c3721cc4b3525d95ce30dd3f625666ba63a115)

https://github.com/user-attachments/assets/d81f0441-c0c4-4d04-a33f-da75b11daa9f

## Now (1d9eab9e5bf1bd5c768d794dc5f5345223588896)

https://github.com/user-attachments/assets/97a8dbef-446b-4d14-9968-ea1f9386a719

### DIFF
--- a/scenes/ink_combat/ink_combat.tscn
+++ b/scenes/ink_combat/ink_combat.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=8 format=4 uid="uid://c764br4tplkb2"]
+[gd_scene load_steps=9 format=4 uid="uid://c764br4tplkb2"]
 
 [ext_resource type="Script" uid="uid://c38tkgfys3d63" path="res://scenes/ink_combat/ink_combat.gd" id="1_v7ikp"]
 [ext_resource type="TileSet" uid="uid://07fq3rspk8ia" path="res://scenes/tileset.tres" id="1_wrx86"]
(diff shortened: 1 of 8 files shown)


JuanFdS:
I just noticed this PR is breaking a property that I think was intended for the interaction label: now it's size depends on the camera's zoom. Its size relative to the player character will always remain the same, which I think in some levels looks good, but, if we were to have a level where the player is too far from the camera, then the interact label would also be small.

On the other hand, having the interact label (and these other kinds of labels) size be camera-independent has one issue: it's hard to grasp how the control will look in the actual game when configuring it in the editor.

Anyways, I think the interact label as it works in this PR would work fine on most levels, but it's true that making its size independent from the camera would be better to support levels with different zooms better. Also, it's the same interact label in all levels, so it isn't something that would be easy to configure per level.
On the other hand, I think that elements like the inkwells labels should be configured per label and have them live in the same canvas layer as the item they are "tagging", because that makes it easier to predict how it'd look on the running game, and we avoid these kind of problems alltogether with them.

I'm closing this to implement an alternative solution that:
- keeps the interact label the same size, independent of the camera.
- moves the other labels somewhere else so they look in the game the same they look in the editor.

PR #115, opened 2025-04-04
dialogue: Use Tiny Swords banner asset for main balloon

wjt:
The main balloon is surrounded by a Panel, which the theme inherited from Godot Dialogue Manager's default balloon scene previously specified to be a StyleBoxFlat with a rounded border.

Replace that style with a 9-patch banner texture from Tiny Swords, with the stretch mode set to tile (it's pixel art!) and its margins fudged a little so that the whole box fits on the screen. There is a bit of a weird interaction with the MarginContainer within - perhaps the margins on that should be tweaked instead – but this looks a lot better already.

The buttons for dialogue choices are still in the default GDM style as of this commit.

Helps https://github.com/endlessm/threadbare/issues/54

----

Screenshot from 2025-04-04 18-45-32
Screenshot from 2025-04-04 18-45-47

I ran out of time today to style the buttons for dialogue choices but IMO this is already a nice improvement.

https://endlessm.github.io/threadbare/branches/endlessm/dialogue-main-balloon-style/

### DIFF
--- a/scenes/ui/dialogue/theme.tres
+++ b/scenes/ui/dialogue/theme.tres
@@ -1,6 +1,7 @@
-[gd_resource type="Theme" load_steps=6 format=3 uid="uid://cvitou84ni7qe"]
+[gd_resource type="Theme" load_steps=7 format=3 uid="uid://cvitou84ni7qe"]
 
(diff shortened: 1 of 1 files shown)


wjt:
Of course this theme ends up being very similar to the override used in the HUD. Maybe we ultimately want one common default theme for controls?

Issue #111, opened 2025-04-04
Improvements on the Combat Level


LauraFoglia:
- Tryout swinging the  the sword in all directions.
- Make the enemies ink drinkers indestructible or respawn.

manuq:
The remaining of this is to have multiple rounds. Since it's tracked in a separate ticket, I will close this one as done.

PR #116, opened 2025-04-04
Simpler fighting

manuq:
Playable at https://endlessm.github.io/threadbare/branches/endlessm/simpler-fighting/#ink_combat/ink_combat

Helps https://github.com/endlessm/threadbare/issues/111

The player now produces a circular hit in all directions. So is now only timing is needed. Previously it was timing plus hitting in a particular direction. A playtest from @jbourqueendless shows that this makes it easier.

Update: I also added some sound effects and made the enemies invulnerable.

[Grabación de pantalla desde 2025-04-04 16-18-47.webm](https://github.com/user-attachments/assets/0655150c-3e0d-47a6-a9af-281644f6e48b)


scenes/ink_combat/ink_combat.gd
```diff
@@ -19,7 +19,16 @@ func _ready() -> void:
 		node.completed.connect(_on_inkwell_completed)
 
 
+func _remove_enemies():
+	for node: Node in on_the_ground.get_children():
+		if node is not InkDrinker:
```
pablitar, line 24: Maybe we can use groups here? Groups allow for querying independently of the hierarchy

scenes/player/animation_player.gd
```diff
@@ -2,15 +2,15 @@
 # SPDX-License-Identifier: MPL-2.0
 extends AnimationPlayer
 
-const FIGHT_ANIMATIONS: Array[StringName] = [&"blow_right", &"blow_left", &"blow_up", &"blow_down"]
+const BLOW_ANTICIPATION_TIME: float = 0.3
```
pablitar, line 5: Since we are using an animation player, maybe this can go in the animation track, as a track that sets the property cancellable or one that calls a method. It's usually easier to maintain since this value could change depending on the animation.

scenes/player/animation_player.gd
```diff
@@ -25,24 +25,18 @@ func _process_walk_idle(_delta: float) -> void:
...
-	var angle: float = player.last_nonzero_axis.angle()
-	if abs(-PI / 2 - angle) <= PI / 4:
-		play(&"blow_up")
-	elif abs(PI / 2 - angle) <= PI / 4:
-		play(&"blow_down")
-	elif abs(PI - angle) <= PI / 4:
-		play(&"blow_left")
-	elif abs(angle) <= PI / 4:
-		play(&"blow_right")
+	if current_animation != &"blow":
```
pablitar, line 40: All this animation logic seems a bit confusing. Maybe some comments can be helpful to understand what's going on?

pablitar (approved):
LGTM!

Left a couple of comments if you wish to address them.

JuanFdS (approved):
Looks great!

scenes/ink_combat/ink_combat.gd
manuq: Oh yes, in fact we have one already!

scenes/player/animation_player.gd
manuq: I just realized that cancellable is not needed anymore, since I removed the first anticipation. I will add remove it and add a note in the commit message.

manuq:
Thanks for the review, I have addressed the comments.

Issue #73, opened 2025-03-28
Initial combat system: multiple rounds


LauraFoglia:
Add 3 challenges to showcase the different mechanics.

manuq:
As discussed in a call, this can be 3 rounds with increasing difficulty.

PR #118, opened 2025-04-07
Simpler fighting walking

manuq:


wjt (commented):
In general I think this is a nice addition to combat – it makes it a bit more interesting/varied/challenging.

Perhaps this is intentional – the radius around the walking enemy includes the centre point of the ring of ink wells. This is annoying (perhaps in a good way) because the player can no longer hang around there and just mash the attack button.

I noticed that when the player repels an ink blob and it hits an enemy, the enemy flashes but doesn't take damage – intentional but feels a little strange.

I accidentally closed the tab and lost a longer comment about this but I think the enemy motion feels weird when the next random target position is either in the sea or very close to the current position. Either way the enemy ends up jogging on the spot for most/all of the walk.

I think one partial solution to the "very close" problem might be to pick the target angle not at random, but excluding (say) 45° either side of the enemy's current angle to the origin, and replacing the `randf()` that scales the range with `randf_range(.25, 1.0)` or similar.

Conceptually the "walking into the sea" issue could be addressed by clipping the line between the enemy and the target to the last point that doesn't collide with the sea.

(I can't quite figure out how it would be done but it would be nice if the target position were visible when Debug → Visible Navigation is enabled in the editor.)

When the battle ends, if there are still some ink blobs that follow the player in the scene, they continue to follow the player, and the player can no longer use the attack action to defend themself so they have to just run!

wjt:
Has conflicts with `main`

manuq:
> In general I think this is a nice addition to combat – it makes it a bit more interesting/varied/challenging.
> 
> Perhaps this is intentional – the radius around the walking enemy includes the centre point of the ring of ink wells. This is annoying (perhaps in a good way) because the player can no longer hang around there and just mash the attack button.

Yes. Also it is a bit chaotic now because there is a single "round" with all the variations (ink follow player, enemy walking, etc). The plan is to have multiple rounds with low difficulty and soft increase of difficulty, where each variation is demonstrated. https://github.com/endlessm/threadbare/issues/73

> I noticed that when the player repels an ink blob and it hits an enemy, the enemy flashes but doesn't take damage – intentional but feels a little strange.

Yes we made the enemies invulnerable. Do you think that they should get hit and respawn when beated? That was the alternative we handled in https://github.com/endlessm/threadbare/issues/111 

> I accidentally closed the tab and lost a longer comment about this
Ouch! GitHub bring my comment back!

> but I think the enemy motion feels weird when the next random target position is either in the sea or very close to the current position. Either way the enemy ends up jogging on the spot for most/all of the walk.

For jogging when colliding (the sea for example) I think I can borrow the collision check from the guard in the stealth game. If you don't mind, I will improve it after this is merged.

> I think one partial solution to the "very close" problem might be to pick the target angle not at random, but excluding (say) 45° either side of the enemy's current angle to the origin, and replacing the `randf()` that scales the range with `randf_range(.25, 1.0)` or similar.

This is a brillant idea. Effectively picking the target position in a semi ring. I just did it as an appended commit. Please check my math. It definitely works better now.

> Conceptually the "walking into the sea" issue could be addressed by clipping the line between the enemy and the target to the last point that doesn't collide with the sea.
> 
> (I can't quite figure out how it would be done but it would be nice if the target position were visible when Debug → Visible Navigation is enabled in the editor.)

I know how that can be done, borrowing from Moddable Pong:
```
	if Engine.is_editor_hint() or get_tree().is_debugging_collisions_hint():
		print(position - _initial_position)
		draw_circle(position - _initial_position, walking_range, Color(0.0, 1.0, 1.0, 0.3))
```
I tried, but the circle moves with the enemy. Probably my math is wrong...

> When the battle ends, if there are still some ink blobs that follow the player in the scene, they continue to follow the player, and the player can no longer use the attack action to defend themself so they have to just run!

Yeah, saw that as well. I think that we should send them a signal so they splash.

wjt:
> Effectively picking the target position in a semi ring. I just did it as an appended commit. Please check my math. It definitely works better now.

I'm not quite sure about the angles, I'll have to spend some time with a pen and paper to understand it. If you're confident with the maths, go forth and merge, and I can always propose another patch if I disagree.

I agree with leaving all those other changes to another PR.

PR #122, opened 2025-04-07
FIX 113 add camera that repositions itself and resizes itself based on camera

JuanFdS:
The logic that handles the size and position of labels was on the nodes that included `interact_area.tscn`. Replaced it by a different scene that handles its own position and also adds the label itself to a `ScreenOverlay` (so it works similar to the `interact_area.tscn` that existed before).

It was also renamed from `interact_area` to `fixed_size_label` because it isn't always used for interactions. However, names suggestions appreciated :D


manuq (approved):
Excellent!

Issue #113, opened 2025-04-04
Interact label is not being correctly placed  when the camera is zoomed


JuanFdS:
Steps to reproduce:
- Open `res://scenes/ink_combat/ink_combat.tscn` (it has the camera zoomed to x2)
- Set the player mode to Cozy, that way it can interact with game elements.

https://github.com/user-attachments/assets/61c8efe1-efa1-410f-be77-aaa18aa33732

Why it happens?

ScreenOverlay has `follow_viewport` enabled set to true. We want that so labels placed on it (like the colors of the inkwells) have spatial position and aren't just fixed on top of everything. However, by having that setting enabled, the ScreenOverlay is already passing a transform to its children, and I think this code was expecting that not to be the case, since it's also taking into account the canvas layer transform:
https://github.com/endlessm/threadbare/blob/85c3721cc4b3525d95ce30dd3f625666ba63a115/scenes/player/player_interaction.gd#L32-L35

manuq:
@JuanFdS this fixes the offset issue but also scales the label with the camera zoom, which is the whole point of having the ScreenOverlay layer for the labels. This is `scenes/main.tscn` Before / after:

https://github.com/user-attachments/assets/f84636db-74c5-455a-b65c-c923e7a3247b

https://github.com/user-attachments/assets/c82daec9-abc0-4733-9d86-3e6c9eabf1ee

JuanFdS:
> [@JuanFdS](https://github.com/JuanFdS) this fixes the offset issue but also scales the label with the camera zoom, which is the whole point of having the ScreenOverlay layer for the labels. This is `scenes/main.tscn` Before / after:
> Grabacion.de.pantalla.desde.2025-04-04.12-17-16.webm
> Grabacion.de.pantalla.desde.2025-04-04.12-18-16.webm

Yeah, I was just realising that 😅 . There are other ways to handle it:
- a) Set follow viewport to false in the combat level + make the inkwell labels reposition themselves automatically as well.
- b) Set the follow viewport to false in the combat level + take the inkwell labels out of the CanvasLayer.

The trade off I see here is:
a) makes sure that the labels will always be readable regardless what the person who's making the level does, b) lets the level-maker decide exactly how that UI should look in the running game (no resizing will happen between what they see in the editor and what they see in the game).

I'm fine with either of them, what do you think? @manuq

manuq:
[...]

PR #127, opened 2025-04-07
Fixed size label: Fix visibility

manuq:
Fix regression when changing the interact label by the fixed size label in 3aa25d56.

Setting visibility to the new label doesn't hide the label container because it may have been reparented to the ScreenOverlay node.

So, add a generic object setter to the FixedSizeLabel and check if the property being set is the visibility. If so, set the label container visibility to the same value.

Fix https://github.com/endlessm/threadbare/issues/113

### DIFF
--- a/scenes/props/fixed_size_label/fixed_size_label.gd
+++ b/scenes/props/fixed_size_label/fixed_size_label.gd
@@ -17,6 +17,15 @@ extends Control
 @onready var label_container: PanelContainer = %LabelContainer
 
 
+func _set(property: StringName, value: Variant) -> bool:
+	if not is_node_ready():
+		return false
+	if property == "visible":
+		label_container.visible = value
+		return true
+	return false
+
+
 func _set_label_text(new_text: String) -> void:
 	label_text = new_text
 	if not is_node_ready():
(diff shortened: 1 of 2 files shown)


pablitar (commented):
Yes please!

scenes/props/fixed_size_label/fixed_size_label.gd
```diff
@@ -17,6 +17,15 @@ extends Control
...
 
+func _set(property: StringName, value: Variant) -> bool:
+	if not is_node_ready():
+		return false
+	if property == "visible":
+		label_container.visible = value
+		return true
+	return false
+
+
```
JuanFdS, line 28: What about making the visibility of the label depend on the visibility of this node?, like, in _process:

```
label_container.visible = visible
```

I think it's simpler and it should work the same

manuq: Yes, we have assignments like that in the `_process` functions everywhere. But I'm not sure, maybe this solution is better in this case? @wjt @pablitar thoughts?

manuq: @JuanFdS I just realized that setting the label visibility in `_process` won't work for player fighting mode, because it disables processing of the PlayerInteraction node (which includes a fixed label instance). I just added a separate commit to hide the label when fighting.

PR #132, opened 2025-04-08
Change player's behavior node visibility when it changes mode

JuanFdS:
Also, make Fixed Size Label hide or show the label that it places on the ScreenOverlay based on whether it is visible in the tree or not.

https://github.com/user-attachments/assets/d3aec36e-88fe-43dc-b553-1372b40071b0

### DIFF
--- a/scenes/player/player.gd
+++ b/scenes/player/player.gd
@@ -32,11 +32,18 @@ func _set_mode(new_mode: Mode) -> void:
 		return
 	match mode:
 		Mode.COZY:
-			player_interaction.process_mode = ProcessMode.PROCESS_MODE_INHERIT
-			player_fighting.process_mode = ProcessMode.PROCESS_MODE_DISABLED
+			_toggle_player_behavior(player_interaction, true)
+			_toggle_player_behavior(player_fighting, false)
 		Mode.FIGHTING:
-			player_interaction.process_mode = ProcessMode.PROCESS_MODE_DISABLED
-			player_fighting.process_mode = ProcessMode.PROCESS_MODE_INHERIT
+			_toggle_player_behavior(player_interaction, false)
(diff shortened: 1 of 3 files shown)


scenes/props/fixed_size_label/fixed_size_label.gd
```diff
@@ -49,13 +40,18 @@ func _ready() -> void:
 	if Engine.is_editor_hint():
 		return
 
+	visibility_changed.connect(self.on_visibility_changed)
```
manuq, line 43: Yes! Much better.

scenes/player/player.gd
```diff
@@ -32,11 +32,18 @@ func _set_mode(new_mode: Mode) -> void:
...
+			_toggle_player_behavior(player_fighting, false)
 		Mode.FIGHTING:
-			player_interaction.process_mode = ProcessMode.PROCESS_MODE_DISABLED
-			player_fighting.process_mode = ProcessMode.PROCESS_MODE_INHERIT
+			_toggle_player_behavior(player_interaction, false)
+			_toggle_player_behavior(player_fighting, true)
+
+
+func _toggle_player_behavior(behavior_node: Node2D, is_active: bool):
+	behavior_node.visible = is_active
```
manuq, line 43: OK, so we are tying the process mode and the visibility together.

manuq (approved):

scenes/player/player.gd
JuanFdS: > OK, so we are tying the process mode and the visibility together.

Yes, at least for these 2 behaviors I thought it made sense. We can see how it evolves with more of these components.
A more future proof way would be for each of those nodes to handle its own _toggle on_ and _toggle off_, but since we only have 2 cases I wasn't sure about changing that yet.

PR #133, opened 2025-04-08
Fix 130 interact label doesnt use pixel art font

JuanFdS:
Set FixedSizeLabel font to m6x11plus pixel font.

Also, removed different places where the size of the label's font was being modified, and left only one: the theme.

Finally, added a custom type to the newly created theme that is specific for FixedSizeLabel

Branched from https://github.com/endlessm/threadbare/pull/132 , so please review that one first.

image
image


manuq (approved):
Excellent!

PR #134, opened 2025-04-08
Stitch Ink Combat level back

JuanFdS:
- Add a collectible in ink combat that is revealed when the challenge is completed, that collectible takes the player to the stealth challenge.

- Add HUD to the ink combat.

- Change puzzle music challenge's collectible's scene path to take the player to the ink combat.

- Remove the ink combat's reward files, since the collectible is now being used as the reward and teleporter.

----------------

Playing the 3 challenges in order:

https://github.com/user-attachments/assets/978b740e-3873-4e31-be94-07769cd324eb


manuq (approved):
All works as expected. The commit message explains the changes very well.

PR #138, opened 2025-04-08
Ink combat narrative

manuq:
Helps https://github.com/endlessm/threadbare/issues/6

### DIFF
--- a/scenes/ink_combat/ink_combat.gd
+++ b/scenes/ink_combat/ink_combat.gd
@@ -3,6 +3,7 @@
 extends Node2D
 
 @export var inkwells_to_win: int = 1
+@export var intro_dialogue: DialogueResource
 
 var inkwells_completed: int = 0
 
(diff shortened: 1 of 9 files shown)


manuq:
Ready for review. Although I think that the text has been changed in the document since I copied it. But I had a thumbs up over Discord that this is the desired behavior.

scenes/ink_combat/ink_combat.gd
```diff
@@ -12,6 +13,12 @@ var inkwells_completed: int = 0
 
 
 func _ready() -> void:
+	DialogueManager.show_dialogue_balloon(intro_dialogue, "", [self])
+	await DialogueManager.dialogue_ended
+	# Add a short delay so the player doesn"t attack when closing the dialogue:
+	await get_tree().create_timer(0.5).timeout
```
wjt, line 19: Not necessarily one for this branch but it would be really nice to fix this more systematically rather than needing a timeout after everywhere that we have dialogue. I think moving player input handling to `_unhandled_input()` would do it.

wjt (approved):
Nice approach – even if we later change the dialogue this is a good start.

I don't think it blocks a merge, but the initial dialogue balloon covers up the enemies (who are speaking):

image

Maybe we can fix this later by attaching dialogue to the speaker (if they are on-screen) or rearranging the scene.

manuq:
> Maybe we can fix this later by attaching dialogue to the speaker (if they are on-screen) or rearranging the scene.
Very good point. Attaching the dialogue to the speaker would work. If we want to keep the current behavior, we can add a way to display the dialogue frame in 3 places: bottom of screen, center of screen, top of screen. In Dialogue Manager each line can have a list of tags. Maybe we can use a tag for this.

PR #139, opened 2025-04-08
Adding support for teleporters that move the player to the same scene and a visible teleporter scene

### LINKED ISSUE #68: Teleporters shouldn't change scene if it's the same scene as the current one

pablitar:
This PR allows teleporters to function without a scene_to_go_to, or having a scene_to_go_to that is the same scene the teleporter is defined. In that case, instead of changing the scene, the teleporter just moves the player to the target spawn point.

It also adds a visible_teleporter scene.

Up until now the teleporters have been put at the edge of the screens so players didn't need a visual cue to identify them.

But with this PR, it becomes possible to have visible teleporters, which function the same as regular teleporters, but have an appearance. They will likely be necessary if learners want to use the teleporters in their scenes.

Fixes #68


scenes/teleporter/teleporter.gd
```diff
@@ -29,21 +29,46 @@ func _ready() -> void:
...
 
 
 func _update_available_spawn_points() -> void:
-	if ResourceLoader.exists(scene_to_go_to, "PackedScene"):
+	if not Engine.is_editor_hint() or not is_inside_tree():
+		return
+
+	var scene_to_go_to_path: String = _scene_to_go_to_path()
+
+	if scene_to_go_to == "" or scene_to_go_to_path == get_tree().edited_scene_root.scene_file_path:
```
manuq, line 65: Empty string is falsey, so you can do:
```suggestion
	if not scene_to_go_to or scene_to_go_to_path == get_tree().edited_scene_root.scene_file_path:
```

scenes/teleporter/teleporter.gd
```diff
@@ -29,21 +29,46 @@ func _ready() -> void:
...
 		# undesired behavior.
 		SceneSwitcher.change_to_file_with_transition.call_deferred(scene_to_go_to, spawn_point_path)
+	else:
+		var spawn_point: SpawnPoint = get_node_or_null(spawn_point_path)
+		if is_instance_valid(spawn_point):
+			spawn_point.move_player_to_self_position(true)
+			self.body_entered.connect(_on_body_entered, CONNECT_ONE_SHOT)
+
+
+func _scene_to_go_to_path():
```
manuq, line 52: I think that is better to prefix this function with "get", as it's obtaining the path. And the return hint would be even better:
```suggestion
func _get_scene_to_go_to_path() -> String:
```

scenes/teleporter/visible_teleporter.gd
```diff
@@ -0,0 +1,7 @@
+@tool
+extends Teleporter
+
+
+func _ready() -> void:
+	super._ready()
+	$BlueFire.play("burning")
```
manuq, line 7: Why not add an `@onready` for this `$BlueFire` node and also add a unique name to it? We have that everywhere else.

```
@onready var blue_fire: AnimatedSprite2D = %BlueFire 
```

manuq (changes requested):
Looking great! Left suggestions. The `assets/particles/circle.png` file needs a matching `assets/particles/circle.png.license` text file with the following content:

```
SPDX-FileCopyrightText: The Threadbare Authors
SPDX-License-Identifier: CC-BY-SA-4.0
```

pablitar:
Changes addressed @manuq !

PR #140, opened 2025-04-08
Added transition properties to teleporters and setup transitions in Frays End and Song Sanctuaries

### LINKED ISSUE #67: Add possibility for selecting different transitions in teleporters
Add a property in the teleporters to change which transition is used when changing scenes. We should keep in mind that some transitions are not symmetrical so we may want different transitions for leave and enter.

### LINKED ISSUE #69: Fade Transition Should Follow the Character
The fade must follow the character when switching screens, whether moving from left to right or right to left.

pablitar:
Two issues have been grouped into this PR because they are closely related.

Three properties have been added to the teleporters. A boolean specifying if the teleporter should use transitions, and a couple of properties to specify an enter_transition and an exit_transition. These properties only appear when the "use_transition" boolean is true.

This PR also sets up transitions of Frays End and Song Sanctuaries teleporters so they match the players movement direction.

Fixes #69 
Fixes #67

### DIFF
--- a/scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
+++ b/scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
@@ -12,12 +12,20 @@ const SPAWN_POINT_GROUP_NAME: String = "spawn_point"
 		_update_available_spawn_points()
 		notify_property_list_changed()
 
+@export var use_transition: bool = true:
+	set(new_val):
+		use_transition = new_val
+		notify_property_list_changed()
+
 var spawn_point_path: NodePath:
 	set(new_val):
 		if new_val == ^"NONE":
 			spawn_point_path = ^""
 		else:
 			spawn_point_path = new_val
+
+var enter_transition: Transition.Effect = Transition.Effect.LEFT_TO_RIGHT_WIPE
+var exit_transition: Transition.Effect = Transition.Effect.RIGHT_TO_LEFT_WIPE
 
 var _available_spawn_points: Array[NodePath] = []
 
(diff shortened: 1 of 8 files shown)


scenes/world_map/teleporter.gd
```diff
@@ -12,13 +12,21 @@ const SPAWN_POINT_GROUP_NAME: String = "spawn_point"
 		_update_available_spawn_points()
 		notify_property_list_changed()
 
+@export var use_transition: bool = true:
+	set(new_val):
+		use_transition = new_val
+		notify_property_list_changed()
```
wjt, line 18: I could be wrong, so let's leave it in for now, but I have a feeling that in practice this property will always be `true` when the destination is a different scene, and `false` when it's not.

scenes/transitions/transitions.tscn
```diff
@@ -2,12 +2,12 @@
 
 [ext_resource type="Script" uid="uid://byb3vg4ylor0d" path="res://scenes/transitions/transitions.gd" id="1_ij4f6"]
 [ext_resource type="Shader" uid="uid://32bpeadb2eyq" path="res://scenes/transitions/transition_layer.gdshader" id="2_rdj0f"]
-[ext_resource type="Texture2D" uid="uid://cpvc4xmg7at7r" path="res://scenes/transitions/Fade.png" id="3_ij4f6"]
+[ext_resource type="Texture2D" uid="uid://dcwmaoqgu5t84" path="res://scenes/transitions/Radial.png" id="3_ij4f6"]
```
wjt, line 5: I don't quite understand why the image is being changed here.

scenes/world_map/teleporter.gd
pablitar: I was thinking there may be some cases in the same scene where you'd want to use a transition.

scenes/transitions/transitions.tscn
wjt: OK, I don't mind either way.

wjt:
Conflicts now need to be resolved.

pablitar:
> Conflicts now need to be resolved.

Conflucts solved @wjt !

scenes/globals/utils.gd
```diff
@@ -0,0 +1,23 @@
...
+
+static func enum_property(name: String, clazz: StringName, an_enum: Dictionary) -> Dictionary:
+	return {
+		"name": name,
+		"type": TYPE_INT,
+		"class_name": clazz,
+		"hint": PROPERTY_HINT_ENUM,
+		"hint_string": get_enum_hint_string(an_enum),
+		"usage": PROPERTY_USAGE_DEFAULT | PROPERTY_USAGE_CLASS_IS_ENUM
+	}
```
manuq, line 23: References to "enum" here are confusing because they are in fact dictionaries. Can this new global be renamed to "property something" rather than a generic "utils"?

pablitar:
Merging this after renaming the Utils class!

Issue #136, opened 2025-04-08
Integrate NPCs/Other characters


LauraFoglia:
Integrate different kind of NPC/Enemies:

- Elder
[Elder (static).png.zip](https://github.com/user-attachments/files/19650534/Elder.static.png.zip)

manuq:
The sprite provided is a single PNG facing to the "camera". I'm asking if we'll also get an idle animation from PixelFrog that faces left / right (like current placeholders). If not I guess we can remove the animation in favor of this single still frame.

LauraFoglia:
Integrate:
- Ink drinkers.
- elder's idle.


[All files.zip](https://github.com/user-attachments/files/19762521/All.files.zip)

PR #147, opened 2025-04-10
Integrate elder

manuq:
Fix https://github.com/endlessm/threadbare/issues/136


wjt (approved):
LGTM, thanks for the fixes here!

PR #152, opened 2025-04-10
Add fade transitions between intro, world map, and quest

wjt:
When testing the game, the abrupt scene-switches at the end of the world map and when entering a quest feel bad, particularly compared to other scene transitions.

Add fade transitions.

The splash-to-intro transition is a bit trickier for reasons that I will explain in a subsequent PR, so is not covered here.


JuanFdS (approved):
💯

Issue #159, opened 2025-04-11
Interaction remains enabled during inter-scene transitions


wjt:
I noticed after #152 was merged that, after talking to the NPC and accepting their quest, it is possible to interact with them again while the fade-out is taking place. Here's a video, with the transition code changed to make the fade-out take 10 seconds so that it's really clear what's happening:

https://github.com/user-attachments/assets/f7bb05c2-d810-4ddd-85c8-21e41bb4c4af

The interaction prompt should not be shown while player input is paused, and it should also not be possible to start the interaction in that case. (Player movement is, correctly, ignored during the transition.)

PR #174, opened 2025-04-11
PlayerInteraction: Disable when inputs are paused

wjt:
Previously, it was possible to trigger an interaction while changing scene. For example, when you interact with an NPC and choose to start a quest, the interaction ending triggers a scene transition. However, during this 1-second transition, the player character is (by definition) in the correct position to trigger an interaction with the NPC, so the “Talk” prompt was re-displayed and pressing ui_accept quickly enough would cause the dialogue to begin again.

The first part needs special handling to not display the interact label when inputs are paused: use PlayerController.inputs_paused() to detect this case and hide the label if it would otherwise be displayed.

The second part is because the input handling used PlayerController.is_action_just_released() rather than .is_action_just_pressed(). Unlike the other is_action_* methods in PlayerController, PlayerController.is_action_just_released() ignores whether input is paused. Switching to
PlayerController.is_action_just_pressed() matches other input actions and also addresses this problem. Strictly speaking this is redundant with checking inputs_paused() above, but you might imagine refactoring the input to a different function in future.

Fixes https://github.com/endlessm/threadbare/issues/159

### DIFF
--- a/scenes/game_elements/characters/player/components/player_interaction.gd
+++ b/scenes/game_elements/characters/player/components/player_interaction.gd
@@ -20,11 +20,11 @@ func _process(_delta: float) -> void:
 		return
 	var interact_area: InteractArea = interact_ray.get_interact_area()
 
-	if not interact_area:
+	if %PlayerController.inputs_paused() or not interact_area:
 		interact_label.visible = false
 		return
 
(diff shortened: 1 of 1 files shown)


JuanFdS (approved):
Awesome 👍🏾

wjt:
Rebased and resolved conflict with #178

Issue #112, opened 2025-04-04
Music Puzzle: improvements


LauraFoglia:
Open to discussion

JuanFdS:
Adding some ideas:

- Interact with the fire to play back the melody that unlocked it.
- After not being able to figure out the next melody (for some time? / after some attempts?), the Bard gives you a hint (this could be either by directly triggering a dialogue, or somehow dragging the atention to the bard NPC and the next time it talks, it gives you the hint).

wjt:
Some ideas from #82:

- Maybe have another NPC basically just explain the symbols?
- Change the symbols? `⏵|⏴` could be just `⏴` by contrast with the first one `⏵`
- Show progress within each step of the puzzle, and feedback when you make a mistake. i.e. have the fires gradually ignite with each correct note, then go out if you play a wrong note.

And from discussion the other day:

- Have the next rock you're supposed to kick wobble occasionally

LauraFoglia:
see feedback #169

wjt:
I will open a series of sub-issues for concrete changes I think we should try, based on the playtesting feedback & discussion.

wjt:
I think we can close this one, with the big rework covered in #232.

PR #178, opened 2025-04-11
MusicalRock: End interaction immediately, while sound plays

wjt:
Previously, the rock interaction lasted as long as the sound effect. This was annoying because the player was frozen on the spot for the whole sound effect, rather than being able to move to the next note.

Don't await the end of the sound effect before ending the interaction.

Helps #112

----

Someone mentioned this in the playtest today; I can't find it in the notes but I think it makes the puzzle feel better. Making this change required fixing a latent bug in PlayerInteraction.


JuanFdS (approved):
💯 great, and nice catch.

manuq (approved):

PR #179, opened 2025-04-11
Scenes: Fix some paths to external resouces

manuq:
Opening the scenes in the editor automatically updates the paths from the
resource UIDs. It looks like the restructuring in commit 13a3a1f0 has left the paths
of a previous uncommited structure.


JuanFdS (approved):
Oops D:, thanks for the fix!

Issue #156, opened 2025-04-10
Fix buildings and trees collision Layers


pablitar:
Currently buildings and trees are on layer 1. They should likely be on layer 5 (walls) or another layer named for them

manuq:
I think that it should be a separate layer, so for example projectiles like ink blobs collide with walls, but not with non-walkable floor. So maybe rename current "walls" collision to "non-walkable" or something, and add a new "walls".

JuanFdS:
> I think that it should be a separate layer, so for example projectiles like ink blobs collide with walls, but not with non-walkable floor. So maybe rename current "walls" collision to "non-walkable" or something, and add a new "walls".

I like this!, but, we should try to define what is a wall and what is non walkable floor.
I think walls should also be "nonwalkable", but the difference would be how projectiles (ink blobs) interact with them. They can go over non walkable floor, but they collide with walls.

And that raises a question, should the stone walls (4th and 6th rows) be `walls` or `non_walkable_floor`? 

![Image](https://github.com/user-attachments/assets/c383ba4f-5d6f-4784-953d-6182c38ed364)

PR #181, opened 2025-04-11
Fix several game elements collisions

JuanFdS:
All these elements are game elements that the player should collide with. Set them all to the collision layer 5 (walls) and, if they had collision mask 1, removed it since it's redundant.

Fixes https://github.com/endlessm/threadbare/issues/156


wjt:
@manuq I'll leave this to your eyes since you have already thought about this topic in #156. Hope that's OK

manuq:
I need to test and understand this change.

pablitar (approved):
LGTM!

manuq (approved):
Yes, these are definitely not "npcs". Good point @JuanFdS .

PR #180, opened 2025-04-11
Replace dialogues in Ink Combat.

JuanFdS:
Replace dialogue lines in ink combat, according to https://github.com/endlessm/threadbare/issues/6

Also:

- Join the 2 ink_combat dialogues into one with two segments.

- Configure the collectible to trigger the 'well_done' segment of that dialogue.

PR #182, opened 2025-04-11
Improve ink combat

manuq:
Here is one simplified combat round. It seems trivial, but is supposed to be one of three. I think this will help to move forward for now.

Helps https://github.com/endlessm/threadbare/issues/111


scenes/quests/lore_quests/quest_001/2_ink_combat/components/ink_combat_logic.gd
```diff
@@ -17,10 +17,23 @@ func start() -> void:
...
+
+func _update_allowed_colors() -> void:
+	var allowed_colors: Array = []
+	for inkwell: Inkwell in get_tree().get_nodes_in_group("inkwells"):
+		if inkwell.is_queued_for_deletion():
+			continue
+		if inkwell.ink_color_name not in allowed_colors:
+			allowed_colors.append(inkwell.ink_color_name)
+	for ink_drinker: InkDrinker in get_tree().get_nodes_in_group("ink_drinkers"):
+		ink_drinker.allowed_colors = allowed_colors
```
wjt, line 31: Suppose you have 3 cyan inkwells and 1 black inkwell. Then the `allowed_colors` array will contain cyan thrice and black once; and so the probability of an enemy throwing cyan will be 75% compared to 25% black since `pick_random()` samples the array with equal probability.

I think that's probably deliberate, and desirable, but maybe it's worth a note somewhere, perhaps a docstring on `var allowed_colors` to say something like:

```gdscript
## The color of each ink blob thrown will be a random choice from this array.
## So if a color appears more than once, this will increase the chance that it is thrown.
```

wjt (approved):

wjt:
Hmm, conflicts here due to the merge of:

- #180

scenes/quests/lore_quests/quest_001/2_ink_combat/components/ink_combat_logic.gd
manuq: Good point, yes I need to get used to explain more through docstrings.

manuq:
> Hmm, conflicts here due to the merge of:
> 
> * [Replace dialogues in Ink Combat. #180](https://github.com/endlessm/threadbare/pull/180)

No problem, I will rebase.

JuanFdS (approved):
Great! 👍🏾

scenes/quests/lore_quests/quest_001/2_ink_combat/components/ink_combat_logic.gd
manuq: The docstring still applies and I'm keeping it.

PR #186, opened 2025-04-14
InkDrinker: Fix UNTYPED_DECLARATION warnings

wjt:
W 0:00:00:802   Variable "current_angle" has no static type.
      <GDScript Error>UNTYPED_DECLARATION
      <GDScript Source>ink_drinker.gd:133


wjt:
Personally I think the inferred types from `:=` are OK – it's clear enough from context. I think explicit type declarations are most important for top-level fields and function signatures.

manuq:
> Personally I think the inferred types from `:=` are OK – it's clear enough from context. I think explicit type declarations are most important for top-level fields and function signatures.

OK! Let's do that then.

PR #188, opened 2025-04-14
Fade from splash to intro narrative

wjt:
Previously, when the splash screen was complete (either due to 5 seconds passing from the end of the animation, or the player pressing a key), the scene would be switched abruptly to the intro, without any transition.

Add a fade.

Compared to #152 there are a few more details here to deal around avoiding double-transitioning and delaying the start of the intro dialogue until the fade is complete.

### DIFF
--- a/scenes/menus/intro/components/intro.gd
+++ b/scenes/menus/intro/components/intro.gd
@@ -12,6 +12,10 @@ var introduction: DialogueResource = preload("res://scenes/menus/intro/component
 
 
 func _ready() -> void:
+	# Wait a moment before starting the dialogue. In particular, this allows
+	# time for any fade-in transition from the previous scene to finish.
+	await get_tree().create_timer(1.0).timeout
+
 	DialogueManager.dialogue_ended.connect(_on_dialogue_ended)
 	DialogueManager.show_dialogue_balloon(introduction, "", [self])
 
--- a/scenes/menus/splash/components/splash.gd
+++ b/scenes/menus/splash/components/splash.gd
@@ -14,7 +14,7 @@ func _ready() -> void:
 
 
 func _process(_delta: float) -> void:
-	if Input.is_action_just_pressed(&"ui_accept"):
+	if Input.is_action_just_pressed(&"ui_accept") or Input.is_action_just_pressed(&"ui_cancel"):
 		switch_to_intro()
 
 
--- a/scenes/menus/splash/components/splash.gd
+++ b/scenes/menus/splash/components/splash.gd
@@ -13,8 +13,9 @@ func _ready() -> void:
 	scene_switch_timer.timeout.connect(switch_to_intro)
 
 
-func _process(_delta: float) -> void:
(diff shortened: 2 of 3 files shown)


scenes/menus/intro/components/intro.gd
```diff
@@ -12,6 +12,10 @@ var introduction: DialogueResource = preload("res://scenes/menus/intro/component
 
 
 func _ready() -> void:
+	# Wait for the scene transition from the splash screen to be complete.
+	while Pause.is_paused(Pause.System.PLAYER_INPUT):
```
pablitar, line 16: This to me seems a little hacky.

Maybe we can add a signal in the `Transitions` node (`Transitions.gd`) that triggers once the transition finishes?I'm a signal called: `transition_finished(is_entering: bool)` where is_entering is true when the transition is entering the scene, and false when leaving.

manuq: The Pause autoload already has a signal (I remember because I requested it in the review).

```
signal pause_changed(system: System, paused: bool)
```

wjt: That signal is awaited in the following line. The problem is that it can't be unconditional because, if you run the scene directly, no transition is running. So unless I'm missing something it has to be conditional even if we add the suggested new signal to `Transitions`.

The other option I considered is to have a method that would be called on the scene by `Transitions`, but again the scene would have to know whether to expect that method call.

If we were using the built-in pause functionality, we could have the scene-switching code pause the new scene until the end of the transition...

wjt: Actually I think it can be made to work if we have the Transitions singleton unconditionally emit that signal with paused=false for each system. I'll experiment tomorrow. And I think pausing the scene after it's constructed may be too late - _ready would have been called.

pablitar: > That signal is awaited in the following line. The problem is that it can't be unconditional because, if you run the scene directly, no transition is running. So unless I'm missing something it has to be conditional even if we add the suggested new signal to Transitions.

You are right! It's a tricky one.

Another option is to delay the dialog just a second after the scene starts. My thinking being that nothing important should happen in the first second you enter a scene. The transitions and everything that happen behind should just be decoration.

Of course, pausing could also be an option. We could add the pause system as a parameter to the methods `change_to_xxx_with_transition` so the caller can choose wich system should pause during transition.

wjt:
Rebased against main; the only change is to wait 1 second before playing the intro, rather than monitoring pausedness.

scenes/menus/intro/components/intro.gd
wjt: OK. In the hope of unblocking this I've replaced the "wait until unpaused" loop with an unconditional 1-second pause before triggering the dialogue. Actually I'm not convinced this is better because we now have the 1-second transition duration hardcoded in a second place; but equally I don't think it's worth exposing the default transition duration in `Transitions`; or doing something more fancy with signals. This works for now. I plan to refactor this script to also play the introduction to the Stella and the Missing Star StoryQuest, so I'm keen to move on :)

manuq (approved):

PR #190, opened 2025-04-14
Add a player direction to spawn point

pablitar:
This commit adds a player direction to the spawn point. When set to other than "UNSPECIFIED", it will change the player look at direction to either left or right.

When players teleport from one scene to the other, sometimes it looks weird if the player looks in the opposite direction. This solves that issue by allowing to specify the look direction in the spawn point.

### DIFF
--- a/scenes/game_elements/characters/player/components/player.gd
+++ b/scenes/game_elements/characters/player/components/player.gd
@@ -104,13 +104,18 @@ func _process(delta: float) -> void:
 	move_and_slide()
 
 
-func teleport_to(tele_position: Vector2, smooth_camera: bool = false) -> void:
+func teleport_to(
+	tele_position: Vector2,
+	smooth_camera: bool = false,
+	look_side: Enums.LookAtSide = Enums.LookAtSide.UNSPECIFIED
+) -> void:
 	var camera: Camera2D = get_viewport().get_camera_2d()
 
 	if is_instance_valid(camera):
 		var smoothing_was_enabled: bool = camera.position_smoothing_enabled
 		camera.position_smoothing_enabled = smooth_camera
 		global_position = tele_position
+		%PlayerSprite.look_at_side(look_side)
 		await get_tree().process_frame
 		camera.position_smoothing_enabled = smoothing_was_enabled
 	else:
--- a/scenes/globals/enums.gd
+++ b/scenes/globals/enums.gd
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name Enums
+
+enum LookAtSide {
+	UNSPECIFIED = 0,
+	LEFT = -1,
+	RIGHT = 1,
+}
--- a/scenes/game_elements/characters/npcs/shared_components/npc.gd
+++ b/scenes/game_elements/characters/npcs/shared_components/npc.gd
@@ -4,17 +4,11 @@
 class_name NPC
 extends CharacterBody2D
 
-enum LookAtSide {
-	FRONT = 0,
-	LEFT = -1,
-	RIGHT = 1,
-}
-
 const DEFAULT_SPRITE_FRAME: SpriteFrames = preload(
 	"res://scenes/game_elements/characters/shared_components/sprite_frames/sprite_frame_01.tres"
 )
 
-@export var look_at_side: LookAtSide = LookAtSide.LEFT:
+@export var look_at_side: Enums.LookAtSide = Enums.LookAtSide.LEFT:
 	set = _set_look_at_side
 
 @export var sprite_frames: SpriteFrames = DEFAULT_SPRITE_FRAME:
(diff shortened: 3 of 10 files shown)


scenes/game_elements/characters/player/components/player.gd
```diff
@@ -39,6 +39,12 @@ var last_nonzero_axis: Vector2
 @onready var player_fighting: Node2D = %PlayerFighting
 @onready var player_sprite: AnimatedSprite2D = %PlayerSprite
 
+enum LookAtSide {
+	UNSPECIFIED = 0,
+	LEFT = -1,
+	RIGHT = 1,
+}
```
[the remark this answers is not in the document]
manuq: :+1: this should be refactored so we don't have 2 implementations.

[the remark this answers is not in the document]
pablitar: I'll do that. I didn't want to mix a bunch of stuff on the same PR, but if you are OK with that, I'll extract the enum.

scenes/globals/enums.gd
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name Enums
```
manuq, line 3: I think that the suggestion was to come up with the following class hierarchy:
- Character
  - Player
  - NPC
 
And this enum can go to Character.

manuq (changes requested):

scenes/globals/enums.gd
pablitar: Signs also use this enum!

manuq: Ah OK, the commit didn't say anything about the signs but I see now. In any case the name `enums.gd` or having a global named `Enums` is too broad, in my opinion. If it lives in a class named `Character` then signs can reference eg. `Character.LookAtSide.LEFT`. Or if a global is preferred, could be `Directions` a better name?

pablitar: But then we would do `Directions.LookAtSide.LEFT`?

There's a limitation in Godot where we cannot have named enums that are available globally. You can have an unnamed enum in a class and access those values as constants, but you cannot use it as a type. So we need the enum to be named and to be inside a class. My thinking was that we could have `Enums` be like a namespace, of all the "global" enums.

wjt: I don't love these kinds of "misc" namespaces but I don't really see a better home for this enum, given the limitations you point out. OK by me!

scenes/menus/splash/splash.tscn
```diff
@@ -42,7 +42,7 @@ curve = SubResource("Curve2D_eqfq0")
...
...2, 982.884, 640.569, 982.121, 640.114, 981.353, 639.656, 980.582, 639.196, 979.807, 638.733, 979.029, 638.268, 978.248, 637.802, 977.465, 637.333, 976.68, 636.863, 975.893, 636.391, 975.104, 635.918, 974.314, 635.445, 973.524, 634.97, 972.733, 634.495, 971.942, 634.019, 971.151, 633.543, 970.36, 633.067, 969.571, 632.591, 968.782, 632.115, 967.995, 631.64, 967.21, 631.165, 966.427, 630.691, 965.646, 630.219, 964.868, 629.747, 964.094, 629.277, 963.322, 628.808, 962.555, 628.341, 961.791, 627.876, 961.032, 627.414, 960.278, 626.953, 959.529, 626.495, 958.785, 626.04, 958.047, 625.588, 957.316, 625.139, 956.59, 624.693, 955.871, 624.25, 955.16, 623.811, 954.455, 623.376, 953.759, 622.945, 953.07, 622.518, 952.39, 622.096, 951.718, 621.678, 951.056, 621.265, 950.403, 620.856, 949.759, 620.453, 949.126, 620.055, 948.502, 619.663, 947.89, 619.276, 947.288, 618.895, 946.698, 618.521, 946.12, 618.152, 945.553, 617.79, 944.999, 617.435, 944.457, 617.086, 943.928, 616.744, 943.413, 616.41, 942.911, 616.082, 942.423, 615.763, 941.95, 615.451, 941.49, 615.147, 941.046, 614.851, 940.618, 614.563, 940.204, 614.284, 939.426, 613.752, 938.713, 613.256, 938.07, 612.798, 937.498, 612.379, 937, 612)
```
wjt, line 45: Don't think you meant to commit this. I should change this tool script to do less when the scene loads!

pablitar: Oops sorry, will revert it!

pablitar:
I think this is ready to merge! @manuq would that be ok?

manuq (approved):
Excellent. Feel free to switch the second commit to imperative mood (see https://github.com/endlessm/threadbare/blob/main/docs/CONTRIBUTING.md#git-commit-messages-and-pull-request-descriptions) and then merge.

manuq (changes requested):
Actually the commits need to be squashed. Because the 2nd one undo changes done in the 1st one.

Also @pablitar I see you changed the preexisting enum from FRONT to UNSPECIFIED. I guess that the unspecified can be interpreted from the null value (eg. NPC look_at_side = null). But anyways if you still prefer that name, fine with me, just explain it.

pablitar:
> Actually the commits need to be squashed. Because the 2nd one undo changes done in the 1st one.
> 
> Also @pablitar I see you changed the preexisting enum from FRONT to UNSPECIFIED. I guess that the unspecified can be interpreted from the null value (eg. NPC look_at_side = null). But anyways if you still prefer that name, fine with me, just explain it.

We currently don't have any sprites that face Front, except for the Elder. The current usage of FRONT is actually to signal "no specific direction", since null values are not allowed in variables typed for enum values. I thought that "UNSPECIFIED" was more descriptive of the usage.

And if eventually we need to add a "FRONT" direction with a corresponding behaviour, we'll still have the "UNSPECIFIED" value which will be useful for places that don't need to speficy a direction.

pablitar:
Commits squashed!

manuq (approved):

Issue #191, opened 2025-04-15
Add skeleton opening scene for Stella


wjt:
The *Stella and the Missing Star* StoryQuest™ begins with a narration scene introducing the story.

Add this scene to the project, with placeholder assets.

PR #196, opened 2025-04-15
Make intro script reusable

wjt:
For the *Stella and the Missing Star* StoryQuest, the GDD calls for a cutscene at the start (and end!) of the quest. This is suspiciously similar to the intro scene for the whole game: dialogue with synched animations but no interactivity.

Refactor the existing Intro scene so that its script can be reused for another cutsceen.

Helps #191


wjt:
Since I also plan to use this for the Stella outro, I wonder if it should be renamed to e.g. `dialogue_cutscene.gd` (and placed into a subdirectory of game_elements)?

manuq:
> Since I also plan to use this for the Stella outro, I wonder if it should be renamed to e.g. `dialogue_cutscene.gd` (and placed into a subdirectory of game_elements)?

Yes, I like the idea. Basically a scene that triggers a dialogue while playing an animation, and can go to the next scene. Dialogue, AnimationPlayer and path to the next scene being the exported properties.

wjt:
- Main game: https://endlessm.github.io/threadbare/branches/endlessm/make-intro-reusable/
- Closing MVP quest cutscene: https://endlessm.github.io/threadbare/branches/endlessm/make-intro-reusable/#uid://cd2p2udmkif80
- Stealth level, to test transition into that cutscene: https://endlessm.github.io/threadbare/branches/endlessm/make-intro-reusable/#uid://db6vhi7s2f37c

manuq (approved):

PR #198, opened 2025-04-15
Update Godot Dialogue Manager to 3.5.1

wjt:
We had to revert the 3.5.0 update in commit
f60f717a7bbf6a2fc31c6aa85e8dac423f96d5b3 due to a regression in exporting. This regression was fixed in 3.5.1.


wjt:
Web export seems to work this time!

Issue #193, opened 2025-04-15
Add Australian shoreline combat scene for Stella quest


wjt:
The second challenge of *Stella and the Missing Star* is set on the Australian shoreline, with a fisherman NPC and a combat quest.

Add a skeleton of this scene, with opening dialogue with NPC, placeholder combat, and collectible.

wjt:
This StoryQuest development is being tracked elsewhere.

PR #203, opened 2025-04-15
Generalize ink combat to a generic fill game

manuq:
So that it can be tweaked. This change is guided by the action game in the Stella storyquest about collecting fishes.

https://github.com/endlessm/threadbare/issues/193


manuq:
This is how far I came today with the renames and generalization. Since the game is playable at this point I am putting it for review.

My plan for the projectile/barrel matching is:
- Add string labels to both projectiles and barrels.
- Add optional color to projectiles.
- Remove the current mapping of labels and colors.

So other games like Stella's can use just labels and custom sprites. If the optional color is set, then the sprites tint (modulate) will change.

manuq:
Back to draft because I'm piling more commits today. This is getting big...

manuq:
OK, since this is getting big I'm putting the first chunk for review. It is mostly renames. The next step will be to expose the sprites so the inkwell and the ink blobs can be themed for Stella.

PR #204, opened 2025-04-15
Add a wind shader to the trees

pablitar:
The trees now have a material that has a shader that simulates wind.

It's done through a fragment shader that offsets the texture sampling based on time and some parameters.

The material also has a noise to define a "wind map" so that trees close to each other have similar wind conditions but there is some variability in the movement. Some parameters are also considered, like max_wind_speed and max_wind_intensity. Those parameters could be increased to simulate a quiet breeze, or a strong gust.


manuq (approved):
Awesome!

Issue #154, opened 2025-04-10
Make it easier to interact with things even if they aren't perfectly aligned with the player


JuanFdS:
There are some interactable game elements that feel a bit hard to interact with because they have to be exactly at the side of the player.

I think we could use a more generous collision from the player to detect interactables, like a [ShapeCast2D](https://docs.godotengine.org/en/stable/classes/class_shapecast2d.html) instead of a [RayCast2D](https://docs.godotengine.org/en/stable/classes/class_raycast2d.html). Or maybe have 2 raycasts instead of one. It still should only let you interact with one element at the time (maybe the one that's closest), but it could make it a bit easier to interact with things in the game.

https://github.com/user-attachments/assets/9c9c0da5-a282-485e-82ef-2d7ccd1c2cae

https://github.com/user-attachments/assets/562f048d-5712-46a9-bb07-13f7abed5b1e

manuq:
Isn't this "just" a matter of enlarging the interact areas? Specially for the rocks.

Also for the rocks, I wouldn't block the player movement while the sound is being played.

JuanFdS:
> Isn't this "just" a matter of enlarging the interact areas? Specially for the rocks.
> 
> Also for the rocks, I wouldn't block the player movement while the sound is being played.

But that means that for every interactable thing, we need to remember that the collision shape should be actually larger than the sprite itself to compensante the player's narrow interaction collision.

PR #214, opened 2025-04-16
Change interaction to use an area

Arebuayon:
Interaction area lets us detect collisions even if they arent exactly in front of the player

This fixes https://github.com/endlessm/threadbare/issues/154


pablitar (approved):
LGTM! Tested it and it works wonders

manuq (approved):
Excellent, yes this rectangle area does the job much better than the raycast.

Issue #205, opened 2025-04-16
Implement new dialogue balloon designs


wjt:
#162 redesigns the dialogue balloons. Implement this design.

wjt:
Making a start

![Image](https://github.com/user-attachments/assets/666d6207-5954-47fa-a722-d71a66389a8b)

wjt:
![Image](https://github.com/user-attachments/assets/4c772feb-4c36-4730-9b5c-2f4929557fc5)

wjt:
![Image](https://github.com/user-attachments/assets/3c3029c0-4ae4-4ce9-84a9-71a28448e3ac)

jofilizola:
The text needs more padding, the overall box can have at least one more 64px module to the hight and width I thing.
I like the buttons! The lables need to be black, because there's no enough contrast now.

wjt:
The two outstanding pieces after #219 are to move the balloon to the bottom-left for in-game dialogue, and to change the speaker name ribbon's colour based on whether the speaker is the player's character or an NPC. I suggest deferring both for now. #158 could track that second change.

PR #216, opened 2025-04-16
Dialogue: Redesign balloon

wjt:
This rebuilds the dialogue balloon scene to be tall rather than wide, anchored to the top-left corner of the viewport, and with the speaker's name pulled out to a ribbon at the top of the scroll. The dialogue-choice buttons are moved into the scroll.

Compared to the previous implementation, derived from Godot Dialogue Manager's example, I found it clearer to use PanelContainers rather than using Panel and then carefully positioning the children. The theme is updated accordingly.

Two styles of ribbon are defined for the speaker's name, BlueRibbon and YellowRibbon. Only one is used in this implementation, but the intention is that, in future, BlueRibbon will be used for the protagonist's name and YellowRibbon for the names of NPCs.

Compared to the design, the balloon is always anchored at the top-left, rather than being in the top-left for narration and bottom-left for dialogue. Again this is deferred to a future change.

Helps https://github.com/endlessm/threadbare/issues/205

### DIFF
--- a/scenes/ui_elements/dialogue/balloon.tscn
+++ b/scenes/ui_elements/dialogue/balloon.tscn
@@ -12,70 +12,25 @@ script = ExtResource("1_36de5")
 [node name="Balloon" type="Control" parent="."]
 unique_name_in_owner = true
 layout_mode = 3
-anchors_preset = 15
-anchor_right = 1.0
-anchor_bottom = 1.0
-grow_horizontal = 2
-grow_vertical = 2
+anchors_preset = 0
 theme = ExtResource("2_pfnde")
 
-[node name="Panel" type="Panel" parent="Balloon"]
-clip_children = 2
-layout_mode = 1
-anchors_preset = 12
-anchor_top = 1.0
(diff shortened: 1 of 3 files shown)


scenes/ui_elements/dialogue/balloon.tscn
```diff
@@ -12,80 +12,54 @@ script = ExtResource("1_36de5")
...
+custom_minimum_size = Vector2(480, 480)
 layout_mode = 2
-mouse_filter = 1
-bbcode_enabled = true
-text = "Character"
-fit_content = true
-scroll_active = false
 
-[node name="DialogueLabel" parent="Balloon/Panel/Dialogue/VBoxContainer" instance=ExtResource("2_a8ve6")]
+[node name="DialogueLabel" parent="Balloon/PanelContainer/VBoxContainer" instance=ExtResource("2_a8ve6")]
```
manuq, line 27: Yes, I agree that PanelContainer is better for positioning the children than Panel here.

manuq (approved):

Issue #121, opened 2025-04-07
Integrate art assets


LauraFoglia:
switch placeholders for the final versions by Pixel Frog
- houses
- trees
- grass tiles

[Files.v1.zip](https://github.com/user-attachments/files/19632492/Files.v1.zip)

LauraFoglia:
the tower 

[Tower.png.zip](https://github.com/user-attachments/files/19637144/Tower.png.zip)

wjt:
As documented in the README, the license should be CC-BY-SA-4.0:

https://github.com/endlessm/threadbare/blob/bab0c312d9207dd22708a267a088dd297dbf35b5/README.md#L8-L9

LauraFoglia:
New assets to integrate:
- bushes
- Thread spools
- new houses and towers structures

[All files.zip](https://github.com/user-attachments/files/19762697/All.files.zip)

manuq:
I guess this is already done? It has the same title as #119 so it's a bit confusing.

LauraFoglia:
@pablitar was going to integrate the new houses and towers structures that were still missing. Let's check this again on todays daily.

JuanFdS:
@JuanFdS will check if this can be archived as duplicate of #119

JuanFdS:
The assets listed here were already integrated

PR #218, opened 2025-04-16
Add new house assets

JuanFdS:
Part of https://github.com/endlessm/threadbare/issues/207, related to https://github.com/endlessm/threadbare/issues/121


scenes/game_elements/props/buildings/house/components/house.gd
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends Node2D
+
+@export var house_texture: Texture2D = preload("./House_Patch_Blue_01.png"):
+	set(new_texture):
+		house_texture = new_texture
+		if house_sprite:
+			house_sprite.texture = house_texture
```
wjt, line 10: Perhaps a stupid question but, how does this work at runtime in the game proper? I thought that all properties of this nature were set before `_ready`, but am I wrong?

JuanFdS: You're not. And it won't work, good catch 😅 , I didn't doublecheck this change we made during the meeting the other, but yep, we're missing setting this in `_ready()`

JuanFdS: Fixed now ✅

PR #219, opened 2025-04-16
Dialogue: Invert text colour in buttons

wjt:
Prevously, the default font colour was used, which is either white or slightly off-white depending on the state of the button. This is not legible against the background colour of the button assets.

Invert each colour. For instance, Button/colors/font_color defaults to hsv(0, 0, 88), so here it is changed to hsv(0, 0, 100 - 88) = hsv(0, 0, 12).

https://github.com/endlessm/threadbare/issues/205#issuecomment-2809899852

Screenshot from 2025-04-16 19-59-39


JuanFdS (commented):
Great ✅

JuanFdS (approved):

Issue #201, opened 2025-04-15
Replace font for the game


jofilizola:
The font we are using (https://managore.itch.io/m6x11) doesn't render well on the web, it has a missing "quote" glyph.

![Image](https://github.com/user-attachments/assets/0a107875-83ee-4dcb-b96e-342395bae6ad)

It also doesn't have all the characters for accents. 
![Image](https://github.com/user-attachments/assets/d75f818b-4702-472d-b29e-6d9d1df90e97)

I propose we replace it (will comment with alternatives in a bit)

wjt:
It does have ôñé – we are using the m6x11plus variant.

It's missing curly quotes and apostrophes: “ ” ‘ ’

![Image](https://github.com/user-attachments/assets/50a730c0-1b0b-4588-919c-6b9d492911cd)

jofilizola:
Ok, I'm just worried learners might write their narrative using inextistent characters.
I would fix that adding the missing chacters, or replace the font... I'm not sure if you can see if the character map is more complete in this collection from Google fonts that has been made for knitting (which I thought was very nice)... If we think replacing could make sense. 

https://fonts.google.com/specimen/Jersey+10/
https://fonts.google.com/specimen/Jersey+15/
https://fonts.google.com/specimen/Jersey+20/
https://fonts.google.com/specimen/Jersey+25/

wjt:
Jersey has pretty good coverage of a variety of quotes and Latin alphabets that I tried quickly on the Google Fonts interface:

![Image](https://github.com/user-attachments/assets/18d5667c-983e-4455-ac47-da790cea0d1e)

It does not cover the [cuatrillo “ꜭ”](https://en.wikipedia.org/wiki/Cuatrillo), a letter used in a 16th-century colonial alphabet for Mayan languages, but I think that's forgivable.

wjt:
We tried adding Jersey and I couldn't get it to render crisply. It's probably me not understanding Godot's text rendering and/or the natural size for each variant properly. So we paused this for now.

wjt:
And we have worked around the lack of em dashes and curly apostrophes by using hyphens and straight apostrophes.

PR #220, opened 2025-04-16
Intro: Use straight apostrophe

wjt:
The current font does not have a glyph for `’`.


wjt:
#201 exists but this is a quick fix to make the game not look broken during the introductory sequence.

JuanFdS (approved):

Issue #221, opened 2025-04-16
Make it easier to have nodes whose textures can be modified


JuanFdS:
We want the learners to be able to change several aspects of the game elements to create their own story quests.

One of the things we'd want them to easily modify is how the game elements look, and the most straightforward way to do this is by exporting a property that makes a node's texture configurable.

However, this requires some boilerplate that's pretty similar every time it happens, as can be seen in:

https://github.com/endlessm/threadbare/blob/23ac23c57c2fece96be211d62d1ba132d3a09626/scenes/game_elements/props/decoration/bush/components/bush.gd#L6-L18
https://github.com/endlessm/threadbare/blob/23ac23c57c2fece96be211d62d1ba132d3a09626/scenes/game_elements/props/decoration/spool/components/spool.gd#L6-L18
https://github.com/endlessm/threadbare/blob/4fa861ac1b93b280d73dd653995e35eeff0b02e2/scenes/game_elements/props/buildings/house/components/house.gd#L6-L16

I'd like to discuss possible ways to address this:
1 - Make all these nodes extend from a common superclass that has that logic. I want to avoid this because it seems a really weak reason to use inheritance.
2 - Just deal with the repetition, is not that much. Try to reduce it a bit more, maybe we can avoid the ready if we write the setter in a cleverer way🤔. I actually prefer this over 1.
3 - Standarize and reduce the boilerplate a bit: I think that we could generalise the setter logic and even avoid the declaration of the `@export`ed sprite by having a `_set()` and a `_get_property_list()` that we can reuse in different scripts. Something like:
```gdscript
func _set(property: StringName, value: Variant) -> bool:
   return HandleConfigurableTextures._set(property, value, self, "PathToMySprite")

func _get_configuration_list():
   return HandleConfigurableTextures._get_configuration_list(self, "PathToMySprite")
```
So at least the boilerplate gets to be exactly the same. However, I'm afraid this would be too opaque for contributors. I still think we should explore something like this to see what we can achieve.
- Other ideas...?

JuanFdS:
Since all these can be considered "decoration" nodes, they could also all share the same exact script, so instead of spools.gd, house.gd, and bush.gd, we could have a decoration.gd that let's us change its sprite. I feel more comfortable with doing this than having a common superclass just for the texture. If any of those nodes need its own special behavior, we can copy paste the setter and variable to a new script and have it do its own thing.

I think this pattern would come up in other kind of nodes as well.

wjt:
[...]

PR #222, opened 2025-04-16
Add bushes and spools

JuanFdS:
Part of #121 
Some logic started repeating with these nodes, created an issue to explore what we can do about it: https://github.com/endlessm/threadbare/issues/221


wjt (approved):
LGTM. I don't particularly mind if this is merged as-is or if you want to make the change you suggested in  https://github.com/endlessm/threadbare/issues/221 to have a `decoration.gd` script that these each use.

PR #223, opened 2025-04-16
Splash: Don't handle input while PLAYER_INPUT is is_paused

wjt:
Previously, if you pressed Escape or Space twice in quick succession on
the splash screen, switch_to_intro() would be called twice. The second
call triggers a warning when trying to disconnect a signal from the
timer a second time.

Ignore input when PLAYER_INPUT is paused (i.e. mid-transition).


JuanFdS (approved):
I think it makes sense 👍🏾

PR #226, opened 2025-04-16
Add and integrate Ink Drinker assets

JuanFdS:
Add new spritesheets for ink drinkers.

Include .aseprite file as well.

Change ink_drinker_frames.tres to use those spritesheets.

Remove all_frames animation since it wasn't used

Fixes https://github.com/endlessm/threadbare/issues/225

https://github.com/user-attachments/assets/2877a8c3-e2f6-44e9-9c5f-fc7b4985a686


manuq (approved):
So cool! Suggestion: the ProjectileMarker node can be moved so the spawned ball matches the new animation. But we can do that after merging this too.

JuanFdS:
> So cool! Suggestion: the ProjectileMarker node can be moved so the spawned ball matches the new animation. But we can do that after merging this too.

Moved it a bit 👍🏾

PR #238, opened 2025-04-17
Animated sprites with configurable textures

JuanFdS:
## Proposed workflow for learner making their own spritesheet

There are actually 2 valid ways of creating a new spritesheet:
- Duplicating it from the editor's filesystem.
- In the inspector of a selected node that exposes a SpriteFrames, click _Make Unique_.

Then, once the new spritesheet is created, it's just either dragging a texture to the one that the user wants to swap.

https://github.com/user-attachments/assets/732af1f1-a11d-4dd1-9340-da68926a93a5

### DIFF
--- a/addons/sprite_frames_exported_textures/sprite_frames_inspector_plugin.gd
+++ b/addons/sprite_frames_exported_textures/sprite_frames_inspector_plugin.gd
@@ -1,0 +1,22 @@
+extends EditorInspectorPlugin
+
+const SpriteFramesEditorProperty = preload(
+	"res://addons/sprite_frames_exported_textures/sprite_frames_editor_property.gd"
+)
+const SpriteFramesHelper = preload(
+	"res://addons/sprite_frames_exported_textures/sprite_frames_helper.gd"
+)
+
+
+func _can_handle(object: Object) -> bool:
+	return object is SpriteFrames
+
+
+func _parse_begin(object: Object) -> void:
+	if object is SpriteFrames:
+		var sprite_frames: SpriteFrames = object
+		for base_texture: Texture2D in SpriteFramesHelper.base_textures_used(sprite_frames):
+			var sprite_frames_editor_property = SpriteFramesEditorProperty.new(
+				base_texture, sprite_frames
+			)
+			add_property_editor(base_texture.resource_path, sprite_frames_editor_property, true)
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -14,6 +14,7 @@ SPDX-License-Identifier = "CC0-1.0"
 # when saving changes to these files.
 [[annotations]]
 path = [
+    "addons/sprite_frames_exported_textures/**",
     "scenes/*.tscn",
     "scenes/**/*.tscn",
     "scenes/**/*.tres",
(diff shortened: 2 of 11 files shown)


pablitar (commented):
I think this is great. We should merge it and test it thoroughly with real world cases!

Feel free to address the comments I made!

manuq:
@JuanFdS did a demo of the current state. It looks very promising!

JuanFdS:
It's working but there are still some quirks I'm trying to fix:
- Sometimes, I get this error in the terminal:
image
Which is related with this code:
```gdscript
			EditorInterface.get_inspector().property_edited.connect(func(property_name: StringName):
				object.notify_property_list_changed(),
				CONNECT_ONE_SHOT
			)
```
- If I make a spriteframe unique, save it and then edit its textures, it seems like if the original spriteframes also had its textures modified, but it isn't the case!, if the project is reloaded, this is fixed. It's probably because there's a shared resource internally that is not being persisted (the AtlasTexture maybe?).

addons/sprite_frames_exported_textures/sprite_frames_inspector_plugin.gd
```diff
@@ -0,0 +1,48 @@
...
+
+func _parse_begin(object: Object) -> void:
+		if object is SpriteFrames:
+			var sprite_frames: SpriteFrames = object
+			if FileAccess.file_exists(sprite_frames.resource_path):
+				for base_texture: Texture2D in SpriteFramesHelper.base_textures_used(sprite_frames):
+					var sprite_frames_editor_property = SpriteFramesEditorProperty.new(
+						base_texture, sprite_frames
+					)
+					add_property_editor(base_texture.resource_path, sprite_frames_editor_property, true)
```
manuq, line 17: Can the property be added for `base_texture` instead of `base_texture.resource_path` ?

addons/sprite_frames_exported_textures/sprite_frames_inspector_plugin.gd
```diff
@@ -0,0 +1,48 @@
...
+			else:
+				var message_label := Label.new()
+				message_label.autowrap_mode = TextServer.AUTOWRAP_WORD
+				message_label.text = "⚠️ SpriteFrames must be saved to a file in order to edit its textures"
```
manuq, line 21: I don't understand this limitation. Can you explain why the SpriteFrames resource has to be saved to a file?

manuq (approved):
@JuanFdS this is working like a charm! I think it's a great addition that works around a current limitation of the Godot editor: unlike TileSet, the SpriteFrames doesn't expose the used textures. Also requiring that the replaced texture is of the same size is a nice touch.

Can you make it work with SpriteFrames that are saved in the scene too? At first glance I don't see what is limiting the addon to only work with SpriteFrames resources saved in its own separate files.

Can you run the linter and other checks in it? In `.pre-commit-config.yaml` we have the checks disabled for all `addons/` because the only addon is `dialogue_manager` which we copy and patch. I have opened https://github.com/endlessm/threadbare/pull/464 for this. If I add that commit on top of this PR and then run `pre-commit run --all-files` I get a number of issues.

PR #250, opened 2025-04-17
Guards improve learner side

### LINKED ISSUE #228: Modifications to configuring "Guard Patrol Paths"
Guard patrol path suggested changes:

- [ ] Can the guard inspector property say “create patrol path” and then, after it is created, the same button now says say “edit patrol path”. 

I think it might be clearer for someone to look there to get the path going if it said "create" at first. 

- [ ] When you create a patrol path: can you have it automatically give a basic straight line patrol path that is visual in the viewport so the learner can see visually what the path is that they now can manipulate? 

Currently a patrol path gets created but it can be confusing as it is a single point in the viewport that is hard to notice (just a small crosshair). If there was a simple path that got create it might be better for the learner to see and understand that there are points to manipulate and add etc.

JuanFdS:
Fixes #228 

-------

[Modify guard's helper button to say Add/Edit instead of just Edit](https://github.com/endlessm/threadbare/pull/250/commits/a44dcbe6994502ae464705419358af3831d64c3c) and [Guard: modify creation of new patrol path to ensure it has 2 points](https://github.com/endlessm/threadbare/pull/250/commits/62e6467462f3939c5a43cfaabe949213838098d2) can be seen here:

https://github.com/user-attachments/assets/5fb46767-da74-4fb6-9aea-a28894e09d68

----------

This feature wasn't mentioned in #228 , but I thought it could be a cool improvement to add as well:
[Guard: highlight patrol path when a guard is selected in the editor](https://github.com/endlessm/threadbare/pull/250/commits/72e68985a6e3138268c11231acc0828e46facf44)

https://github.com/user-attachments/assets/7d0c4f05-4513-4426-9cdd-a39223041a0f

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -26,7 +26,12 @@ const LOOK_AT_TURN_SPEED: float = 10.0
 @warning_ignore("unused_private_class_variable")
 @export_tool_button("Add/Edit Patrol Path") var _edit_patrol_path: Callable = edit_patrol_path
 ## The path the guard follows while patrolling.
-@export var patrol_path: Path2D
+@export var patrol_path: Path2D:
+	set(new_value):
+		patrol_path = new_value
+		if Engine.is_editor_hint():
+			queue_redraw()
+
 ## The wait time at each patrol point.
 @export_range(0, 5, 0.1, "or_greater", "suffix:s") var wait_time: float = 1.0
 ## The speed at which the guard moves.
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -409,6 +409,9 @@ func edit_patrol_path() -> void:
 		get_parent().add_child(patrol_path)
 		patrol_path.owner = owner
 		patrol_path.global_position = global_position
-		patrol_path.curve = Curve2D.new()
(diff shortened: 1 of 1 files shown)


scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -409,6 +417,38 @@ func edit_patrol_path() -> void:
 		get_parent().add_child(patrol_path)
 		patrol_path.owner = owner
 		patrol_path.global_position = global_position
-		patrol_path.curve = Curve2D.new()
+		var patrol_path_curve: Curve2D = Curve2D.new()
+		patrol_path.curve = patrol_path_curve
 		patrol_path.name = "%s-PatrolPath" % name
+		patrol_path_curve.add_point(Vector2.ZERO)
+		patrol_path_curve.add_point(Vector2.UP.rotated(randf_range(0, PI * 2.0)) * 150.0)
```
manuq, line 424: I'm not really sure about the second point being random. I think a horizontal line or a vertical line will be more predictable. But maybe I'm wrong and this is fun!

JuanFdS: Mmmm, yeah, let's keep it simple and just make a fixed horizontal path and have the students modify it.

wjt (approved):
Really nice work!

While testing this I found an edge case where, if you change the patrol path assigned to the guard, the old path is still highlighted. I pushed a patch to queue_redraw() in that case – feel free to merge if you think my fix is OK!

Issue #162, opened 2025-04-11
Design patterns for dialog and narration


jofilizola:
- Identify the cases (4 cases that will described)
- mock up the 4 cases

Playtesting:
https://github.com/endlessm/threadbare/issues/158

jofilizola:
@wjt I lost the 4 cases on the task that I was just creating and deleted, could you please help me with bullet points of what you just said in the meeting?

wjt:
The 4 cases I can think of are:

1. Dialogue between 2 or more characters on the screen (e.g. talking to an Elder to start a quest; talking to the bard in the [music challenge](https://endlessm.github.io/threadbare/#quests/lore_quests/quest_001/1_music_puzzle/main)):

![Image](https://github.com/user-attachments/assets/df345455-753a-4916-950d-af253a1bbbe3)

2. Dialogue where the speaker is at the bottom of the screen, e.g. in the [ink combat challenge](https://endlessm.github.io/threadbare/#quests/lore_quests/quest_001/2_ink_combat/ink_combat) – you can't see the ink drinkers because their speech covers them:

![Image](https://github.com/user-attachments/assets/201285e0-948a-4c46-b015-130dbda227ec)

3. An unseen character talks to you, e.g. I think the current design (not implemented) is that, when you collect a collectable at the end of a challenge, the elder speaks to you, to congratulate you and send you on to the next stage of the quest

4. Pseudo-dialogue that just directly tells the player a story, like in the [intro scene](https://endlessm.github.io/threadbare/#menus/intro/intro) or in the mockup for the start of the stella quest:

![Image](https://github.com/user-attachments/assets/13043122-36e4-449f-aca5-9d7ad000a478)

![Image](https://github.com/user-attachments/assets/f11b1f5e-6dd0-4c75-93cb-2890b4294ad1)

jofilizola:
Dialog boxes proposals (for MVP we can improve later) 

For cases 1 and 2 are the same, anchored on the bottom left:

[Prototype on Figma - Music Puzzle - clicks](https://www.figma.com/proto/C0WOJs9EWkDNLwGOnfDuHh/Threadbare-Storyboards?page-id=170%3A176&node-id=172-527&viewport=-775%2C535%2C0.53&t=S4oigOtaIrPDsYMj-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=172%3A527&show-proto-sidebar=1)

[...]

PR #252, opened 2025-04-20
Dialogue: Adjust ribbon colour based on who is speaking

### LINKED ISSUE #230: Change dialogue speaker banner based on current speaker
The banner should be blue if the protagonist is speaking and yellow otherwise.

wjt:
Previously, the speaker name ribbon was always blue.

With this change, the ribbon is blue if the player's character is speaking, or yellow otherwise, as designed in #162.

This is a little trickier than expected because the player controls different characters in StoryQuests (e.g. Stella in *Stella and the Missing Star*) compared to in the base game & its LoreQuests (StoryWeaver). To make this work, add a `player_name` property to the `Player` class, and pass the `Player` instance through to dialogue, similar to how the `npc_name` and NPC being interacted with is passed to the dialogue. Then, in our custom dialogue balloon, look for the Player in the list of game states, and fish out its name if found.

Fixes #230

### DIFF
--- a/scenes/game_elements/characters/player/components/player_interaction.gd
+++ b/scenes/game_elements/characters/player/components/player_interaction.gd
@@ -10,6 +10,8 @@ var is_interacting: bool:
 @onready var interact_marker: Marker2D = %InteractMarker
 @onready var interact_label: FixedSizeLabel = %InteractLabel
 
+@onready var player: Player = self.owner as Player
+
 
 func _get_is_interacting() -> bool:
 	return not interact_zone.monitoring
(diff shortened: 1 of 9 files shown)


wjt:
image
Screenshot from 2025-04-20 14-06-33

scenes/game_elements/characters/player/components/player_interaction.gd
```diff
@@ -28,7 +30,7 @@ func _process(_delta: float) -> void:
 		interact_zone.monitoring = false
 		interact_label.visible = false
 		interact_area.interaction_ended.connect(_on_interaction_ended, CONNECT_ONE_SHOT)
-		interact_area.start_interaction(interact_zone.is_looking_from_right)
+		interact_area.start_interaction(player, interact_zone.is_looking_from_right)
```
JuanFdS, line 33: For another moment, but if we are passing the player as a parameter, I think we won't need to pass which side the player's looking at as well.

JuanFdS (approved):
👍🏾

PR #254, opened 2025-04-21
StoryQuestStarter: Remove “Maybe later” dialogue branch

wjt:
This is functionally the same as the “No!” option and consensus during
an in-person playtest was that we should remove it.


JuanFdS (approved):
👍🏾

Issue #263, opened 2025-04-22
Add “Next” hint to dialogue balloons


wjt:
Once a line of dialogue has been spoken, if there are no response choices to be made, currently the player needs to guess that they have to press `Space`, `Enter`, or joypad `A` to progress to the next line. Playtesting has revealed that some players do not understand this, and so sit at the first line of dialogue in the intro scene forever, until they accidentally press a key to advance.

@jofilizola designed the following:

![Image](https://github.com/user-attachments/assets/f7add2de-cc08-4a2a-9d77-bc4de333be12)

She believes that, rather than having a visible description of which key needs to be pressed (space/enter), we can make it a flat button. Players who don't figure out that it's space/enter (both of which are normal "accept" shortcuts in non-game UIs) will click it so as not to get stuck; other players will figure it out.

PR #264, opened 2025-04-22
Dialogue: Add visible “next” Button

wjt:
Dialogue: Add visible “next” Button

In playtesting we have found that not all players realise that they need
to press a key to progress the dialogue.

Add a button that, when pressed, acts like the input action for "next"
(Enter, Space, Xbox joypad A) has been entered.

Style variations for the button added in a new style called FlatButton.

Fixes https://github.com/endlessm/threadbare/issues/263


manuq:
Great! Joana approved, with a change: to add bottom marging to the button. So I added it to a MarginContainer and adjusted the script accordingly.

PR #268, opened 2025-04-22
Updated dialogue

hydrolet:
Updated dialogue for each scene.


wjt:
I revised this patch to fix some linter errors, and (with @Stregoica777's agreement) to remove Tom the Elder entirely from the stealth quest rather than making his dialogue be empty.

PR #271, opened 2025-04-22
Redesign Frays End

PlayMatters:
Added villagers, rocks, bushes, trees top and bottom, river separation for start, bridge, pier, new pathways, and redesigned tileset accordingly.


wjt:
I rebased this to fix a conflict with #268, remove an unintended (I think) change from the sokoban scene, and a spurious edit to splash.tscn that is due to a bug in that tool script (to be fixed).

scenes/tileset.tres
```diff
@@ -29,12 +29,17 @@ tile_data = PackedInt32Array(0, 327681, 3, 1, 393217, 3, 2, 458753, 3)
...
 0:0/0 = 0
 0:0/0/z_index = -1
 0:0/0/physics_layer_0/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
+0:0/0/physics_layer_1/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
 0:0/1 = 1
 0:0/1/z_index = -1
+0:0/1/physics_layer_0/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
+0:0/1/physics_layer_1/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
+0:0/2 = 2
+0:0/2/physics_layer_0/polygon_0/points = PackedVector2Array(-32, 3.85675, 32, 4.40771, 32, 32, -32, 32)
```
wjt, line 42: I understand that the purpose of adding the second alternative is to have a "half-height" collision shape for the water at the bottom of the bridge.

However I think that your edit to the first alternative is wrong. I am pretty sure that the whole point of the first water alternative tile is that it has no collision shape, so it can be placed under bridges. I have checked and this tile is the one that is under the bridges in the stealth quest, and sure enough they aren't walkable after this change.

wjt:
I am addressing some issues in this PR. Converting to draft until it is ready.

scenes/tileset.tres
```diff
@@ -29,10 +29,13 @@ tile_data = PackedInt32Array(0, 327681, 3, 1, 393217, 3, 2, 458753, 3)
...
 texture = ExtResource("1_glnwb")
 texture_region_size = Vector2i(64, 64)
-0:0/next_alternative_id = 2
+0:0/next_alternative_id = 3
 0:0/0 = 0
 0:0/0/z_index = -1
 0:0/0/physics_layer_0/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
+0:0/2 = 2
+0:0/2/z_index = -1
+0:0/2/physics_layer_0/polygon_0/points = PackedVector2Array(-32, 3.85675, 32, 4.40771, 32, 32, -32, 32)
```
wjt, line 38: Argh, this conflicts with d4c8aa459fe865812e4d3ee287c64eca5681563d.

scenes/tileset.tres
```diff
@@ -291,45 +291,35 @@ texture_region_size = Vector2i(64, 64)
 texture = ExtResource("4_qml6a")
 texture_region_size = Vector2i(64, 64)
 0:0/0 = 0
-0:0/0/z_index = -1
```
wjt, line 291: @JuanFdS in commit dd6ba63b6c956dd917cfa6d53bff1ee1de7fc0d4 you set this z-index on bridge tiles. Do you remember why?

Here's a screenshot from the stealth level with this z_index (main) and without (this PR). I think it looks better without the z_index as proposed here.

## z_index = -1 (current `main` branch)

image


## No z_index (proposed change)

image

scenes/tileset.tres
```diff
@@ -29,10 +29,13 @@ tile_data = PackedInt32Array(0, 327681, 3, 1, 393217, 3, 2, 458753, 3)
...
 texture = ExtResource("1_glnwb")
 texture_region_size = Vector2i(64, 64)
-0:0/next_alternative_id = 2
+0:0/next_alternative_id = 3
 0:0/0 = 0
 0:0/0/z_index = -1
 0:0/0/physics_layer_2/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
+0:0/2 = 2
+0:0/2/z_index = -1
+0:0/2/physics_layer_0/polygon_0/points = PackedVector2Array(-32, 3.85675, 32, 4.40771, 32, 32, -32, 32)
```
wjt, line 38: @JuanFdS Following your change in commit d4c8aa459fe865812e4d3ee287c64eca5681563d, should this half-collidable water tile also use physics layer 2, in case it gets used in a level with ink blobs?

wjt (approved):
I have reviewed this PR as part of splitting it into more self-contained commits, and explained each change to the tileset (which was my main question). With the exception of perhaps changing the physics layer on the new water alternative, I think this is good to go.

scenes/tileset.tres
```diff
@@ -29,12 +29,17 @@ tile_data = PackedInt32Array(0, 327681, 3, 1, 393217, 3, 2, 458753, 3)
 [sub_resource type="TileSetAtlasSource" id="TileSetAtlasSource_sugp2"]
 texture = ExtResource("1_glnwb")
 texture_region_size = Vector2i(64, 64)
-0:0/next_alternative_id = 2
+0:0/next_alternative_id = 3
 0:0/0 = 0
 0:0/0/z_index = -1
 0:0/0/physics_layer_0/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
+0:0/0/physics_layer_1/polygon_0/points = PackedVector2Array(-32, -32, 32, -32, 32, 32, -32, 32)
```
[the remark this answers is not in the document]
wjt: In-person conversation confirmed that this was a mistake.

scenes/tileset.tres
wjt: In-person conversation confirmed that this was a mistake, and I've fixed this.

wjt:
I'm happy with this branch. I'll land it, and if it turns out the `z_index = -1` that this PR removes was intentional we can follow up.

PR #272, opened 2025-04-22
Add writing sound to dialogue

### LINKED ISSUE #210: Placeholder Sounds for Intro Scene
I found a couple of sounds with a CC0 license to use as placeholders in the intro scene — the one with the solid image of the woven world map gradually losing color.

One is a pencil writing on what seems to be wood, which sounds really good. ⬇️
https://freesound.org/people/Anthousai/sounds/337085/

The other is metal knitting needles, which makes more sense with the game's plot, but I don’t like it as much. ⬇️
https://freesound.org/people/HanulSkyGirl/sounds/505486/ 

It might be a good idea to test how they work in the scene until we have the final music. My only concern is that if we use these sounds for every single dialogue line, it might become annoying or tiring.

Would be great to test both options and see how it feels in context.

Arebuayon:
This adds a subtle sound effect of a pencil on paper that is played while dialogue is being "written" onto the balloon.

Fixes #210


manuq (changes requested):
@Arebuayon thanks for working on this! Here we tested the branch, discussed and decided that:
- Put the intro pencil writing sound in every dialogue, even when the player talks to NPCs. The voice sounds too robotic.
- Can you pause/resume the pencil writing instead of starting it every time the dialogue line changes? The track is long but since it starts always from the beginning, it currently sounds a bit monotonous.

manuq:
I'm getting this crash testing the PR, I wonder if it's because of the spaces in the `"res://" path.
Captura desde 2025-04-24 15-56-55

Arebuayon:
> I'm getting this crash testing the PR, I wonder if it's because of the spaces in the `"res://" path. Captura desde 2025-04-24 15-56-55

Thanks! I’ve updated the path to remove the spaces, try if its working now

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -211,4 +222,40 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+
+
+func _on_spoke(_letter: String, _index: int, _speed: float) -> void:
+	if _has_played_sfx:
+		return
+
+	_has_played_sfx = true
+
+	if talk_sound_player.stream != SFX_MAP["pencil"]:
+		talk_sound_player.stream = SFX_MAP["pencil"]
```
manuq, line 236: There is a single stream (the pencil SFX), so it makes more sense to put it directly in the `TalkSoundPlayer` node.

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -211,4 +222,40 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+
+	talk_sound_player.pitch_scale = 1.0
+	talk_sound_player.volume_db = 2
```
manuq, line 239: We are not changing the pitch or the volume, please just set it in the `TalkSoundPlayer` node and remove these lines.

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -211,4 +222,40 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+	talk_sound_player.play()
+	talk_sound_player.seek(_paused_position)
```
manuq, line 241: `play()` accepts a `from_position` parameter. So this can be simplified:
```suggestion
	talk_sound_player.play(_paused_position)
```

scenes/ui_elements/dialogue/balloon.tscn
```diff
@@ -56,5 +56,7 @@ response_template = NodePath("ResponseExample")
 layout_mode = 2
 text = "Response example"
 
+[node name="TalkSoundPlayer" type="AudioStreamPlayer" parent="."]
```
manuq, line 59: The bus of this AudioStreamPlayer is "master" but there exists 2 other buses: Music and SFX. I guess we want this in the SFX bus.

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -211,4 +222,40 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+
+
+func _on_dialogue_finished_typing() -> void:
+	if talk_sound_player.playing:
+		_paused_position = talk_sound_player.get_playback_position()
+		talk_sound_player.stop()
+	else:
+		_paused_position = 0.0
```
manuq, line 249: I think the `else` will never happen, because you have a callback `_on_talk_sound_finished` that automatically plays the SFX when finished, basically creating a custom loop.

This brings me a question, can the stream be set to loop and then you just pause/resume instead of using a separate variable `_paused_position`?

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -211,4 +222,40 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+
+
+func _on_new_dialogue_line():
+	_has_played_sfx = false
+	talk_sound_player.pitch_scale = 1.0
```
manuq, line 254: Please remove this line, the pitch doesn't change.

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -211,4 +222,40 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+	talk_sound_player.stop()
+
+
+func _on_talk_sound_finished() -> void:
+	if dialogue_label.is_typing:
+		talk_sound_player.seek(0)
+		talk_sound_player.play()
```
manuq, line 261: The seek is not needed, because `play()` has `from_position` parameter that defaults to 0.0.

manuq (changes requested):
I tested it and this is what we want. Some changes requested.

scenes/ui_elements/dialogue/components/sounds/voice.ogg.import
```diff
@@ -0,0 +1,19 @@
+[remap]
+
+importer="oggvorbisstr"
+type="AudioStreamOggVorbis"
+uid="uid://b7u2a347rljyt"
+path="res://.godot/imported/voice.ogg-428062faaec10b1bc73a91763dfb06ad.oggvorbisstr"
```
wjt, line 6: Can you remove the unused voice audio?

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -222,4 +227,15 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
...
+
+
+func _on_spoke(_letter: String, _index: int, _speed: float) -> void:
+	if talk_sound_player.playing:
+		return
+	talk_sound_player.play()
+
+
+func _on_dialogue_finished_typing() -> void:
+	talk_sound_player.stream_paused = true
```
wjt, line 241: I don't think you need to connect to these signals: start the stream playing in `apply_dialogue_line` in this block:

```
	if not dialogue_line.text.is_empty():
        # resume audio here
		dialogue_label.type_out()
		await dialogue_label.finished_typing
        # pause it here
```

wjt (changes requested):

scenes/ui_elements/dialogue/components/balloon.gd
wjt: The reason to connect to the `spoke` signal would be if you were to also connect to the `paused_typing` signal and pause the sound during the pause. If I understand correctly the signal to indicate that dialogue has resumed in that case is `spoke`.

wjt (approved):
Very nice!

PR #273, opened 2025-04-22
ci: Notify Discord of PRs merged to main via webhook

cassidyjames:
We want to highlight work that gets merged into the game in the Discord so we can celebrate learner achievements. This also provides a lower-noise feed for anyone interested in following along with Threadbare development beyond just releases.

GitHub's built-in webhooks are not granular enough for our needs; they deliver _all_ PR-related events which causes too much noise. Instead, this workflow uses a GitHub Action to deliver a notification to Discord via webhook only when a pull request is merged into main.

- Requires a `DISCORD_WEBHOOK_MERGE` repository secret
- Uses https://github.com/marketplace/actions/actions-for-discord
- Docs: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows

### DIFF
--- a/.github/workflows/discord.yml
+++ b/.github/workflows/discord.yml
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 name: Discord Notification
 
-on: 
+on:
   pull_request:
     types:
       - closed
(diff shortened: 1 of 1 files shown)


cassidyjames:
I've added the requisite `DISCORD_WEBHOOK` repository secret.

cassidyjames:
Merging since it only touches a new GitHub CI workflow and passes tests.

PR #288, opened 2025-04-23
Fix combat next scene

PlayMatters:
The round 1 telepor had a UID different than the round 2.


scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1.tscn
```diff
@@ -109,7 +109,7 @@ position = Vector2(0, 65)
 position = Vector2(0, -25)
 collision_layer = 4
 script = ExtResource("5_lmnu5")
-scene_to_go_to = "uid://cgemeabsewy1f"
+scene_to_go_to = "uid://bj1lfa32hmtqi"
```
wjt, line 112: I think this change is wrong. The current UID (`cge...`) is the correct one for round 2:

```
wjt@camille:threadbare (main)$ head -n1 scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_2.tscn
[gd_scene load_steps=11 format=4 uid="uid://cgemeabsewy1f"]
```

`bj1lfa32hmtqi` does not appear in the source tree in `main`.

wjt (changes requested):

wjt:
I think I hit the same issue just now with the Lore Quest Elder not taking me to the music puzzle but instead back to Fray's End. When I opened frays_end.tscn in the editor, it complained about a broken reference to the music puzzle. Then when I fixed that, and opened the music puzzle scene, it changed the music puzzle scene's UID!

After quitting the editor, resetting the git tree (but not erasing `.godot`), and reopening the project in the editor, everything worked correctly. And the web build works fine.

So I think this is caused by an editor bug when a file is renamed behind its back while the project is opened.

PR #290, opened 2025-04-23
edit storyquest starter dialogue

hydrolet:
Adjust the ending to inform the player that they will become a different character


wjt (approved):
Nice improvement!

PR #296, opened 2025-04-23
Fix collectible that starts revealed

manuq:
In 77834b0b the collectible animations were improved, and the sprite starts with a transparent modulate color to avoid a flicker effect. But the result of this is that if the collectible starts already revealed (which is the case in the stealth game), the collectible is transparent forever because the revealed animation doesn't play.

So in addition to set visibility when already revealed at the start, set modulate color to white.

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -59,6 +59,7 @@ func _get_configuration_warnings() -> PackedStringArray:
 
 func _ready() -> void:
 	_update_based_on_revealed()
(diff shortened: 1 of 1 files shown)


wjt (changes requested):
This fixes the initially-revealed collectible, but at the cost adding a glitch at the start of the animation when an initially-hidden collectible is revealed:

[revealed-collectible-regression.webm](https://github.com/user-attachments/assets/17e4f056-85f4-4ac7-9d83-c347103c3f46)

I think this is because `reveal()` first sets `revealed = true` which triggers this new code which sets `modulate` to `WHITE`. Then, the `reveal` animation is played; in frame 0 of that animation, modulate is set back to `TRANSPARENT` and then animated to `WHITE`.

Issue #283, opened 2025-04-23
Iterate on round 1 of combat


wjt:

manuq:
@wjt this is the issue for round one, once you have a branch from Stephen's content changes.

PR #299, opened 2025-04-23
Ink combat: Improve round 1

wjt:
Add an entrance and exit to the scene.

Add more scenery to give more of a sense of this being part of a larger space.

Fixes https://github.com/endlessm/threadbare/issues/283


wjt:
This is actually @PlayMatters' work – I pushed it because I was fixing a tricky rebase conflict. I removed an unnecessary rename of the TileMapLayers node to TileMapLayers2.

Issue #284, opened 2025-04-23
Adjust elder dialogue when returning to Fray's End after completing a quest


wjt:
When player gets back to Fray's End after completing a quest, dialogue with elders needs to reflect this state and direct the player to interact with the loom.

PR #304, opened 2025-04-23
Direct StoryWeaver to Loom after completing quest

wjt:
Previously, after completing a quest and returning to Fray's End, it was
possible to talk to an elder and start a new quest before taking the
threads you have just collected to the Eternal Loom.

Add new dialogue for both elders in this case. Update the misnamed
story_quest_starter.gd script (which in practice is actually the elder
script) to detect when the loom is ready for an offering and play the
alternative dialogue in this case.

Change the default _previous_look_at_side to UNSPECIFIED so that this is
preserved on the elder even though we have overridden the method that
would set it at the start of the dialogue.

Fixes https://github.com/endlessm/threadbare/issues/284


wjt:
Setting this as draft because it will conflict with @manuq's branch at #302. It's ready for review. I'll refresh it once that branch lands.

wjt:
Ready for review!

JuanFdS (approved):
Great 👌🏾 
I just have a question, what happens with the sokoban then? 🤔 , will it be post MVP?

manuq:
> Great 👌🏾 I just have a question, what happens with the sokoban then? 🤔 , will it be post MVP?

Yes, it will be reintroduced when ready.

PR #306, opened 2025-04-24
Revise second ink combat challenge

PlayMatters:
The main change is to add doors to enter and leave the stage.


manuq:
After conversation with Joana, I changed the red to be magenta (print colors). I did not touch the other changes from Stephen.

PR #313, opened 2025-04-24
Add wobble hint in music puzzle

Arebuayon:


wjt (changes requested):
I think the wobble should not be continuous, just a brief wobble (as if the rock is played) which repeats periodically.

Talking to the musician should stop the hints.

Rather than pushing a second commit to fix linter warnings, please rebase to squash it into the previous commit.

wjt (approved):
This looks good to me.

@pablitar what do you think? Can you reconcile this with your megabranch?

PR #314, opened 2025-04-24
create a very simplified stealth maze as learner template scene 

jbourqueendless:
I am adding a bare bones stealth level with very limited numbers of stealth maze components for them to build their own level off of. This would be an additional resource for them to build their storyquests with.


scenes/quests/story_quests/stealth_template/stealth_template.tscn
```diff
@@ -0,0 +1,129 @@
...
+color = Color(0, 0, 0, 0)
+metadata/_edit_lock_ = true
+
+[node name="ScreenOverlay" type="CanvasLayer" parent="."]
+
+[node name="HUD" parent="." instance=ExtResource("7_bqwsu")]
+
+[node name="CollectibleItem" parent="." instance=ExtResource("8_wu3if")]
+position = Vector2(2335, 955)
+scene_to_go_to = "uid://cd2p2udmkif80"
```
wjt, line 128: @jbourqueendless This is the wrong target scene. I suggest that you either link it to nothing (until all 3 templates are landed and we can string the 5 scenes together), or to the outro.

wjt (changes requested):
Please rebase onto main and move into the new template directory that @hydrolet added in #305

wjt (approved):
LGTM!

PR #315, opened 2025-04-24
Add more hints for music puzzle

Arebuayon:


wjt (approved):
I love it. The hint progression is just what I imagined and I really like the text you have written. I revised the dialogue to remove the `PLACEHOLDER` notes, and fix a couple of typographical things (e.g. the font we are using doesn't have the ellipsis `…` glyph).

PR #327, opened 2025-04-25
Update to Godot Dialogue Manager 3.6.1

wjt:
We previously had 3.5.1 in the tree. Release notes for the updates in this
commit:

- https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v3.6.1
- https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v3.6.0


wjt:
There are a few nice improvements here -- but the primary reason I propose this change is because I want to patch GDM to stop updating the list of translatable files in the project file: https://github.com/nathanhoad/godot_dialogue_manager/issues/887

PR #328, opened 2025-04-25
Don't add dialogue to translatable files

wjt:
Up till now, adding or removing dialogue files has caused project.godot to be
updated. This has repeatedly caused conflicts even with our small team working
on the game, and would lead to chaos when learners start adding their own
dialogue and we try to merge their changes.

The game is not currently localized, and you could imagine that we might not
want to localize all learners' dialogue files, only the core game and LoreQuest
dialogue.

Using a new feature I patched into the addon (and sent upstream), disable this
behaviour, and delete the existing list from project.godot.


wjt:
Stacked on #327

manuq (approved):

PR #330, opened 2025-04-25
Join bonfire and sign in a unique entity that also has clearer feedback when it has been turned on

pablitar:
A new entity called BonfireSign has been added. This is because the bonfire and the clues are matched 1:1, so it makes sense to have them be the same entity.

The BonfireSign can be configured with different signs, currently 4, but it could be expanded for more.

image


wjt:
Just looking at the screenshot: I love it.

scenes/quests/lore_quests/quest_001/2_ink_combat/components/ink_combat_example.tscn
```diff
@@ -1,7 +1,7 @@
 [gd_scene load_steps=12 format=4 uid="uid://512qlyin7f7m"]
 
 [ext_resource type="Script" uid="uid://dx4aqh50o2kro" path="res://scenes/quests/lore_quests/quest_001/2_ink_combat/components/ink_combat_example.gd" id="1_3llvl"]
-[ext_resource type="Script" uid="uid://cp54mgi54nywo" path="res://scenes/quests/lore_quests/quest_001/2_ink_combat/components/fill_game_logic.gd" id="3_gntf4"]
+[ext_resource type="Script" path="res://scenes/quests/lore_quests/quest_001/2_ink_combat/components/fill_game_logic.gd" id="3_gntf4"]
```
wjt, line 4: I don't think you meant to commit this and the hunk below here.

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/bonfire_sign/bonfire_sign.gd
```diff
@@ -1,22 +1,48 @@
...
 func ignite() -> void:
+	_is_ignited = true
 	fire.play(&"burning")
 	interact_area.disabled = false
+	update_sign_sprite()
 
 
 func _on_interact_area_interaction_started(_player: Player, _from_right: bool) -> void:
 	await puzzle.play_demo_melody_of_fire(self)
 	interact_area.end_interaction()
```
wjt, line 39: This signal became disconnected from the InteractArea so interacting with a lit fire breaks the game.

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/bonfire_sign/bonfire_sign.tscn
```diff
@@ -0,0 +1,182 @@
...
+animation = &"forward_off"
+
+[node name="Fire" type="AnimatedSprite2D" parent="."]
+unique_name_in_owner = true
+position = Vector2(0, -88)
+sprite_frames = SubResource("SpriteFrames_p0tgm")
+
+[node name="CollisionShape2D" type="CollisionShape2D" parent="."]
+scale = Vector2(1.4, 1)
+shape = SubResource("CircleShape2D_kbib7")
```
wjt, line 172: I think this collision area might be a bit small – I can walk through the edge of the sign. But I couldn't get stuck, so maybe it's OK.

wjt (changes requested):
Conceptually I think this makes much more sense than the symbols being separate to the bonfires, and I like how your hint for the 3rd tune has both the "backwards" and "+1" indicated clearly.

I noticed a couple of minor issues I noticed while testing this.

scenes/quests/lore_quests/quest_001/2_ink_combat/components/ink_combat_example.tscn
pablitar: Sorry about that!

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/bonfire_sign/bonfire_sign.gd
pablitar: oh

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/bonfire_sign/bonfire_sign.tscn
pablitar: The idea is that only the base collides. You can go "behind" the sign, and that's OK. Do you think I should change that?

pablitar:
Changes addressed! Merging!

Issue #339, opened 2025-04-25
Pause player input during cinematic


wjt:
@jbourqueendless is using the Cinematic node in a stealth quest template to show some instructions to the level designer. It's a bit outside what the node what was intended for, but it does work.

However, Cinematic doesn't pause player input while the dialogue is running, so the player can be moved around while the dialogue is playing.

PR #340, opened 2025-04-25
Cinematic: Pause player input

wjt:
A forthcoming stealth maze template uses a Cinematic to trigger some dialogue at the start of the scene. However, currently player input is not paused during cinematics.

Pause player input in this case.

Fixes https://github.com/endlessm/threadbare/issues/339

### DIFF
--- a/scenes/ui_elements/cinematic/cinematic.gd
+++ b/scenes/ui_elements/cinematic/cinematic.gd
@@ -23,6 +23,7 @@ extends Node2D
 
 
 func _ready() -> void:
+	Pause.pause_system(Pause.System.PLAYER_INPUT, self)
 	DialogueManager.show_dialogue_balloon(dialogue, "", [self])
 	await DialogueManager.dialogue_ended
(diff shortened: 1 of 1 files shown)


wjt:
The other way to fix this would be to pause player input in the dialogue balloon code. I guess that probably makes more sense: we use the same keystrokes/buttons to control dialogue as to control the player.

wjt:
Or – handle player input in `_unhandled_input()`. I think this is what the Godot Dialogue Manager author recommends.

JuanFdS:
Just a note, iirc same happens in the ink combat!

> Or – handle player input in `_unhandled_input()`. I think this is what the Godot Dialogue Manager author recommends.

I think this would be the way to go, so if something consumes the input, that's it.

wjt:
I have a patch that I've not pushed yet that does that change. It seems to work but not tested too thoroughly

wjt:
Superseded by #346

PR #343, opened 2025-04-25
Upload dialogue import files that were automatically updated

JuanFdS:
I think this was caused by uploading dialogue manager's version https://github.com/endlessm/threadbare/pull/327


wjt (approved):
Yes, I saw this too. I hoped it would only happen when the dialogue files were edited but then I found it was happening apparently spontaneously... Thanks for sorting this out!

PR #344, opened 2025-04-25
Background music

JuanFdS:


assets/music/Threadbare_Main_Full_Loud.ogg.license
```diff
@@ -0,0 +1,2 @@
+SPDX-FileCopyrightText: The Threadbare Authors
+SPDX-License-Identifier: CC-BY-SA-4.0
```
wjt, line 2: I would instead add this directory to `REUSE.toml` at the top level, along with the other first-party assets.

assets/music/Threadbare_Main_Full_Quiet.ogg.import
```diff
@@ -0,0 +1,19 @@
+[remap]
+
+importer="oggvorbisstr"
+type="AudioStreamOggVorbis"
```
wjt, line 4: What parameters did you use to convert them to Ogg Vorbis?

scenes/world_map/frays_end.tscn
```diff
@@ -60,7 +62,7 @@ noise = SubResource("FastNoiseLite_shbkg")
 resource_local_to_scene = true
 shader = ExtResource("34_opcsp")
 shader_parameter/cloud_texture = SubResource("NoiseTexture2D_thm8h")
-shader_parameter/offset = Vector2(58.1011, 29.0506)
+shader_parameter/offset = Vector2(411.066, 205.533)
```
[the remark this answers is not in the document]
wjt: Is it because it has `resource_local_to_scene`?

wjt (approved):
I like this approach of defining our own `BackgroundMusic` node so that we can make it do the necessary magic to preserve music between scenes.

Conflicts need resolution. We may want to tweak which soundtracks are used in which scenes but we can come to that in due course.

assets/music/Threadbare_Main_Full_Quiet.ogg.import
JuanFdS: Oh, I just opened them in Audacity and exported as .ogg, I don't know much about this, what do you recommend?

scenes/world_map/frays_end.tscn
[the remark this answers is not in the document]
wjt: Addressed in #362 – PR needs to be updated to remove this hunk.

assets/music/Threadbare_Main_Full_Quiet.ogg.import
wjt: That's probably fine, I was just curious what the quality settings were. We can always revise it.

JuanFdS: The settings I used were the default:
- Sample Rate: 44100hz
- Quality: 5

Issue #103, opened 2025-04-03
Handle interact "cooldown" in a reliable way


JuanFdS:
When a player interaction triggers dialogue, we had an issue:
The interact key is the same as the key to go through the messages (which is fine) and when the player pressed that key to _finish_ the dialogue, it started all over again because the input was read again by the PlayerInteraction node, triggering a new interaction and so on.

The way it's fixed right now is by having a timer after the dialogue ends that completes the interaction:

https://github.com/endlessm/threadbare/blob/0a01ec0fe764ec882aab0f8be3d6099009c1269d/scenes/npcs/talker/talker.gd#L31-L35

I think there are 2 things that could be improved:
- Instead of relying on timers, rely on godot's input handling system. I think if we want learners to copy some of the patterns in the code, implementing it in such a way that doesn't depend on timing would be better.
- If possible, have the logic that avoids the "double interaction" in `player_interaction.gd`. Otherwise, we'd need to repeat that code everytime we have interactions that trigger dialogue.

wjt:
GDM has an FAQ that [says this](https://github.com/nathanhoad/godot_dialogue_manager/blob/main/docs/FAQ.md#how-do-i-stop-my-player-from-moving-while-dialogue-is-showing):
> One of the most common causes is that you've implemented player movement inside _process instead of in _unhandled_input.

We indeed handle input in `_process`. GDM takes care to mark input as handled when it has consumed the event, so I think that would be the solution here – move player input handling to the `_unhandled_input` phase, and then no extra logic will be needed to avoid double-interactions because GDM will consume the input in `_input`.

PR #346, opened 2025-04-26
Player: Handle input in _unhandled_input() not _process() 

wjt:
Previously, player input (directions, interacting, and repelling ink
blobs) was all done in the _process() method of Player, PlayerFighting
and PlayerInteraction. This meant that we had to take special steps to
avoid processing input during and at the end of dialogue.

Godot Dialogue Manager takes care to mark all input events as handled
during dialogue, and documents in its FAQ that player input should be
handled in _unhandled_input() rather than _process().

https://github.com/nathanhoad/godot_dialogue_manager/blob/main/docs/FAQ.md#how-do-i-stop-my-player-from-moving-while-dialogue-is-showing

Handle player input in _unhandled_input(). Two cases are simple:

- Player: add an input_vector field, update it from _unhandled_input(),
and use it in _process()
- PlayerFighting; rename _process() to _unhandled_input().

PlayerInteraction is slightly more complicated because we still
need to update the visibility of the interaction label in _process(): if
you keep an arrow key held down, the player's position (and potential
interaction) will change without any new unhandled input. So the
visibility of the label continues to be managed in _process(), while the
interact action itself is triggered in _unhandled_input().

This change fixes an issue when a Cinematic node is used to trigger introductory
dialogue in a scene which also has an interactive Player, as we are doing in the
StoryQuest template's stealth template. Previously the player could be moved
during the introductory dialogue. Now it cannot.

It also allows us to remove some artificial short pauses after interactions and dialogue finish.

Fixes https://github.com/endlessm/threadbare/issues/339
Fixes https://github.com/endlessm/threadbare/issues/103

### DIFF
--- a/scenes/game_elements/characters/player/components/player.gd
+++ b/scenes/game_elements/characters/player/components/player.gd
@@ -37,8 +37,6 @@ const DEFAULT_SPRITE_FRAME: SpriteFrames = preload(
 @export var sprite_frames: SpriteFrames = DEFAULT_SPRITE_FRAME:
 	set = _set_sprite_frames
 
-var last_nonzero_axis: Vector2
-
 @onready var player_interaction: PlayerInteraction = %PlayerInteraction
 @onready var player_fighting: Node2D = %PlayerFighting
 @onready var player_sprite: AnimatedSprite2D = %PlayerSprite
@@ -87,9 +85,6 @@ func _process(delta: float) -> void:
 		return
 
 	var axis: Vector2 = %PlayerController.get_vector(&"ui_left", &"ui_right", &"ui_up", &"ui_down")
-
-	if not axis.is_zero_approx():
-		last_nonzero_axis = axis
 
 	var speed: float
 	if %PlayerController.is_action_pressed(&"running"):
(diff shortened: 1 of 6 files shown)


scenes/game_elements/characters/player/components/player_interaction.gd
```diff
@@ -20,21 +20,27 @@ func _get_is_interacting() -> bool:
...
+
+
+func _unhandled_input(_event: InputEvent) -> void:
+	if is_interacting:
 		return
 
-	if %PlayerController.is_action_just_pressed(&"ui_accept"):
+	var interact_area: InteractArea = interact_zone.get_interact_area()
+	if interact_area and %PlayerController.is_action_just_pressed(&"ui_accept"):
+		get_viewport().set_input_as_handled()
```
manuq, line 39: :+1:

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -37,6 +37,8 @@ const DEFAULT_SPRITE_FRAME: SpriteFrames = preload(
 @export var sprite_frames: SpriteFrames = DEFAULT_SPRITE_FRAME:
 	set = _set_sprite_frames
 
+var input_vector: Vector2
```
manuq, line 40: This variable name could have the private variable convention `_input_vector` (prepend underscore).

manuq (approved):

scenes/game_elements/characters/player/components/player.gd
wjt: I didn't do this because I imagined that, in future, this could be set from `PlayerController` (so that that node can be paused independently of the rest of the player) and because now it would be possible to animate the player's movement in cutscenes by adjusting `input_vector`.

PR #349, opened 2025-04-26
set outro character to template sprite

hydrolet:
Update the character in the outro scene to be the template sprite, so that learners can eventually replace with their own.


wjt (approved):

wjt:
Egg

wjt:
🥚

PR #350, opened 2025-04-27
Rework sokoban, add new levels, add new visuals.

PhoenixStroh:
A complete rework of the existing Sokoban. This commit does a few things from the original sokoban implementation:

1. Simplify the user interface and replace resources with nodes.

This meant changing out the use of resources in the inspector for rules and goals, to using node children and inheritance to create custom rules and goals. This removed a lot the nested resources and makes it easier/more intuitive how a sokoban is functioning.

This also meant removing some of the internal resources like sokoban_fragment.gd and turning them into inner classes inside pattern.gd.

Also the group resource was removed and replayed with ids and tags. Each piece has an id that uniquely identifies it, and a list of tags that it can share with other pieces. This then can be used in rules or goals for complex behavior.

2. Add Query class to Board2D

Instead of an increasing list of parameters for each query function for Board2D, I made a Query class that can be constructed and added to to create a match against pieces. This should be expandable if needed in the future.

3. Add go_to_scene on rule_engine.gd

When goals are completed, the rule engine will try to automatically load the next scene. Currently this is set up to do three increasingly difficult puzzles and then go back to fray's end.

4. Add Sokoban Dialogue to explain how to play

5. Add Undo and Reset abilities

The rule engine now remembers every step the player can take, and can undo steps or reset the puzzle entirely with "z" and "r".

6. Automatically assign layers to pieces

Pieces do not need to be assigned a layer anymore, and will automatically assume their layer based on which TileMapLayer it is a part of. This means that pieces that are manually placed into the scene will need to be placed as a child of a TileMapLayer to work properly.

7. Add new levels

There are currently 3 sets of levels, each with an A, B, and C section. The A section starts with just one thread, then B adds a second, and finally C has all three threads. After C, the level will go to fray's end. Currently, the only level set 2 is set to work from the base game, but that could be randomized in another pull request.

8. Add new visuals

There are now new visuals for every tile for the sokoban. On top of this, there is a shader to darken every other tile to make it easier to distinguish the different tiles. There is a new piece varient, the "spool" for each color that acts like a wall.

Example of the new visuals!
image


scenes/eternal_loom_sokoban/components/spools/spool_yellow.png.import
```diff
@@ -0,0 +1,34 @@
...
+type="CompressedTexture2D"
+uid="uid://ce6ehvjr3vs6q"
+path="res://.godot/imported/spool_yellow.png-ed5c85fb376cee6d61d7c44c160fdd16.ctex"
+metadata={
+"vram_texture": false
+}
+
+[deps]
+
+source_file="res://scenes/eternal_loom_sokoban/components/spools/spool_yellow.png"
```
wjt, line 13: Did @PlayMatters draw these spools?

scenes/eternal_loom_sokoban/components/goals/goal_all_on.gd
```diff
@@ -0,0 +1,25 @@
...
+
+	for board_target in board.get_pieces(target_query):
+		if board.is_empty(board_target.grid_position, key_query):
+			result = false
+
+	for child in get_children():
+		if child is Goal:
+			result = child.is_completed(board) and result
+
+	return result
```
wjt, line 25: This copy-pastes the base `Goal` class's implementation of this method. I would write:

```suggestion
	var key_query := Board2D.Query.new().add_id_or_tag(key, rule_engine)
	var target_query := Board2D.Query.new().add_id_or_tag(target, rule_engine)

	for board_target in board.get_pieces(target_query):
		if board.is_empty(board_target.grid_position, key_query):
			return false

	return super(board, rule_engine)
```

scenes/eternal_loom_sokoban/components/system/rules/goal.gd
```diff
@@ -0,0 +1,23 @@
...
+class_name Goal
+extends Node
+## A match pattern used in a Sokoban puzzle.
+##
+## A Goal will be checked after every turn
+## to see if there is a match. If so,
+## it will return true to the RuleEngine.
+
+# Virtual function, when called should return if goal is matched or failed
+@warning_ignore("unused_parameter")
```
wjt, line 13: There are no unused parameters so this shouldn't be necessary.

```suggestion
```

scenes/eternal_loom_sokoban/components/system/rules/goal.gd
```diff
@@ -0,0 +1,23 @@
...
+
+
+func is_completed(board: Board2D, rule_engine: RuleEngine) -> bool:
+	var result := true
+
+	for child in get_children():
+		if child is Goal:
+			result = child.is_completed(board, rule_engine) and result
+
+	return result
```
wjt, line 23: Personally I might write:

```suggestion
## Returns true if this goal, and all child goals (if any) are matched; false otherwise.
## Subclasses should override this method, chaining up with [code]super(board, rule_engine)[/code]
func is_completed(board: Board2D, rule_engine: RuleEngine) -> bool:
	for child in get_children():
		if child is Goal:
			if not child.is_completed(board, rule_engine):
				return false

	return true
```

As you have it is fine, but I think you could take the docstring.

You could also consider having the recursive step not be a virtual function (error-prone if a subclass fails to chain up), and instead have (sketch, names are bad):


```gdscript
## Don't override this
func is_completed(board, rule_engine) -> bool:
    if not this_is_completed(board, rule_engine):
        return false
    # existing loop here

## Override this in subclasses
func _this_is_completed(board, rule_engine) -> bool:
    return true
```

scenes/eternal_loom_sokoban/levels/sokoban_template.tscn
```diff
@@ -0,0 +1,56 @@
+[gd_scene load_steps=6 format=4 uid="uid://byvwl6rx7pyps"]
```
wjt, line 1: What's the difference between `sokoban_template`, `eternal_loom_template`, and `eternal_loom_sokoban`?

scenes/eternal_loom_sokoban/levels/sokoban_level_2b.tscn
```diff
@@ -0,0 +1,57 @@
+[gd_scene load_steps=5 format=4 uid="uid://c6qwiy01e6jqd"]
+
+[ext_resource type="Script" uid="uid://dppvw2f8yh4vu" path="res://scenes/eternal_loom_sokoban/components/system/board/board_2d.gd" id="1_k8gxf"]
+[ext_resource type="Material" uid="uid://jpl3v3f7elq6" path="res://scenes/eternal_loom_sokoban/components/floor/tiled_mat.tres" id="2_2kj74"]
+[ext_resource type="TileSet" uid="uid://dand23uvn70pg" path="res://scenes/eternal_loom_sokoban/components/eternal_loom_tileset.tres" id="2_p3sau"]
+[ext_resource type="PackedScene" uid="uid://btpeaqx2nur3q" path="res://scenes/eternal_loom_sokoban/components/eternal_loom_ruleset.tscn" id="4_vc6vr"]
+
+[node name="SokobanLevel2b" type="Node2D"]
```
wjt, line 8: I'm not very good at this game but, is this really solvable?!

image

The half-spools are barriers so I can't see how it's possible to get through to the left-hand side without permanently blocking the yellow thread goal.

wjt (changes requested):
I confess that I haven't read every line of code… this is pretty cool!

scenes/eternal_loom_sokoban/components/spools/spool_yellow.png.import
PhoenixStroh: No, these were drawn by Rita Brederson. I wasn't sure if in contracting for Endless OS the rights fell under Endless or Rita. All the assets for the sokoban except for the player was done by Rita Brederson.

PhoenixStroh: Should the license be set differently?

scenes/eternal_loom_sokoban/levels/sokoban_level_2b.tscn
PhoenixStroh: Good catch! I broke this puzzle while adding the threads! I'll fix that one.

scenes/eternal_loom_sokoban/levels/sokoban_template.tscn
PhoenixStroh: So currently there are two different rulesets, one for "classic sokoban" where it's just about getting every target filled with a crate (in our case clump). And then the eternal loom ruleset requires three different targets to be filled up. So they have two different templates in case someone wants to make a new level for the eternal loom, or just make a classic sokoban puzzle.

`eternal_loom_sokoban` was just the file that was linked to the main game, so I left that alone. That would either be where we put the "randomly choose level" feature, or we put that feature in fray's end on the eternal loom exterior scene and get rid of `eternal_loom_sokoban`.

wjt: FWIW it's not currently linked to the main game - I unlinked it because I wanted to test the case of entering a quest after completing a quest.

A to-be-answered question is, can a quest specify a sokoban puzzle (or even provide its own), or does the game specify one (or a random one)?

Personally I would put the code to pick a random puzzle into the loom's script.

PR #351, opened 2025-04-27
Revise keybindings for pausing & skipping dialogue

wjt:
Previously, Esc was the only keyboard binding for both skipping dialogue and pausing the game. This led to confusion (why does the same key perform two unrelated actions) and poor discoverability (at least one player tried P for pause, and another tried Tab for skip; neither tried Esc).

Introduce `dialogue_next` and `dialogue_skip` actions so that the bindings can be adjusted separately to the `ui_accept` and `ui_cancel` actions used for Controls.

Replace the `Esc` binding for `dialogue_skip` with `Tab`.

Add `P` and the `Pause` key as additional pause bindings.


scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -197,6 +197,7 @@ func _on_balloon_gui_input(event: InputEvent) -> void:
 		)
 		var skip_button_was_pressed: bool = event.is_action_pressed(skip_action)
 		if mouse_was_clicked or skip_button_was_pressed:
+			print("Skip was pressed")
```
manuq, line 200: Is this print line intentional?

scenes/menus/splash/components/splash.gd
```diff
@@ -16,7 +16,7 @@ func _ready() -> void:
 func _input(event: InputEvent) -> void:
 	if (
 		not Pause.is_paused(Pause.System.PLAYER_INPUT)
-		and (event.is_action_pressed(&"ui_accept") or event.is_action_pressed(&"ui_cancel"))
+		and (event.is_action_pressed(&"dialogue_next") or event.is_action_pressed(&"dialogue_skip"))
```
manuq, line 19: :+1:

manuq (approved):

scenes/ui_elements/dialogue/components/balloon.gd
wjt: Oops, no

wjt: The reason this was in here is that I couldn't figure out why `Esc` was still skipping dialogue even when I had changed the binding. The answer is that `dialogue_label.gd` also has code for this:

```gdscript
func _unhandled_input(event: InputEvent) -> void:
	# Note: this will no longer be reached if using Dialogue Manager > 2.32.2. To make skip handling
	# simpler (so all of mouse/keyboard/joypad are together) it is now the responsibility of the
	# dialogue balloon.
	if self.is_typing and visible_ratio < 1 and InputMap.has_action(skip_action) and event.is_action_pressed(skip_action):
		get_viewport().set_input_as_handled()
		skip_typing()
```

The comment is not true: when I had changed the skip action on the balloon but not the label, this code was reached when I triggered `ui_cancel`.

wjt: (As you will have seen, I also changed the skip action on the label.)

PR #352, opened 2025-04-27
Collect item: Move label to the top

manuq:
So the label doesn't interfere with the sprite image.

Move it above the sprite, just like the rocks in the music puzzle game.

Fix https://github.com/endlessm/threadbare/issues/338

Captura desde 2025-04-27 09-30-27


manuq:
Doing this I noticed that the label in the rocks of the music puzzle is also over the sprite, preventing the wobble visual hint to be seen. I'll move the rock label to the bottom for now, as the whole scene is being redesigned.

Captura desde 2025-04-27 09-41-20

wjt (approved):

Issue #276, opened 2025-04-22
An ink-drinker can attempt to shoot ink even when there are no ink wells, causing a debugger error


JuanFdS:
The issue can be seen at around 0:28

https://github.com/user-attachments/assets/e81284eb-c8bb-4a0a-a7e1-4fa438c9cd27

manuq:
This is easily replicated by removing all barrels of any combat scene and then play the scene.

manuq:
Turns out that there is another way to reproduce this error: when the barrels to win number in the game logic is bigger than the barrels added to the scene. In this case, the barrels are gone, so the array to pick a label is empty, but the game logic hasn't called `remove()` on the enemies (which correctly stops the timer to spawn projectiles).

PR #354, opened 2025-04-27
Fill game: Fix enemy throwing without target label

manuq:
If the is no target label to add to a projectile, simply don't throw.

Fix https://github.com/endlessm/threadbare/issues/276

### DIFF
--- a/scenes/quests/lore_quests/quest_001/2_ink_combat/components/fill_game_logic.gd
+++ b/scenes/quests/lore_quests/quest_001/2_ink_combat/components/fill_game_logic.gd
@@ -20,6 +20,11 @@ func start() -> void:
 	_update_allowed_colors()
 
 
+func _ready() -> void:
+	var filling_barrels: Array = get_tree().get_nodes_in_group("filling_barrels")
+	barrels_to_win = clampi(barrels_to_win, 0, filling_barrels.size())
+
+
 func _update_allowed_colors() -> void:
 	var allowed_labels: Array[String] = []
 	var color_per_label: Dictionary[String, Color]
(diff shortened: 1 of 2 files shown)


scenes/quests/lore_quests/quest_001/2_ink_combat/components/throwing_enemy/components/throwing_enemy.gd
```diff
@@ -180,6 +180,8 @@ func _on_timeout() -> void:
 	var player: Player = get_tree().get_first_node_in_group("player")
 	if not is_instance_valid(player):
 		return
+	if not allowed_labels:
+		return
 	_is_attacking = true
 	animated_sprite_2d.play(&"attack anticipation")
 	await animated_sprite_2d.animation_finished
```
wjt, line 187: I don't think this will work because a few lines later we `await` the end of the `attack anticipation` animation. By that point the `allowed_labels` may have changed, if the player filled the last inkpot during the anticipation animation.

You could move this check down:

```suggestion
	_is_attacking = true
	animated_sprite_2d.play(&"attack anticipation")
	await animated_sprite_2d.animation_finished
	if not allowed_labels:
		return
```

I think that would work.

wjt (changes requested):

scenes/quests/lore_quests/quest_001/2_ink_combat/components/throwing_enemy/components/throwing_enemy.gd
manuq: Oh yes, I just figure out that myself when trying to fix something else. The `allowed_labels` may be empty after the `await`. Thanks.

manuq:
Found another case: https://github.com/endlessm/threadbare/issues/276#issuecomment-2833514205

Ready for review again!

scenes/quests/lore_quests/quest_001/2_ink_combat/components/fill_game_logic.gd
```diff
@@ -20,6 +20,11 @@ func start() -> void:
 	_update_allowed_colors()
 
 
+func _ready() -> void:
+	var filling_barrels: Array = get_tree().get_nodes_in_group("filling_barrels")
+	barrels_to_win = clampi(barrels_to_win, 0, filling_barrels.size())
```
wjt, line 25: Do we want to emit `goal_reached` immediately (or perhaps deferred) in the case where `barrels_to_win` is `0`? Or log a warning?

wjt:
I'll merge this to fix the bug. I think warning about `barrels_to_win` being `0` would be a nice improvement but it's not a new issue!

Issue #291, opened 2025-04-23
Clarify focused/unfocused dialogue responses


wjt:
@hydrolet noted that, in dialogue with two response choices, it is hard to tell which one is highlighted:

![Image](https://github.com/user-attachments/assets/064177c0-d556-44ba-9a77-079592a96a81)

One of them has a golden rectangle around it that looks like a focus rectangle; the other has a colourful background. (For the record, the `Yes.` option is selected.)

wjt:
I'm implementing a different style of button for the main menu #33 where only the focused button has a background.

PR #357, opened 2025-04-27
Dialogue: Use flat buttons for response menu

wjt:
It has turned out that the previous button style makes it hard to tell which choice is focus when there are only 2 choices: is it the blue button with a grey border, or the dimmer yellow-brown button with a gold border?

Adjust the FlatButton style variation to show no background (actually a fully-transparent modulation of the same texture) when unfocused, and the flat sign texture when focused. Set the response buttons to use this.

Add a new FlatNextButton style, derived from FlatButton, that implements the previous FlatButton style of having a visible background even when unfocused, and set the "next >" button to use it. (This is because this button is not actually focused: the dialogue balloon itself grabs focus, for reasons I don't fully understand.)

I chose to define the styles this way around because I plan to use the same style of button where only the focused one has a visible background for the main menu.

Fixes https://github.com/endlessm/threadbare/issues/291


manuq:
Oh yes, much better!

Issue #277, opened 2025-04-22
Ink drinkers can't turn right


JuanFdS:
Currently, the side the ink drinkers are looking at is fixed.

It isn't terrible, but I think it'd be better if we could configure them to either always face the player, or to manually configure where they are facing.

![Image](https://github.com/user-attachments/assets/e198e4a0-79f5-46e3-94b1-bc7d579806da)

PR #358, opened 2025-04-27
Fill game: Face enemies toward player

manuq:
In two moments: when the enemy is ready, and when the enemy attacks.

The alternative is to do this every frame, but I think that's not needed.

Note: The way the enemy is flipped horizontally is by scaling the whole scene, not by using the AnimatedSprite2D flip_h property. This is because the marker used to place the spawned projectile needs to be flipped as well. Alternatively, we could store the initial marker position as the left position, and a mirrored marked position as the right position. Since scaling works and no regressions were found, I'd say let's try this instead.

Fix https://github.com/endlessm/threadbare/issues/277

### DIFF
--- a/scenes/quests/lore_quests/quest_001/2_ink_combat/components/throwing_enemy/components/throwing_enemy.gd
+++ b/scenes/quests/lore_quests/quest_001/2_ink_combat/components/throwing_enemy/components/throwing_enemy.gd
@@ -95,6 +95,12 @@ func _ready() -> void:
 	_initial_position = position
 	if Engine.is_editor_hint():
 		return
+	var player: Player = get_tree().get_first_node_in_group("player")
+	if is_instance_valid(player):
+		var direction: Vector2 = projectile_marker.global_position.direction_to(
+			player.global_position
+		)
+		scale.x = 1 if direction.x < 0 else -1
 	if autostart:
 		start()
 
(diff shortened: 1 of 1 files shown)


wjt (changes requested):
Unfortunately I found a regression in the third round of the game:

https://endlessm.github.io/threadbare/branches/endlessm/face-ink-drinker/#quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3

If you stand to the right of the InkDrinkers, then the projectiles from the top one get stuck on top of the rocks.

image

This is essentially the same thing discussed on https://github.com/endlessm/threadbare/pull/353#issuecomment-2833529234 . I think if we want to do this, we need to find a way to have a one-way collision shape that allows the projectiles to come "down" from the top of the rocks, but not get back "up" again.

manuq:
> Unfortunately I found a regression in the third round of the game:
> 
> https://endlessm.github.io/threadbare/branches/endlessm/face-ink-drinker/#quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3
> 

Oh thanks for testing with the actual rounds.

> This is essentially the same thing discussed on [#353 (comment)](https://github.com/endlessm/threadbare/pull/353#issuecomment-2833529234) . I think if we want to do this, we need to find a way to have a one-way collision shape that allows the projectiles to come "down" from the top of the rocks, but not get back "up" again.

Yeah, @JuanFdS and I discussed a bit more in https://github.com/endlessm/threadbare/pull/379 . I am heading towards: Adding characters in elevations is not supported, so we should redesign the level instead.

manuq:
We decided to hack the combat 3 so the player can collide with the elevation but not the ink projectiles (switching walls to water, for example). So that shouldn't be a blocker for this particular PR.

PR #363, opened 2025-04-28
Allow configuring Guards' spritesheets in the inspector

### LINKED ISSUE #341: Let learners change spriteframes for guards through godot's inspector

JuanFdS:
Fixes #341 

https://github.com/user-attachments/assets/876545e8-bb90-4063-a5cc-ef75246a93f2

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -20,7 +20,19 @@ enum State {
 	RETURNING,
 }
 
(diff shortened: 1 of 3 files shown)


wjt (approved):
This looks fine to me. I guess the reason you are not using https://github.com/endlessm/threadbare/blob/main/scenes/globals/property_utils.gd to export the child AnimatedSprite2D's sprite_frames property in a generic way is that (a) it's not much code; (b) you want to check that the expected animations are present?

PR #364, opened 2025-04-28
Projectiles: avoid collisions with walls if they spawn in an elevated terrain

JuanFdS:
The collision with walls is turned back on once they left that elevated terrain.

This is an attempt to fix the issue mentioned here:
https://github.com/endlessm/threadbare/pull/358#pullrequestreview-2798689464

> 
> image
>
> This is essentially the same thing discussed on https://github.com/endlessm/threadbare/pull/353#issuecomment-2833529234 . I think if we want to do this, we need to find a way to have a one-way collision shape that allows the projectiles to come "down" from the top of the rocks, but not get back "up" again.

## Other possible solutions

- One way collisions: It can't be configured **_which_** way the collision should go :(, at least not at the moment. https://github.com/godotengine/godot-proposals/issues/7483
- Instead of using the rigid body's collision, use a raycast that detects walls, then, we can handle that by making the blob bounce (as it does when it's hit by an attack). This would give us more control about when it collides and when it doesn't, and also **_what it does_** when it collides. I've just thought of this option when writing this PR, and I'm starting to think this one might be a better solution than the one in this PR 🤔 
- Have the ink blob not hit any walls for a small amount of time when it's spawned. This could go well with a spawn effect that makes it look like the blob was thrown into the air for some frames (I guess we can achieve it by scaling it up and then back to its regular size). However, this has some drawbacks: If the projectile spawned in an elevated terrain but a bit far from the walls, this won't work. Also, if there are 2 elevated terrains near each other, the projectile might enter into the other one.

## This solution

Pro: it is working ✅, at least for ink combat's third round.
Cons:
- it's a bit complex.
- it requires to either scan through all nodes to find tile map layers, or have some setup like groups to find the map layers that we want to check against.
- it doesn't support multiple levels of elevation.
- it turns off ALL wall collisions, so, if for example we had walls or houses in an elevated terrain, the blob would not collide with them.


JuanFdS:
I think we should (and I will try to) explore the raycast alternative, but sharing in case this is useful for now @wjt @manuq

PR #365, opened 2025-04-28
Transitions: Hide by default

wjt:
Previously, the Transitions scene (a CanvasLayer with a ColorRect covering the viewport) was initially visible, although the ColorRect itself has a material that makes it fully transparent. At the start of any transition, it is explicitly made visible, then hidden at the end of the transition.

This means that on the first scene played, an invisible CanvasLayer swallows all clicks on the scene; and so if you play an individual scene, mouse input does not work until you switch to another scene. It is desirable that clicks (and other input) are blocked during a transition, but not outside one.

Set the scene's visible property to false by default.


JuanFdS (approved):
👍🏾

PR #366, opened 2025-04-28
Update dialogue import files

wjt:
I noticed that two dialogue import files were not updated in #343, perhaps because they were merged after that PR.

One of the files was unused. Remove that one and update the other.


wjt:
I'm going to merge this without review because it's so annoying to have to keep reverting these changes. Post-merge feedback welcome of course.

PR #372, opened 2025-04-29
Music puzzle redesign

pablitar:


scenes/quests/lore_quests/quest_001/1_music_puzzle/components/music_puzzle.gd
```diff
@@ -26,6 +26,12 @@ var _is_demo: bool = false
 func _ready() -> void:
 	if not Engine.is_editor_hint():
 		xylophone.note_played.connect(_on_note_played)
+		
+	for i in fires.size():
+		if fires[i].is_ignited:
+			_current_melody = i + 1
+			_position = 0
+			break
```
wjt, line 34: If there are 4 fires and the first 2 are ignited, this `break` is wrong.

wjt (commented):
Personally I think this is a really good idea: different people teaching you the different symbols and the base melodyy.

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/music_puzzle.gd
pablitar: Fixed!

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/trees.gd
```diff
@@ -0,0 +1,15 @@
...
+const WOOL_PERCENTAGE: float = 0.4
+
+const TREE_WOOL_GREEN_01 = preload(
+	"res://scenes/game_elements/props/tree/components/Tree_Wool_Green_01.png"
+)
+const TREE = preload("res://scenes/game_elements/props/tree/components/Tree.png")
+
+@export_tool_button("Randomize Trees Appearances") var a = func():
+	for child in get_children():
+		child.set("tree__texture", TREE if randf() > WOOL_PERCENTAGE else TREE_WOOL_GREEN_01)
```
manuq, line 15: This works for now. If we add more weighted choices it would be better to use [RandomNumberGenerator.rand_weighted()](https://docs.godotengine.org/en/stable/tutorials/math/random_number_generation.html#weighted-random-probability). Example:

```
var fruits = ["apple", "orange", "pear", "banana"]
var probabilities = [0.5, 1, 1, 2];

var random = RandomNumberGenerator.new()
print(fruits[random.rand_weighted(probabilities)])
```

JuanFdS:
I think the changes are great!, I have just one concern: there are 3 melodies in the level:
- The Magical Melody (the one that when solved gets you to the next level)
- The River Song
- The Forest Song

And 4 xylophones (2 for the magic melody, and one for each of the others).

I'm not sure how the player is supposed to realise that the song they have to play in the first xylophone they find is the same as in the second one, but not the same as the rest. The only thing that hinted me about that is the sign with the name of the melody, but I'm afraid some players might not notice that.

I'm wondering if we could add something else that links those 2 xylophones together. Maybe some visual cue like the color of the hint rocks, or how the rocks are placed in the xylophone, I'm not sure 🤔 .

Still, I think this is worth merging and if we need to make those changes later.

PR #373, opened 2025-04-29
Add title scene with main menu

### LINKED ISSUE #33: Main menu

wjt:
This screen shows a logo for the game, plus buttons to start the game or exit.

There are two hidden buttons for future options & credits scenes, which I added here to test the appearance of the menu with more than two buttons.

The splash scene is adjusted to switch to this scene rather than the intro: the title screen goes logically in between the two.

Fixes #33


JuanFdS (approved):
Looks good!,

Just 2 comments:
- What do you think about having the hover color on the button be a different one than the default?

https://github.com/user-attachments/assets/2172fa60-0302-4097-918d-4329035821db

- I just realised we have 2 themes (`res://scenes/ui_elements/dialogue/components/theme.tres` and `res://scenes/ui_elements/shared_components/theme.tres`) 👀 , I think that in a different PR we should probably join them into just one to avoid confusion.

wjt:
I'll add a hover state.

wjt:
White looks OK actually...

I have to say I don't understand why, in the dialogue balloon, the "next" button has a different stylebox for hover:

image
image

while on the main menu it does not:

image
image

PR #375, opened 2025-04-29
Intro: Fix unsightly line break before ellipsis

wjt:
Previously there was a regular (breaking) space character before the ellipsis at the end of the last line of the intro dialogue. Because of the dimensions of the dialogue balloon, the font size, and the text being displayed, it just so happens that a line break gets inserted between the word "Until" and the ellipsis. This looks bad.

My personal preference is not to use a space before an ellipsis when it indicates that the speaker is trailing off at the end of a sentence:

> Until...

However style guides vary considerably on this, and it seems more common in US English style guides to require a space before it.

Use a non-breaking space character between "Until" and "..." to avoid wrapping text in between the two. Use the BBCode tag for an arbitrary Unicode character to make the nature of the space visible. (Annoyingly, while Godot's BBCode has special tags for other invisible characters (e.g. `[zwnj]`), it does not have a `[nbsp]` tag.)


wjt:
Here's how it currently looks:

image

but annoyingly... on desktop `main` looks fine. And on the web, this change has no effect. OK, I give up, I have bigger fish to fry.

Issue #368, opened 2025-04-28
Game hangs when switching to large scenes (particularly Fray's End)


wjt:
Fray's End now takes several seconds to load (on my 2019 Dell XPS laptop when running in debug mode in the editor). This causes a visible pause at the end of the intro before the scene transition starts. (This is a regression introduced in commit 8f8ec378a9aa9609721b1c03e5f14ac80e258d7d because prior to that commit the intro scene referenced fray's end with `preload()`, but Cinematic (and Collectible) can't do this when there is a scene loop (fray's end -> challenge 1..3 -> fray's end).)

A quick fix is to move the `load()` call after the transition when using `change_to_file_with_transition`. This means the scene will be loaded at the blank point of the transition – there's still a pause but it feels better.

A less quick fix is to use `ResourceLoader.load_threaded_request` to start loading the scene at the start of the transition, but not wait for it to complete until the blank point. Since the transition is 1 second long, this should shave a good amount off the perceived loading time.

The best fix I can think of is for Cinematic (and other scene-switching nodes) to kick off the threaded request as soon as the target is known (in the case of Cinematic that's in _ready, in the case of the Elders it's when you choose the relevant dialogue response branch), so that by the time the transition runs the following scene has probably already been fully loaded.

wjt:
I have a WIP change to load the scene in the background so that it's ready when the switch actually arrives. It works quite nicely but it's a little fiddly.

wjt:
I did a little bit of experimentation and the slow part of loading Fray's End is generating the noise texture for the clouds.

wjt:
With #377 this is not so perceptible. I think I will close this off and open a separate task for making fray's end load more quickly – it should be possible to generate the clouds asynchronously.

PR #377, opened 2025-04-29
Load next scene asynchronously during transition

wjt:
Previously, the target scene was loaded in full before starting the fade-out transition.

Instead, start loading the scene asynchronously at this point, then start the transition. Only wait for the scene to be fully loaded at the blank point of the transition. This means the transition starts immediately, and if the scene takes longer than 1 second to load, the pause is at the blank point: a natural place for it.

Helps #368


JuanFdS (approved):
Great!

PR #379, opened 2025-04-29
Make projectile collisions with walls be one way

JuanFdS:
The collision with walls is turned back on once they left that elevated terrain.

This is an attempt to fix the issue mentioned here:
https://github.com/endlessm/threadbare/pull/358#pullrequestreview-2798689464

> 
> image
>
> This is essentially the same thing discussed on https://github.com/endlessm/threadbare/pull/353#issuecomment-2833529234 . I think if we want to do this, we need to find a way to have a one-way collision shape that allows the projectiles to come "down" from the top of the rocks, but not get back "up" again.

## How this works

Instead of handling collisions through the rigidbody, handle those collisions through raycast. That way, we can decide what to do when we detect that the raycast is colliding. This lets us set collision exceptions with some nodes (like the tilemap layers) if needed.
When a projectile spawns, it detects if any point in the vertices of its bounding box is in an elevated terrain tile and adds it to the exceptions list.
Then, every frame it checks if the projectile's bounding box has escaped that tilemap layer (no vertices are in any tile of the layer), and if that's the case, it erases the layer from the exceptions list.

## Other possible solutions (❌ were explored but couldn't make them work)

- ❌ One way collisions: It can't be configured **_which_** way the collision should go :(, at least not at the moment. https://github.com/godotengine/godot-proposals/issues/7483
- ❌Using `RigidBody.add_collision_exception_with`: that would be great for this use case but sadly it can't be used with objects that are not CollisionObject2D, and tilemap layers aren't: https://github.com/godotengine/godot/issues/17090
- Turn walls collision mask off and on depending on whether the projectile is on top of a tilemap layer with elevated terrain: https://github.com/endlessm/threadbare/pull/364 . I lean more toward the solution in this PR instead of that one because this one allows colliding with other things that are walls and just turns off collisions with the tilemap layer.
- Have the ink blob not hit any walls for a small amount of time when it's spawned. This could go well with a spawn effect that makes it look like the blob was thrown into the air for some frames (I guess we can achieve it by scaling it up and then back to its regular size). However, this has some drawbacks: If the projectile spawned in an elevated terrain but a bit far from the walls, this won't work. Also, if there are 2 elevated terrains near each other, the projectile might enter into the other one.


manuq:
@JuanFdS thank you so much for attempting so many ways to fix this!

I placed an enemy behind a wall but is not doing what I would expect:

[Grabación de pantalla desde 2025-04-29 16-48-58.webm](https://github.com/user-attachments/assets/cdd3d88f-5cc7-4c29-afa6-0eec72a43659)

I don't think there is a general way to fix this issue unless we add a "height" or "z" or "elevation" property to world objects. I have an idea for how to do it but it would be a major change. At this point I would rather change the level design than trying to fix this in a complex way. What do you think?

PR #383, opened 2025-04-29
Dialogue: Don't stretch scroll texture

wjt:
Previously, the background texture was set to "tile_fit" in both the horizontal and vertical axes. In the case where the width or height is not an exact multiple of the texture width/height (64px), this causes it to be stretched slightly so that the edges fit exactly to the border region.

Since the dialogue balloon resizes between lines of dialogue and each time the "next >" button is shown and hidden, this causes the background texture to visibly change constantly. I think it's ugly.

Instead, use "tile", which tiles the texture without deformation, even if that means truncating the last tile. I personally can't see the seam where it joins the outer section, and this means that the background never changes in a visible section of the balloon.

### DIFF
--- a/scenes/ui_elements/dialogue/components/theme.tres
+++ b/scenes/ui_elements/dialogue/components/theme.tres
@@ -115,8 +115,8 @@ texture_margin_left = 64.0
 texture_margin_top = 64.0
 texture_margin_right = 64.0
 texture_margin_bottom = 64.0
-axis_stretch_horizontal = 2
-axis_stretch_vertical = 2
+axis_stretch_horizontal = 1
+axis_stretch_vertical = 1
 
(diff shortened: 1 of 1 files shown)


wjt:
Before:

[Screencast from 2025-04-29 17-30-51.webm](https://github.com/user-attachments/assets/726db7ca-bd32-4b9c-b958-a9cdd66aeecf)

After:

[Screencast from 2025-04-29 17-33-32.webm](https://github.com/user-attachments/assets/a862f26a-fb8d-4c0c-ac58-883d54e8fa65)

It's subtle but I think it's better.

JuanFdS (approved):
Hadn't noticed it before but yes!, it is better this way

PR #384, opened 2025-04-29
Simplify blocking input during transitions

### LINKED ISSUE #374: Splash screen and transitions can be paused
Currently, if you hit Esc or Pause at the splash screen, during a scene transition, or at the main menu (#33), the game is paused. I think it should not be.

wjt:
During a scene transition, we want to stop processing player input. Previously, this was handled using a global Pause scene which was told to pause PLAYER_INPUT during a transition; and then every site in the game that processes player input has to manually check whether input is occurring. I have found this to be error-prone: for example, the game can be paused midway through a transition, and there have been several cases where we have forgotten to check the PLAYER_INPUT state when processing input.

Commits 58e81c9 and 6d1a909 solved this problem for player input during dialogue in a different way:

- Player input is processed in _unhandled_input();
- The dialogue UI marks all input as handled in its _input() implementation.

This suggested to me the approach implemented in this commit:

- In Transition, implement _input() and consume all input while a transition is active.

- In PauseOverlay (which implements pausing the whole game), check for the pause event in _unhandled_input(), so that pause is blocked during
  transitions.

- Similarly, on the splash screen, handle skipping in _unhandled_input() (so that trying to skip a second time during the fade-out has no effect).

- In the splash screen, treat the "pause" action as an additional way to skip to the next scene.

- In the main menu, remove workarounds to prevent the player from interacting with the menu after choosing "start", which are now unnecessary.

- Remove the Pause global and the now-unnecessary PlayerInteraction indirection that checked the PLAYER_INPUT pause state.

The one place where removing the PLAYER_INPUT pause state leaves things worse is the main menu. For the buttons on the menu to be usable with the keyboard or joypad, we need to make the default button grab focus. However, it is misleading to focus the button during the fade-in transition, as the game will not actually accept input from the player in this case. To fix this, add a method & signals to the transitions node so that the main menu can check for a running transition and wait for it to end. This is something we have wanted before: the intro scene currently hardcodes a 1-second pause to wait for the transition to finish.

Fixes #374


wjt:
This effectively reverts #89. We might live to regret this! but my feeling (as someone who has not shipped a large Godot game before...) is that the pause system as it stands does not solve the problem that we have (player input being processed when it shouldn't be), and the simpler implementation here solves that problem using tools that Godot already provides (the input processing hierarchy).

scenes/globals/scene_switcher/transitions/transitions.gd
```diff
@@ -54,7 +54,7 @@ func _do_tween(
 func leave_scene(
 	_transition_effect: Effect = Effect.FADE,
 	duration: float = 1.0,
-	easing: Tween.EaseType = Tween.EaseType.EASE_IN,
+	easing: Tween.EaseType = Tween.EaseType.EASE_OUT,
```
manuq, line 67: Yes indeed!

manuq (approved):
This makes the code much easy to follow.

Issue #360, opened 2025-04-28
Rethink approach to asset license tracking


wjt:
I don't think the approach of placing a `.license` file next to every asset is really working for first-party assets. I wonder if we could instead do this:

1. Split `res://assets` into `res://assets/first_party` and `res://assets/third_party/`; or alternatively/equivalently declare that everything in `res://assets` must be third-party and move the first-party assets to the scenes that use them
2. Now declare in `REUSE.toml` every asset in `res://scenes` and the first-party-asset directory is CC-BY-SA-4.0 and © The Threadbare Authors, rather than requiring `.license` files. (This is what we already do for `.tscn` etc. files in `res://scenes`.)
3. During code review, check the provenance of assets

We have to do 3 anyway, even with the current system, so the `.license` files are just noise.

(By contrast I think that requiring a 2-line comment at the start of every actual source file is actually working fine.)

JuanFdS:
> Now declare in REUSE.toml every asset in res://scenes and the first-party-asset directory is CC-BY-SA-4.0 and © The Threadbare Authors, rather than requiring .license files. (This is what we already do for .tscn etc. files in res://scenes.)

Yes!, I think that would make it a lot smoother to integrate new assets.

PR #385, opened 2025-04-29
Split assets into first_party and third_party folders

JuanFdS:
Fixes https://github.com/endlessm/threadbare/issues/360


wjt (approved):

wjt:
LGTM. This is actually also less error-prone. For example:

```
diff --git a/assets/particles/circle.png.license b/assets/particles/circle.png.license
deleted file mode 100644
index 5ae7ad7..0000000
--- a/assets/particles/circle.png.license
+++ /dev/null
@@ -1,2 +0,0 @@
-# SPDX-FileCopyrightText: The Threadbare Authors
-# SPDX-License-Identifier: MPL-2.0
```

MPL is a code license; we license original non-code assets under CC-BY-SA-4.0.

wjt:
I'll fix the conflict.

wjt:
Actually the change to REUSE is wrong - assets in the main source tree should be CC-BY-SA-4.0. I'll fix...

Issue #389, opened 2025-04-30
Make template quest winnable


manuq:
We have intro, stealth (aka "Maze") and outro. We can give the remaining collectibles through dialogue so the quest can be completed.

PR #390, opened 2025-04-30
Template connect scenes

manuq:
https://github.com/endlessm/threadbare/issues/389


wjt (approved):
Perfect, I was going to suggest doing this.

Issue #260, opened 2025-04-22
Integrate ink drinker defeated animation


manuq:
When the ink drinker is beated by the player, a frame-by-frame animation should be played. Currently the character fades out.

manuq:
Still non actionable. Waiting for PixelFrog to provide the assets.

manuq:
This is now actionable, I'm attaching the defeated animation.

[Ink Drinker.zip](https://github.com/user-attachments/files/19930052/Ink.Drinker.zip)

PR #398, opened 2025-04-30
Guards: Modify how variables are being exported

JuanFdS:
While working in https://github.com/endlessm/threadbare/issues/260, I realised:
- we could have used `PropertyUtils.expose_children_properties` in https://github.com/endlessm/threadbare/pull/363 (so answering https://github.com/endlessm/threadbare/pull/363#pullrequestreview-2799972287, I actually forgot about that when I made that PR 😅 ).
- the way variables were exported in guards was different than in some other scenes (using category instead of export).
- `PropertyUtils.expose_children_properties` had a bug that made it not work for some properties names.

This PR addresses those 3 things and adds a helper to PropertyUtils to create groups.

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -26,15 +26,7 @@ const DEFAULT_SPRITE_FRAMES = preload(
 
 const LOOK_AT_TURN_SPEED: float = 10.0
 
-@export_category("Appearance")
-@export var sprite_frames: SpriteFrames = DEFAULT_SPRITE_FRAMES:
-	set(new_value):
-		sprite_frames = new_value
-		if animated_sprite_2d:
-			animated_sprite_2d.sprite_frames = sprite_frames
-		update_configuration_warnings()
-
-@export_category("Patrol")
+@export_group("Patrol")
 @warning_ignore("unused_private_class_variable")
 @export_tool_button("Add/Edit Patrol Path") var _edit_patrol_path: Callable = edit_patrol_path
 ## The path the guard follows while patrolling.
@@ -49,7 +41,7 @@ const LOOK_AT_TURN_SPEED: float = 10.0
 ## The speed at which the guard moves.
 @export_range(20, 300, 5, "or_greater", "or_less", "suffix:m/s") var move_speed: float = 100.0
 
-@export_category("Player Detection")
+@export_group("Player Detection")
 ## Whether the player is instantly detected upon being seen.
 @export var player_instantly_detected_on_sight: bool = false
 ## Time required to detect the player.
@@ -61,7 +53,7 @@ const LOOK_AT_TURN_SPEED: float = 10.0
 		if detection_area:
 			detection_area.scale = Vector2.ONE * detection_area_scale
 
-@export_category("Debug")
+@export_group("Debug")
 ## Enables movement in the editor for debugging.
 @export var move_while_in_editor: bool = false
 ## Toggles visibility of debug info.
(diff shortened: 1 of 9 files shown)


scenes/globals/property_utils.gd
```diff
@@ -40,7 +40,7 @@ static func expose_children_property(
 		var property_dict: Dictionary = get_property_dict(child, property)
 		if not property_dict.is_empty():
 			property_dict = property_dict.duplicate()
-			property_dict["name"] = "%s__%s" % [child.name.to_snake_case(), property]
+			property_dict["name"] = "%s__%s" % [child.name, property]
```
wjt, line 49: Does this break any properties currently saved in the game? I think yes:

```
scenes/game_elements/props/buildings/house/house_1.tscn:11:house_sprite__texture = ExtResource("2_qufpk")
scenes/game_elements/props/buildings/house/house_2.tscn:11:house_sprite__texture = ExtResource("2_1mgcp")
scenes/game_elements/props/decoration/bush/bush.tscn:8:bush_sprite__texture = ExtResource("1_p46yk")
scenes/game_elements/props/decoration/spool/spool.tscn:11:spool_sprite__texture = ExtResource("2_dh24t")
...
```

scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -488,3 +482,20 @@ func _draw() -> void:
...
+	)
+	return properties
+
+
+func _get(property_name: StringName) -> Variant:
+	return PropertyUtils.get_child_property(self, property_name)
+
+
+func _set(property_name: StringName, value: Variant) -> bool:
+	return PropertyUtils.set_child_property(self, property_name, value)
```
manuq, line 501: Using PropertyUtils we lose adding a documentation comment to the exported property. For example in Player SpriteFrames, I have documented which animations and how many frames each animation should have.

Reading all this, I'm not so sure how much is PropertyUtils saving us, compared to losing the documentation comment. It also makes things more complex by having to traverse all the tree, in my opinion.

scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -105,14 +97,19 @@ var state: State = State.PATROLLING:
...
+			warnings.push_back("AnimatedSprite2D__sprite_frames must be set.")
+
+		for required_animation in ["alerted", "idle", "walk"]:
+			if sprite_frames and not sprite_frames.has_animation(required_animation):
+				warnings.push_back(
+					(
+						"AnimatedSprite2D__sprite_frames is missing the following animation: %s"
+						% required_animation
+					)
+				)
```
manuq, line 112: Nice, I should add a warning to the Player SpriteFrames as well.

scenes/game_elements/characters/enemies/guard/components/guard.gd
JuanFdS: Oh, I tried encoding that requirement as a warning, so it's harder to miss. You are right tho about the documentation (!), I think there's no way to add documentation to properties added through _get_property_list, which is a shame

manuq: Yeah, I think that this project being for educational purposes, we should put the effort on our (developer) side when exporting each of these properties, and keep each property documented. Is also not apparent to me how much does PropertyUtils really save to us, developers. I'd get rid of PropertyUtils and move Decoration and Teleporter back to normal exporting.

JuanFdS: Mmm, okay, I think this is a trade off between making it easier to mod almost any texture in the game vs code simplicity. Both are things that are desirable for the learners side.

-----------

With the PropertyUtils approach, we can write that once and any animated sprite owned by this scene will be exposed, which might be nice if we have scenes with more than one. Also, it'd be the same exact boilerplate we could have in other scenes (I actually just copied these 3 functions to the `throwing_enemy.tscn` and that was it), even when the AnimatedSprite2D nodes might be called different or be in different parts of the scene hierarchy. So, I think it has some value. We could be even go higher level and have something like:

```gdscript
func _get_property_list() -> Array[Dictionary]:
        return PropertyUtils.expose_children_sprites(self)
```

That automatically exposes the texture property from any `Sprite` and the `sprite_frames` property from any `AnimatedSprite2D`. It'd be even more magic (pejorative) but it would make it simple to make a node's textures configurable (just have these 3 functions and `@tool`).

-----------

On the other hand, it's not obvious at all how this works and requires deeper knowledge of godot, so we might not want the learners to deal with that just so they have automatic exports for the spritesheets. The alternative would be to repeat the boilerplate for each node that we want to re-export properties and write the necessary glue code:
https://github.com/endlessm/threadbare/blob/b4fc8f370a29e3a186cea5234cee0f8f1002570c/scenes/game_elements/characters/enemies/guard/components/guard.gd#L33-L35
https://github.com/endlessm/threadbare/blob/b4fc8f370a29e3a186cea5234cee0f8f1002570c/scenes/game_elements/characters/enemies/guard/components/guard.gd#L131-L132

-----------

I'm fine either way, what do you think should be the priority in this case? @manuq

scenes/globals/property_utils.gd
JuanFdS: You are right. Opening those scenes again corrected it. Then, for the scenes that included these decorations, I went over them and corrected the property manually in https://github.com/endlessm/threadbare/pull/398/commits/5bd21a60426af0ac26f24bb73cb4a976a3b71c46

JuanFdS:
Closed since we are moving out of exporting variables automagically with PropertyUtils in the sake of simplicity

PR #399, opened 2025-04-30
Indicate checkpoints visually

### LINKED ISSUE #240: Indicate checkpoints visually
@PlayMatters passed on feedback from young people who playtested the game: they said they wanted checkpoints in the stealth puzzle. When shown that there are already checkpoints, they remarked that they wanted them to be visible.

wjt:
Previously, there was no visible indication that there are checkpoints in the stealth level, or that you have reached one. In playtesting, some players suggested that we should add checkpoints to the stealth level, not realising that they were already present.

Add a sprite to the checkpoint scene. Hide it until the player enters the checkpoint area.

Add a small collision shape to it, so that the player can't walk through it. Deactivate that collision shape when the sprite is invisible.

Enable y-sorting on the parent node of all the checkpoints in the two stealth levels currently in the source tree.

For now, a scarecrow is used to indicate the checkpoint, as a placeholder. I have not added any kind of animation or sound effect when the checkpoint is reached.

You can interact with the scarecrow to trigger a brief bit of dialogue that explains, in-game, what it does. The (first line of) dialogue is taken from #240.

Adjust a few checkpoint areas and positions, as explained in each commit message.

Fixes #240

### DIFF
--- a/scenes/game_elements/props/checkpoint/components/checkpoint.gd
+++ b/scenes/game_elements/props/checkpoint/components/checkpoint.gd
@@ -4,6 +4,10 @@ class_name Checkpoint
 extends Area2D
 ## A place where the player respawns if the current scene is reloaded.
 
+const DIALOGUE: DialogueResource = preload(
+	"res://scenes/game_elements/props/checkpoint/components/checkpoint.dialogue"
+)
+
 ## The point where the player will spawn.
 @onready var spawn_point: SpawnPoint = %SpawnPoint
 
@@ -13,10 +17,13 @@ extends Area2D
 ## The collision shape for the sprite, when this checkpoint is activated
 @onready var collision_shape: CollisionShape2D = %CollisionShape
 
+@onready var interact_area: InteractArea = %InteractArea
+
 
 func _ready() -> void:
 	sprite.visible = false
 	body_entered.connect(func(_body: Node2D) -> void: self.activate())
+	interact_area.interaction_started.connect(_on_interaction_started)
 
 
 ## Makes this the active checkpoint.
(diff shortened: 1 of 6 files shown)


wjt:
https://endlessm.github.io/threadbare/branches/endlessm/visual-checkpoint/#uid://db6vhi7s2f37c

wjt:
image

image

scenes/game_elements/props/checkpoint/components/checkpoint.gd
```diff
@@ -1,17 +1,27 @@
...
+## The sprite displayed when this checkpoint is activated
+@onready var sprite: Sprite2D = %Sprite
+
+## The collision shape for the sprite, when this checkpoint is activated
+@onready var collision_shape: CollisionShape2D = %CollisionShape
+
 
 func _ready() -> void:
+	sprite.visible = false
+	collision_shape.set_deferred(&"disabled", true)
```
JuanFdS, line 25: Why not disable the collision shape in the saved scene?

JuanFdS (approved):
I love the idea!, I think in a next PR we could add a small animation for the sprite that makes it "pop up" a bit more.

Also, what do you think about making both the sprite and the dialogue configurable?, so learners can make their own checkpoints in story quests?

scenes/game_elements/props/checkpoint/components/checkpoint.gd
wjt: Good question... I think I wanted it to be visible in the scene.

wjt:
I left the dialogue and sprite non-configurable for the same reason: they are placeholders.

scenes/game_elements/props/checkpoint/components/checkpoint.gd
wjt: I adjusted this to set it to disabled in the scene.

PR #407, opened 2025-04-30
Update credits.tscn

hydrolet:
Add Joana


wjt:
Thanks - I made the same change in parallel in #406

PR #410, opened 2025-04-30
Integrate ink drinker defeated animation

JuanFdS:
Added a new DYING state so the animation in the AnimationSprite2D wouldn't get overriden in _process().

Deleted the now unused remove animation from AnimationPlayer, in favor of the new dead animation in AnimationSprite2D.

Fixes https://github.com/endlessm/threadbare/issues/260

https://github.com/user-attachments/assets/97a17f43-0105-4153-9238-f8d9b1a54169


scenes/quests/lore_quests/quest_001/2_ink_combat/components/throwing_enemy/components/throwing_enemy.gd
```diff
@@ -5,11 +5,7 @@
 class_name ThrowingEnemy
 extends CharacterBody2D
 
-enum State {
-	IDLE,
-	WALKING,
-	ATTACKING,
-}
+enum State { IDLE, WALKING, ATTACKING, DYING }
```
manuq, line 8: I think that `DEFEATED` is a better word than `DYING`. Same for the file name.

manuq:
@JuanFdS to keep things going I will rename dying for defeated and merge.

PR #411, opened 2025-04-30
Credits: Add Laura Foglia

wjt:


wjt:
(This and #406 demonstrate that we can't automatically generate this list from the Git commit history, though in these cases the mistake is mine.)

LauraFoglia (approved):
Thanks!

Issue #335, opened 2025-04-25
During the different combat rounds, have the same background music playing without resetting between scenes


JuanFdS:

JuanFdS:
One thing I'm thinking is:
- If we go from scene A to scene B, and they both use the same background music, then the expected behavior would be to have the music play through the transition without stopping or resetting, right?
- If we go from scene A to scene B, and they have different background musics, then the expected behavior would be that as soon as scene B's music starts, scene A stops. It could be crazier and have some crossfade or stuff like that, but I think that at least stopping it would be okay.
- If we go from scene A to scene B, scene A has background music and scene B doesn't, the expected behavior would be for the music to stop when scene B starts, right?

wjt:
Yes, those three are what I imagined. Perhaps in the case where the music is changing, the music could fade out as the scene fades out, and fade in when the new scene fades in.

PR #413, opened 2025-04-30
Make background music continue between levels

JuanFdS:
Add a new global, MusicPlayer, that holds an AudioStreamPlayer that is accessible from everywhere.

Change the code in BackgroundMusic so it uses that AudioStreamPlayer instead of an internal one. Also, change the code so it checks if the AudioStream it will play is the same as the one that is currently playing, if that's the case, do nothing.

When transitioning out of a level, the BackgroundMusic node starts a tween that mutes the global AudioStreamPlayer by tweening its volume_db to -100 (linear_to_db(0.00001)). On _ready() of the next BackgroundMusic node, the AudioStreamPlayer volume_db would tween back to 0.0. This solves 2 things:
- We have fade out and fade in between scenes.
- If a level doesn't have background music, we don't have to do anything special in that level because the audio stream will be muted because of the volume_db. The only gotcha with this is that if we go from `level A` with `audio stream 1`, to `level B` with no music, to `level C` with `audio stream 1` again, the audio stream will resume where it was playing instead of starting again.

Fixes https://github.com/endlessm/threadbare/issues/335

# 🔊 

https://github.com/user-attachments/assets/cf51b608-292e-4825-8207-5635d215a3f8

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -22,6 +22,7 @@ GameState="*res://scenes/globals/game_state/game_state.tscn"
 SceneSwitcher="*res://scenes/globals/scene_switcher/scene_switcher.gd"
 Transitions="*res://scenes/globals/scene_switcher/transitions/transitions.tscn"
 PauseOverlay="*res://scenes/globals/pause/pause_overlay.tscn"
+MusicPlayer="*res://scenes/globals/music_player/music_player.tscn"
 
 [debug]
 
--- a/scenes/game_elements/props/background_music/background_music.tscn
+++ b/scenes/game_elements/props/background_music/background_music.tscn
@@ -3,8 +3,5 @@
 [ext_resource type="Script" uid="uid://c8405c212rbn6" path="res://scenes/game_elements/props/background_music/components/background_music.gd" id="1_g25to"]
 
 [node name="BackgroundMusic" type="Node"]
+process_mode = 3
 script = ExtResource("1_g25to")
-
-[node name="AudioStreamPlayer" type="AudioStreamPlayer" parent="."]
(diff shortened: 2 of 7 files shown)


wjt (approved):
This is much better than the previous abrupt changes!

In future it might be nice to try a different fade in/out strategy. The problems I see here are:

- In between the rounds of combat, it would be nice if the music could just keep playing, rather than fading out and in again;
- Fading in the first second of every track means we're guaranteed to miss the first couple of beats of each melody.

Another strategy might be:

- Keep music playing until the blank point, without a fade-out.
- At the blank point of the transition, check whether the new track is the same as before:
  - if so, do nothing
  - else, fade out the old during the transition into the new scene, and start the new (if any) at full volume at the end of the fade-out.
  - 
One for another week!

JuanFdS:
> This is much better than the previous abrupt changes!
> 
> In future it might be nice to try a different fade in/out strategy. The problems I see here are:
> 
>     * In between the rounds of combat, it would be nice if the music could just keep playing, rather than fading out and in again;
> 
>     * Fading in the first second of every track means we're guaranteed to miss the first couple of beats of each melody.
> 
> 
> Another strategy might be:
> 
>     * Keep music playing until the blank point, without a fade-out.
> 
>     * At the blank point of the transition, check whether the new track is the same as before:
>       
>       * if so, do nothing
>       * else, fade out the old during the transition into the new scene, and start the new (if any) at full volume at the end of the fade-out.
> 
> 
> One for another week!

Also, another thing that could be done is have a crossfade with a second AudioStreamPlayer as a child of MusicPlayer

Issue #76, opened 2025-03-28
Music and sounds


LauraFoglia:
Sound effects and music should be added when the art style is defined.

PR #416, opened 2025-04-30
Add sound effects

Arebuayon:
Related to https://github.com/endlessm/threadbare/issues/76


wjt (changes requested):
The reason that CI is failing is that you have not provided licensing information for these third-party assets. It's necessary to do so because we need to know whether we can legally redistribute them, and whether we have to provide attribution. (Even if the license is CC0, we need to record that fact!)

I think some of these sounds are from freesound. Please place those into `res://assets/third_party/freesound`. (The `third_party` directory was added in #385.) Then for each file, add a .license file alongside it, e.g. if you were adding `assets/third_party/freesound/inspiration.ogg`, please add 
`assets/third_party/freesound/inspiration.ogg.license` with the following contents (assuming the sound was https://freesound.org/s/524574/):

```
SPDX-FileCopyrightText: _bliind
SPDX-License-Identifier: CC0-1.0

Inspiration4.wav by _bliind -- https://freesound.org/s/524574/
```

There are also some merge conflicts against main that you will need to resolve.

wjt:
Functionally speaking: great. In Fray's End, the sounds of the axe, hammer, and pages turning add life to the level. (We might consider in future adjusting the elders so their pages don't turn in sync; similarly, we could adjust the axeman and hammerman to not be in sync. Not necessary for this PR.)

The walking sound is great. It would be nice to speed it up slightly while running, but not essential.

scenes/quests/lore_quests/quest_001/2_ink_combat/components/throwing_enemy/components/throwing_enemy.gd
```diff
@@ -156,7 +156,7 @@ func _process(_delta: float) -> void:
 			return
 		State.IDLE:
 			if animated_sprite_2d.animation not in [&"attack anticipation", &"attack"]:
```
wjt, line 158: You have changed the name of `attack_anticipation` so I think you need to change this too.

I also noticed that the inkdrinkers do not reliably disappear when the level is solved.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -83,8 +86,10 @@ func reveal() -> void:
 func _on_interacted(player: Player, _from_right: bool) -> void:
 	z_index += 1
 	animation_player.play("collected")
+	collectible_sound.stream = collect_sound
+	collectible_sound.play()
 	await animation_player.animation_finished
-
+	await collectible_sound.finished
 	GameState.add_collected_item(item)
```
wjt, line 93: I don't think we should wait for the sound to finish before adding the item to the HUD. It looks weird to me – the item disappears, there's a pause, and then the HUD updates.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -28,10 +28,13 @@ class_name CollectibleItem extends Node2D
 		notify_property_list_changed()
 ## The dialogue title from where [member collected_dialogue] will start.
 var dialogue_title: StringName = ""
-
+var collect_sound = preload(
+	"res://scenes/game_elements/props/collectible_item/components/sounds/MemoryCollectable.ogg"
```
wjt, line 32: I would prefer for these sounds to live in the third-party assets folder.

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/bonfire_sign/bonfire_sign.tscn
```diff
@@ -186,4 +187,15 @@ action = "Listen"
...
+[node name="FireSignSound" type="AudioStreamPlayer2D" parent="."]
+stream = ExtResource("13_rd7vt")
+volume_db = -3.0
+max_distance = 300.0
+parameters/looping = true
+
+[node name="FireStart" type="AudioStreamPlayer2D" parent="."]
+stream = ExtResource("13_6s45h")
+volume_db = -1.0
+max_distance = 300.0
```
wjt, line 199: I wonder if the distance wants to be a little bit larger – I think I couldn't hear one of the fires start because I was at a rock far away from it.

scenes/game_elements/props/door/components/door.gd
```diff
@@ -7,9 +7,14 @@ extends Toggleable
...
 		update_opened_state()
+@onready var ring_sound: AudioStreamPlayer = $RingSound
+@onready var door_sound: AudioStreamPlayer2D = $DoorSound
 
 
 func open() -> void:
+	ring_sound.play()
+	door_sound.play()
+	await door_sound.finished
 	set_toggled(true)
```
wjt, line 18: I would play the door sound while the door opens, not before it opens.

Arebuayon: Oh, sorry — I just saw this. I’m aware the licenses were missing; my idea was to upload everything first so you could try it out and help decide which sounds to keep or discard.

But when I tested the web build on https://endlessm.github.io/threadbare/branches/endlessm/add-sound-effects/, a lot of the sounds were broken (mostly at the stealth level, not sure why, since they work fine in the editor), so I wanted to fix that before sharing the build with everyone.

Thanks a lot for testing it out, though! Your feedback already gave me plenty to work with

wjt (changes requested):
Needs another rebase to fix merge conflicts, I'm afraid. (For the future I suggest making changes like this that cut across dozens of scenes in a number of smaller PRs.)

JuanFdS:
Fixed the issue in #552, #550, #547, #543, #529, #528, #498, #494 and #485.

Issue #212, opened 2025-04-16
Make Combat level generic


manuq:
The minigame should be generalized to be:
- An action game involving enemies that throw projectiles.
- Player can hit the projectiles to redirect them into barrels.
- The barrels may match or not the projectiles.
- Goal is to fill a certain amount of barrels.

The components (projectile, barrels, etc) should be themeable.

The current ink color-matching game should be an special case of this.

Acceptance criteria:
- Have a template combat level created.

manuq:
I'm going to continue with this now. I'll move the combat components out of the quest folder and into `game_elements` folder so they are reusable. And add ways to tweak the projectile and enemy assets.

PR #418, opened 2025-05-01
Combat template

manuq:
Part of https://github.com/endlessm/threadbare/issues/212


manuq:
I have fixed the conflict with the enemy defeated animation.

scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
```diff
@@ -8,7 +8,7 @@ extends CharacterBody2D
 enum State { IDLE, WALKING, ATTACKING, DEFEATED }
 
 const PROJECTILE_SCENE: PackedScene = preload(
-	"res://scenes/quests/lore_quests/quest_001/2_ink_combat/components/projectile/projectile.tscn"
+	"res://scenes/game_elements/props/projectile/projectile.tscn"
```
wjt, line 11: It doesn't have to be changed here but I noticed that in the Sokoban code, @PhoenixStroh used `uid://` paths for paths embedded in GDScript source code, presumably so that they don't have to be updated when the source tree is rearranged. It's a bit more opaque, but you can hover to see what the `uid://` resolves to and ctrl-click to jump to it. I did the same in the splash screen when I inserted the title screen after it. We might try to get into the habit of doing this in future.

image

wjt (approved):

scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
manuq: Oh good point, I didn't know about it. Yes, I'll switch to uid paths.

Issue #419, opened 2025-05-01
Consistently use Memory, Imagination, Spirit order


wjt:
In our narrative, the threads should be described and collected in the order Memory, Imagination, Spirit. (In StoryQuests a different order may be used.) Currently there are a variety of different orders in the game:

- Some dialogue refers to Memory, Spirit and Imagination
- The LoreQuest challenges use the order yellow (imagination), red (spirit), green (memory)

We should reorder these. I would also like to take the opportunity to consistently use the Oxford/serial comma ("Memory, Imagination, and Spirit" rather than "Memory, Imagination and Spirit"). Currently the dialogue uses a mixture. I believe the most common US English literary style guides use the serial comma.

PR #420, opened 2025-05-01
Consistently use Memory, Imagination, Spirit order

wjt:
Previously we had the threads described and shown in a variety of different orders. In our own narrative we want to always use the order Memory, Imagination, Spirit.

Update dialogue, the order of collectible threads in the LoreQuest, and dialogue files to consistently use this order.

Additionally, display the HUD in the music puzzle, so that the collected thread has somewhere to go, and fix inconsistent use of the serial comma in dialogue.

Fixes https://github.com/endlessm/threadbare/issues/419


wjt:
Since much of the team is out for a national holiday today, I'm going to merge this without review – post-hoc review welcome.

PR #429, opened 2025-05-01
Splash: Use blurple fabric background

wjt:
This better matches one of the standard presentations of our logo.

Reuse the fabric background from the title screen, modulated with the Endless
brand colour. The result is not quite the right tint but closer to it.

Remove the placeholder grey fabric background image.


cassidyjames (approved):
We chatted about this on Slack—it's an improvement over what's there and makes sense. 👍🏻

PR #430, opened 2025-05-01
Fray's End: Add a new flower and tweak a Crochethemum

wjt:
image


manuq:
@wjt what is this PR about? I see the flowers in the screenshot already in main.

wjt:
I don't think it's worth refreshing this PR – I would rather spend the time implementing some kind of generic "flower garden" scene that loads up randomly selected flowers from a directory.

PR #433, opened 2025-05-02
Parameterise player name in default dialogue

wjt:
I noticed a few places where the name "StoryWeaver" is hardcoded in dialogue that will, by default, be used in StoryQuests if certain components are used there without customisation – namely, stealth challenge checkpoints and default NPC dialogue.

Update all of these to get the player name dynamically, and make it possible to customise the checkpoint dialogue in any case.

### DIFF
--- a/scenes/game_elements/props/checkpoint/components/checkpoint.gd
+++ b/scenes/game_elements/props/checkpoint/components/checkpoint.gd
@@ -4,9 +4,9 @@ class_name Checkpoint
 extends Area2D
 ## A place where the player respawns if the current scene is reloaded.
 
-const DIALOGUE: DialogueResource = preload(
-	"res://scenes/game_elements/props/checkpoint/components/checkpoint.dialogue"
-)
+## Dialogue to trigger when the player interacts with the checkpoint. If empty, the player will not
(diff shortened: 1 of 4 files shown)


scenes/game_elements/props/checkpoint/components/checkpoint.gd
```diff
@@ -4,9 +4,9 @@ class_name Checkpoint
 extends Area2D
 ## A place where the player respawns if the current scene is reloaded.
 
-const DIALOGUE: DialogueResource = preload(
-	"res://scenes/game_elements/props/checkpoint/components/checkpoint.dialogue"
-)
+## Dialogue to trigger when the player interacts with the checkpoint. If empty, the player will not
+## be able to interact with the checkpoint.
+@export var dialogue: DialogueResource = preload("uid://bug2aqd47jgyu")
```
manuq, line 9: :+1:

manuq (approved):
The replacement looks good. At first glance I didn't see where is the `player_name` being set in these changes. But looking at the existing code, I see that the player is being passed in existing baloons as namespace, so `Player.player_name` can be referenced from the dialogue.

PR #439, opened 2025-05-02
Add ambient sounds to Song Sanctuaries

### LINKED ISSUE #423: Ambient sounds
- [ ] Forest ambient sounds for Fray's End and Music Puzzle
- [ ] River ambient sound for Music Puzzle

pablitar:
Added a forest sound and a river sound.

In order to make the sounds get louder when the player gets near forest and/or river, this commit also adds a player follower, which is a node that follows the player, constrained in a path.

Fixes #423


pablitar:
Test here: https://endlessm.github.io/threadbare/branches/endlessm/ambient-sounds-in-song-sanctuaries/#quests/lore_quests/quest_001/1_music_puzzle/music_puzzle

wjt (changes requested):
Try standing here:

image

Then pause and play a few times. At least on the web, the two copies of the same sound play very slightly out of phase with each other and sound awful, with mechanical distortion. You can get a similar effect (but less bad) from the forest sound:

image

I think the issue is that you have the same sounds playing twice simultaneously. Why not have each sound just once, with a continuous path for each? Sure, in the case of the forest you'd be cheating because the sound's playerfollower would cross the grass that the player walks across, but that's not the end of the world.

wjt:
Audio recording from the web build:

[untitled.webm](https://github.com/user-attachments/assets/7a9b74c2-80ce-4005-a628-18aaf1eb7742)

pablitar:
Oh. That's sad. The reason I made it like that was because of panning: The sound is panned to left or right depending on the position of the sound and the camera. This can be disorienting if the sound jumps from one side to the other.

I disabled panning for the sounds and replaced the sound with only one instance per type of sound. Could you check if it works correctly now? @wjt

pablitar:
Hey @wjt! I went ahead and rebased this on main. If you are OK with it, I'd love to merge it!

wjt (approved):
Thanks!

PR #441, opened 2025-05-02
Player template: Move it to the template quest

manuq:
Learners will duplicate the template quest. With this change, the sprite frames of the player are ready to be modified by the learner after the folder duplication


manuq:
Ready for review. I checked what is Candy Collective doing: we have desaturated versions of the assets in "World One", the world that learners are supposed to duplicate. So I guess is fine going with the duplication, even if learners don't mod the player (and upcoming assets).

wjt:
The difference in candy collective is that the assets in the world directory are used automatically, without the learners needing to make any other change. Whereas here they will also need to change the scenes to reference the copy of the assets.

That kind of magic makes sense in such a simple game... I'm not so sure it fits here - the game is more complicated, and we are expecting people to learn more about godot as part of working on this -  but you could imagine some system of looking at up the player spites from a particular path relative to the scene being played...

wjt (approved):

PR #447, opened 2025-05-02
Replace trees wind noise with a static png

pablitar:
Noise textures have long loading times. It makes sense to use a noise texture if you are changing its parameters. If not, it's better to use a static texture.

This commit replaces tree noise with a static png, which should improve loading times in every scene that has trees.


wjt (approved):
LGTM but has merge conflicts due to #450

pablitar:
Conflicts solved! Merging!

Issue #437, opened 2025-05-02
Music puzzle reset logic has regressed


wjt:
Suppose that the sequence of notes that you have to play is: C D E G. And suppose that you play C D C D E G.

What *should* happen is:

1. C: correct! next expected note is D
2. D: correct! next expected note is E
3. C: wrong! reset to start of puzzle;
    1. and try matching C again: correct! next expected note is D
4. D: correct!
5. ... etc, the puzzle is solved

The point of the logic in branch (3)(i) is that the puzzle should have no hidden state: no matter what you have played before, if you play the correct melody you should succeed.

commit bf96e5f721f24b39ef6c9b2fc13e7a9b7f56738e broke this: it removed the code that does branch (3)(i).

wjt:
I think this is quite important because it makes the puzzle unfair. In a playtest we saw @Stregoica777 play the correct melody, but the puzzle didn't accept her solution; so then she (reasonably enough) doubted that she had understood the puzzle correctly.

PR #451, opened 2025-05-02
Music Puzzle: On wrong note, match again at start of melody

wjt:
Suppose that the sequence of notes that you have to play is C D E G, and you play C D C D E G. This should be accepted as a valid solution, because although the third note played is wrong for the 3rd note of the melody, it's the correct first note. But this regressed in commit bf96e5f.

After failing to match the note in the middle of the melody, match it again at the start.

Fixes https://github.com/endlessm/threadbare/issues/437


manuq (approved):
Yes this fixes the regression.

PR #456, opened 2025-05-03
Combat template

manuq:
Helps https://github.com/endlessm/threadbare/issues/212


scenes/game_logic/fill_game_logic.gd
```diff
@@ -14,7 +14,9 @@ func start() -> void:
 	var player: Player = get_tree().get_first_node_in_group("player")
 	if player:
 		player.mode = Player.Mode.FIGHTING
-	get_tree().call_group("throwing_enemy", "start")
+	for throwing_enemy: ThrowingEnemy in get_tree().get_nodes_in_group("throwing_enemy"):
+		if not throwing_enemy.autostart:
+			throwing_enemy.start()
```
wjt, line 20: This works but you might have considered making `start()` idempotent, i.e. a no-op if it has already been called.

scenes/game_logic/fill_game_logic.gd
```diff
@@ -25,6 +26,10 @@ func start() -> void:
 func _ready() -> void:
 	var filling_barrels: Array = get_tree().get_nodes_in_group("filling_barrels")
 	barrels_to_win = clampi(barrels_to_win, 0, filling_barrels.size())
+	if intro_dialogue:
+		DialogueManager.show_dialogue_balloon(intro_dialogue, "", [self])
```
wjt, line 30: This is an existing issue but I think we might in future want to pass the player in as extra state, in case the dialogue has them speaking.

wjt (approved):

Issue #458, opened 2025-05-05
Add eternal loom


LauraFoglia:
![Image](https://github.com/user-attachments/assets/2f5270ba-9d39-4dfe-b128-a46e167d0020)

[Structures_All colors.zip](https://github.com/user-attachments/files/20037550/Structures_All.colors.zip)

hydrolet:
pr here: https://github.com/endlessm/threadbare/pull/465

PR #465, opened 2025-05-05
add eternal loom 

hydrolet:
Add eternal loom to loom scene

And adjust Fray's End by moving the Elder and tree to better fit the loom.

Fix https://github.com/endlessm/threadbare/issues/458


manuq (changes requested):
This is great and works as expected! I will only comment about the files structure.

@jbourqueendless and @JuanFdS came with the following structure for the scenes that compose the game. This is intended to avoid confusion among learners that want to instantiate these scenes. The new `eternal_loom.tscn` is one such component so I think it should follow this structure:
- Each folder should have a single file in it, with the same name (plus the `.tscn` extension).
- Any other files like assets or scripts should be added into a subfolder named `components/`

So in this case, can you please do the following?
- Create a subfolder `scenes/game_elements/props/buildings/eternal_loom/components/`
- Move all PNG files there. I don't think the sub-sub-folder with the color names is needed. You can append the color in the file name. For example: `EternalLoomPurple.png`.

The current `eternal_loom.tscn` file is OK and follows this structure!

hydrolet:
committed my changes, would love a review! thanks!

manuq (changes requested):
@hydrolet I see now that the `eternal_loom` had the wrong structure before you attempted to replace the assets (script and dialogue file not inside a `components/` subfolder). And I see you are trying to fix that too. Great!

It looks like you moved and renamed the files correctly in Godot. I can see the scene file now referencing `scenes/game_elements/props/eternal_loom/components/eternal_loom_blue.png`.

But you probably forgot to `git add` that new folder with all the files inside. If I go to the "Files changed" tab here in Github I currently see 4 files being removed and 2 files being modified. This is probably because of the missing `git add scenes/game_elements/props/eternal_loom/components/`.

manuq:
I used "squash and merge" and adjusted the commit message. Thank you @hydrolet !

PR #467, opened 2025-05-06
Settings: Reduce default music volume to -15 dB

wjt:
The music is very loud relative to the other sounds in the game.

Reduce the default volume for the music channel.

Fixes https://github.com/endlessm/threadbare/issues/443

### DIFF
--- a/scenes/globals/settings/settings.gd
+++ b/scenes/globals/settings/settings.gd
@@ -6,6 +6,9 @@ const SETTINGS_PATH := "user://settings.cfg"
 
 const VOLUME_SECTION := "Volume"
 const MIN_VOLUME := -30.0
+const DEFAULT_VOLUMES: Dictionary[String, float] = {
+	"Music": -15.0,
+}
 
 var _settings := ConfigFile.new()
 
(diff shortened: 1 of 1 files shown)


JuanFdS:
> This works as a quick fix but I'm not sure it's a great solution. Regardless of what the levels are behind the scenes or how this is achieved, I think it would be better if the music & sfx were evenly balanced so that setting both sliders to the maximum would sound good.
> 
> Some ways to achieve this:
> 
>     * Normalise the sfx samples so that they are louder
> 
>     * Master the music at a quieter max amplitude (I think the different pieces of music sound reasonably well balanced compared to each other so personally this wouldn't be my preference)
> 
>     * Apply an offset to the music channel setting so that setting it to its maximum actually sets the channel to (say) -6 dB
> 
> 
> but maybe just reducing the default volume is OK for now.

I think the best solution would be to mix the music and sfx from the start. It seems there isn't any way in godot to tweak the volume when importing .ogg 🤔 (.wav can be normalized). For now I think this is good enough.

JuanFdS (approved):

Issue #468, opened 2025-05-06
Create Tree Variants


PlayMatters:
We need a setting in the tree scene to allow change in the sprite frames for each colour. We also need one sprite frame per tree variation.

Issue #396, opened 2025-04-30
Add fabric tree & bush colour variations


wjt:

manuq:
Reopening because bush variations are missing.

PR #471, opened 2025-05-06
New trees

manuq:
Fix https://github.com/endlessm/threadbare/issues/468
Fix https://github.com/endlessm/threadbare/issues/396


manuq:
I'm still missing the bushes. But the trees are done.

scenes/game_elements/props/tree/components/tree.gd
```diff
@@ -3,6 +3,31 @@
 @tool
 extends Decoration
```
JuanFdS, line 4: Should trees still extend from the `Decoration` class?

Checking the logic from `decoration.gd`, tree isn't using any of that.

I think we could have achieved a similar result to what's done in this PR by editing `decoration.gd` to this (and it'd work for the next decorations that we need to animate without having to add new code for each)* instead of adding a new property and its handling to `tree.gd`:

```gdscript
func _get_property_list() -> Array[Dictionary]:
	return PropertyUtils.expose_children_property(self, "texture", "Sprite2D") + PropertyUtils.expose_children_property(self, "sprite_frames", "AnimatedSprite2D")
```

*it wouldn't work exactly the same because there's logic for default spriteframes and what to do in case of a null spriteframe, but that could be moved to `decoration.gd` instead of repeating it in each node.
*if this fix was applied: https://github.com/endlessm/threadbare/pull/398/commits/5bd21a60426af0ac26f24bb73cb4a976a3b71c46

However, we are moving out of the automagic way of handling these kind of scenes, right?

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/trees.gd
```diff
@@ -3,13 +3,14 @@
...
+const ALTERNATIVE_SPRITEFRAMES = preload("uid://djwymcffy83")
+const SPRITEFRAMES = preload("uid://d36eq8tqdaxdy")
 
 @export_tool_button("Randomize Trees Appearances") var a = func():
 	for child in get_children():
-		child.set("tree__texture", TREE if randf() > WOOL_PERCENTAGE else TREE_WOOL_GREEN_01)
+		child.set(
+			"sprite_frames",
+			SPRITEFRAMES if randf() > ALTERNATIVE_PERCENTAGE else ALTERNATIVE_SPRITEFRAMES
+		)
```
JuanFdS, line 16: For a different PR, but, I think it'd be really cool if this was configurable, maybe with an exported dictionary that goes from sprite frames to weight?

JuanFdS (approved):
Great work! ✅

scenes/game_elements/props/tree/components/tree.gd
manuq: Oh good point, I wasn't even paying attention to the Decoration class.

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/trees.gd
manuq: I think the way the scenes are currently dressed is by changing trees one by one. I kept this randomizer because it was an easy change, otherwise I would have removed the script.

PR #472, opened 2025-05-06
Add new rock sprite to music puzzle

pablitar:
This commit adds the new rock sprites to music puzzle. Since they are animated, the sprite had to be converted into an animated sprite. The animation player was removed.

Since Rocks could be a decoration, their sprites were left in the assets folder, so they can easily be reused.

Also: Moved the musical rock xylophone to the "components" folder

Fixes #393

PR #473, opened 2025-05-06
Musical rock fixes

manuq:
Fixes for https://github.com/endlessm/threadbare/pull/472 that I realized after merging.


JuanFdS (approved):
This is so the animation isn't cut off before it finished a loop?
And the FPS makes sense, all should be at 10fps I think.

manuq:
> This is so the animation isn't cut off before it finished a loop? 

Yes, without this the last frame can stop abruptly when the audio stream ends, not last the amount of milliseconds as the other frames. Also I think is better to complete the looping animation.

PR #477, opened 2025-05-07
Switch all res:// preload references to UID paths

manuq:
This makes the scripts immune to file renames.

It is also less readable, but the Godot editor understands these paths and you can Ctrl-click on them in the script editor to see the resource open in the Inspector. From there you can also use the top bar menu to show the file in the Filesystem dock. Or, if it is a scene, it opens in a new tab directly, then right-clicking in the root node it can be shown in the Filesystem dock.

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -20,9 +20,7 @@ enum State {
 	RETURNING,
 }
 
-const DEFAULT_SPRITE_FRAMES = preload(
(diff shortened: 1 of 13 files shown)


manuq:
> Did you do these by hand or is there a way to convert a `res://` to a `uid://` in the editor?

I grepped the project to find all cases. Then I went one by one opening the script in the editor, then Ctrl-click on the `res://` path which:
- If it's a scene, it opens in a new editor tab. Then I can right click the root node I can click Show in Filesystem.
- If it's a resource that's not a scene, it opens in the Inspector. From there clicking in the tree dots at the top bar I can click Show in Filesystem.

Then from the Filesystem dock, I can right click the file and Copy UID.

Issue #421, opened 2025-05-01
The character got stuck in the attack animation At the end of the combat challenge


jofilizola:
It happened 2 times, but I could not reproduce this, a third time. The screenshots shows the asset with transparency, and the animation keeps going. It doesn't stop the game, so it's polish.

![Image](https://github.com/user-attachments/assets/f507f829-b52a-418a-973a-72c0e15318cc)

PR #478, opened 2025-05-07
Fix attack animation getting stuck when changing modes

JuanFdS:
Previously, if the player's AnimationPlayer was set to the blow animation and it changed mode from FIGHTING to COZY, it would stay looping in the blow animation.

Changed the code so it can never be in a fighting animation if the mode is COZY by matching by the player_mode first.

Then, changed the logic of the fighting mode so it handles the cases in which the player should walk because they are not attacking that frame.

Finally, adding a comment to explain how if the blow animation is played, it's only cut off if it is at the anticipation phase.

Fixes https://github.com/endlessm/threadbare/issues/421


manuq (approved):
Excellent!

PR #480, opened 2025-05-07
Tree: Randomize animation speed

wjt:
Previously, the leaf animations on trees were started from random offsets but otherwise played in sync. This looks unnatural.

Scale the speed of the leaf animation on each tree by a random factor between 0.5× and 1.0×.


manuq:
Hmm, I'm not so sure about this! The frame-by-frame animations were animated intended to play at 10 frames per second, just like the ones for Tiny Swords. I would raise this to @PlayMatters who is playing the art director role in the project.

wjt:
Preview this at:

https://endlessm.github.io/threadbare/branches/endlessm/randomize-tree-animation-speed/#world_map/frays_end

PR #481, opened 2025-05-07
Determine needed fills from animation

JuanFdS:
Previously, the amount of times a filling barrel needed to be filled was hardcoded to 3, and the animation for it could be configured, but it would always need to have 4 frames.

Now, that requirement was relaxed, the FillingBarrel exposes a SpriteFrames instead of a texture, and the amount of times it needs to be filled depends on the amount of frames in the SpriteFrames' animation.

### DIFF
--- a/scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
+++ b/scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
@@ -6,8 +6,19 @@ extends StaticBody2D
 
 signal completed
 
-const NEEDED: int = 3
-const DEFAULT_TEXTURE: Texture2D = preload("uid://b2nmajpf8dlh")
+const DEFAULT_SPRITE_FRAMES: SpriteFrames = preload("uid://dlsq0ke41s1yh")
+const FILLING_NAME_ANIMATION: StringName = &"filling"
+
+## Determines how the FillingBarrel looks.[br]
+## It is required that the [member sprite_frames] have an animation called filling.[br]
+## The barrel default sprite will be the first frame of that animation, and it
+## advances frames as it is being filled.
+@export var sprite_frames: SpriteFrames = DEFAULT_SPRITE_FRAMES:
+	set = _set_sprite_frames
+
+## The amount of times the barrel needs to be filled.[br]
+## When the barrel is filled that many times, it emits [signal completed].
+@export var needed_amount: int = 3
 
 ## Projectiles with this label fill the barrel.
 @export var label: String = "???"
@@ -16,14 +27,9 @@ const DEFAULT_TEXTURE: Texture2D = preload("uid://b2nmajpf8dlh")
 @export var color: Color:
 	set = _set_color
 
-## Optional custom texture for the barrel. An inkwell texture is used by default.
-## The texture must have 4 vertical frames, from empty to filled.
-@export var texture: Texture2D:
-	set = _set_texture
-
 var _amount: int = 0
 
-@onready var sprite_2d: Sprite2D = %Sprite2D
+@onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
 @onready var animation_player: AnimationPlayer = %AnimationPlayer
 @onready var collision_shape_2d: CollisionShape2D = %CollisionShape2D
 @onready var hit_box: StaticBody2D = %HitBox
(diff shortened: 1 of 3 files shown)


scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -6,8 +6,30 @@ extends StaticBody2D
 
 signal completed
 
-const NEEDED: int = 3
-const DEFAULT_TEXTURE: Texture2D = preload("uid://b2nmajpf8dlh")
+const DEFAULT_SPRITE_FRAMES: SpriteFrames = preload(
+	"res://scenes/game_elements/props/filling_barrel/components/filling_barrel_sprite_frames.tres"
```
manuq, line 10: We have switched to UID paths everywhere (see commit 158c776c). Can you change this one too? Here is how I did it: https://github.com/endlessm/threadbare/pull/477#issuecomment-2858246741

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -6,8 +6,30 @@ extends StaticBody2D
...
+## That amount is calculated as the amount of frames in the filling animation of
+## [member sprite_frames].[br]
+## When the barrel is filled that many times, it emits [signal completed].
+@export_custom(
+	PROPERTY_HINT_NONE,
+	"",
+	PROPERTY_USAGE_READ_ONLY | PROPERTY_USAGE_EDITOR | PROPERTY_USAGE_NO_INSTANCE_STATE
+)
+var needed_amount: int:
+	get = _needed
```
manuq, line 32: I find a bit tricky that the amount depends on the amount of frames. A learner should be able to set the amount number directly in the inspector. Ideally the frames would match, but if not they should be mapped to the current amount. There is a `&"fill"` squash & stretch animation anyways which will indicate the player that the amount has changed. I havent tested but this should do the trick:

```
var total_frames: int = animated_sprite_2d.sprite_frames.get_frame_count(animated_sprite_2d.animation)
animated_sprite_2d.frame = floor((_amount / needed_amount) * total_frames)
```

What do you think?

manuq (changes requested):
Excellent, is great to have the asset changed to an AnimatedSprite2D and already exported.

I think it would be better for a learner to set the amount regardless of the animation frames. What do you think?

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
JuanFdS: yes, I think that works too!, I like that it lets a learner change the current barrels to make them easier without having to deal with textures or animations. I think it could look a bit weird if the needed amount is > the frame count, but it's true that we have the fill animation so it won't look like nothing happened.

JuanFdS:
Addressed the comments @manuq

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -74,3 +82,31 @@ func fill() -> void:
...
+	return animated_sprite_2d.sprite_frames.get_frame_count(FILLING_NAME_ANIMATION)
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	return _sprite_frames_errors()
+
+
+func _sprite_frames_errors() -> PackedStringArray:
+	if not sprite_frames:
+		return ["sprite_frames must be set."]
```
manuq, line 100: This is not needed because the setter ensures that it will be `DEFAULT_SPRITE_FRAMES`.

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -34,35 +40,37 @@ func _set_color(new_color: Color) -> void:
...
+	if animated_sprite_2d:
+		animated_sprite_2d.sprite_frames = sprite_frames
+	update_configuration_warnings()
 
 
 func _ready() -> void:
 	_set_color(color)
-	_set_texture(texture)
+	_set_sprite_frames(sprite_frames)
+	if _are_sprite_frames_valid():
```
manuq, line 60: I don't think we should add these kinds of validations. It will be better for a learner to see the error in the editor, than to hide the error, so they can fix it.

manuq (changes requested):
Excellent! Just 2 more small requests.

JuanFdS:
Done! @manuq

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -74,3 +81,28 @@ func fill() -> void:
 func _disable_collisions() -> void:
 	hit_box.process_mode = Node.PROCESS_MODE_DISABLED
 	collision_shape_2d.disabled = true
+
+
+func _total_frames() -> int:
+	if not _are_sprite_frames_valid():
+		return 0
+
+	return animated_sprite_2d.sprite_frames.get_frame_count(FILLING_NAME_ANIMATION)
```
manuq, line 90:
```suggestion
	return animated_sprite_2d.sprite_frames.get_frame_count(FILLING_NAME_ANIMATION)
```

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -74,3 +81,28 @@ func fill() -> void:
...
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	return _sprite_frames_errors()
+
+
+func _sprite_frames_errors() -> PackedStringArray:
```
manuq, line 97:
```suggestion
func _get_configuration_warnings() -> PackedStringArray:
```

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -74,3 +81,28 @@ func fill() -> void:
...
+		return ["sprite_frames is missing the following animation: %s" % FILLING_NAME_ANIMATION]
+
+	if sprite_frames.get_frame_count(FILLING_NAME_ANIMATION) < 2:
+		return ["sprite_frames %s animation must have at least 2 frames" % FILLING_NAME_ANIMATION]
+
+	return []
+
+
+func _are_sprite_frames_valid() -> bool:
+	return _sprite_frames_errors().is_empty()
```
manuq, line 108:
```suggestion
```

manuq (changes requested):
I'm suggesting to remove the validation entirely.

JuanFdS:
Ohhhhh okay, fixed it! @manuq

manuq (approved):

PR #482, opened 2025-05-07
simplify dialogue for learner stealth template when item is collected

jbourqueendless:
Reduced the text to just give a slightly more simplified hint or breadcrumb for the learner as to where to go next. I also referred to the label is "Next Scene" instead of "Scene to Go To" because we will be adjusting that label going forward.


scenes/quests/story_quests/template/1_stealth/stealth.tscn
```diff
@@ -1,6 +1,6 @@
 [gd_scene load_steps=17 format=4 uid="uid://coi4lvxir542y"]
 
-[ext_resource type="Script" path="res://scenes/game_logic/stealth_game_logic.gd" id="1_eujsg"]
+[ext_resource type="Script" uid="uid://dnp0tjloec2d7" path="res://scenes/game_logic/stealth_game_logic.gd" id="1_eujsg"]
```
wjt, line 3: This change is probably correct but shouldn't be in the same commit. I will check this and split the commit.

PR #484, opened 2025-05-07
Add the hiding mushroom prop

pablitar:
The hiding mushroom is a decorative prop that hides or reveals itself when the player moves nearby.

It can also be configured to "do nothing", so its behavior may be controlled through a custom script.

This PR also adds some hiding mushrooms to Frays End to show how they work.

https://github.com/user-attachments/assets/58b1f9cf-392a-4cf4-a9eb-04f05fde72cf

Fixes https://github.com/endlessm/threadbare/issues/394


scenes/world_map/frays_end.tscn
```diff
@@ -1,7 +1,7 @@
-[gd_scene load_steps=44 format=4 uid="uid://cufkthb25mpxy"]
+[gd_scene load_steps=45 format=4 uid="uid://cufkthb25mpxy"]
 
 [ext_resource type="TileSet" uid="uid://07fq3rspk8ia" path="res://scenes/tileset.tres" id="1_4evgk"]
-[ext_resource type="Script" uid="uid://8div00rgvsds" path="res://scenes/world_map/components/frays_end.gd" id="1_4gse6"]
+[ext_resource type="Script" path="res://scenes/world_map/components/frays_end.gd" id="1_4gse6"]
```
wjt, line 4: I don't think we want to remove the UID here?

pablitar: No idea why that changed. Godot shenanigans. I'll revert that and merge. Thanks!!

Issue #440, opened 2025-05-02
Enter Sokoban from eternal loom


wjt:
Before commit 1792b4204bdea9009e140484704ee3279a906891, interacting with the loom switched to a Sokoban puzzle. I removed that linkage because the target scene was a placeholder which couldn't be completed.

Now that the Sokoban puzzles are playable, we should reinstate this.

PR #486, opened 2025-05-07
Link Sokoban puzzles to the Eternal Loom

wjt:
When the player interacts with the Eternal Loom while in possession of the threads of Memory, Imagination, and Spirit (after completing a quest) we previously played an animation of the threads entering the loom, and then some dialogue spoken by the Elders.

After the animation plays, switch scenes to a series of three Sokoban puzzles where the player repairs the loom using the threads they have collected. Once they have completed all three puzzles, return to Fray's End and play the Elder dialogue.

Fixes https://github.com/endlessm/threadbare/issues/440

### DIFF
--- a/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
+++ b/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
@@ -4,24 +4,51 @@ class_name EternalLoom
 extends Node2D
 
 const ETERNAL_LOOM_INTERACTION: DialogueResource = preload("uid://yafw7bf362gh")
-const ETERNAL_LOOM_SOKOBAN_PATH = "res://scenes/eternal_loom_sokoban/eternal_loom_sokoban.tscn"
+
+## Scenes that are the first of three Sokoban puzzles. A random one will be used
+## each time the player successfully interacts with the Loom.
+const SOKOBANS := [
+	"uid://b8mywvmgsxqb",
+	"uid://11cdlcqge3fu",
+	"uid://b64uft76tbblp",
+]
 
 @onready var interact_area: InteractArea = %InteractArea
 @onready var loom_offering_animation_player: AnimationPlayer = %LoomOfferingAnimationPlayer
 
 
-func _ready():
+func _ready() -> void:
 	interact_area.interaction_started.connect(self._on_interacted)
+
+	if GameState.incorporating_threads:
+		if Transitions.is_running():
+			await Transitions.finished
+
+		DialogueManager.show_dialogue_balloon(
+			ETERNAL_LOOM_INTERACTION, "threads_incorporated", [self]
+		)
+		await DialogueManager.dialogue_ended
+		GameState.incorporating_threads = false
 
 
 func _on_interacted(player: Player, _from_right: bool) -> void:
-	DialogueManager.show_dialogue_balloon(ETERNAL_LOOM_INTERACTION, "", [self, player])
+	var have_threads := is_item_offering_possible()
+
+	var title := "have_threads" if have_threads else "no_threads"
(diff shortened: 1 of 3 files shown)


wjt:
This is a little annoying to test because you have to actually complete a quest in order to get the threads; and then you have to solve three Sokoban puzzles! I recommend going through the tutorial StoryQuest — it is shorter than the LoreQuest.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -4,24 +4,48 @@ class_name EternalLoom
...
+	# A slight abuse of the spawn point mechanism – this is set by the last round of the Sokoban
+	# puzzle to indicate that we are returning from solving that puzzle.
+	if GameState.current_spawn_point == get_tree().current_scene.get_path_to(self):
+		if Transitions.is_running():
+			await Transitions.finished
+
+		DialogueManager.show_dialogue_balloon(
+			ETERNAL_LOOM_INTERACTION, "threads_incorporated", [self]
+		)
+		await DialogueManager.dialogue_ended
```
JuanFdS, line 32: I'm not so sure about this 🤔, we are using the current_spawn_point for something that isn't a spawn point.
We don't have an EternalLoom spawn point in fray's end, but if we wanted to add one and teleport the player back from a finished quest to that point, we could confuse it with this.

What do you think about these alternatives?:
- Adding a new variable to GameState, a boolean like: `just_completed_sokoban` or `just_incorporated_threads`. That is set to true in the sokoban and, after "consuming it" here is set to false again.
- Making it possible to pass callbacks to the SceneSwitcher, so, after the new scene is introduced and the transition fades out, the callback is ran. That way we could run this from the sokoban. However, I prefer this code being **_here_** in the eternal loom than in the scene that sends the player here, I think it could grow in a way that needs access to Fray's End props or the loom.

scenes/game_elements/props/eternal_loom/components/eternal_loom_interaction.dialogue
```diff
@@ -1,13 +1,21 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
-~ start
+~ blurb
```
JuanFdS, line 3: I had to look up if `blurb` meant something 😅 , is it referring to this?:
https://dictionary.cambridge.org/es/diccionario/ingles/blurb

Like, in this context, a short introduction?

JuanFdS (changes requested):
Left some comments, I'm not so sure about using the spawn point.
I think it has one thing that's cool: any scene could be used as a way to solve the loom, as long as it sends you back to that spawn point. However, I think we could make it more explicit for this mechanic even if that forces us to include some extra node or logic in the scenes that complete the loom.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
wjt: I think you're probably right that this should live in the game state. I guess I used this spawn point because it was already there :) I'll rework it.

scenes/game_elements/props/eternal_loom/components/eternal_loom_interaction.dialogue
wjt: Great point. "Introduction" would be a better name (and certainly a less obscure word)

wjt: I went with "welcome"

manuq (approved):
I went through it (using template, not lore) and works as expected. I was lucky that level 1 was picked, which I know how to solve.

PR #492, opened 2025-05-07
Player: Warn when SpriteFrames animations are wrong

manuq:
Inspired by the same warnings for the guards.

Note: This effectively adds a warning to the player template SpriteFrames aka "egg guy" because it's missing the attack_02 animation on purpose.

### DIFF
--- a/scenes/game_elements/characters/player/components/player.gd
+++ b/scenes/game_elements/characters/player/components/player.gd
@@ -13,6 +13,12 @@ enum Mode {
 	FIGHTING,
 }
 
+const REQUIRED_ANIMATION_FRAMES: Dictionary[StringName, int] = {
+	&"idle": 10,
+	&"walk": 6,
+	&"attack_01": 4,
+	&"attack_02": 4,
+}
 const DEFAULT_SPRITE_FRAME: SpriteFrames = preload("uid://dtoylirwywk0j")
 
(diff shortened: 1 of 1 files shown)


scenes/game_elements/characters/player/components/player.gd
```diff
@@ -13,6 +13,12 @@ enum Mode {
 	FIGHTING,
 }
 
+const REQUIRED_ANIMATION_FRAMES: Dictionary[StringName, int] = {
+	&"idle": 10,
+	&"walk": 6,
+	&"attack_01": 4,
+	&"attack_02": 4,
+}
```
JuanFdS, line 21: Why do we require these specific amount of frames in these animations?, I know that for the attack animation we use the frame number to decide if it can be cancellable, but why do we need the rest? 🤔

manuq: > Why do we require these specific amount of frames in these animations?, I know that for the attack animation we use the frame number to decide if it can be cancellable, but why do we need the rest? 🤔

Because the frames of these animations are used in the timeline of the AnimationPlayer.

JuanFdS: Ohhh I see, okay then!, I think it's good to merge

PR #496, opened 2025-05-07
Add walk particles to player

### LINKED ISSUE #245: Walking & Running particles

pablitar:
As the player walks or runs, it now emits dust particles.

https://github.com/user-attachments/assets/cdd89b58-0e52-4752-9dbd-35effe66865e

Fixes #245


manuq:
Nice touch! I think this needs designer eyes. I find it a bit too rough, it could be more subtle. Smaller, more transparent, with an asset that has a single dust instead of 2.

pablitar:
> Nice touch! I think this needs designer eyes. I find it a bit too rough, it could be more subtle. Smaller, more transparent, with an asset that has a single dust instead of 2. Also it would be nice to use this effect to better differentiate walking from running.

Most of it can be tuned by changing particle parameters! The only thing that's dependant of the sprite is it having two dusts instead of 1, but that can easily be adjusted.

wjt:
Superseded by #652

PR #498, opened 2025-05-07
Add sounds for stealth challenge

JuanFdS:
Added sounds for:
- Guard footsteps: plays only while they walk.
- Guard's torch: plays continuously, can only be heard if you are really near them.
- Guard's alerted sound: plays when they see you and when they alert everybody. If it's already playing it doesn't play twice.
- When the guard is alerting the rest, they swing the torch, that's also added as a sound.

Related to https://github.com/endlessm/threadbare/issues/76.

## Learners side: Make sounds configurable

Add exported variables that change the audio streams used in the guard, with descriptions that I hope help the learners or collaborators figure out what they could use there.

If the audiostream is cleared, then the guard won't emit that sound.

Also, changed the way the sounds were played in the AnimationPlayer so it doesn't override the AudioStream from the AudioStreamPlayer.

https://github.com/user-attachments/assets/016f44f6-0d39-44f1-a668-d77ef487d820

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -32,6 +32,20 @@ const LOOK_AT_TURN_SPEED: float = 10.0
 			animated_sprite_2d.sprite_frames = sprite_frames
 		update_configuration_warnings()
 
+@export_category("Sounds")
+## Sound played when a guard's [enum State] enters DETECTING or ALERTED.
+@export var alerted_sound_stream: AudioStream = preload("uid://dgpeb1dtmfqud"):
+	set = _set_alerted_sound_stream
+## Sound played when a guard's moving from one point to the next.
+@export var footsteps_sound_stream: AudioStream = preload("uid://4ec6e2alns71"):
+	set = _set_footsteps_sound_stream
+## Sound played continuously.
+@export var idle_sound_stream: AudioStream = preload("uid://c7om8kdork2rx"):
+	set = _set_idle_sound_stream
+## Sound played in bursts after the guard entered [enum State] ALERTED.
+@export var alert_others_sound_stream: AudioStream = preload("uid://bwif2oo6ymiu2"):
+	set = _set_alert_other_sound_stream
+
 @export_category("Patrol")
 @warning_ignore("unused_private_class_variable")
 @export_tool_button("Add/Edit Patrol Path") var _edit_patrol_path: Callable = edit_patrol_path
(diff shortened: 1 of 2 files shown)


scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -32,6 +32,44 @@ const LOOK_AT_TURN_SPEED: float = 10.0
 			animated_sprite_2d.sprite_frames = sprite_frames
 		update_configuration_warnings()
 
+@export_category("Sounds")
+## Sound played when a guard's [enum State] enters DETECTING or ALERTED.
+@export var alerted_sound_stream: AudioStream = preload(
+	"res://assets/sound/third_party/enemies/guard/AlertSound.ogg"
+):
```
manuq, line 39: As discussed on Discord, let's put the current sound as default referenced by UID in a constant above. Then in the setter if the new value is null, set it to the default.

scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -32,6 +32,44 @@ const LOOK_AT_TURN_SPEED: float = 10.0
...
+	set(new_value):
+		alerted_sound_stream = new_value
+		if not is_node_ready():
+			await ready
+		_alert_sound.stream = new_value
```
manuq, line 44: I find it a bit more readable when the setters are named functions, than when they are unnamed like in here.

manuq (changes requested):
Looks great! Thanks for exporting the sounds too. Left a comment on having the new sounds as defaults.

scenes/game_elements/characters/enemies/guard/components/guard.gd
JuanFdS: > As discussed on Discord, let's put the current sound as default referenced by UID in a constant above. Then in the setter if the new value is null, set it to the default.

@manuq  I think we haven't reached an agreement on what should a null value mean 👀 
> I think null sound meaning no sound makes more sense because godot already provides a way to return to the default value, so, if we make it so null means default, we'd have 2 ways of representing the same: leaving it cleared or leaving it to the default value.

https://github.com/user-attachments/assets/805b766e-ac0e-43e0-94ce-2fa827ba615c

manuq: @JuanFdS ah it is already like the sprite frames, the only difference is that sprite frames has the `preload(...)` in a constant. Then it is fine as it is right now! The only thing you need is to use `uid://` paths.

JuanFdS: > @JuanFdS ah it is already like the sprite frames, the only difference is that sprite frames has the `preload(...)` in a constant. Then it is fine as it is right now! The only thing you need is to use `uid://` paths.

Changed it to use uids! ✅ @manuq

PR #504, opened 2025-05-07
Fill game logic: Pass player to dialogue

manuq:
In case a learner references the player name through the placeholder in the dialogue. Like in other parts of the game, pass the player as namespace to the dialogue.

This was suggested in https://github.com/endlessm/threadbare/pull/456#discussion_r2075014885.


wjt (approved):
Nice.

Seeing this I wonder if the logic that finds the player's name in the balloon could simply do this too...!

PR #506, opened 2025-05-08
Redesign LoreQuest music puzzle

PlayMatters:
Redesign music puzzle to simplify dialogue, choices, and solutions, and allow the player to interact with the signs before they have been lit.
Redesign tilemap to reduce complexity.
Redesign aesthetics and add ruined buildings to feel more worn and broken, as the narrative suggests.

These changes were made in response to feedback of complexity both in tilemap design and dialogue, as well as a need to meet the aesthetics of the narrative for this element of the game.


assets/third_party/tiny-swords/Terrain/Water/Rocks/Rocks_01.png.import
```diff
@@ -1,34 +0,0 @@
-[remap]
```
wjt: Did you mean to rename one of the Tiny Swords rocks? I don't think so – I'll move it back.

wjt (commented):
I want to make a couple of changes here but broadly speaking this looks pretty good!

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/dialogues/musician.dialogue
```diff
@@ -37,70 +36,75 @@ Musician: Hello again, StoryWeaver. Would you like to hear about the stones agai
...
+Musician: You have opened the portal to the Ink Well. Let's go there to collect ink so songs and stories can be written again.
+Musician: But we must be careful, InkDrinkers dwell there!
+Musician: The melody awakened the memory of song. Collect the Memory threads before you go!
 => END
 
-~ item_hint
-Musician: (PLACEHOLDER) You did it! The melody worked. I left a special item near the stones, grab it before you go!
+~ well_done
+StoryWeaver: Ink Well, here we come!
 => END
```
wjt, line 111: There is no guarantee that the player interacts with the musician after completing the puzzle but before interacting with the thread. If the player goes straight to the thread, they will miss all the exposition about the Ink Well.

We can address this later – the trick might be to play the musician's dialogue automatically when the puzzle is solved.

PR #515, opened 2025-05-08
Simplify learner templates

jbourqueendless:
I reduced the amount of text and simplified the guidance within the dialogue in the intro template, checkpoint, and combat template scenes just to keep the "instruction" there to a bare minimum of bread crumbs or hints. This is to go along with a previous PR where I adjusted the language of the collectible dialogue for the stealth template scene as well.

Within the combat scene and stealth scene I also generalized the player name property to just be **Player Name** instead of Storyweaver and a made a few nodes sound more generic to encourage more open creativity/ideation for learners about the potential these components have for their unique StoryQuests. 

- Barrels are targets
- ThrowingEnemy is a ThrowingNPC
- I left "Guards" alone in stealth but will review with others on this one as well.

### DIFF
--- a/scenes/game_elements/props/checkpoint/components/checkpoint.dialogue
+++ b/scenes/game_elements/props/checkpoint/components/checkpoint.dialogue
@@ -1,6 +1,11 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 ~ start
-KnitWitch: This is a checkpoint that can be used to save player progress.
-{{player_name}}: Thanks!
+KnitWitch: I have knitted your progress so far, {{player_name}}.
+%9
+%
+	{{player_name}}: Wow, you can talk!
+	KnitWitch: Wow, so can you. What a coincidence.
+	{{player_name}}: I didn't think scarecrows could talk.
+	KnitWitch: Maybe they just never had anything to say to you. Ever think of that?
 => END
--- a/scenes/game_elements/props/checkpoint/components/checkpoint.dialogue
+++ b/scenes/game_elements/props/checkpoint/components/checkpoint.dialogue
@@ -1,11 +1,6 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 ~ start
-KnitWitch: I have knitted your progress so far, {{player_name}}.
-%9
-%
-	{{player_name}}: Wow, you can talk!
-	KnitWitch: Wow, so can you. What a coincidence.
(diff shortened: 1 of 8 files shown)


scenes/game_elements/props/checkpoint/components/checkpoint.dialogue
```diff
@@ -1,11 +1,6 @@
...
 ~ start
-KnitWitch: I have knitted your progress so far, {{player_name}}.
-%9
-%
-	{{player_name}}: Wow, you can talk!
-	KnitWitch: Wow, so can you. What a coincidence.
-	{{player_name}}: I didn't think scarecrows could talk.
-	KnitWitch: Maybe they just never had anything to say to you. Ever think of that?
+KnitWitch: This is a checkpoint that can be used to save player progress.
+{{player_name}}: Thanks!
```
wjt, line 5: This change also affects the LoreQuest – is that intended?

jbourqueendless: No sorry! I meant to adjust the dialogue there only in the template scene. Forgot the dialogue was shared.
Should I just create a separate template dialogue for this?

wjt: Yes, I think that would be best!

jbourqueendless: Just pushed updates based on feedback from @wjt .

wjt: You created a different file but didn't revert this change to the file used in the LoreQuest. I'll fix this up.

scenes/quests/story_quests/template/2_combat/combat.tscn
```diff
@@ -82,7 +83,7 @@ revealed = false
 next_scene = "uid://x3wm2ce0ax8i"
 item = SubResource("Resource_a51xm")
 collected_dialogue = ExtResource("2_3qn31")
-dialogue_title = &"well_done"
+dialogue_title = &""
```
wjt, line 86: This change is wrong – this means that the same dialogue will play when you collect the collectible as when the scene starts. Again I can fix this.

manuq:
After the fact review: Isn't it better to rename ThrowingEnemy to ThrowingNPC in the actual scene? It may be confusing to learners that it's now called "NPC" in the template but going to the instanced scene it is called "Enemy".

manuq:
I ask because I keep calling it enemy in https://github.com/endlessm/threadbare/pull/554

PR #508, opened 2025-05-08
Storyweaver color variations

manuq:
This also integrates the player defeated animation. Currently unused. I guess it can be used in the Stealth minigame, when the player is cought.

Fixes https://github.com/endlessm/threadbare/issues/507

PR #517, opened 2025-05-08
Player: Fix storyweaver red color variant

manuq:
The SpriteFrames resource file was copied externally from the blue, so it had the same UID as the blue variant.

Fix it by:
- Duplicating the resource file to a temporary file name in Godot (this changes UID and internal IDs)
- Remove the red file.
- Rename the temporary name to the red file name.

See https://github.com/endlessm/threadbare/pull/508#issuecomment-2863282680


wjt:
It's super annoying that Godot doesn't have a "fix duplicate UIDs" UI or any obvious way to dot this. I wouldn't have thought of what you did!

wjt (approved):

PR #527, opened 2025-05-08
Add book props and add books to frays end

pablitar:
This PR also adds the flip & offset properties as exposed properties in decoration.gd.

In order to allow for more variation, decorations can have their flip and offset properties so they can be tuned by level editors.

The flip_h and flip_v properties allow to reuse the same decoration, but adding a bit more variation. The offset, on the other hand, allows to stack decorations without messing with the y-sort.


manuq (approved):

scenes/game_elements/props/buildings/house/house_1.tscn
```diff
@@ -9,6 +9,9 @@ size = Vector2(64.0002, 80)
 [node name="House_1" type="Node2D"]
 script = ExtResource("1_qufpk")
 house_sprite__texture = ExtResource("2_qufpk")
+house_sprite__flip_h = false
+house_sprite__flip_v = false
+house_sprite__offset = Vector2(0, 0)
```
manuq, line 14: I think this diff in the decoration scenes happen because this is being exposed through the PropertyUtils, not through regular exporting. If the properties are the default ones, Godot doesn't save them in the scene.

pablitar: Hmm... I think this can be fixed by using the [`_property_can_revert`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-property-can-revert)and the [`_property_get_revert`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-property-get-revert) methods.

We can add an issue if it's important to solve it!

PR #529, opened 2025-05-08
Add player footsteps

JuanFdS:
-  Add sound asset for player footsteps 
- Player: add sfx for walking 
  - Foot sfx plays twice in a loop of the walk animation.
  - AudioStream is exposed in the inspector so learners can swap it for another.
- Player: walk animation goes at double speed when running 

Related to #76 

https://github.com/user-attachments/assets/9504e287-f7cc-4a10-9aeb-8aa88b08a270


assets/third_party/sounds/player/Foot.ogg.license
```diff
@@ -0,0 +1,4 @@
+﻿SPDX-FileCopyrightText: d00121058
+SPDX-License-Identifier: CC0-1.0
+
+FX_001_footstep_grass_l.wav by d00121058 -- https://freesound.org/people/d00121058/sounds/390760/
```
wjt, line 4: Are you sure this is the right sound? To me the sound in the PR sounds very different to this linked sound.

JuanFdS: I may have mixed it up with the one in the stealth challenge!, let me check

JuanFdS: It seems the license was wrong in the original PR too D:, I'll check with @Arebuayon , if I can't find the original license I'll get a new sound

wjt (approved):

PR #530, opened 2025-05-08
Eliminate MusicalRockXylophone

### LINKED ISSUE #519: Remove MusicalRockXylophone
Currently the six MusicalRock nodes live inside a MusicalRockXylophone scene. It doesn't really do anything except sit between the `MusicPuzzle` and the rocks, proxying method calls and signals. And it makes it harder to customise the rocks themselves.

Instead, have the rocks sit directly below the MusicPuzzle node, removing the Xylophone abstraction.

wjt:
This was a wrapper around a fixed layout of 6 rocks, and served only to:

- Discover its rock children
- Proxy play() calls through to the correct rock
- Proxy note_played signals outwards

The fixed set of rocks was a barrier to varying the puzzle in different
levels.

Inline the child rocks into the two scenes that used the
musical_rock_xylophone scene. Move the rock discovery, playing, and
listening out to MusicPuzzle.

To allow the rocks to be grandchildren of MusicPuzzle rather than
direct children (so they can be grouped together into an intermediate
node if desired), introduce a new sequence_object group, place the rock
scene into it, and use this when discovering rocks. The name hints at a
future where the rocks may not be rocks, and the sequence may not
necessarily be a musical melody.

Fixes #519

This also incorporates a fix for an existing issue.


project.godot
```diff
@@ -58,6 +58,7 @@ throwing_enemy=""
 filling_barrels=""
 projectiles=""
 guard_enemy=""
+sequence_object=""
```
manuq, line 61: Unrelated but I see we have a mix of singular and plural in current groups.

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/music_puzzle.gd
```diff
@@ -47,6 +49,16 @@ func _ready() -> void:
 			hint_levels[i] = 0
 
 
+func _find_rocks() -> void:
+	_rocks.clear()
+
+	for object: Node in get_tree().get_nodes_in_group(&"sequence_object"):
+		if self.is_ancestor_of(object) and object is MusicalRock:
```
manuq, line 56: Nice combination of filtering by group and by children.

manuq (approved):

PR #533, opened 2025-05-08
Add tile rips to tileset

### LINKED ISSUE #426: Add Tile Rips to tilemap
[Tile Rips.zip](https://github.com/user-attachments/files/20001058/Tile.Rips.zip)

pablitar:
Adds the tile rips to the default tileset so it can be used in levels.

No fancy terrain and automatic placement has been configured, as that behavior cannot be deactivated, and may be detrimental to normal use.

Fixes #426


manuq (approved):
Excellent!

Issue #520, opened 2025-05-08
Remove automatic recolouring of musical rocks


wjt:
Currently the rocks have their `modulate` set based on the note that they play, in code. This makes them harder to customise. Remove this logic, and instead set the modulation on each rock directly.

wjt:
This is not quite as trivial as I had imagined because the "Kick" interact placeholder also gets modulated if we modulate the whole rock:

![Image](https://github.com/user-attachments/assets/edd7942d-f9b7-4c34-a2a5-bbff28a01a3f)

(Currently the AnimatedSprite2D is modulated.)

wjt:
The labels are fake in the sense that they're not actually shown at runtime, so the effect there is fine:

![Image](https://github.com/user-attachments/assets/dc5bdb53-49cf-400f-ae6d-cd3b118b97eb)

Maybe it's OK and we can live with this? The better alternative would be to draw six colours of rock but that's one for another day.

wjt:
I tried implementing this using `PropertyUtils` to export the AnimatedSprite2D's `modulate` property. It does work but it seems like a lot of complexity, and the resulting property is undocumented. I think I'll put up with the labels being modulated, or recolour the rocks myself.

PR #536, opened 2025-05-09
MusicalRock: Remove automatic modulation by note

wjt:
Previously, the colour of each MusicalRock was set in the rock's script, which set the modulate property on the sprite. The colour was chosen by picking 7 evenly-spaced hues for the 7 possible notes, and the same saturation and value for each.

This was convenient, especially since we only have one rock asset. But it makes it harder to have different appearances in different instances of the musical-rock/sequence puzzle. For example, you may want the objects to be four different shapes, not recoloured versions of the same asset.

Remove the logic that recolours the rock based on its note. Instead, modulate each rock in the scene that uses it, using the same colour that was previously generated from the note in code.

This looks a bit weird in the editor because the rock's collision shapes and "kick" interaction label also get modulated. But at runtime the collision shapes are invisible, and the label is not actually used – it just sets the position where the label belonging to the player will be positioned. And I consider modulating the sprite to be a temporary measure; it would be better to recolour the rocks.

(I [tried using PropertyUtils](https://github.com/endlessm/threadbare/pull/537) to export the modulate property of the AnimatedSprite2D. It solved this problem but in my opinion the cure was worse than the disease: it added more complexity to what is otherwise a very simple scene, and the resulting "Sprite Modulate" property in the inspector had no documentation.)

Fixes https://github.com/endlessm/threadbare/issues/520

### DIFF
--- a/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
+++ b/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
@@ -8,10 +8,7 @@ signal note_played
 const NOTES: String = "ABCDEFG"
 
 ## Note
-@export_enum("A", "B", "C", "D", "E", "F", "G") var note: String = "C":
-	set(_new_value):
-		note = _new_value
-		_modulate_rock()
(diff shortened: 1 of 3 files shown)


JuanFdS (approved):
👍🏾

PR #537, opened 2025-05-09
MusicalRock: Remove automatic modulation by note (PropertyUtils version)

wjt:
Previously, the colour of each MusicalRock was set in the rock's script, which set the modulate property on the sprite. The colour was chosen by picking 7 evenly-spaced hues for the 7 possible notes, and the same saturation and value for each.

This was convenient, especially since we only have one rock asset. But it makes it harder to have different appearances in different instances of the musical-rock/sequence puzzle. For example, you may want the objects to be four different shapes, not recoloured versions of the same asset.

Remove the logic that recolours the rock based on its note. Instead, export a `sprite__modulate` property (proxying the AnimatedSprite2D's modulate property) and, in each scene that uses the rocks, set this property to the value that was previously generated from the note.

musical_rock.gd must be made a `@tool` script in order for PropertyUtils to work; so guard the code in _ready() which modifies the AudioStreamPlayer2D's stream.

I renamed the AnimatedSprite2D to Sprite because otherwise the generated property name is animated_sprite_2d__modulate, rendered as Animated Sprite 2D Modulate in the inspector, which is long and ugly.

Fixes https://github.com/endlessm/threadbare/issues/520

### DIFF
--- a/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
+++ b/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
+@tool
 class_name MusicalRock
 extends StaticBody2D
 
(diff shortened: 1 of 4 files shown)


wjt:
This PR is just by way of comparison; I prefer the approach in #536 that doesn't use PropertyUtils.

PR #538, opened 2025-05-09
Music puzzle: Represent steps as nodes

### LINKED ISSUE #522: Specify music puzzle sequences numerically
Rather than referring to rocks by note, refer to them by their 0-based index. This will make it easier to reuse the puzzle in more generic contexts.

wjt:
Conceptually this music/sequence puzzle is a series of steps. Each step
has a sequence of nodes that the player must interact with in that
order, and a sign which provides a hint and lights up when that step is
solved. When all steps have been solved, the puzzle as a whole is
solved.

Previously, the sequences were represented as a string containing the
names of musical notes, which corresponded to a note property on
MusicalRock nodes. This makes the puzzle specific to musical melodies,
and is also error-prone: nothing prevented the level designer entering
the string "Sausages" as a "melody".

There was also a requirement that there be the same number of melodies
as bonfire signs, enforced with a warning in the editor.

Introduce a SequencePuzzleStep node, which contains the sequence for
this step and the corresponding bonfire sign. The sequence is
represented as an array of MusicalRock node references. This makes it
much more difficult to implement an unsolvable puzzle, and provides a
natural way to refer to the objects in the sequence directly. It also
makes it harder to get a mismatched number of signs and sequences. (But
not impossible: you could assign the same sign to multiple sequences.)

BonfireSign previously hardcoded a NodePath to its parent to find the
puzzle, which was convenient but error-prone – instead use a normal
reference, and disable interactivity if it is not provided.

Update both instances of the puzzle for these changes, and tweak the
template dialogue.

Fixes #522

### DIFF
--- a/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
+++ b/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
@@ -3,6 +3,7 @@
 class_name MusicalRock
 extends StaticBody2D
 
+## Emitted when the rock is struck by the player.
 signal note_played
 
 const NOTES: String = "ABCDEFG"
@@ -22,12 +23,13 @@ func _ready() -> void:
 
 
 func _on_interaction_started(_player: Player, _from_right: bool) -> void:
+	note_played.emit()
 	play()
 	interact_area.interaction_ended.emit()
 
 
+## Act as if the rock was struck by the player. Does not emit [member note_played].
 func play() -> void:
-	note_played.emit()
 	animated_sprite.play(&"struck")
 	audio_stream_player_2d.play()
 	await audio_stream_player_2d.finished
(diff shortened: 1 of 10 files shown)


wjt:
https://endlessm.github.io/threadbare/branches/endlessm/music-puzzle-steps-as-nodes/#quests/lore_quests/quest_001/1_music_puzzle/music_puzzle

https://endlessm.github.io/threadbare/branches/endlessm/music-puzzle-steps-as-nodes/#uid://7hoy2p14t6kc

wjt:
There should be no visible changes in the scenes – that's the point :)

scenes/quests/lore_quests/quest_001/1_music_puzzle/components/xylophone/musical_rock.gd
```diff
@@ -22,12 +23,13 @@ func _ready() -> void:
 
 
 func _on_interaction_started(_player: Player, _from_right: bool) -> void:
+	note_played.emit()
```
manuq, line 21: :+1: I remember we talked about doing this instead of having a "is demo" flag. Better.

manuq (approved):
Much better! This makes the puzzle much less prone to user input errors.

PR #540, opened 2025-05-09
Delete musical_rock_xylophone scene

wjt:
This should have been removed in commit c3745a38a58617622db5aacd07035481d524ad37.


wjt:
I'm going to go ahead and merge this without review since the scene I'm deleting is unused and also broken.

Issue #539, opened 2025-05-09
Integrate animated eternal loom


JuanFdS:
[Eternal Loom.zip](https://github.com/user-attachments/files/20122119/Eternal.Loom.zip)

PR #541, opened 2025-05-09
Eternal Loom: replace static image for animation

JuanFdS:
Fixes https://github.com/endlessm/threadbare/issues/539


wjt (approved):
This looks great!

As a follow-up we may later want to:

1. Have the loom only animate once you have taken some threads to it (though how would this work if you play more than 1 quest?)
2. Use the colour variations during the incorporation animation

JuanFdS:
> This looks great!
> 
> As a follow-up we may later want to:
> 
>     1. Have the loom only animate once you have taken some threads to it (though how would this work if you play more than 1 quest?)

I think we could track the completed quests in the GameState or something like that to keep a more permanent state of the game across levels.

> 
>     2. Use the colour variations during the incorporation animation

Yeah!, that's a good idea!

PR #543, opened 2025-05-09
Fix misplaced sound assets

JuanFdS:
While working on the different PRs for https://github.com/endlessm/threadbare/issues/76, I misplaced some files, reorganized them so they all are in this directory: `assets/third_party/sounds`.

Related to https://github.com/endlessm/threadbare/issues/76


wjt (approved):
Thanks! I noticed this the other day but didn't get around to moving the files!

PR #546, opened 2025-05-09
Fix bonfire issues

JuanFdS:


wjt:
You'd have to check with Steven but I think it's intentional that you can listen to all the unlit bonfires. I think the bug is that you shouldn't be able to listen to the second bonfire until you've lit the first.

JuanFdS:
> You'd have to check with Steven but I think it's intentional that you can listen to all the unlit bonfires. I think the bug is that you shouldn't be able to listen to the second bonfire until you've lit the first.

Ohhhh, good point. @PlayMatters , how should these work?

wjt:
I confirmed with @PlayMatters that the intended behaviour in the LoreQuest puzzle is:

- The current step of the puzzle should be interactive
- And any already solved steps

So initially only the first altar should be interactive; once solved, the first and second should be interactive; etc.

wjt:
I dropped that patch from the series

Issue #497, opened 2025-05-07
Ink Combat: Storyweaver Becomes Transparent


LauraFoglia:
Near the end of combat in instance 3, when I was hit by two blobs of ink, the Storyweaver character suddenly became transparent. I was able to finish the level while remaining invisible, including picking up the thread for that level.

https://github.com/user-attachments/assets/31f8b85e-ee27-4645-95ce-5dea4e5ab3e2

wjt:
I think this may have been fixed by https://github.com/endlessm/threadbare/pull/478 ?

LauraFoglia:
we think it's not the same, spoke with Juan before creating the issue just to be sure. Maybe related but not quite the same I believe-

JuanFdS:
> I think this may have been fixed by [#478](https://github.com/endlessm/threadbare/pull/478) ?

I think this is a different one, we should check if is still possible to reproduce or if it was fixed, but, I think that what happens in this issue is that the `got_hit` animation sets the modulate of the player, and when swapping modes, the modulate remains at that value.

hydrolet:
This happened to my daughter when playtesting last night.

PR #548, opened 2025-05-09
Fix: storyweaver becomes transparent when exiting combat

JuanFdS:
The issue could happen when a player switched from fighting mode to cozy mode while the got hit animation was playing.

Fixed by setting the modulate to 1 when exiting the fighting mode.

Fixes https://github.com/endlessm/threadbare/issues/497

### DIFF
--- a/scenes/game_elements/characters/player/components/player_fighting.gd
+++ b/scenes/game_elements/characters/player/components/player_fighting.gd
@@ -7,6 +7,9 @@ var is_fighting: bool = false
 @onready var hit_box: Area2D = %HitBox
 @onready var got_hit_animation: AnimationPlayer = %GotHitAnimation
 @onready var air_stream: Area2D = %AirStream
+@onready var player_sprite: AnimatedSprite2D = %PlayerSprite
+
+@onready var player: Player = self.owner as Player
 
 
 func _ready() -> void:
@@ -32,3 +35,10 @@ func _on_body_entered(body: Node2D) -> void:
 
 func _on_air_stream_body_entered(body: Projectile) -> void:
 	body.got_hit(owner)
+
+
+func _notification(what: int) -> void:
+	match what:
+		NOTIFICATION_DISABLED:
+			got_hit_animation.play(&"RESET")
+			got_hit_animation.advance(0)


scenes/game_elements/characters/player/components/player_fighting.gd
```diff
@@ -32,3 +35,11 @@ func _on_body_entered(body: Node2D) -> void:
...
 func _on_air_stream_body_entered(body: Projectile) -> void:
 	body.got_hit(owner)
+
+
+func _notification(what: int) -> void:
+	match what:
+		## When fighting is disabled, reset the modulate that could have been
+		## modifeid in the got_hit animation.
+		NOTIFICATION_DISABLED:
+			player_sprite.modulate.a = 1.0
```
manuq, line 45: This works, but I wonder if we can make use of the RESET animation that Godot creates.
```suggestion
		NOTIFICATION_DISABLED:
			got_hit_animation.play(&"RESET")
```
I'll give it a try.

JuanFdS: I tried that first and it didn't work! @manuq . I guess it was because the animation player was disabled as well. If you think of a way to playing that before disabling feel free to change it tho!

And btw, in case it helps: for testing this I temporarily changed one of the ink combats to have just 1 ink well with needed_amount = 1, and stretched the got hit animation to 20 seconds so it was a lot easier to reproduce.

JuanFdS: Mmmm, also, another way of fixing it I'm thinking of right now 🤔 : we could have that animationplayer to have process **_ALWAYS_** instead of inherits, so it will always finish the animation! That should work too, what do you think?

manuq: Yep, I wasn't able to reproduce the bug by calling RESET. @JuanFdS with your permission, I'm going to edit your commit.

JuanFdS: > Yep, I wasn't able to reproduce the bug by calling RESET. @JuanFdS with your permission, I'm going to edit your commit.

are you sure? @manuq , I applied that change and was able to reproduce it:

https://github.com/user-attachments/assets/a9deb752-cfe2-400f-ad65-f1ac0669a113

manuq: > > Yep, I wasn't able to reproduce the bug by calling RESET. @JuanFdS with your permission, I'm going to edit your commit.
> 
> are you sure? @manuq , I applied that change and was able to reproduce it:
> 
>  Grabacion.De.Pantalla.2025-05-09.183657.mp4

Ohh! Let me check...

manuq: @JuanFdS thanks for catching the error. I made it work by calling `got_hit_animation.advance(0)`. From the documentation of AnimationPlayer, this is needed, otherwise it plays in the next processing, which won't happen for a disabled node.

I really prefer using RESET than doing what the reset animation does, because the animation may change to animate other properties, and then we'll have to change the script.

Yes, I think the other alternative is to set the AnimationPlayer node to always process. This works fine though. I confirmed by changing "got hit" to last 10 seconds and stay transparent.

[Grabación de pantalla desde 2025-05-09 19-07-03.webm](https://github.com/user-attachments/assets/ff5d30dd-dc6a-4d33-b52b-6ef5a52d1d9f)

PR #551, opened 2025-05-09
Add defeated animation when caught in stealth level

Arebuayon:


scenes/game_elements/characters/player/components/player.gd
```diff
@@ -11,6 +11,7 @@ enum Mode {
 	COZY,
 	## Player is engaged in combat. Player can use combat actions.
 	FIGHTING,
+	CAUGHT
```
manuq, line 14: I wonder if this really has to be a new player mode. For the stealth game the player is "cozy". The only thing that's needed when caught is to play an animation, and stop handling player input.

scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -396,17 +396,19 @@ func _is_sight_to_point_blocked(point_position: Vector2) -> bool:
 ## Otherwise, it returns null.
 func _player_in_sight() -> Node2D:
 	if instant_detection_area.has_overlapping_bodies():
-		return instant_detection_area.get_overlapping_bodies().front()
+		for body in instant_detection_area.get_overlapping_bodies():
+			if body is Player and not body.caught:
+				return body
```
manuq, line 401: Why has this changed? It is not explained.

scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -396,17 +396,19 @@ func _is_sight_to_point_blocked(point_position: Vector2) -> bool:
...
 
 	if not detection_area.has_overlapping_bodies():
 		return null
 
-	var player: Node2D = detection_area.get_overlapping_bodies().front()
+	for body in detection_area.get_overlapping_bodies():
+		if body is Player and not body.caught:
+			if not _is_sight_to_point_blocked(body.global_position):
+				return body
```
manuq, line 409: Same here, I wonder why this change is needed.

manuq (changes requested):
I would like to give this a try myself.

scenes/game_elements/characters/enemies/guard/components/guard.gd
Arebuayon: Hi! I changed this because, to allow defeat animations to play, I could no longer rely on PROCESS_MODE_DISABLED to stop everything. This created a problem: if one guard detected the player and others saw them in the area, multiple guards would trigger alerts for an already detected player.

Iterating through all the overlapping bodies might seem unnecessary if only the player should be in that detection layer, but I added this logic as a safety precaution. If someone accidentally adds another node to the detection layer or group, this prevents errors such as trying to access .caught in a scene that doesn't define it

manuq: Thanks for the explanation @Arebuayon !

manuq:
Converting this PR to draft while I iterate on it.

PR #553, opened 2025-05-09
Fray's end: fix nodes' parents

JuanFdS:
Some NPCs weren't children of the NPCs node, and there was a tree that was a child of another tree. Fixed those cases.


manuq (approved):
Excellent!

Issue #511, opened 2025-05-08
Add guard enemy spritesheet template


manuq:
Just like player "egg guy" template.

Issue #510, opened 2025-05-08
Add throwing enemy spritesheet template


manuq:
Just like the player "egg guy" template.

PR #554, opened 2025-05-12
More character templates

manuq:
Fix https://github.com/endlessm/threadbare/issues/510
Fix https://github.com/endlessm/threadbare/issues/511


wjt (approved):
I love the enemy's grumpy expression!

PR #559, opened 2025-05-12
MusicalRock: Allow customising sprite frames

### LINKED ISSUE #524: Expose SpriteFrames for MusicalRock
The audio stream is already customisable but the SpriteFrames are not.

wjt:
This follows a similar pattern used elsewhere in the project, including adding a warning if the expected animations are not defined.

This is primarily intended for StoryQuests but you could imagine we could also have different-coloured versions of the base rock asset, rather than using the modulate property as we currently do.

This PR adds a sprite_frames property to the rock, and uses it to add a new generic asset to the StoryQuest template.

Fixes #524


manuq (approved):
Nice. For the modulate, I wonder if we should expose an optional "Color" property in MusicalRock that internally sets the SpriteFrames modulate. Like in the filling barrel https://github.com/endlessm/threadbare/blob/aba018f1804f6d361be840dca36071d944eeb706/scenes/game_elements/props/filling_barrel/components/filling_barrel.gd#L26-L28

wjt:
Nice idea!

PR #567, opened 2025-05-13
Shaker: Vibrate controller if in use

wjt:
I couldn't resist adding this after testing the camera shake effect.

Stacked on:

- #566


manuq:
Nice touch! I need to try it with joycons.

scenes/game_elements/fx/shaker/components/shaker.gd
```diff
@@ -62,6 +66,8 @@ func _ready() -> void:
 func shake(intensity: float = shake_intensity, time: float = duration) -> void:
 	noise.seed = randi()
 	started.emit()
+	if _last_controller_id >= 0:
+		Input.start_joy_vibration(_last_controller_id, 0.5, 0.5, time)
```
wjt, line 70: I've only tested this with the controller I have plugged into my PC which is an 8BitDo Xbox controller clone. The two parameters are the magnitude of the weak and strong motors, each between `0` and `1`. `0.5` is just a number. It feels OK.

https://docs.godotengine.org/en/stable/classes/class_input.html#class-input-method-start-joy-vibration

wjt:
One potential problem with adding this Easter egg is that once you have joypad vibration, you need a setting to disable it! But perhaps someone will contribute that.

But I imagine that this simple logic to determine whether a keyboard or joypad is in use could be extracted to some other component, which would then make the HUD input hints show different images.

manuq (approved):
I didn't have a chance to test this yet but looks good!

PR #569, opened 2025-05-13
Remove property utils

manuq:
See main motivations in commit 0815083d .


scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -28,29 +28,19 @@ class_name CollectibleItem extends Node2D
...
-				"hint_string": ""
-			}
-		)
-
-	return properties
+func _validate_property(property: Dictionary) -> void:
+	match property.name:
+		"dialogue_title":
+			if not collected_dialogue:
+				property.usage |= PROPERTY_USAGE_READ_ONLY
```
wjt, line 47: Does the `property` get reset each time this function is called? Or do you need to clear the flag when `collected_dialogue` is `true`?

scenes/game_elements/props/decoration/decoration.gd
```diff
@@ -4,19 +4,59 @@
...
+		return
+	if texture != null:
+		sprite_2d.texture = texture
+
+
+func _set_flip_h(new_flip_h: bool) -> void:
+	flip_h = new_flip_h
+	if not is_node_ready():
+		return
+	if flip_h != null:
```
wjt, line 38: Can a `bool` be `null`?

scenes/game_elements/props/decoration/decoration.gd
```diff
@@ -4,19 +4,59 @@
...
+		return
+	if flip_v != null:
+		sprite_2d.flip_v = flip_v
+
+
+func _set_offset(new_offset: Vector2) -> void:
+	offset = new_offset
+	if not is_node_ready():
+		return
+	if offset != null:
```
wjt, line 54: Ditto here: I didn't think a `Vector2` can be `null`.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
manuq: I think that the vaildation happens each time any property changes:
[Grabación de pantalla desde 2025-05-13 14-43-20.webm](https://github.com/user-attachments/assets/fe33e000-4cda-45f2-8651-91df7fbe778c)

wjt: Got it!

wjt (approved):
LGTM

Issue #512, opened 2025-05-08
Make projectile template-able


manuq:
The projectile asset currently can't be changed.

PR #573, opened 2025-05-13
Template for combat projectile

manuq:
Fix https://github.com/endlessm/threadbare/issues/512


scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
```diff
@@ -250,6 +254,7 @@ func shoot_projectile() -> void:
 		projectile.node_to_follow = player
 	projectile.small_fx_scene = projectile_small_fx_scene
 	projectile.big_fx_scene = projectile_big_fx_scene
+	projectile.trail_fx_scene = projectile_trail_fx_scene
 	projectile.speed = projectile_speed
 	projectile.duration = projectile_duration
```
wjt: I don't think you should change this in this PR, but, in effect what all this code is doing is papering over the fact that `PROJECTILE_SCENE` is a constant. Another approach would be to have instances of combat define an inherited scene (or a copy of `projectile.tscn`, duplicating its structure but referencing the same scripts), set properties on that, and then have the throwing enemy have a `projectile_scene: PackedScene` export.

wjt: As an aside, I think [this](https://www.reddit.com/r/godot/comments/129sgc8/what_is_the_intended_workflow_for_inheriting/) is one of the threads that put me off using inherited scenes though:

> My advice is to avoid it if possible.

> scene inheritence is best used sparingly, but does come in handly if you have a bunch of things that are behaviorally identical but have a few different properties

>> What is the intended workflow for inheriting scenes?
>
> Not using the feature.

scenes/game_elements/props/projectile/components/inkblob_spriteframes.tres
```diff
@@ -2,50 +2,50 @@
 
 [ext_resource type="Texture2D" uid="uid://sgprgh3jbreo" path="res://scenes/game_elements/props/projectile/components/ink_blob.png" id="1_enimy"]
```
wjt, line 3: I like how the generated `id` here is almost "enemy".

wjt (approved):

scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
manuq: You are right! I don't know how I reached this state, but the enemy should export a PackedScene for the projectile scene directly.

Issue #576, opened 2025-05-14
Remove “curly quotes” in dialogue


wjt:
As tracked on https://github.com/endlessm/threadbare/issues/201, the font we are currently using doesn't have glyphs for curly quotes and apostrophes. And on the web platform build of Godot, fallback to the system font does not work. (This is quasi-documented at https://docs.godotengine.org/en/latest/tutorials/ui/gui_using_fonts.html#system-fonts and tracked in https://github.com/godotengine/godot/issues/84590.)

For the time being, we need to avoid using such characters in dialogue. We noticed a few during a recent playthrough.

PR #577, opened 2025-05-14
Template: Use straight double quotes in dialogue

wjt:
The font we are currently using does not have glyphs for “curly quotes” and on the web platform Godot does not fall back to system fonts.

Fixes https://github.com/endlessm/threadbare/issues/576


wjt:
@jbourqueendless you introduced these in #515. I'm curious about the workflow that leads to you typing “fancy quotes”. Are you composing the dialogue in some external editor that replaces your quotes?

(I'm the kind of person who knows how to type curly quotes and makes a point of doing so even at the cost of slowing down my typing. Maybe you are too :-)

manuq:
Funny! I did a playtest myself last night and wake up to fix this. You beated me.

manuq (approved):

Issue #568, opened 2025-05-13
Reduce size of exported game


wjt:
The exported `.pck` file is 29M compressed, 34MB when unpacked. This seems like a lot to me given the current scope of the project. Much of this size is a small number of large assets, and a large number of small assets!

Possible approaches here:

* Find large, uncompressed assets and compress them
* Find large, compressed assets and compress them more aggressively without compromising audible/visible quality
* Remove unused assets from the tree (but it is useful to have a library of assets that contributors can use, so e.g. I don't think it would be valuable to remove assets from Tiny Swords that aren't used - individually they're small, and previously-unused assets have been put into use in contributed storyquests)
* I believe that every asset in the source tree is included in the exported `.pck` file. Perhaps there is a way to only include referenced assets - but without having to manually update a list?

I think a good target would be to cut the `.pck` file size by 10%, though more is welcome.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Optimization & Performance | Problem Solving | 150 |
| Engineering: Optimization & Performance | Creativity & Design Thinking | 75 |
| Engineering: Optimization & Performance | Technical Literacy | 150 |

PR #579, opened 2025-05-14
Convert river sound to Ogg Vorbis

wjt:
Previously the sound of the river in the LoreQuest music puzzle was a 17 MB WAV file. By way of comparison, the project source tree is 57 MB in total, so this is 30% of the size of the project. (It was converted at export time to QOA, but in the context of hundreds of learners cloning the source tree, 17 MB per clone adds up fast.)

Convert it to Ogg Vorbis with the default quality setting (3):

    oggenc \
        -a 'Pfannkuchn' \
        -t 'Small River 1 - Fast - Semi Close'  \
        ./assets/third_party/sounds/ambient/459405__pfannkuchn__small-river-1-fast-semi-close.wav

resulting in a more reasonable 750 KB file.

https://github.com/endlessm/threadbare/issues/568


manuq (approved):
Yes please!

PR #583, opened 2025-05-14
Adjust intro template placeholder dialogue 

jbourqueendless:
Adjust intro placeholder dialogue to better match learner StoryQuest contributor workflow.

### DIFF
--- a/scenes/quests/story_quests/template/0_intro/intro.dialogue
+++ b/scenes/quests/story_quests/template/0_intro/intro.dialogue
@@ -4,9 +4,10 @@
 ~ start
 do animation_player.play(&"walk_on")
 do animation_player.animation_finished
-Change this text by selecting the Cinematic node and opening the Dialogue resource in the Inspector.
+Inside your StoryQuest intro folder, double click intro.dialogue to edit the text.
 Press Enter to add a new line. Each line becomes a new dialogue box in-game.
 Text can be placed during and after the walking animations. Try experimenting!
 do animation_player.play(&"walk_off")
-Set the next scene using the "Next Scene" property on the Cinematic node.
+Select Cinematic node; drag your StoryQuest dialogue to the dialogue field in the Inspector.
+Set the next scene using the "Next Scene" field in the Inspector.
 => END


scenes/quests/story_quests/template/0_intro/intro.dialogue
```diff
@@ -4,9 +4,10 @@
 ~ start
 do animation_player.play(&"walk_on")
 do animation_player.animation_finished
-Change this text by selecting the Cinematic node and opening the Dialogue resource in the Inspector.
+Inside your StoryQuest intro folder, double click intro.dialogue to edit the text.
```
manuq, line 7: I was about to tell you about using rich text for setting mono font in the file name like in:
```
Inside your StoryQuest intro folder, double click [code]intro.dialogue[/code] to edit the text.
```
But it doesn't work because of our current font! So I said nothing.

By the way, I really think that it's better to access the dialogue by opening the intro scene, selecting the Cinematic node, then opening the Dialogue from the Inspector (what it was previously mentioned).

manuq (approved):
Looks good to me. I made an observation but I'm no teacher :)

scenes/quests/story_quests/template/0_intro/intro.dialogue
wjt: I guess the thinking behind this change is that it's harder to edit the wrong dialogue file if you go via the filesystem.

The other approach would be to swap the instruction round:

1. Drag the correct dialogue (your copy, not the template) onto the cinematic node
2. Now it doesn't matter whether you go via the inspector or the filesystem

jbourqueendless:
Yeah. @manuq the way you are describing is also my natural flow to finding the dialogue and how it might work for me in a scene.

However, as the template now stands that may accidentally encourage someone to open the intro.dialogue , not realizing that the template scene they have copied for themselves still connects to the "template" intro.dialogue and that they really should be replacing with their own file. 

So this is an attempt to steer someone away from accidentally changing the template dialogue file.

manuq:
> Yeah. @manuq the way you are describing is also my natural flow to finding the dialogue and how it might work for me in a scene.
> 
> However, as the template now stands that may accidentally encourage someone to open the intro.dialogue , not realizing that the template scene they have copied for themselves still connects to the "template" intro.dialogue that they really should be replacing with their own file.
> 
> So this is an attempt to steer someone away from accidentally changing the template dialogue file.

I see @jbourqueendless. Then the problem (as Will stated above) is that the duplicate of the template folder has the dialogue pointing to the original template.

PR #584, opened 2025-05-14
Lore on top of template

manuq:
Fix https://github.com/endlessm/threadbare/issues/585

@wjt @jbourqueendless I wonder if we want this. The template settings (assets, etc) are the default even for Lore. And Lore "dresses" eg. the player to be the StoryWeaver.

[Grabación de pantalla desde 2025-05-14 11-25-13.webm](https://github.com/user-attachments/assets/21ded949-ddef-4bdc-a6f5-e02d69eb9f72)

I have only changed a few things to demo the concept (player name, player SpriteFrames, throwing enemy SpriteFrames). And applied them to only one Lore scene, while clearing one template scene. If we decide to do this, we better do it soon as it is a change that will affect all the project!

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -20,7 +20,7 @@ enum State {
 	RETURNING,
 }
 
-const DEFAULT_SPRITE_FRAMES = preload("uid://c4kqsgl2pcofo")
+const DEFAULT_SPRITE_FRAMES = preload("uid://ovu5wqo15s5g")
 
 const LOOK_AT_TURN_SPEED: float = 10.0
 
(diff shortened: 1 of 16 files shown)


wjt:
Personally, I think yes, this is good. It will be more awkward for us to work with, but will make it more likely that generic or learner-made assets will be used in storyquests.

wjt:
Having discussed with @hydrolet and @jbourqueendless: they also like the idea of this change, but only if it will not disrupt our other work getting things ready for workshops.

manuq:
Ready for review. I will do sounds in a separate commit.

manuq:
@wjt thanks for checking! Ready for review again.

wjt (approved):
LGTM!

Issue #586, opened 2025-05-14
Eternal Loom collectibles: Use sprites for the world


manuq:
Current animation uses the sprites that were drawn specifically for the HUD:

https://github.com/user-attachments/assets/4d4799dd-1d03-487f-b30d-c124e499b536

PR #587, opened 2025-05-14
Eternal Loom: Use world sprites for collectibles

manuq:
The animation for returning the collected Memory, Imagination and Spirit to the Eternal Loom was using the sprites for the HUD. Use the sprites for the world instead.

Fix https://github.com/endlessm/threadbare/issues/586


wjt (approved):
Looks better! It also makes conceptual sense because those versions are what the player picked up.

PR #591, opened 2025-05-15
Fix large file check in GitHub Actions CI

### LINKED ISSUE #589: Add CI check for large files not stored in Git LFS
In #579 I removed a 17MB WAV file from the repo. Unfortunately `*.wav` was not set up to be stored in LFS so that file is in the history forever and every clone will have to fetch it.

We should:

- track WAV files in LFS
- add a CI check that no file in the repo (not LFS) is 1MB or greater. Surely precommit has a hook for this.

wjt:
As outlined in #589, a misconfiguration/misfeature of `pre-commit` meant that we committed a 17 MB file to the repo despite having a `pre-commit` hook that should have prevented this.

Reconfigure the hook to prevent this in future. Migrate existing (and future) `.wav` files to Git LFS. Add a couple more nice-to-have pre-commit hooks.

Fixes #589

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,8 @@ repos:
   - id: check-added-large-files
     args:
       - --enforce-all
+  - id: check-case-conflict
+  - id: check-illegal-windows-names
   - id: check-yaml
   - id: end-of-file-fixer
     exclude: \.svg$


wjt:
I pushed an extra patch that adds a 2MB empty file, and it was correctly caught:

https://github.com/endlessm/threadbare/actions/runs/15041922991/job/42275429130?pr=591

```
check for added large files..............................................Failed
- hook id: check-added-large-files
- exit code: 1

massive.txt (2048 KB) exceeds 500 KB.
```

.pre-commit-config.yaml
```diff
@@ -13,6 +13,8 @@ repos:
     exclude: \.svg$
   - id: check-yaml
   - id: check-added-large-files
+    args:
+      - --enforce-all
```
manuq, line 12: :+1:

.gitattributes
```diff
@@ -8,5 +8,6 @@
 *.ogg filter=lfs diff=lfs merge=lfs -text
 *.png filter=lfs diff=lfs merge=lfs -text
 *.ttf filter=lfs diff=lfs merge=lfs -text
+*.wav filter=lfs diff=lfs merge=lfs -text
```
manuq, line 11: Oh yes, the missing one!

manuq (approved):

Issue #526, opened 2025-05-08
Rename music puzzle components & move to a new home


wjt:
* MusicPuzzle: SequencePuzzle
* MusicalRockXylophone: deleted #519 
* BonfireSign: SequenceHint (?)
  - is_ignited: is_solved
* Note: becomes an integer, covered by #522

wjt:
As for where thing should live:

- scenes/game_logic/sequence_puzzle.gd
- scenes/game_logic/sequence_puzzle_step.gd
- scenes/game_logic/props/sequence_puzzle_object/sequence_puzzle_object.tscn
  - scenes/game_logic/props/sequence_puzzle_object/components/sequence_puzzle_object.gd
- scenes/game_logic/props/sequence_puzzle_hint_sign/sequence_puzzle_hint_sign.tscn
  - scenes/game_logic/props/sequence_puzzle_hint_sign/components/sequence_puzzle_hint_sign.gd

It's kind of an awkwards split but it matches the other minigames.

hydrolet:
this works for us!

PR #592, opened 2025-05-15
Template-ify music puzzle bonfire sign

### LINKED ISSUE #521: Make music puzzle altars customisable
The idea we hatched is that each altar should have 3 animations:

1. The idle animation (which in the current puzzle would be a single frame, with the symbol precomposed onto the altar)
2. The interact animation (optional, not present in the current puzzle)
3. The "solved" animation (which in the current puzzle would have the lit-up symbol, fire animation, and altar precomposed)

wjt:
Previously, the BonfireSign had a hardcoded asset, with a fixed set of hints that could be shown on the front. This limited the possibilities for alternative sequence puzzles – even if they were to use the same shifted-melody concept but with different permutations!

Clean up some unused code. Merge the bonfire assets into one file, exporting 4 sprite sheets for the 4 different hints. Add a more generic asset of a signpost as the default. Introduce the optional ability to have the sign play an animation while demonstrating the sequence. Fix the wobble duration of the rocks so that the demonstration can be synced to the animation. Allow customising the sound effects, with the defaults being silence instead of the fire sounds.

The name is now very misleading but I will fix that for the whole component in #526.

Fixes #521


scenes/quests/lore_quests/quest_001/1_music_puzzle/components/bonfire_sign/bonfire_sign.gd
```diff
@@ -56,7 +61,15 @@ func ignite() -> void:
...
 func _on_interact_area_interaction_started(_player: Player, _from_right: bool) -> void:
-	await puzzle.play_demo_melody_of_fire(self)
+	if demonstrate_sequence.has_connections():
+		demonstrate_sequence.emit()
+	else:
+		demonstration_finished()
+
+
+## Should be called by the handler of [signal demonstrate_sequence] when the demonstration is
+## complete.
```
manuq, line 91: :+1:

manuq (approved):
Excellent! All 7 commits make sense.

Issue #594, opened 2025-05-15
Bug: Ink combat round 2 impossible


manuq:
The ink blobs are colliding with the water that separates the player from the inkwells, making it impossible to pass the level. Looks like it's due to a changes in the water tiles in https://github.com/endlessm/threadbare/pull/588

https://github.com/user-attachments/assets/cf837644-0513-4769-bb5a-fedfb47043a0

manuq:
Found the issue: these polygons added to the water tile:

![Image](https://github.com/user-attachments/assets/69dc28df-448f-4b73-bc5c-034726138b29)

@PlayMatters were those intentional?

PR #596, opened 2025-05-15
Tileset: Fix collisions in water tile

manuq:
Remove 3 polygons introduced in 506314a3 by accident.

For understanding the regression, I compared the collision shapes in all tree water tiles. There are 3 physics layers in the tileset:

- Layer 0 is for "walls"
- Layer 1 is for "sight occluders"
- Layer 2 is for "non walkable floor"

There are also 3 water tiles, the original plus 2 alternative tiles. So this means I compared a 3x3 matrix. Checking before/after, I realized that 3 polygons were added to physics layer 0 of the original tile. That was the only change. Removing the polygons fixes the regression.

For future reference, the 3 water tiles differ only in physics layer 2 "non walkable floor":
- Original tile: Full rectangle. So it's not walkable.
- Alternative 1: Empty. So the player can walk over it.
- Alternative 2: Bottom half rectangle. So the player can walk over the top half only (used in Fray's End for the pier)

Fix https://github.com/endlessm/threadbare/issues/594


wjt (approved):

manuq:
> This fix seems correct to me. The shapes that were added in [506314a](https://github.com/endlessm/threadbare/commit/506314a31e06ac02ce8b83b056c6eab71a60018f) look to be a simple error: the shapes are those of the top-left mine elevation tile (plus a random bonus polygon).

Ah, good finding!

PR #607, opened 2025-05-15
Rename music puzzle to sequence puzzle

### LINKED ISSUE #526: Rename music puzzle components & move to a new home

wjt:
The music puzzle logic and its element have now been made generic enough to be used for various different sequence puzzles. So we can now rename all the scenes, scripts, and some lingering music and rock references in the API, and move the files out of the LoreQuest into generic directories.

Fixes #526

----

One remaining piece is that I can make the musician a generic NPC but I'll do that as a follow-up.


manuq:
> One remaining piece is that I can make the musician a generic NPC but I'll do that as a follow-up.

I wonder if we really need this. Asking over Discord.

manuq (approved):
All good!

Issue #597, opened 2025-05-15
Template: Set no sounds as default


manuq:
Currently some game elements export sounds, and has the "Lore" sounds by default. For example, the Throwing Enemy has a constant "flying" sound and a "spit" sound when throwing. Then the template removes these sound effects.

Let's do the opposite: the template has these properties empty (so no sound), and the scenes in Lore set the specific sounds.

The only exception is the musical objects of the puzzle game.

PR #621, opened 2025-05-15
Template mute sounds

manuq:
I know that ideally the 1st commit should be split. But I ran out of time today. I can do it tomorrow!

Fixes https://github.com/endlessm/threadbare/issues/597

### DIFF
--- a/scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
+++ b/scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=77 format=4 uid="uid://db6vhi7s2f37c"]
+[gd_scene load_steps=81 format=4 uid="uid://db6vhi7s2f37c"]
 
 [ext_resource type="Script" uid="uid://dnp0tjloec2d7" path="res://scenes/game_logic/stealth_game_logic.gd" id="1_f6lvl"]
 [ext_resource type="PackedScene" uid="uid://2rbpl811wlv1" path="res://scenes/game_elements/props/background_music/background_music.tscn" id="1_t8tj0"]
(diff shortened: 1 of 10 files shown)


scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
```diff
@@ -855,6 +863,10 @@ curve = SubResource("Curve2D_vnsq3")
 [node name="Guard2-GoingInCircles" parent="EnemyGuards" node_paths=PackedStringArray("patrol_path") instance=ExtResource("8_lbv6i")]
 position = Vector2(1567, 101.949)
 sprite_frames = ExtResource("23_5q0mi")
+alerted_sound_stream = ExtResource("24_agwy0")
+footsteps_sound_stream = ExtResource("25_dsau0")
+idle_sound_stream = ExtResource("26_w5e5c")
+alert_others_sound_stream = ExtResource("27_lp8w2")
```
wjt, line 869: This works for now, but I wonder whether we might want to define an inherited scene, or a copy of the scene, to avoid having to remember to set these on each guard in the level. (Ditto InkDrinkers.) Specifically I imagine we may some day want to replace the asset on the guards; or replace the stomach-churning spitting noise on the InkDrinker.

wjt (approved):
Fine by me – I think the split is OK.

scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
manuq: Yes, we are facing the same problems learners will have when using the template elements. I haven't tried inherited scenes myself yet. I have that pending after your feedback on projectiles.

It wasn't that bad to use copy/paste in the editor to do this PR. But is true that the same override applied everywhere is a lot of redundance and prone to errors. So yes, let's try inherited scenes and see how it goes.

manuq:
Ouch I left this unmerged and now it conflicts. I will redo it.

Issue #623, opened 2025-05-15
Integrate new project icon


manuq:
To replace the placeholder.

@jofilizola will provide the asset.

PR #631, opened 2025-05-15
New project icon

manuq:
Replace the placeholder with a new icon.

Fixes https://github.com/endlessm/threadbare/issues/623


wjt (approved):
Looks great!

(At some point we will need other variants, e.g. some platforms require icons to be round, or defined by a central safe region plus a mask.)

PR #632, opened 2025-05-15
HintSign: Make demonstrating sequence optional

### LINKED ISSUE #525: Make hint sign demoing sequence optional
Currently, interacting with a hint sign always demoes the sequence on the objects.

Now that it is possible to specify an animation for the sign itself when interacted with, we may not want the rocks to also play the sequence. Make this configurable.

wjt:
Previously, interacting with a hint sign would always demonstrate the sequence. Now that it is possible to supply a `hint` animation to play on the sign itself (as is done in the template), some level designers may prefer to not demonstrate the sequence and make the puzzle more than a memory/imitation game.

For example, the template puzzle has two signs. Each sign shows a different series of symbols that have a natural correspondence to a coloured object (e.g. green = leaf). You could imagine changing this puzzle so that the player learns the correspondence from the first sign (which demonstrates the sequence of objects) and then has to interpret the symbols alone on the second sign. (They can replay the first sign if they have forgotten.)

I also adjusted the interact prompt on the sign.

Fixes #525

### DIFF
--- a/scenes/game_elements/props/sequence_puzzle_hint_sign/components/sequence_puzzle_hint_sign.gd
+++ b/scenes/game_elements/props/sequence_puzzle_hint_sign/components/sequence_puzzle_hint_sign.gd
@@ -14,14 +14,25 @@ const DEFAULT_SPRITE_FRAMES: SpriteFrames = preload("uid://b5pj1pt7r6hdg")
 ## The animations which must be defined for [member sprite_frames]. The [code]idle[/code]
 ## animation is used when [member is_solved] is false;
 ## [code]solved[/code] is used when [member is_solved] is true.
-## Optionally, a [code]hint[/code] animation can be defined, which will be played when the player
-## interacts with the sign to see a demonstration of the sequence.
+##
+## A [code]hint[/code] animation can be defined, which will be played when the player
+## interacts with the sign. This is optional if [member interact_demonstrates] is [code]true[/code]
+## (the default) and required otherwise.
 const REQUIRED_ANIMATIONS: Array[StringName] = [&"idle", &"solved"]
 
 ## Animations for this object. The SpriteFrames must have specific animations.
 ## See [constant SequencePuzzleHintSign.REQUIRED_ANIMATIONS].
 @export var sprite_frames: SpriteFrames:
 	set = _set_sprite_frames
+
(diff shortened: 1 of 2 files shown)


scenes/game_elements/props/sequence_puzzle_hint_sign/components/sequence_puzzle_hint_sign.gd
```diff
@@ -136,8 +147,12 @@ func _set_solved_ambient_sound(new_sound: AudioStream) -> void:
 
 func _get_configuration_warnings() -> PackedStringArray:
 	var warnings: PackedStringArray
+	var required_animations: Array[StringName] = REQUIRED_ANIMATIONS.duplicate()
+
+	if not interact_demonstrates:
+		required_animations.append(&"hint")
```
manuq, line 153: I find it a bit hard to read because "hint" is supposed to be optional but is added to a variable named "required". What about?

```suggestion
	var all_animations: Array[StringName] = REQUIRED_ANIMATIONS.duplicate()

	if not interact_demonstrates:
		all_animations.append(&"hint")
```

scenes/game_elements/props/sequence_puzzle_hint_sign/components/sequence_puzzle_hint_sign.gd
```diff
@@ -136,8 +147,12 @@ func _set_solved_ambient_sound(new_sound: AudioStream) -> void:
...
 
-	for animation in REQUIRED_ANIMATIONS:
+	for animation in required_animations:
```
manuq, line 155:
```suggestion
	for animation in all_animations:
```

scenes/game_elements/props/sequence_puzzle_hint_sign/sequence_puzzle_hint_sign.tscn
```diff
@@ -36,7 +36,7 @@ unique_name_in_owner = true
 position = Vector2(-2, 1)
 collision_layer = 0
 disabled = true
-action = "Listen"
+action = "Examine"
```
manuq, line 39: :+1:

manuq (approved):
Nice to see the original implementation back.

Issue #488, opened 2025-05-07
Sokoban: Add controller bindings for Undo and Reset


wjt:
Currently there are no controller bindings for these actions. It would be nice to add them. Personally I would expect the Xbox `B` button to mean "undo", but I'm not sure what button should be "reset".

Issue #403, opened 2025-04-30
On-screen control hints assume keyboard


cassidyjames:
This threw my kid off a bit when we were playtesting with a controller—they assumed they had to use the keyboard but I reassured them the controller indeed worked. We should probably swap to controller-style hints when a controller is detected. We could also dynamically swap the hints based on whether a controller or keyboard was last used, e.g. in case someone has a controller passively connected but is using their keyboard.

wjt:
I added some code recently which tracks whether the last input device was a controller or a keyboard:

https://github.com/endlessm/threadbare/blob/c5b84f49432bc21fa12a935dcba31a79949956ab/scenes/game_elements/fx/shaker/components/shaker.gd#L45-L47


https://github.com/endlessm/threadbare/blob/c5b84f49432bc21fa12a935dcba31a79949956ab/scenes/game_elements/fx/shaker/components/shaker.gd#L89-L93

I imagine we could move this to a different global, with a signal for when the input method changes, and a property to get the current controller device id (`-1` for "it's not a joypad"). Then the HUD scenes would gain a second property, the image to use for the joypad input hint, and swap in response to the signal.

Then we have the fun of showing A, × or B for the "interact" action depending on whether the controller uses xbox, sony, or nintendo labels. https://github.com/nathanhoad/godot_input_helper looks like an appealingly small addon, by the same person as Dialogue Manager, which could save writing out a lot of code of our own.

manuq:
> https://github.com/nathanhoad/godot_input_helper looks like an appealingly small addon, by the same person as Dialogue Manager, which could save writing out a lot of code of our own.

It also does input mapping, nice!

LauraFoglia:
@manuq podríamos asignar a las chicas al issue? ya lo estan trabajando

manuq:
@LauraFoglia yes, sorry!

PR #635, opened 2025-05-15
Add controller bindings for Sokoban undo and reset

wjt:
Undo is Xbox B, which is consistent with this often meaning "back" in menus (including the menus in this game).

Reset is Xbox X, which is quite arbitrary but perhaps the mnemonic can be that "X" is destructive?

Fixes https://github.com/endlessm/threadbare/issues/488


manuq (approved):
Looks good. The UI still shows keys only "Z", "R", "Space", "Shift" but that's a separate issue.

wjt:
I think that is adequately tracked by https://github.com/endlessm/threadbare/issues/403

Issue #580, opened 2025-05-14
Invert mushroom disappearing logic


PlayMatters:
The animated mushrooms (Hushrooms) appear when the should disappear. This needs to be fixed so that when players are near a hushroom, they disappear into then ground, returning once the player leaves.

PR #642, opened 2025-05-16
hiding_mushroom: Simplify behaviour

wjt:
Previously, the behaviour of the mushroom was configurable: it could be configured to start hidden, or start visible; and it could be configured to stay hidden/revealed once its state changes. I found the logic confusing, particularly the dynamic redefinition of the "stay hidden/revealed" property based on the value of is_hidden. And the behaviour of the four instances in Fray's End was incorrect: they are supposed to start revealed, and hide when the player gets near.

Remove all configuration from the mushroom. Make it always follow the behaviour of being visible unless the player is nearby.

Fixes https://github.com/endlessm/threadbare/issues/580

### DIFF
--- a/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
+++ b/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
@@ -3,21 +3,6 @@
 @tool
 extends Node2D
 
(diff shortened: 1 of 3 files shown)


manuq (approved):
Thanks for this fix. Yes I found all the configuration making the code hard to follow.

PR #644, opened 2025-05-16
Guard: Update resource paths

wjt:
This is mostly due to commit a668de8e5eee52d0f49a28a044f50cd4b29da5d1 ("Fix misplaced sound assets").


manuq:
Yes, saw this diff appearing too. I think there are other cases in other scenes. I wonder if there is a better way to find them than be opening the scenes one by one in the editor. It will be problematic for learners using Git for the first time if they find uncommited changes done by the editor.

wjt:
My best idea is to have a CI/pre-commit job that uses a headless Godot to open, save, and close every `.tscn` in the project.

wjt:
I tried and failed to find a way to do that with the command line flags directly but I bet it could be done by running a tool script from the command line.

Issue #645, opened 2025-05-16
Sokoban 3a is too tall for the viewport


wjt:
![Image](https://github.com/user-attachments/assets/7d167d28-2159-41d6-b3ff-f55c7c8e316e)

I think we can just remove one row of the narrow central section without making the puzzle any easier or harder.

wjt:
![Image](https://github.com/user-attachments/assets/ff5d54df-3771-430f-9426-0bb62e106cb2)

A little better.

PR #647, opened 2025-05-16
Sokoban: Reduce vertical size of level 3a

wjt:
This level was previously so tall that the bottom boundary was not visible on the screen.

Remove one row from the centre of the puzzle. This does not fundamentally change the puzzle, but it makes all walkable cells be visible, and at least the bottom/top of the top/bottom boundary (respectively).

Fixes https://github.com/endlessm/threadbare/issues/645


wjt:
Before:

Image

After:

Image

PR #650, opened 2025-05-16
StoryQuest Template: Add template_ prefix to all files

wjt:
In discussion with Heather and Justin, it is useful to consistently name all files that should be renamed with a prefix that should be replaced with the short name of the quest, e.g. replacing template_ with stella_, to make it clearer that the correct file is being edited.

Rename all files and directories in the StoryQuest template to contain template_ as a prefix.

The one exception is quest.tres which must not be renamed – that is how the storybook discovers quests.


scenes/menus/storybook/components/quest.gd
```diff
@@ -21,10 +21,10 @@ extends Resource
 
 @export_group("Advanced")
 
-## The path to the first scene of the quest. If you are following the structure of the template,
-## leave this as the default path. The path can be absolute (beginning with [code]res://[/code])
+## The path to the first scene of the quest.
+## The path can be absolute (beginning with [code]res://[/code])
 ## or relative to the folder where this quest resource is saved.
```
wjt, line 26: It's no longer useful to have this be a relative path because learners will need to change it because they are renaming the intro scene… as a follow up commit later I will simplify this.

scenes/game_elements/characters/enemies/guard/guard.tscn
```diff
@@ -5,7 +5,7 @@
 [ext_resource type="Texture2D" uid="uid://8t6ihmesvghq" path="res://scenes/game_elements/characters/enemies/guard/components/field_of_view.png" id="3_0hjcv"]
 [ext_resource type="Texture2D" uid="uid://c75ofhy3swhj3" path="res://scenes/game_elements/characters/enemies/guard/components/exclamation_mark.png" id="3_pnkgp"]
 [ext_resource type="Script" uid="uid://cxsi2xqcdyw7g" path="res://scenes/game_elements/characters/enemies/guard/components/light.gd" id="4_lptvm"]
-[ext_resource type="SpriteFrames" uid="uid://ovu5wqo15s5g" path="res://scenes/quests/story_quests/template/components/guard_enemy_template.tres" id="5_mswbt"]
+[ext_resource type="SpriteFrames" uid="uid://ovu5wqo15s5g" path="res://scenes/quests/story_quests/template/template_components/template_guard_enemy.tres" id="5_mswbt"]
```
manuq, line 8: Have you agreed to also rename the `components/` folder to `template_components/` ?

manuq (approved):
Awesome @wjt @jbourqueendless @hydrolet ! Hopefully this rename will make clear to learners which files are modding. I'm not so sure about the `components/` folder though. But feel free to merge.

scenes/game_elements/characters/enemies/guard/guard.tscn
wjt: That was discussed briefly & agreed, though we were in a bit of a hurry and it was kind of an arbitrary choice.

manuq: OK! Then let's merge.

PR #652, opened 2025-05-16
Player: Add dust particles when running

manuq:
Add a small spritesheet with a dust cloud. Use it to add a new GPUParticles2D node to the player scene.

Add a script to the particles, so they don't emit when the player is not running. And use the player velocity to set the direction of the particles.

Make the particles material semi transparent, and with an "Add" blend mode.

Add a bit of randomness to the particle effect. In scale and in initial speed.

Fixes https://github.com/endlessm/threadbare/issues/245


manuq:
This is an alternative to https://github.com/endlessm/threadbare/pull/496

I made it more subtle, and only for running.

[Grabación de pantalla desde 2025-05-16 17-24-36.webm](https://github.com/user-attachments/assets/503d4114-b47f-4679-b36a-d6dad7595257)

wjt:
This looks good. (Conflicts with #551... I wish `load_steps` could be removed from the scene header, it all but guarantees conflicts when otherwise independent changes are made to a scene, and if it's wrong, the only thing that changes is that the progress bar is less accurate. I found https://github.com/godotengine/godot-proposals/issues/11802 proposing to remove it entirely for exactly this reason, hooray.)

While testing this I noticed that you get weird results with an analogue stick if you hold down the "run" button but only slightly incline the joystick:

[Screencast from 2025-05-19 11-12-51.webm](https://github.com/user-attachments/assets/f7a632d2-09e8-4944-8242-9106b8b877b1)

But this is not a new problem with the particles - the run animation also looks silly! I wonder if (thinking aloud) we could change the definition of "is running" to be "velocity is greater than the walk speed". Separate branch.

wjt:
Rebased to fix conflict:

```
diff --cc scenes/game_elements/characters/player/player.tscn
index 0f8eae0,41a24dd..0000000
--- a/scenes/game_elements/characters/player/player.tscn
+++ b/scenes/game_elements/characters/player/player.tscn
@@@ -1,4 -1,4 +1,4 @@@
- [gd_scene load_steps=27 format=3 uid="uid://iu2q66clupc6"]
 -[gd_scene load_steps=33 format=3 uid="uid://iu2q66clupc6"]
++[gd_scene load_steps=34 format=3 uid="uid://iu2q66clupc6"]
  
  [ext_resource type="Script" uid="uid://bwllxup305eib" path="res://scenes/game_elements/characters/player/components/player.gd" id="1_g2els"]
  [ext_resource type="SpriteFrames" uid="uid://vwf8e1v8brdp" path="res://scenes/quests/story_quests/template/template_player_components/template_player.tres" id="2_blfj0"]
```

Issue #590, opened 2025-05-15
Disable (some) CI checks on forks?


wjt:
I'm on the fence about this but I have the feeling that showing failing checks for e.g. trailing whitespace in dialogue files will not provide any benefit to learners. And I think expecting learners to set up pre-commit to catch problems locally before they are committed is unlikely to happen. (Much of our own team does not have pre-commit set up.)

So maybe we should set those workflows not to run on forks.

wjt:
Interestingly, by default GitHub disables Actions in a fork. You can go to the Actions tab and click an "I understand" button to enable them. The reason is security: if forking a repo made actions run automatically in my org, then the actions from upstream can potentially access my organisation's secrets.

wjt:
I think I can see how to do this if I am prepared to hardcode `endlessm/threadbare` in the workflow files. That will break if we move this to a different organisation, e.g. https://github.com/Endless-Access-Community, but ... we'll notice that pretty fast so it's probably Ok.

PR #653, opened 2025-05-16
Rename NPC to KnitWitch in template and edit narrative text in template scenes

jbourqueendless:
Rename NPC to KnitWitch in template to match lore. Edit narrative text in template scenes to clarify tasks for learners.


manuq:
@jbourqueendless this was conflicting with https://github.com/endlessm/threadbare/pull/621 that I just merged. Then there were some lint issues, automatically fixed by running `pre-commit run --all-files`. I think it is too much to ask learners to do that, but I see Will is already considering removing them from forks: https://github.com/endlessm/threadbare/issues/590

PR #660, opened 2025-05-19
Music puzzle: Adjust altar and collectible positions

### LINKED ISSUE #578: LoreQuest music puzzle collectible appears offscreen
We noticed during a recent playtest that, when you play the final note of the LoreQuest music puzzle, you are very likely to be in a position that means the green thread is off camera when it appears:

Image

It is *possible* to position yourself such that you can see the thread while playing the rock but you're unlikely to do so organically:

Image

Our friends at FannyPack noted that their experience has taught them that if something changes off-camera, the player sometimes doesn't notice it even when they move around to a position where it's visible.

We might be able to rearrange the scene slightly to make it visible. (And/or make it so that completing the puzzle triggers the musician's "well done!" interaction that explicitly tells you to collect the thread.)

wjt:
image

Fixes #578


wjt:
https://endlessm.github.io/threadbare/branches/endlessm/lorequest-music-puzzle-collectible-position/#quests/lore_quests/quest_001/1_music_puzzle/music_puzzle

ramcq:
Driveby - this still has the property that if you're moving sideways from the higher notes to the lower notes - when solving the 3rd and 4th altars - you can accidentally run into the 1st/2nd altar rather than being in the activation zone for the lower stones. So you accidentally play the earlier melody rather than finishing the one you're trying to play. Maybe that's a "speedrunning because I know the tunes" problem, so take it with a pinch of salt.

wjt:
I can keep moving them higher and higher... 😁 

Actually I had a thought last night: maybe the puzzle (and whole island???) should be vertically flipped, to have the altars below the stones. Or perhaps transposed to have the altars on the left and the keyboard on the right, because the screen is wider than it is tall.

wjt:
The puzzle is not perfect with this change but I think it's a little better, and it does address the primary issue I was trying to solve (the collectible appears offscreen) as well as some cleanups to the scene as a whole.

Issue #509, opened 2025-05-08
Change ink combat defend animation


PlayMatters:
Adopt the second attack mechanism (shield thrust) as the means to deflect ink in the combat missions.

PlayMatters:
https://github.com/user-attachments/assets/44ca5c09-d5a1-4cb5-b15e-2673f382f3c2

manuq:
Here are some alternatives:
- Rename Attack 02 to Attack 01 in StoryWeaver, leaving the template untouched https://github.com/endlessm/threadbare/pull/661
- Add back Attack 02 to the template, and use it for ink repel https://github.com/endlessm/threadbare/pull/662
- Don't do this change at this point.

@wjt what do you think we should do? If doing the second option, I better sync with @jbourqueendless and the rest of the Stella devs.

manuq:
Removing from MVP. We decided to do this after, when we have tag releases so we can add to the main branch without conflicting with our programs.

PR #661, opened 2025-05-19
StoryWeaver: Use different animation for attack

manuq:
Use the animation named "Attack 02" in the Aseprite source file for the SpriteFrames animation "Attack 01".

Ideally we would change the AnimationPlayer in the Player scene to just use Attack 02. But we can't do that without also adding back "Attack 02" to the template for the player. That "Attack 02" animation was removed from the player template upon request, to avoid confusion in learners, as the animation wasn't used.

So, rename animation "Attack_01" to "Attack_unused" in StoryWeaver SpriteFrames in all color variations. And rename "Attack_02" to "Attack_01".

This is confusing, but may be the only thing we can do at this point.

Fixes https://github.com/endlessm/threadbare/issues/509

### DIFF
--- a/scenes/game_elements/characters/player/components/player.gd
+++ b/scenes/game_elements/characters/player/components/player.gd
@@ -21,7 +21,6 @@ const REQUIRED_ANIMATION_FRAMES: Dictionary[StringName, int] = {
 	&"idle": 10,
 	&"walk": 6,
 	&"attack_01": 4,
-	&"attack_02": 4,
 	&"defeated": 11,
 }
 const DEFAULT_SPRITE_FRAME: SpriteFrames = preload("uid://vwf8e1v8brdp")
--- a/scenes/game_elements/characters/shared_components/sprite_frames/storyweaver_blue.tres
+++ b/scenes/game_elements/characters/shared_components/sprite_frames/storyweaver_blue.tres
@@ -6,22 +6,6 @@
 [ext_resource type="Texture2D" uid="uid://ceiyxl0dy2cg1" path="res://scenes/game_elements/characters/player/components/storyweaver_blue_idle.png" id="3_lo7nh"]
 [ext_resource type="Texture2D" uid="uid://dmvl71i6783rk" path="res://scenes/game_elements/characters/player/components/storyweaver_blue_walk.png" id="4_3nn5s"]
 
-[sub_resource type="AtlasTexture" id="AtlasTexture_007i1"]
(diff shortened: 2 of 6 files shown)


manuq:
@wjt ready for review. I have removed the requirement of the unused animation, because the template was warning about it.

wjt:
https://endlessm.github.io/threadbare/branches/endlessm/storyweaver-attack-animation/#quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1

wjt (approved):
LGTM.

You could consider renaming the animations in the Aseprite source file to match?

manuq:
> LGTM.
> 
> You could consider renaming the animations in the Aseprite source file to match?

Honestly I don't want to go with the inconsistency that this leaves. The Aseprite source of the template has Attack 01 matching the Aseprite source of the StoryWeaver Attack 01. Same for Attack 02. So if I change Storyweaver.aseprite and not player_template.aseprite it will become inconsistent. Should we leave this out for now, until we can modify the template again? Also, when will be the next window to change the template? After Stella is merged? After the first program ends?

wjt:
Ahh, now I understand. I had misunderstood that the template only has one animation. But there are two in its Aseprite source file: Attack 01 is a swipe, Attack 02 is a shield. Similar to StoryWeaver.aseprite.

Then we have `storyweaver_blue.tres` SpriteFrames (resp. other colours) with matching `attack_01` and `attack_02` animations. The `template_player.tres` SpriteFrames only defines `attack_01` – but both `attack_01` and `attack_02` are exported as PNGs alongside it.

So for consistency we have the following options:

1. Swap the template's attack animations (in the spriteframes, exported PNGs, and Aseprite file) along with StoryWeaver's;
2. Define `attack_02` in the template SpriteFrames; remove `attack_01`; and make the Player scene use `attack_02`;
3. Keep using the existing swipe-of-needle animation.

Options 1 and 2 will (very slightly...) invalidate the course materials that are being prepared/have been prepared. Option 3 means leaving #509 open.

I think that unless 1 or 2 are unexpectedly acceptable (and the relevant people have other things to think about this week) we should take option 3...

> Also, when will be the next window to change the template? After Stella is merged? After the first program ends?

I think it's probably too late for the first program now but we may need to consider having programs work from a `release` branch (doesn't currently exist), or the latest tag (we don't have one), rather than from `main` in future so that we don't have this tight coupling.

PR #669, opened 2025-05-20
Combat tilemap

hydrolet:
Edit tilemap, character assets, narrative, and mechanics to make the combat level of the StellaQuest.

There are three things I would like help with:

1. Highest priority: I don't know how to make it so Stella does not walk right over the fisherman.
2. Medium priority: I want the fisherman to be an NPC that has a dialog interaction with Stella (instead of him telling her about the constellation in the narrative box). 
3. Low priority: I would like there to be four different projectiles (I have the assets already). One is the starfish, but the other three are regular fish. They should not fill the boat (only the starfish should). This is a stretch goal for me, so it's fine if we can't make it happen. :)


hydrolet:
Oh, and I will need to update the Stella attack animation when it's ready, but I don't need help with that. :)

hydrolet:
Annnnnd I didn't mean to push it to main. I meant to push it to my forked stellaquest repo. Apologies!

hydrolet:
How do I re-commit the same thing to that repo? 🫣

wjt:
The branch is already in the other repo. Open a new pull request starting on that other repo.

manuq:
> There are three things I would like help with:
> 
> 1. Highest priority: I don't know how to make it so Stella does not walk right over the fisherman.

The easiest would be to add a StaticBody2D to the scene, with a CollisionPolygon2D node as child. Set the collision layer of the StaticBody2D to layer 5 "walls". This is how is made in Fray's end so the player can't walk over the forest:

Captura desde 2025-05-20 08-54-44
 

> 2. Medium priority: I want the fisherman to be an NPC that has a dialog interaction with Stella (instead of him telling her about the constellation in the narrative box).

The throwing enemies are not talker NPCs (@wjt we may want to revisit that, using composition for behevior?). But you can add an Intro Dialogue in the FillGameLogic node. See the 1st combat round scene as example `scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1.tscn`:

Captura desde 2025-05-20 09-00-26

> 3. Low priority: I would like there to be four different projectiles (I have the assets already). One is the starfish, but the other three are regular fish. They should not fill the boat (only the starfish should). This is a stretch goal for me, so it's fine if we can't make it happen. :)

Not currently possible, sorry. Should we make this possible for the next release?

Issue #638, opened 2025-05-16
Change base resolution of game?


wjt:
Currently we configure the game to run in a 1920×1080 viewport. I believe this is the most common PC display size but it does mean that the game gets downscaled to run on Steam Deck which is 1280×800 and looks bad as a result.

But we also use a Camera2D set to 200% in most scenes (170% scaling in the LoreQuest stealth challenge, from memory). So the effective game resolution is 960×540 (~1130×635 in the stealth quest). And all our pixel assets are scaled twice (once for the camera, and again because the viewport will not be exactly 1920×1080 in any case since we don't currently maximise!).

I wonder if it would be better to configure the viewport to a lower resolution, and remove the scaling from the camera?

manuq:
Yes the camera is almost 2x in scenes:
```
$ git grep -l 'node name="Camera2D"' scenes/ | while IFS= read -r file; do awk '/node name="Camera2D"/,/^$/' $file; done | grep zoom
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(1.5, 1.5)
zoom = Vector2(1.5, 1.5)
zoom = Vector2(1.5, 1.5)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(1.6, 1.6)
zoom = Vector2(1.7, 1.7)
zoom = Vector2(1.8, 1.8)
zoom = Vector2(2, 2)
zoom = Vector2(1.7, 1.7)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
zoom = Vector2(2, 2)
```

Why I went with the current resolution:
- 1080p is a common resolution and aspect ratio in modern PCs and consoles (88629313)
- In d358358a I added a 2X camera and texture filter to "Nearest" to the root node of that `scenes/main.tscn` scene. Back then, I thought this was going to be a pixel-perfect game, where you don't want the pixels to be scaled at discrete numbers (a zoom 1.7 should be forbidden in this case).
- This still allows having UI in high resolution. In particular, some pixel perfect games still have hi-res fonts for readability / accesibility.

[...]

PR #674, opened 2025-05-20
Add optional animation for quest to storybook

### LINKED ISSUE #557: Specify player sprite in quest.tres
It would be nice to be able to add the SpriteFrames for the protagonist to the `quest.tres` for a StoryQuest so that we can have some visually-interesting element for it in the storybook, and introduce the concept that the playable character in a storyquest is not StoryWeaver but some other story-specific character.

wjt:
For the purposes of the storybook, a quest is defined by a Quest resource with various fields that populate the storybook page.

To add some visual variety to an otherwise text-heavy page, and potentially hint that the protagonist in a StoryQuest will not be StoryWeaver, allow a quest to specify a SpriteFrames, and an animation from that SpriteFrames, which will be shown on the storybook page if present.

To do this, I had to implement a custom TextureRect that can play an animation from a SpriteFrames. This was a surprising gap in Godot's Controls to me but there are many open bugs related to this.

I also cleaned up some bits and bobs in the quest resource class and storybook scene.

Fixes #557 

https://endlessm.github.io/threadbare/branches/endlessm/quest-animation-in-storybook/#menus/storybook/storybook

### DIFF
--- a/scenes/menus/storybook/components/animated_texture_rect.gd
+++ b/scenes/menus/storybook/components/animated_texture_rect.gd
@@ -1,0 +1,94 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name AnimatedTextureRect
+extends TextureRect
+## A [TextureRect] which shows a single animation from a [SpriteFrames].
+##
+## This is useful because the only built-in way to display an animation in a
+## TextureRect is to use an [AnimatedTexture], which is documented to be
+## deprecated and broken; and because we use this to display an animation from
+## an in-game sprite, so an appropriate SpriteFrames will already exist.
+
+## A sprite frame library containing the animation. If unset, no texture is shown.
+@export var sprite_frames: SpriteFrames = preload(
+	"res://scenes/quests/story_quests/template/template_player_components/template_player.tres"
+):
+	set(new_value):
+		sprite_frames = new_value
+		notify_property_list_changed()
+		_update_animation()
+
+## The animation from [member sprite_frames] to display. If this animation is
+## not defined in [member sprite_frames], no texture is shown.
+@export var animation_name: StringName = &"idle":
+	set(new_value):
+		animation_name = new_value
+		_update_animation()
+
+var _frame: int = 0
+var _time_to_next_frame: float = 0.0
+
+
+func _validate_property(property: Dictionary) -> void:
+	match property["name"]:
+		"animation_name":
+			if sprite_frames:
+				property.hint = PROPERTY_HINT_ENUM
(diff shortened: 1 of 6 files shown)


scenes/menus/storybook/storybook.tscn
```diff
@@ -67,6 +74,14 @@ size_flags_vertical = 1
 text = "by Jane Bloggs, Erika Mustermann and Jóna Jónsdóttir"
 autowrap_mode = 2
 
+[node name="Animation" type="TextureRect" parent="PanelContainer/HBoxContainer/RightPage/VBoxContainer"]
```
wjt, line 77: Because we generally have a 200% scale factor in the game (c.f. #638) this ends up looking a little small. It's not completely trivial to scale a control but I think we would do it by setting a custom minimum size to 2× the width & height of the texture.

scenes/menus/storybook/components/quest.gd
```diff
@@ -19,12 +18,30 @@ extends Resource
...
+
+
+func _validate_property(property: Dictionary) -> void:
+	match property["name"]:
+		"animation_name":
+			if sprite_frames:
+				property.hint = PROPERTY_HINT_ENUM
+				property.hint_string = ",".join(sprite_frames.get_animation_names())
+			else:
+				property.usage |= PROPERTY_USAGE_READ_ONLY
```
wjt, line 44: This is of course adapted from `teleporter.gd` – it really makes it nicer to use. I was surprised that Resource subclasses could be tool scripts but I guess why not!

scenes/menus/storybook/storybook.tscn
manuq: Yes, for TextureRect that's how you adjust the size too. By the way this is a good video explanation https://www.youtube.com/watch?v=5Hog6a0EYa0&t=529s . I come back to that video every time I use Control nodes, I wish it was a written tutorial :)

manuq (approved):
The AnimatedTextureRect is clever and small enough. Hopefully Godot fixes the gap at one point.

scenes/menus/storybook/storybook.tscn
wjt: Great reference, thanks!

scenes/menus/storybook/components/animated_texture_rect.gd
```diff
@@ -0,0 +1,94 @@
...
+	)
+
+
+func _process(delta: float) -> void:
+	if Engine.is_editor_hint() or _frame < 0:
+		return
+
+	# It is unlikely but possible that an animation frame's duration may be so
+	# short that it should be skipped entirely.
+	while delta >= _time_to_next_frame:
```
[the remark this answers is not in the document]
wjt: I don't think it will happen but I only had to change `if` to `while` for it to work!

PR #687, opened 2025-05-21
Elder: Fix end of dialogue

wjt:
In commit 1edfb44e78faab2a08621f467e9e442ccd2df978, I moved talker.gd's connection to DialogueManager.dialogue_ended to
_on_interaction_started(). What I missed is that elder.gd overrides this function without chaining up. So the elders' _on_dialogue_ended() is never called, making it impossible to enter a quest.

Copy the same new code into elder.gd.


wjt:
I'm going to merge this without review because the game is unplayable on `main`, but post-hoc review is welcome.

PR #688, opened 2025-05-21
Sequence redo

hydrolet:
redesign the sequence puzzle for stella.


hydrolet:
@wjt I somehow lost the examine abilities on my second and third rock hints. help?

also, I changed the label to "tap" to accommodate all scenes game-wide. rocks, tortoises, etc etc.

hydrolet:
@ma

PR #689, opened 2025-05-21
Change "kick" to "tap" in sequence puzzles

wjt:
This is a change from @hydrolet in https://github.com/StellaQuest/threadbare/pull/31. It's OK to kick a rock, but not a tortoise. By using a more neutral term for the default interact verb, we reduce the number of cases where someone will have to figure out how to override it.

I updated the dialogue accordingly. Unfortunately "kick" is in the API and we can't change that.


manuq (approved):
Better!

PR #690, opened 2025-05-21
Refactor storybook.gd somewhat

wjt:
This splits some hardcoded override data for the template storyquest into a resource file, and factors out the section of the scene that displays a quest to a separate scene.

(I was experimenting with switching to a one-quest-per-page model.)


manuq (approved):
Nice, factoring out the page makes it easier to read.

Issue #489, opened 2025-05-07
Sokoban: Design gameplay hints


wjt:
I have always found Sokoban puzzles very difficult and these are no exception. It might be nice at some point to provide the player with some hints if they're stuck.

### XP Summary (total 300)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Progression & Motivation | Problem Solving | 75 |
| Game Design: Progression & Motivation | Creativity & Design Thinking | 150 |
| Game Design: Progression & Motivation | Technical Literacy | 75 |

manuq:
Adding hints would need game design. My alternative proposal is to offer a way to skip the puzzles: https://github.com/endlessm/threadbare/pull/691

wjt:
I've repurposed this issue as a task to do that game design thinking: what is the right design for hints? Once we have a design for hints we can create follow-on tasks to bring them to fruition.

wjt:
In our game, you could imagine having the shadow of a piece in a position that it should be moved to (either temporarily or permanently). Mockups:

In 1a you have to put the dust blobs in particular places:

<img width="668" height="377" alt="Image" src="https://github.com/user-attachments/assets/ab5803d2-5b64-448f-ade1-ce9fde7ebfdb" />

In 1b you have to put the yellow thread in an unexpected place, temporarily, to be able to manoeuvre the green behind it, so that you can place the yellow then green.

As I mock this up I realise that we need to preserve the colours because the threads all have the same shape – and this may be an accessibility issue for people with colour-blindness. I'll open another issue for that. But ignoring that for a moment, here's hint 1:

<img width="627" height="560" alt="Image" src="https://github.com/user-attachments/assets/15f61a15-33f8-43d7-aa97-aa11a08ed797" />

and here's hint 2 (with the yellow thread moved out of the way):

<img width="627" height="560" alt="Image" src="https://github.com/user-attachments/assets/e2497510-5d96-40d0-85d7-30ce97cee1c3" />

sebastianangulod:
I plan to work on this

wjt:
[Spooky Express](https://spooky.express/) is a grid-based puzzle game whose technical artist developed these Sokoban puzzles! The goal is to lay out a train line that takes the vampire to the coffin, etc., and then to the exit. Here is the hint interface from that game:

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/05b035f2-c9e6-4510-8150-38d9dd58ab0c" />

[...]

Issue #432, opened 2025-05-02
Add a way to abandon a quest


wjt:
As the game grows, it will become more likely that some quests will include challenges that are too hard for some players to complete. Right now, there is no way to abandon a quest, short of quitting and reopening the game.

It might be nice to add an "Abandon Quest" option to the pause menu (#8), available while on a quest (including when you have returned to Fray's End but not yet taken the threads you collected to the Loom). If activated, this would prompt for confirmation, then clear any inventory items from the current quest and return you to Fray's End. In particular, as someone who is bad at Sokoban, I think it should work in the Sokoban puzzles :)

wjt:
@PhoenixStroh this is my memory of your idea from a conversation earlier in the week – is this what you were thining9?

wjt:
Reopening this because "Exit to Title" doesn't actually do what I was thinking here:

- If on a quest:
- Have a Pause menu item “Abandon Quest”
- This should:
    - Clear your inventory of any collected threads
    - Return you to Fray's End

PR #691, opened 2025-05-21
Make sokoban puzzles skippable

manuq:
A "press X to skip" will appear in the HUD, next to the hints for the undo and reset. Only after 2 minutes of being stuck or after 10 retries.

Helps https://github.com/endlessm/threadbare/issues/489


scenes/eternal_loom_sokoban/components/hud/sokoban_hud.tscn
```diff
@@ -57,3 +58,20 @@ layout_mode = 2
 theme = ExtResource("3_uhi48")
 theme_type_variation = &"HintLabel"
 text = "Reset"
+
+[node name="HBoxContainer3" type="HBoxContainer" parent="VBoxContainer"]
+layout_mode = 2
```
wjt: IMO this should be above the other two, because otherwise the hud shifts upwards when the new action becomes visible:

[Screencast from 2025-05-22 14-05-46.webm](https://github.com/user-attachments/assets/c43b5996-74a7-42d5-be52-cf70f0548a38)

wjt (approved):
IMO this is a reasonable interim solution. But I'd be interested in other opinions.

The alternative would be to add an "abandon quest" feature #432 but that would rob the player of the ability to see the closing dialogue with the loom...

manuq:
> The alternative would be to add an "abandon quest" feature #432 but that would rob the player of the ability to see the closing dialogue with the loom...

Yes, IMO #432 is a must have. But the skip can happen even with that, as it makes the sokobans optional so you can complete the game, instead of having to quit.

manuq:
https://github.com/endlessm/threadbare/issues/432 has been implemented. Even with that, it may be good to have this way to skip the sokoban puzzles, effectively making them optional. Considering that their difficulty is super high compared to all the rest. As a way to have more players satisfied and able to finish the game.

PR #694, opened 2025-05-22
Update combat fish next scene

hydrolet:


hydrolet:
removed the red fish modulation, added second fish instead
set the correct next scene

PR #695, opened 2025-05-22
ci: Disable discord & linter workflows in forks

### LINKED ISSUE #590: Disable (some) CI checks on forks?

wjt:
Fixes #590

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -14,7 +14,11 @@ concurrency:
 jobs:
   flatpak:
     name: Build Flatpak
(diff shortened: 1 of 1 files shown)


wjt:
I pushed this to StellaQuest (well, the first two patches) and it seems to work as expected:

https://github.com/StellaQuest/threadbare/actions/runs/15186950344

PR #697, opened 2025-05-22
Add a single rock to the intro scene template

jbourqueendless:
Add a single rock to the intro template as an exemplar of a simple background or world "component". This rock.tscn can be a reference for how to create a simple art asset using:

StaticBody2d
Sprite2D
CollisionShape2D

This type of scene may be an approachable "first scene from scratch" example.


wjt (commented):
I think this is in the wrong folder and you meant to replace the one in `res://scenes/quests/story_quests/template/0_template_intro`?

jbourqueendless:
Trying that again

PR #698, opened 2025-05-22
Storybook: Fix template first scene

manuq:
The first scene property was pointing to the same template_quest.tres file. Point it to the 01 intro scene instead.


wjt:
Oh I was sure I fixed an issue like this.. thanks!

PR #700, opened 2025-05-23
Pause overlay: Add button to go back to title screen

manuq:
And move the Resume button to the top, as it is the default focused option.

Currently this is a way to abandon and lose all your progress, because nothing is persisted. But is also a way to restart the experience if you get stuck in the Sokobans, for example.

In the future, the progress should be persisted, and the title screen should be able to offer a Resume + Restart options instead of Start, if there is progress.

Fixes https://github.com/endlessm/threadbare/issues/432

### DIFF
--- a/scenes/globals/pause/pause_overlay.gd
+++ b/scenes/globals/pause/pause_overlay.gd
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: MPL-2.0
 extends CanvasLayer
 
+const TITLE_SCENE: PackedScene = preload("uid://stdqc6ttomff")
+
 @onready var pause_menu: Control = %PauseMenu
 @onready var resume_button: Button = %ResumeButton
 @onready var options: Control = %Options
@@ -38,3 +40,10 @@ func _on_options_back() -> void:
 
 func _on_resume_button_pressed() -> void:
 	toggle_pause()
+
+
+func _on_title_screen_button_pressed() -> void:
+	toggle_pause()
+	SceneSwitcher.change_to_packed_with_transition(
+		TITLE_SCENE, ^"", Transition.Effect.FADE, Transition.Effect.FADE
+	)
--- a/scenes/globals/pause/pause_overlay.tscn
+++ b/scenes/globals/pause/pause_overlay.tscn
@@ -58,13 +58,6 @@ theme_type_variation = &"PlayerRibbon"
 layout_mode = 2
 text = "Game Paused"
 
-[node name="OptionsButton" type="Button" parent="TabContainer/PauseMenu/VBoxContainer"]
-unique_name_in_owner = true
-layout_mode = 2
-size_flags_horizontal = 4
-theme_type_variation = &"FlatButton"
-text = "Options"
-
 [node name="ResumeButton" type="Button" parent="TabContainer/PauseMenu/VBoxContainer"]
 unique_name_in_owner = true
 layout_mode = 2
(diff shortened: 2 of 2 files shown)


scenes/globals/pause/pause_overlay.tscn
```diff
@@ -58,26 +58,34 @@ theme_type_variation = &"PlayerRibbon"
...
 text = "Options"
 
-[node name="ResumeButton" type="Button" parent="TabContainer/PauseMenu/VBoxContainer"]
+[node name="TitleScreenButton" type="Button" parent="TabContainer/PauseMenu/VBoxContainer"]
 unique_name_in_owner = true
 layout_mode = 2
 size_flags_horizontal = 4
 theme_type_variation = &"FlatButton"
-text = "Resume"
+text = "Back to title screen"
```
manuq, line 80: I'm not very sure about this label. Other option I handled: "Back to main menu", "Back".

scenes/globals/pause/pause_overlay.gd
```diff
@@ -38,3 +40,10 @@ func _on_options_back() -> void:
 
 func _on_resume_button_pressed() -> void:
 	toggle_pause()
+
+
+func _on_title_screen_button_pressed() -> void:
+	toggle_pause()
+	SceneSwitcher.change_to_packed_with_transition(
+		TITLE_SCENE, ^"", Transition.Effect.FADE, Transition.Effect.FADE
```
wjt, line 48: FWIW I expected the button to take you back to Fray's End (and only be visible if you're not in Fray's End) rather than to the title screen. But this would also work!

scenes/globals/pause/pause_overlay.tscn
wjt: I don't think "Back" is a destructive-enough sounding word.

If we're going back to the title, "(Quit/Exit) to Title (Screen)" would read better to me. (I think these should be in Title Case.)

If we're going back to Fray's End then perhaps "Abandon Quest" would work?

scenes/globals/pause/pause_overlay.tscn
```diff
@@ -58,26 +58,34 @@ theme_type_variation = &"PlayerRibbon"
 layout_mode = 2
 text = "Game Paused"
 
+[node name="ResumeButton" type="Button" parent="TabContainer/PauseMenu/VBoxContainer"]
+unique_name_in_owner = true
+layout_mode = 2
+size_flags_horizontal = 4
+theme_type_variation = &"FlatButton"
+text = "Resume"
```
wjt, line 66: Good point about moving this up.

scenes/globals/pause/pause_overlay.tscn
wjt: I checked in Spelunky 2 (other games are available, just by way of example).  It uses "Quit Game" on the main menu as the "return to the OS" action, and "Exit to Title" in the pause menu.

scenes/globals/pause/pause_overlay.gd
manuq: Oh I see, yes both can work. I think going back to the title screen would be better if we ultimately persist state, and for having the "credits" and "exit game" options reachable.

wjt: I see - so in the case where you have some state saved, you can Continue and not play the intro again

wjt (approved):

Issue #709, opened 2025-05-26
Even without sound effects or music, footsteps sound can be heard


LauraFoglia:
When starting threadbare I went to options and silenced both music and sound effects, but then when playing I could still hear the steps as the story weaver walked.

PR #710, opened 2025-05-26
Player: move walk sound to sfx bus

JuanFdS:
Previously, that audio stream player had its bus set to Master, meaning it could still be heard even if the SFX volume was turned down.

Fixes https://github.com/endlessm/threadbare/issues/709


manuq:
Thanks!

Issue #670, opened 2025-05-20
Going back/forward or refreshing mid-quest breaks the quest progression


wjt:
Currently scene_switcher always updates the hash in the URL when you change scene. This is very convenient for linking to a specific scene but it has unforeseen consequences. For example, if a player accidentally refreshes the page or triggers the browser's Back button/gesture while playing the ink combat challenge, then when the refresh finishes or they go Forward, then they will be playing the ink combat challenge but minus the thread they collected in the music puzzle. As a result, once they get back to Fray's End they will only have 2 threads and will be unable to operate the Eternal Loom.

But then, if they embark on another quest, they'll start the quest with 2 threads and end it with 5. I don't really know what will happen in that case!

wjt:
Some thoughts...

The scene-in-URL-hash feature is a developer tool rather than something for end-users. We could do something like:

  - Build branches in debug configuration and tags/releases in release configuration
  - In the game, if in debug configuration, preserve the current behaviour.
  - Otherwise, in release configuration, if no URL hash is provided, disable the feature (until some hidden keystroke is pressed?)

On losing threads across refresh, you could imagine fixing this by having persistent state so that if you refresh the game the title screen gives you a Continue option that takes you back to where you left off.

We could also consider simplifying the collectible system. It's currently very general but there are actually only 4 things we need to track:

- Which quest are we playing (or no quest)
- How many threads have been collected (0, 1, 2, 3)
- Has the player activated the loom?
- What's the current scene?

manuq:
Since we are already after MVP, we should consider reopening https://github.com/endlessm/threadbare/issues/9 for persisting state. It will also be helpful for the "Abandon quest" https://github.com/endlessm/threadbare/issues/432

manuq:
This should be fixed by the last commit of https://github.com/endlessm/threadbare/pull/716

PR #716, opened 2025-05-27
Save game progress

manuq:
Resolves https://github.com/endlessm/threadbare/issues/9
Fixes https://github.com/endlessm/threadbare/issues/670

## For testers:
- Try starting the game, enter a quest, then at one point exit to the main title, then exit game (or close the tab if in web). Then start the game again (or open it in a web tab). Your progress (as in below) should be kept.
- Try in particular during the sokoban games and see if the ending dialogue is shown.
- Try doing it from the web and from the editor.
- Try running a single scene from the editor.
- Try running a single scene from a given URL in web.

## What player progress is saved

- The current scene and the starting point of the player in that scene (called spawning point).
- The chosen quest. Although not used for anything! see doubts below.
- The amount of items collected.
- If the player is incorporating the threads to the Eternal Loom. In other words, if in the middle of the Sokoban and an ending dialogue should be shown after.

So to be clear:

- The exact player position is not persisted. The player will enter the scene from last starting point. For example, from the bottom of Fray's End at the beginning of the game.
- The Memory, Imagination and Spirit collectibles will be restored in that order, not in the order the quest provides it. Although current quests provide them in that order too.

## Doubts & questions for reviewers:

Persist the chosen quest at all? It felt right to me to persist the chosen quest as offered by the Elders. It's actually the resource path of that quest, which I guess can be considered its unique identifier. However that is not needed at all in current loading.

What to do when running a specific scene from the Godot Editor? What I did was to disable the load/save from/to the configuration file in that case. The game state is still used (you can exit to the title screen and continue) but nothing is persisted.

How should this work in the web? Currently the scene switcher has a handy way to pair the URL hash with the current scene. I still need to check if this PR messes with that awesome debugging tool. UPDATE: I think I managed to consider the web hash by the last commit:
- If the URL hash contains the same scene path as the last save, consider that the player wants to continue their progress.
- If the URL hash contains another scene, assume that the URL was given for testing / debugging a particular scene. So act as if the player has ran the scene in isolation from the Editor. That is, do not persist state.


scenes/globals/game_state/game_state.gd
```diff
@@ -1,6 +1,5 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
-@tool
```
wjt, line 3: Huh, are autoloads available to `@tool` scripts in the editor??

scenes/globals/game_state/game_state.gd
```diff
@@ -23,27 +27,60 @@ signal collected_items_changed(updated_items: Array[InventoryItem])
...
+func _update_inventory_state() -> void:
+	var amount: int = clamp(inventory.size(), 0, InventoryItem.ItemType.size())
+	_state.set_value(INVENTORY_SECTION, INVENTORY_ITEMS_AMOUNT_KEY, amount)
+
+
+func _restore() -> void:
+	var amount_in_state: int = _state.get_value(INVENTORY_SECTION, INVENTORY_ITEMS_AMOUNT_KEY, 0)
+	var amount: int = clamp(amount_in_state, 0, InventoryItem.ItemType.size())
+	for index in range(amount):
+		var item := InventoryItem.with_type(index)
```
wjt, line 155: I'm a little worried that it is going to be confusing if the threads are restored in a different order. I guess we can just say that it's a bug to have them in the wrong order in a quest though!

scenes/globals/game_state/game_state.gd
```diff
@@ -37,8 +41,25 @@ func _ready() -> void:
 	_restore()
 
 
-func start_quest() -> void:
-	clear_inventory()
+func start_quest(quest: Quest) -> void:
+	_do_clear_inventory()
+	_update_inventory_state()
+	_state.set_value(QUEST_SECTION, QUEST_PATH_KEY, quest.resource_path)
```
wjt, line 72: I think this is a useful thing to have in the (persisted) game state in order to implement "abandon quest".

scenes/globals/game_state/game_state.gd
```diff
@@ -24,6 +24,12 @@ const QUEST_SPAWNPOINT_KEY := "current_spawn_point"
 const GLOBAL_SECTION := "global"
 const GLOBAL_INCORPORATING_THREADS_KEY := "incorporating_threads"
 
+## Scenes to skip from saving.
+const TRANSIENT_SCENES := [
```
wjt, line 28: How about the splash?

scenes/menus/title/components/main_menu.gd
```diff
@@ -16,6 +18,10 @@ func _ready() -> void:
 	# custom code in the HTML shell that we do not have yet.
 	quit_button.visible = OS.get_name() != "Web"
 
+	continue_button.visible = GameState.can_restore()
+	if GameState.can_restore():
+		start_button.text = "Restart"
```
wjt, line 23: It will be a bit hard to test the initial no-saved-state state because the only in-game way to clear the saved state is to start a new game, and entering Fray's End saves state. But there is a way:

- Restart
- Quit during intro

or just delete the file.

scenes/globals/game_state/game_state.gd
```diff
@@ -39,10 +39,20 @@ const TRANSIENT_SCENES := [
...
 var _state := ConfigFile.new()
 
 
 func _ready() -> void:
+	var initial_scene_uid := ResourceLoader.get_resource_uid(
+		get_tree().current_scene.scene_file_path
+	)
+	var main_scene_uid := ResourceLoader.get_resource_uid(
+		ProjectSettings.get_setting("application/run/main_scene")
+	)
```
wjt, line 52: This is ingenious!

scenes/globals/scene_switcher/scene_switcher.gd
```diff
@@ -38,6 +38,18 @@ func _restore_from_hash() -> void:
...
+				# Continue if the path matches the saved scene. This would happen
+				# if the player reloads the page while playing.
+				GameState.restore()
+			else:
+				# Otherwise, treat it as the player is debugging a scene from the web.
+				# In that case, do not persist progress and clear the game state.
+				# This is the same behavior as if the scene is ran from the editor
+				# for testing or debugging.
+				GameState.persist_progress = false
+				GameState.clear()
```
wjt, line 51: This is clever!

I think we may also want to disable the url-hash feature in some cases, but this should do the right thing in the case where it is enabled. Good idea.

wjt, line 51: This is clever!

I think we may also want to disable the url-hash feature in some cases, but this should do the right thing in the case where it is enabled. Good idea.

wjt (approved):

scenes/globals/game_state/game_state.gd
manuq: Yes. The other option I handled (and forgot to mention) is: Remove the `@export var item: InventoryItem` from the collectible item, and let the game set it according to the current amount of items. That means current scenes that have the collectible item instantiated won't have the final thread color, and learners won't be able to set it. But maybe that's fine?

scenes/menus/title/components/main_menu.gd
manuq: That's exactly how I tested it. Sorry, I forgot to mention it.

scenes/globals/game_state/game_state.gd
wjt: Ah. And in fact we never *switch* to the main scene, so `set_scene` is never called for it, so I think it will never be saved. And if it is saved it's harmless enough.

scenes/globals/game_state/game_state.gd
wjt: I think what we have here is OK for now. Let's see how it works out in practice.

Issue #714, opened 2025-05-27
Integrate Storyvore assets


manuq:
Replace the current character used in the Lore quest stealth game for the guards, with this new asset.

[Storyvores.zip](https://github.com/user-attachments/files/20454619/Storyvores.zip)

Issue #680, opened 2025-05-21
Guard Detection Bug Stealth Game


jgbourque:
With detection time at the default 1.0 seconds I am having an issue where the guards are taking much longer than 1 second to detect my player. This is happening in the main branch.

I have tried closing down Godot and reloading project files as well but nothing seems to change this issue for me.

https://drive.google.com/file/d/1EpDkC8rBIdY6Jlt0X2RghooXCsLQcTry/view?usp=sharing

<img width="952" alt="Image" src="https://github.com/user-attachments/assets/4fbbf9c3-6aee-4298-8d48-6a5305d33b63" />

### XP Summary (total 350)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

manuq:
We debugged a little bit:
- Happens to @jbourqueendless in main as well
- We have the same Godot version 4.4.1 stable, except he is on Windows and I'm on Endless OS
- Doesn't happen in the web build
- He tried reloading the project, removing the `.godot/` folder then reloading Godot. None of this worked.

wjt:
@jgbourque did you ever figure this one out?

jgbourque:
So it seems like this first guard does something different now where it does detect me but then sometimes will not give much chase and or officially end the run and instead is just alert to my presence. It is still behaving in a way that feels like as long as I don't physically come in contact with that guard I am fine.
[video capture](https://drive.google.com/file/d/1-YsJAD-TDAsGALyxBQEVQSeDxDXRiQ1M/view?usp=sharing)

jgbourque:
@manuq What data do you need that might help debug from the playtest? Is there something in Godot that you need to see while I am running the scene? Let me know and I can rerecord a capture or add any data that might help.

manuq:
Closing since this wasn't reported anymore.

manuq:
Reopening. This is "a Justin's bug" because so far we saw it only on @jgbourque computer.

manuq:
For the record, I wasn't able to reproduce this in a high refresh rate monitor. So this is still a Justin's bug.

PR #721, opened 2025-05-29
Guard: Add Storyvore

manuq:
Resolves https://github.com/endlessm/threadbare/issues/714

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -244,12 +244,11 @@ func _update_player_awareness(player_in_sight: Node2D, delta: float) -> void:
 
 
 func _update_animation() -> void:
-	if state == State.ALERTED and animation_player.current_animation == &"alerted":
+	if state == State.ALERTED:
 		return
 
 	if velocity.is_zero_approx():
-		animation_player.stop()
-		animated_sprite_2d.play(&"idle")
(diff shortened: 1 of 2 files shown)


manuq:
I have redone the sprite frames after PixelFrog was kind enough to flip the assets horizontally. The characters animation is a bit odd, as you can see below. But it looks like an existing bug. The idle animation is not being played, and the attack is choppy.

[Grabación de pantalla desde 2025-06-02 18-21-32.webm](https://github.com/user-attachments/assets/e317c47b-ee80-4f1b-a7ff-72b832fdb5c7)

The stealth also looks very CPU intense so the recording is bad. It is worth measuring performance so I'll try using the Profiler tab in the Debugger dock with it. Maybe https://github.com/endlessm/threadbare/issues/680 is about that as well.

manuq:
I have appended commits to fix the (existing) bug of the idle animations not being played. And since Storyvore and guard template have animations of different length, I took the decision to make it work for the one with longer frames (Storyvore). So the animations in the template for attack and idle will have still frames at the end before looping. This doesn't look bad. It is just an odd situation that the final character arrived after we shipped the template.

[Grabación de pantalla desde 2025-06-03 10-42-19.webm](https://github.com/user-attachments/assets/9ee3b870-cd74-495d-9c5a-77c6cddad68f)

wjt (approved):
LGTM!

PR #722, opened 2025-06-02
Duplicate template and rename to threadcraft

DonSeja06:


JuanFdS:
I think this PR was meant to target the fork itself instead of the main repo

Issue #692, opened 2025-05-21
Hide template StoryQuest when not launched from editor


wjt:
Once we have at least one non-template StoryQuest merged here, I think we should hide the template StoryQuest in exported builds. (If the player has launched the project from the Godot Editor, it's fine to show it.)

wjt:
I discussed this with @hydrolet in person. She doesn't think we should hide the template quest in web builds: it's very useful to show prospective partners as much as possible of what learners will see without them having to download Godot, clone the project, etc.

I think I filed this in response to a playthrough by @saspiers where she pointed out that the dialogue instructs you to do things (e.g. "go to the inspector") that you can't do in an exported build. So a suggestion was that we could tweak the dialogue if the game is not running in the editor, to skip instructions that the player can't follow. But that would add complexity to the template and also undermine the "this is what learners will see".

So perhaps we can address this by making the template's description say something like "Best used in the Godot editor"; and also the problem will be less of a problem as we have more and more StoryQuests, particularly if we group them into different sections.

wjt:
I filed #750 to add some of the info needed to group quests by how complete they are. #751 adds the serving suggestion.

Issue #702, opened 2025-05-23
Publish latest release as top-level page


wjt:
Currently https://endlessm.github.io/threadbare/ is `main`.

Now that we have a release, I think it would be better to have the latest release at the top-level, and `main` under https://endlessm.github.io/threadbare/branches/ like any other branch.

PR #727, opened 2025-06-04
Storybook: Hide template in exported builds

manuq:
Allow to play the template story quest only when running the project from the editor.

Resolves https://github.com/endlessm/threadbare/issues/692

### DIFF
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -44,7 +44,9 @@ func _enumerate_quests() -> Array[Quest]:
 			else:
 				quests.append(quest)
 
-	if has_template:
+	# Add the template quest only when running the project in the editor. It doesn't
+	# make much sense to play the template standalone in exported builds.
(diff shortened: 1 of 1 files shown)


manuq:
Can be tested directly from: https://endlessm.github.io/threadbare/branches/endlessm/hide-template-in-builds/#world_map/frays_end

wjt:
I think this makes sense but I would like to check with the learning team – do they rely on being able to demo the template in a web browser?

Another approach we could take is:

- Implement #702 
- Export in release mode only when the git tree we're building is a release
  - https://github.com/endlessm/threadbare/blob/main/.github/workflows/export.yml#L67 we currently use release mode when exporting to the web
- Make this conditional on being a release build

wjt:
I left some notes on #692 of which the summary is: the learning team wants the template to be in exported builds.

manuq:
> I left some notes on #692 of which the summary is: the learning team wants the template to be in exported builds.

Excellent! Thanks for checking.

Issue #381, opened 2025-04-29
Consolidate themes


wjt:
We have two themes in the project, one for dialogue and menus and the other for a few other labels:

scenes/ui_elements/dialogue/components/theme.tres
scenes/ui_elements/shared_components/theme.tres

We should consolidate these.

wjt:
In #427 I removed one of the 3 uses of `scenes/ui_elements/shared_components/theme.tres`. I think we're in striking distance of having just one theme, based on the dialogue theme but moved to a more descriptive filename.

PR #732, opened 2025-06-04
Fullscreen option

manuq:
This adds a fullscreen ON/OFF checkbox to the Options menu, that nobody asked but I wanted to implement for a while:

Captura desde 2025-06-04 13-51-22

Like the volume settings, the change is persisted.

Note: I did not want to deal with the theme for now, there is https://github.com/endlessm/threadbare/issues/381 for that.

Note 2: Ideally all settings should be in a single GridContainer, so the labels and controls are aligned. I tried but I wasn't able to make the "Sound Settings" and "Video Settings" labels occupy 2 columns of the grid.

### DIFF
--- a/scenes/globals/settings/settings.gd
+++ b/scenes/globals/settings/settings.gd
@@ -10,6 +10,9 @@ const DEFAULT_VOLUMES: Dictionary[String, float] = {
 	"Music": -15.0,
 }
 
+const VIDEO_SECTION := "Video"
+const VIDEO_WINDOW_MODE_KEY := "Window Mode"
+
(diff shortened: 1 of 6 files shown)


manuq:
Works in web too: https://endlessm.github.io/threadbare/branches/endlessm/fullscreen-option/

wjt:
On the web platform the fullscreen setting can't be restored without customising the HTML shell because a page can only enter fullscreen in response to user input https://docs.godotengine.org/en/latest/tutorials/export/exporting_for_web.html#full-screen-and-mouse-capture but I think that's fine.

Issue #231, opened 2025-04-17
Document Git LFS requirement


wjt:
Someone cloned & opened the project and hit the following:

![Image](https://github.com/user-attachments/assets/0301f2c6-d525-4ceb-9c21-83fb08e3db88)

I believe this is because they do not have Git LFS configured. This needs to be documented somewhere.

wjt:
Another documentation-adjacent need: I think we should have an editor plugin in the project which (on startup) detects the case of the project being cloned without Git LFS support and shows a warning, along with a link to further documentation.

wjt:
https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage has quite good documentation for all 3 major platforms (Windows, macOS, Linux)

wjt:
I have created https://github.com/endlessm/threadbare/discussions/772 as a target for a warning message shown in the editor when a new plugin I'm writing detects that the project has been cloned without LFS support.

PR #773, opened 2025-06-12
Warn if the project has been cloned without Git LFS

wjt:
A common problem is to clone the project on a system that does not have Git LFS installed, and then be confused by error messages reading "Load failed due to missing dependencies".

Add an editor plugin that detects this case, by checking whether icon.png at the root of the repository (accessed by UID, following our convention) contains a Git LFS pointer. In this case, show a pop-up window with a warning. When the game developer clicks Learn More… in the dialogue, take them to an appropriate thread on our discussion forum.

https://github.com/endlessm/threadbare/issues/231


addons/git_lfs_checker/plugin.gd
```diff
@@ -0,0 +1,62 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends EditorPlugin
+
+const SOME_LFS_MANAGED_FILE: String = "uid://bc70phmq55dkf"
```
manuq, line 6: Nice to use the icon, which is unlikely to be removed.

addons/git_lfs_checker/plugin.gd
```diff
@@ -0,0 +1,62 @@
...
+	var file := FileAccess.open(SOME_LFS_MANAGED_FILE, FileAccess.READ)
+	if not file:
+		var error := FileAccess.get_open_error()
+		push_warning(
+			"Failed to open ", SOME_LFS_MANAGED_FILE, " to check for Git LFS: ", error_string(error)
+		)
+		return false
+
+	var lfs_header := LFS_HEADER.to_ascii_buffer()
+	var header := file.get_buffer(lfs_header.size())
```
manuq, line 40: :+1:

addons/git_lfs_checker/plugin.gd
```diff
@@ -0,0 +1,62 @@
...
+	_dialog.dialog_autowrap = true
+	_dialog.dialog_text = WARNING_MESSAGE.dedent().strip_edges()
+	_dialog.ok_button_text = OK_BUTTON_TEXT
+	EditorInterface.get_editor_main_screen().add_child(_dialog)
+	_dialog.confirmed.connect(_show_lfs_documentation)
+	_dialog.popup_centered()
+
+
+func _show_lfs_documentation() -> void:
+	OS.shell_open(LFS_DOCUMENTATION_URL)
```
manuq, line 56: Even opening the documentation, good touch.

PR #777, opened 2025-06-12
Components prototype

manuq:
This is for discussion in https://github.com/endlessm/threadbare/discussions/774

### DIFF
--- a/scenes/game_elements/characters/enemies/guard_new/guard_new.tscn
+++ b/scenes/game_elements/characters/enemies/guard_new/guard_new.tscn
@@ -1,145 +1,23 @@
-[gd_scene load_steps=12 format=3 uid="uid://dp5wc4xuxruon"]
+[gd_scene load_steps=8 format=3 uid="uid://dp5wc4xuxruon"]
 
-[ext_resource type="Script" uid="uid://mwgpxjf6tdx4" path="res://scenes/game_logic/character.gd" id="1_qws0c"]
+[ext_resource type="Script" uid="uid://c1fydd8vlq6mx" path="res://scenes/game_elements/characters/enemies/guard_new/components/guard_new.gd" id="1_40xag"]
 [ext_resource type="SpriteFrames" uid="uid://xolbc1j4a62f" path="res://scenes/game_elements/characters/enemies/guard/components/storyvore_frames_blue.tres" id="1_x211u"]
 [ext_resource type="Script" uid="uid://csqb1r1pasyi6" path="res://scenes/game_logic/can_walk_along_path.gd" id="3_a8qh0"]
-[ext_resource type="Script" uid="uid://b6cesrvjlada7" path="res://scenes/game_logic/can_talk.gd" id="3_e14if"]
 [ext_resource type="Texture2D" uid="uid://c75ofhy3swhj3" path="res://scenes/game_elements/characters/enemies/guard/components/exclamation_mark.png" id="3_l4ps3"]
-[ext_resource type="PackedScene" uid="uid://dbam7hbkdj2vw" path="res://scenes/game_elements/props/interact_area_new/interact_area_new.tscn" id="4_40xag"]
 [ext_resource type="Script" uid="uid://cf66tvbhbaxd0" path="res://scenes/game_logic/can_follow_target.gd" id="6_xja8d"]
 
 [sub_resource type="CapsuleShape2D" id="CapsuleShape2D_pvecv"]
 height = 58.0
 
-[sub_resource type="GDScript" id="GDScript_xja8d"]
-script/source = "extends Node
-
-const GIVE_UP_TIME: float = 5.0
-const TURN_SPEED: float = 10.0
-
-enum State {
-	## Going along the path.
-	PATROLLING,
-	## Player is in sight, it takes some time until the player is detected.
-	DETECTING,
-	## Player was detected.
-	ALERTED,
-	## Player was in sight, going to the last point where the player was seen.
-	INVESTIGATING,
-	### Lost track of player, walking back to the patrol path.
-	#RETURNING,
-}
-
-@export_category(\"Target Detection\")
-## Time required to detect the target.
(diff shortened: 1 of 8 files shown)


demo.tscn
```diff
@@ -0,0 +1,36 @@
...
+[node name="ScreenOverlay" type="CanvasLayer" parent="."]
+
+[node name="World" type="Node2D" parent="."]
+y_sort_enabled = true
+
+[node name="Guard" parent="World" instance=ExtResource("2_0e48y")]
+position = Vector2(242, 238)
+
+[node name="CanWalkAlongPath" parent="World/Guard" index="3" node_paths=PackedStringArray("walking_path")]
+walking_path = NodePath("../../../Path2D")
```
wjt: Ah I see. You need to be able to reach "inside" the guard in order to set the walk path. And in order to do that, you need to enable Editable Children...

But in general that's what you would need to do anyway to change the sprite/hitbox/etc.

scenes/game_logic/can_interact.gd
```diff
@@ -0,0 +1,75 @@
...
+	else:
+		interact_label.visible = true
+		interact_label.label_text = interact_area.action
+		interact_marker.global_position = interact_area.get_global_interact_label_position()
+
+	if not is_zero_approx(character.velocity.x):
+		if character.velocity.x < 0:
+			scale.x = -1
+		else:
+			scale.x = 1
```
wjt: FTR this can be written as 
```suggestion
		scale.x = signf(character.velocity.x)
```

demo.tscn
manuq: Yes, I guess the guard will need to pass through the Path2D using an export.

scenes/game_logic/can_interact.gd
manuq: I didn't know, thanks!

scenes/game_logic/can_follow_target.gd
```diff
@@ -0,0 +1,73 @@
...
+var update_target_time: float = 0.5
+
+@export var node_to_follow: Node2D:
+	set = _set_node_to_follow
+
+var _update_target_timer: Timer = Timer.new()
+var _target_position: Vector2
+
+# TODO: Add configuration warning to the node if the parent is not a Character.
+@onready var character: Character = get_parent()
```
wjt, line 32: Something I saw in one of Firebelley Games' videos is using `owner` here rather than `get_parent()`. `owner` means the scene that owns the node. So if you have a scene frays_end.tscn which contains an instance P of player.tscn, and player.tscn contains a node N with `can_follow_target.gd` attached, then IIUC when you play frays end, N.owner is P.

Using owner rather than parent would allow structuring the character scenes with the behaviours as descendents of the root (not just a direct child) but it would mean (I believe!) that you couldn't add them to an instance in another scene. (In the example above, if you added a CanTalk as a child of P, its owner would be frays end.)

scenes/game_elements/characters/enemies/guard_new/guard_new.tscn
```diff
@@ -0,0 +1,213 @@
...
+	match previous_state:
+		State.DETECTING:
+			_detecting_timer.stop()
+		State.INVESTIGATING:
+			_investigating_giveup_timer.stop()
+
+	match new_state:
+		State.PATROLLING:
+			can_walk_along_path.process_mode = Node.PROCESS_MODE_INHERIT
+			can_follow_target.process_mode = Node.PROCESS_MODE_DISABLED
```
wjt: Don't you want to always disable `can_follow_target` unless in the `INVESTIGATING` state? Or, hmm, I guess not because there is not a 1:1 mapping between the behaviour nodes and the states.

I think this structure will make it hard to add new states. Suppose someone wanted to add a ranged attack behaviour to a guard in their storyquest – perhaps triggered when the guard starts chasing the player but can't get across a river. They wouldn't be able to do that without copying this whole function and then adding new cases to it. It would be nice in principle to be able to do that.

scenes/game_logic/can_follow_target.gd
manuq: Actually `owner` was how I had it at the beginning before switching to `get_parent()`. It is how we currently do some components of the player too like `player_fighting.gd`, as I did originally.

I think there are pros and cons. As you said using owner is good to add the component anywhere inside the game entity scene. But it is a bit inflexible: you must save your entity in a scene first, and you can't try out adding components in a level like Fray's End.

scenes/game_elements/characters/enemies/guard_new/guard_new.tscn
manuq: You have a good point here. The GuardLogic script doesn't have a 1:1 mapping between states and behavior nodes. My hope was that some behavior could be split into separate components, while some more trivial or too-specific behavior could stay in this specific game entity script. So in the state setter, I have:
- `match previous_state` where I do the "leave" step of states that don't map to a behavior node.
- `match new_state` where I do the "enter" step of states that don't map to a behavior node, plus disabling/enabling processing of the behavior nodes.

So yeah this is not a pure or "finite state machine". It is more similar to what we have already. I think that it would be worth exploring the pros of having a StateMachine class (like Firebelley's) that do the state transitions automatically. Maybe using callables for the "inline" states?

Just for reference of what we talked about today: State machines and reusable components are different things, while related:

- The goal of wrapping behavior in components is reusability. Which can become valuable as learners try to implement new game mechanics.
- The goal of FSMs is to split the increasing complexity of `_process` or `_physics_process` as the player controls or an NPC AI becomes advanced.

Our current states in Player, Guard and ThrowingEnemy are based on enums and a variable:
```
enum State { PATROLLING, DETECTING, ALERTED, ... }
var state: State = State.PATROLLING
```
Which matches "Implementing a Finite State Machine with one variable" in this [GDQuest tutorial](https://www.gdquest.com/tutorial/godot/design-patterns/finite-state-machine/).

Issue #791, opened 2025-06-16
Teleporter's Spawn Point Path list ignores instanced scenes


wjt:
### Description

`teleporter.gd` has some custom code to populate the "Spawn Point Path" property in the inspector by introspecting the target scene and finding spawn points within it. But while making an example for the video in #790 I noticed that it doesn't quite work right. After I added a new `SpawnPoint` node to Fray's End, here is the list of spawn points offered:

![Image](https://github.com/user-attachments/assets/02a90a98-3e45-4eda-a796-b720656b4154)

But there is another spawn point in Fray's End: `SpawnPointAfterIntro`. It is not shown in this list.

The difference between `SpawnPointAfterIntro` and the `SpawnPoint` I added is that the former is an instanced scene while the latter is a bare node:

![Image](https://github.com/user-attachments/assets/0885aabf-88a0-4d63-b62a-489330f049e8)

There's nothing much to speak of in `spawn_point.tscn`, just a `Marker2D` node with the `spawn_point.gd` script attached.

https://github.com/endlessm/threadbare/blob/4e40d710c9358463b154e0f4ce94997e74d2745e/scenes/game_elements/props/spawn_point/spawn_point.tscn

It was added in commit a5b36dc208bfb119cdff7c836cd23a14964bd2cd ("Reorganize project files"), with (I think) the goal of consistently providing a scene that can be instantiated for each component.

We should either remove the `spawn_point.tscn` scene, or (better IMO) fix the [logic in teleporter.gd](https://github.com/endlessm/threadbare/blob/4e40d710c9358463b154e0f4ce94997e74d2745e/scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd#L99-L110) to detect the instantiated-scene form of spawn points as well as the bare node form.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #792, opened 2025-06-16
Teleporter: Detect instances of spawn_point.tscn

wjt:
Teleporter has (among other properties) a `next_scene` property, which is the resource path of the scene to switch to when the player enters the teleporter, and a `spawn_point_path` property, which is the `NodePath` within `next_scene` of a `SpawnPoint` (a custom subclass of `Marker2D`). The tricky thing here is that normally an `@export`ed `NodePath` refers to paths within the current scene, but here it refers to a path in a different scene.

To make this property easier to use in the editor, there is custom code which introspects the target of the `next_scene` property, listing all spawn points within, and then exposes the `spawn_point_path` property as an enum listing all those paths, plus a NONE entry to not use a spawn point.

Spawn points in the remote scene are found by their group. `spawn_point.gd` is also a `@tool` script and adds itself to the `spawn_point` group.

However, as well as `spawn_point.gd` we also have a `spawn_point.tscn` scene (which just contains a Marker2D with `spawn_point.gd` attached). When inspecting a `next_scene` with an instance of `spawn_point.tscn`, then the entry in the `SceneState` for `next_scene` does not list the spawn point as being in the spawn_points group. For example, in `frays_end.tscn` we have:

    [node name="SpawnPointAfterIntro" parent="." instance=ExtResource("37_thm8h")]
    position = Vector2(62, 1747)

As a result, if you have a teleporter in another scene that targets `frays_end.tscn`, `SpawnPointAfterIntro` does not appear in the drop-down list.

To fix this, when enumerating nodes in the `SceneState`, check if each one is an instanced scene; if so, get the SceneState for that scene, and also consider the groups for its root node.

Fixes https://github.com/endlessm/threadbare/issues/791

### DIFF
--- a/scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
+++ b/scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
@@ -6,12 +6,27 @@ extends Area2D
 
 const SPAWN_POINT_GROUP_NAME: String = "spawn_point"
 
+## Scene to switch to when the player enters this teleport. If empty, the player
+## will teleport within the current scene, to the position specified by [member
+## spawn_point_path].
 @export_file("*.tscn") var next_scene: String:
 	set(new_value):
 		next_scene = new_value
 		_update_available_spawn_points()
 		notify_property_list_changed()
 
+## Which SpawnPoint in [member next_scene] the player character should start at;
+## or blank/NONE to start at the default position in the scene.
+@export var spawn_point_path: NodePath:
+	set(new_val):
+		if new_val == ^"NONE":
+			spawn_point_path = ^""
+		else:
+			spawn_point_path = new_val
+
+@export_group("Transition")
+
+## Whether to use a visual transition effect when the player enters the teleporter.
 @export var use_transition: bool = true:
 	set(new_val):
 		use_transition = new_val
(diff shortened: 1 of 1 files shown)


wjt:
I added a couple more patches to also improve the documentation of these exported properties.

scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
```diff
@@ -102,8 +102,13 @@ func _update_available_spawn_points() -> void:
...
 		for i: int in scene_state.get_node_count():
-			if SPAWN_POINT_GROUP_NAME in scene_state.get_node_groups(i):
-				var node_path_as_string: String = String(scene_state.get_node_path(i))
+			var path := scene_state.get_node_path(i)
+			var node_groups := scene_state.get_node_groups(i)
+			var instance := scene_state.get_node_instance(i)
+			if instance:
+				node_groups.append_array(instance.get_state().get_node_groups(0))
+			if SPAWN_POINT_GROUP_NAME in node_groups:
+				var node_path_as_string: String = String(path)
```
manuq: Does the editor warn about untyped declaration if you remove the type here?
```
var node_path_as_string := String(path)
```

scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
```diff
@@ -23,7 +23,9 @@ const SPAWN_POINT_GROUP_NAME: String = "spawn_point"
 ## Transition to use when the player leaves this teleport.
 @export var exit_transition: Transition.Effect = Transition.Effect.RIGHT_TO_LEFT_WIPE
 
-var spawn_point_path: NodePath:
+## Which SpawnPoint in [member next_scene] the player character should start at;
+## or blank/NONE to start at the default position in the scene.
+@export var spawn_point_path: NodePath:
```
manuq: :+1:

manuq (approved):

scenes/game_elements/props/teleporter/teleporter/components/teleporter.gd
wjt: Good point. No, that works and is more legible. I've amended the patch.

PR #793, opened 2025-06-16
Fray's End: Simplify serialised flowers in scene

wjt:
This is the result of loading frays_end.tscn and saving it without making any changes.

I don't know why these had the redundant type & script information when they are instanced scenes.


wjt:
@W1ndl0ve thank you for taking a look!

PR #814, opened 2025-06-18
PR template: remove

### LINKED ISSUE #811: Remove PR template
It doesn't appear to be serving its purpose of preventing PRs from being mistakenly opened upstream, and could cause confusion for learners. Instead, @endlessm/learning is going to focus on improving documentation by creating a specific guide for opening a PR against a forked repo, and then we can link to that guide any time a PR is mistakenly opened against upstream.

cassidyjames:
Fixes #811


wjt:
OK, so my understanding is not quite right. I'll follow up on #815

Issue #704, opened 2025-05-23
Editing a release reruns its build action


wjt:
I edited the v0.1 release after publishing it. This triggered a new run of the "Build and Export Game" workflow:

![Image](https://github.com/user-attachments/assets/f825dcd0-985a-4bc9-8f65-f675fc923bc7)

and it failed because it tried to re-attach the build artifacts.

I think the workflow should be triggered only on the `published` type of the `release` event https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release.

wjt:
Also the Flatpak build didn't run because it is only run for a push to main.

I think we could move the `Build Flatpak` job to the `Build and Export Game` workflow, dependent on the previous job running, and make it conditional on being a release (or maybe also a merge to `main`) if not in a fork. It essentially already has that condition so this would just be a simplification.

PR #858, opened 2025-06-23
ci: Only run release CI when release is published

wjt:
Previously, the export workflow was run for any `release` event. But there are several types of `release` events, so previously this workflow would be triggered multiple times when making a release. This is wasted effort because the same artifact can't be attached to the release twice.

Instead, only run the job for the `published` type of `release` event.

Fixes https://github.com/endlessm/threadbare/issues/704

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -5,6 +5,8 @@ on:
   workflow_dispatch:
   pull_request:
   release:
+    types:
+      - published
   push:
     branches:
       - '*'


.github/workflows/export.yml
```diff
@@ -5,6 +5,8 @@ on:
   workflow_dispatch:
   pull_request:
   release:
+    types:
+      - published
```
manuq, line 9: :+1:

manuq (approved):

PR #866, opened 2025-06-24
Add web and Flatpak builds to releases

### LINKED ISSUE #865: Attach web build to releases
Currently, we add the `.pck` file to releases as an asset, but we do not add the web build (or Flatpak build for that matter).

If we did this, we would be able to (in future) offer a playable list of past releases; and/or change the layout of the GitHub Pages site so that the top-level page is the latest (pre-)release while the `main` branch is under `/branches/`.

wjt:
Previously, the only file attached to releases as an asset was the `.pck` file.

I would like to change the GitHub Pages site to have the latest release at its root, and the `main` branch under `branches` like every other branch. To do this, amalgamate-pages will need to be able to get the web build asset from the latest release.

First, some preparatory work:

- Make the `wget` output less verbose – I noticed this while testing the PR on a throwaway fork which did not have Godot & the export templates cached.
- Merge the Flatpak build with the main build/export flow – this fixes a problem I noticed at https://github.com/endlessm/threadbare/issues/704#issuecomment-2905585396
- Attach Flatpak build as a release asset
- Disable Flatpak build on `main`, unless explicitly requested – this saves needless computation

And then the main event: attach the web build to releases. To do this we repack the files from the build artifact in the same format (a zip file) to ease the implementation on the `amalgamate-pages` side.

Fixes #865

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -144,3 +144,19 @@ jobs:
         run: |
           mv threadbare.flatpak 'threadbare-${{ github.ref_name }}-x86_64.flatpak'
           gh release upload '${{ github.ref_name }}' 'threadbare-${{ github.ref_name }}-x86_64.flatpak' --repo '${{ github.repository }}'
+
+      - name: Download web bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: web
+          path: web
+
+      - name: Attach web bundle to release
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          pushd web
+          zip -r "../threadbare-${{ github.ref_name }}-web.zip" .
+          popd
+
+          gh release upload '${{ github.ref_name }}' 'threadbare-${{ github.ref_name }}-web.zip' --repo '${{ github.repository }}'


wjt:
Some test builds:

- https://github.com/endlessm/threadbare/actions/runs/15847963492 shows the new checkbox to enable the Flatpak build in the workflow when needed
- https://github.com/wjt/threadbare/releases/tag/test5 is a release on my throwaway fork which shows the 2 new assets. This is not built from the exact same git tree – I did a little bit of rebasing/squashing – but I hope it demonstrates that the release flow (which can't be tested on a PR) works!

.github/workflows/export.yml
```diff
@@ -45,13 +45,13 @@ jobs:
           mkdir -p build && cd build
 
           # Download Godot Engine itself
-          wget https://github.com/godotengine/godot/releases/download/${GODOT_VERSION}-stable/Godot_v${GODOT_VERSION}-stable_linux.x86_64.zip && \
+          wget --progress=dot:giga https://github.com/godotengine/godot/releases/download/${GODOT_VERSION}-stable/Godot_v${GODOT_VERSION}-stable_linux.x86_64.zip && \
```
manuq, line 52: Good to know about this setting to output less dots for big downloads. Well explained too.

.github/workflows/export.yml
```diff
@@ -90,10 +94,7 @@ jobs:
   flatpak:
     name: Build Flatpak
     needs: build
-    if: ${{
-      !github.event.repository.fork &&
-      (github.ref == 'refs/heads/main' || github.event_name == 'release')
-      }}
+    if: ${{ inputs.build_flatpak || github.event_name == 'release' }}
```
manuq, line 97: :+1:

.github/workflows/export.yml
```diff
@@ -144,3 +144,19 @@ jobs:
...
+        with:
+          name: web
+          path: web
+
+      - name: Attach web bundle to release
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          pushd web
+          zip -r "../threadbare-${{ github.ref_name }}-web.zip" .
```
manuq, line 159: So is the `../threadbare-${{ github.ref_name }}-web.zip` actually a directory? Is that what you describe in the commit message as "when you use `actions/download-artifact` the zip file is helpfully unpacked"?

wjt: No, it's a zip file.

What I'm getting at is: if you go to https://github.com/endlessm/threadbare/actions/runs/15847963492, you will see there are three artifacts. Each of them is a zip file. [threadbare-x86_64.flatpak](https://github.com/endlessm/threadbare/actions/runs/15847963492/artifacts/3391034161) for example is actually a zip file, which contains a file called `threadbare.flatpak`. All build artifacts are always zip files.

In this job, we want to fetch an artifact and attach it as an asset to a release. When you use [actions/download-artifact](https://github.com/actions/download-artifact), then rather than getting a zip file in the target directory (`path` parameter, defaults to current directory), the zip is unpacked and all its files placed into the target directory.

In the case of the pck and flatpak files, this is actually what we want. We want to attach the pck file and flatpak file directly to the release, not wrapped up in a zip file. Unlike build artifacts, release assets are files of whatever type you like.

But in the case of the web build, it is a collection of files, which we instruct `actions/download-artifact` to place into a directory called `web`. So we need to put it into some archive format. `amalgamate-pages` currently expects to consume zip files – that's what build artifacts are – so I chose to pack the files back into a zip file. (It's a little annoying that you can't tell `actions/download-artifact` not to unpack the zip file, but there we go.)

manuq: Got it, thanks for expanding on the artifacts behavior.

.github/workflows/export.yml
```diff
@@ -3,6 +3,10 @@
 name: "Build and Export Game"
 on:
   workflow_dispatch:
+    inputs:
+      build_flatpak:
+        description: 'Also build a Flatpak bundle'
+        type: boolean
```
cassidyjames, line 9: TIL you can do this. How handy!

PR #868, opened 2025-06-24
Guard refactor

manuq:
Helps https://github.com/endlessm/threadbare/issues/680


manuq:
I wanted to go with a bigger refactor, after my learnings checking the guard's behavior. For now, here is a little improvement that hopefully makes the script a bit more readable too. Instead of constantly checking for overlapping bodies in 2 areas for each guard, use the body entered and body exited signals.

PR #872, opened 2025-06-25
Ignore macOS system files; and zip and rar files

wjt:
I spotted some inadvertently-committed files in a learner branch. We can be kind to new Git users by proactively ignoring files that we don't want in the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ build/
 .DS_Store
 ._*
 __MACOSX
+
+# Archive formats
+*.zip
+*.rar
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 # Build & export artifacts
 build/
 
-# From Tiny Swords pack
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,6 +8,12 @@
 # Build & export artifacts
 build/
 
-# From Tiny Swords pack
+# macOS system files
+.AppleDouble
 .DS_Store
-.aseprite
```
manuq, line 13: How is it that we had  `.aseprite` before but we managed to commit a lot of them? Like `scenes/game_elements/characters/enemies/guard/components/Storyvores.aseprite`

wjt: Because this only ignores files whose name is `.aseprite`, not files whose names end with `.aseprite`. If this line had said `*.aseprite` then those files would have been ignored.

manuq: Ouch, of course!

PR #873, opened 2025-06-25
Project: Add color codes to filesystem folders

manuq:
Set colors to the scenes/quests subfolders:

- Lore Quests are purple
- Story Quests are green
- Template for Story Quests is yellow

Captura desde 2025-06-25 09-01-27


manuq:
@jofilizola and @Stregoica777 agreed on these colors to start.

PR #887, opened 2025-06-26
WIP: floor eating challenge

wjt:


wjt:
I pushed a new variant of this challenge:

https://endlessm.github.io/threadbare/branches/endlessm/world-eater#uid://bm4ewr8p48x0i

The difference is that there's a visible enemy who chases you and destroys the world beneath them as they go. I used the Storyvore as a placeholder but the real enemy would need to be something that floats and has some visible effect beneath them.

I made it erase a 3×3 grid rather than the cross shape of the other experiment.

I also dressed the level a bit, making the path and the sanctuary island be raised on cliffs, adding water around the sanctuary. It looks a bit better.

Unfortunately what we actually need is a kind of void tile that does not block movement, because it looks bad that the void has seams where it joins the walkable world.

I wonder if we could do this:

1. No collisions on the void;
2. Use clifftops with edge collisions under the grass/path in order to stop the player walking off the edge.

The same might work for water. In that way we wouldn't have to have the 3 types of water (walkable, half-walkable, non-walkable) though of course we'd need some kind of semi-visible edge tiles and an invisible half-blocking tile for the bridge in fray's end...

wjt:
I'm focusing on polishing https://endlessm.github.io/threadbare/branches/endlessm/world-eater/#experimental/enemy_destroys_floor/enemy_destroys_floor. It now has a sprinkling of plot and an effect where the void retreats once you talk to a (placeholder) monk.

Remaining steps:

- [ ] Add more props (trees etc.) under the initial void area, revealing them when you interact with the monk
- [ ] Maybe try to split up the chasing script a little?
- [ ] Then squash

I have some other changes on this branch so I paused to split out the water rock decoration: https://github.com/endlessm/threadbare/issues/970

PR #888, opened 2025-06-26
Guard: Use correct class for Player

manuq:
The collision methods like body entered return a Node2D. But we must check that the player is the one being returned because the stealth game logic treats the node as a player, changing the mode to Player.Mode.DEFEATED when detected by the guards.

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -5,7 +5,7 @@ class_name Guard extends CharacterBody2D
 ## Enemy type that patrols along a path and raises an alert if the player is detected.
 
 ## Emitted when the player is detected.
-signal player_detected(player: Node2D)
+signal player_detected(player: Player)
 
 enum State {
 	## Going along the path.
@@ -89,7 +89,7 @@ var state: State = State.PATROLLING:
 	set = _change_state
 
 # The player that's being detected.
-var _player: Node2D
+var _player: Player
 
 ## Area that represents the sight of the guard. If a player is in this area
 ## and there are no walls in between detected by [member sight_ray_cast], it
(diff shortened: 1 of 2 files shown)


scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -504,22 +504,28 @@ func _set_alert_other_sound_stream(new_value: AudioStream) -> void:
 
 
 func _on_instant_detection_area_body_entered(body: Node2D) -> void:
+	if not body is Player:
+		return
 	state = State.ALERTED
-	player_detected.emit(body)
+	player_detected.emit(body as Player)
```
wjt, line 510: I believe that an `is` check narrows the type of the variable; i.e. I believe that inside an `if` block that has checked the type of `body` to be `Player`, then `body` has the type `Player`. Can you try this and see if it type-checks?

```gdscript
if body is Player:
    state = State.ALERTED
    player_detected.emit(body)
```

scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -504,22 +504,28 @@ func _set_alert_other_sound_stream(new_value: AudioStream) -> void:
...
 
 
 func _on_detection_area_body_entered(body: Node2D) -> void:
-	_player = body
+	if not body is Player:
+		return
+	_player = body as Player
```
wjt, line 516: But in this case I think the early return is clearer than indenting the whole function.

wjt (changes requested):

scenes/game_elements/characters/enemies/guard/components/guard.gd
wjt: I re-read https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html and I think I might have imagined this...

Some languages have this feature and I thought gdscript was one of them!

wjt: The funny thing is, I can't get any error or warning to be logged by Godot for passing the wrong type of argument to `.emit()`!

wjt (approved):

scenes/game_elements/characters/enemies/guard/components/guard.gd
manuq: Yes, it wasn't warning before as well.

PR #902, opened 2025-06-29
Minijuego3

AlexanderV24:


scenes/quests/story_quests/Where_we_at/Minijuego_Terminal.tscn
```diff
@@ -0,0 +1,117 @@
+[gd_scene load_steps=5 format=3 uid="uid://bpnb6hfx88ac"]
+
+[ext_resource type="Script" path="res://scenes/quests/story_quests/Where_we_at/5_Where_we_at_mistery/minijuego_terminal.gd" id="1_dmjd5"]
```
wjt, line 3: This scene can't be loaded because this file is not committed.

Issue #747, opened 2025-06-09
Automate instantiating StoryQuest template


wjt:
We currently ask learners to duplicate `scenes/quests/story_quests/template` to `scenes/quests/story_quests/my_new_storyquest` (name of their choice); then rename all the files within to replace `template` with `my_new_storyquest`, and relink all the scenes together & update the resources in the scene files/sprite frames/etc. to refer to assets/resources in the copy.

IMO we should automate this: it's error-prone manual work that stands in the way of learning; and doesn't teach anything useful, except perhaps replacing the resource references which teaches you something about what Duplicate does/doesn't do, but it's probably too early in a learner's journey with Godot for them to learn this lesson.

wjt:
I think this could be implemented incrementally:

1. Add a plugin for a menu item which literally just does the "duplicate step"
2. Teach it to rename the files
3. (The hard bit) Teach it to recursively relink resources that point to the template to the copies in the instantiated template
4. Give it fields to populate the `quest.tres` file

Stregoica777:
I actually do think that to duplicate and rename by hand teaches you something: you need to know Godot's UI and the path to get to the templates, you need to inspect what you are renaming, etc. That plus patience. 
But that's my opinion. If we didn't have the time crunch we manage with the program, I would def put a lot more emphasis on making learners do the process manually, but since we DO have a really tight schedule... well... maybe we can skip this part to pay attention to other things.

PR #909, opened 2025-06-30
Add plugin to automate the creation of StoryQuests

manuq:
Resolves https://github.com/endlessm/threadbare/issues/747


addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,89 @@
...
+	)
+	assert(error == OK)
+
+	error = DirAccess.make_dir_recursive_absolute(
+		storyquest_path + "/4_template_outro".replacen("template", filename)
+	)
+	assert(error == OK)
+
+	var template_intro_scene_subpath := "/0_template_intro/template_intro.tscn"
+	var intro_scene_subpath := template_intro_scene_subpath.replacen("template", filename)
```
wjt: I talked with @hydrolet last week about the redundant `template_` prefix in each subdirectory and file. If I remember the conclusion correctly: it's desirable for the **template** files to have a clear marker in their filename that indicates that they shouldn't be edited, e.g. as today:

```
scenes/quests/story_quests/template/0_template_intro/template_intro.tscn
```

But when duplicated for a quest, we can remove the `template_` component of the subdirectory and filenames:

```
scenes/quests/story_quests/magical_unicorn_paradise/0_intro/intro.tscn
```

@hydrolet can you confirm my memory of this discussion?

manuq: Sounds good, I'll wait for @hydrolet to confirm before changing it.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,130 @@
...
+	# EditorInterface.get_resource_filesystem().scan()
+
+	var template_intro_dialogue_uid := ResourceUID.id_to_text(
+		ResourceLoader.get_resource_uid(TEMPLATE_PATH + template_intro_dialogue_subpath)
+	)
+	var intro_dialogue_uid := ResourceUID.id_to_text(
+		ResourceLoader.get_resource_uid(storyquest_path + intro_dialogue_subpath)
+	)
+
+	prints(template_intro_dialogue_uid, intro_dialogue_uid)
```
manuq: My current blocker is that `intro_dialogue_uid` is giving `"uid://<invalid>"`. I think it is because the new dialogue has to be imported. I tried `Engine.get_meta("DMCache").mark_files_for_reimport(...)` (as I saw in DialogueManager source) and `EditorInterface.get_resource_filesystem().scan()` without much success. More digging tomorrow.

manuq: Sorted out!

addons/storyquest_bootstrap/plugin.gd
hydrolet: confirmed!

addons/storyquest_bootstrap/storyquest_dialogue.gd
```diff
@@ -0,0 +1,48 @@
...
+	_title = new_text
+	var errors: PackedStringArray
+	if _title.length() < min_title_length:
+		errors.append("The title lenght must be bigger than %d." % min_title_length)
+
+	_filename = _title.to_snake_case()
+	if FileAccess.file_exists(storyquests_path + _filename):
+		errors.append("The StoryQuest folder %s already exists." % storyquests_path + _filename)
+
+	prints(storyquests_path + _filename, FileAccess.file_exists(storyquests_path + _filename))
```
manuq: Current blocker: For some reason the check for the StoryQuest folder gives false here when the folder exists. But then the assert in `plugin.gd` fails.

wjt: I suggest using `storyquests_path.path_join(_filename)` rather than `+` (and removing the trailing `/` from the directory name). I think it's clearer than having to think each time `+` is used about whether the separator is part of the left-hand side.

I guess that the problem is that `file_exists()` is false for a directory, because it's not a file. https://docs.godotengine.org/en/stable/classes/class_diraccess.html#class-diraccess-method-dir-exists

wjt: Yep https://github.com/godotengine/godot/blob/4d1f26e1fd1fa46f2223fe0b6ac300744bf79b88/drivers/unix/file_access_unix.cpp#L313-L335

manuq: Thanks for the help! Strange because the same check in `plugin.gd` works: https://github.com/endlessm/threadbare/pull/909/files#diff-9061eab365cca8846dfb586144d5aee36d5d9673b2816d061030cc7567e83b48R64

My guess is that it works there because it's an `EditorPlugin` and has access to the project files.

I added the check in 2 places for safety, the UI is for displaying a user-facing warning. And right before creating the directory, it is asserted.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,165 @@
...
+	},
+	"outro":
+	{
+		"scene_path": "/4_template_outro/template_outro.tscn",
+		"components_path": "/4_template_outro/template_outro_components/",
+		"dialogue_path": "/4_template_outro/template_outro_components/template_outro.dialogue",
+	}
+}
+
+var _storyquest_dialogue: Window
```
wjt: I really hate to be a pedant but https://github.com/endlessm/threadbare/blob/main/docs/CONTRIBUTING.md#language documents that `dialogue` refers to in-game speech (matching the Godot Dialogue Manager addon) while `dialog` refers to dialog windows (matching Godot's own API like `popup_dialog`). https://en.wiktionary.org/wiki/dialog and https://en.wiktionary.org/wiki/dialogue both mention this distinction too.

manuq: Not pedant, that is great feedback, thanks!

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,165 @@
...
+			assert(error == OK)
+
+		# The following should be enough instead of a full scan, but it doesn't work:
+		# EditorInterface.get_resource_filesystem().update_file(storyquest_path + dialogue_path)
+		# EditorInterface.get_resource_filesystem().reimport_files(
+		# 	[storyquest_path + dialogue_path]
+		# )
+		EditorInterface.get_resource_filesystem().scan()
+		if EditorInterface.get_resource_filesystem().is_scanning():
+			await EditorInterface.get_resource_filesystem().resources_reimported
```
wjt: Is it guaranteed that this is emitted once for the whole set of files? I guess not but perhaps in practice it is enough.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,165 @@
...
+		var image_uid: String
+		if image_path:
+			image_uid = ResourceUID.id_to_text(ResourceLoader.get_resource_uid(image_path))
+
+		var scene_contents = FileAccess.get_file_as_string(
+			TEMPLATE_PATH + template_info[i]["scene_path"]
+		)
+
+		scene_contents = scene_contents.replace(template_scene_uid, storyquest_uids[i])
+		scene_contents = scene_contents.replace(template_dialogue_uid, dialogue_uid)
```
wjt: I'm interested that you're implementing this with search-and-replace on the raw `.tscn` file contents rather than manipulating the scene object and then saving it back – any particular reason?

addons/storyquest_bootstrap/storyquest_dialogue.gd
wjt: Oh, I have never seen `prints` before! I have wanted that function many times.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,167 @@
...
+			storyquest_path + directory.replacen("template_", "")
+		)
+		assert(error == OK)
+
+		var scene_path: String = template_info[i]["scene_path"].replacen("template_", "")
+		var dialogue_path: String = template_info[i]["dialogue_path"].replacen("template_", "")
+
+		var template_dialogue: DialogueResource = ResourceLoader.load(
+			TEMPLATE_PATH + template_info[i]["dialogue_path"]
+		)
```
wjt, line 101: That's surprising because presumably DialogueManager knows how to save its resources. I think it's probably fine to do it this way for now but it'd be interesting to understand later.

addons/storyquest_bootstrap/plugin.gd
manuq: From doing lots and lots of trial and error, this is what works. Following the documentation, after creating a new file with scripting or externally, it should be enough to call [`EditorInterface.get_resource_filesystem().update_file(file_path)`](https://docs.godotengine.org/en/4.4/classes/class_editorfilesystem.html#class-editorfilesystem-method-update-file) and then `EditorInterface.get_resource_filesystem().reimport_files([file_path])`.

The documentation of `update_file()` says:
> This will not import the file. To reimport, call [reimport_files()](https://docs.godotengine.org/en/4.4/classes/class_editorfilesystem.html#class-editorfilesystem-method-reimport-files) or [scan()](https://docs.godotengine.org/en/4.4/classes/class_editorfilesystem.html#class-editorfilesystem-method-scan) methods.

The documentation of `reimport_files()` say:
> If the file type changed or the file was newly created, use [update_file()](https://docs.godotengine.org/en/4.4/classes/class_editorfilesystem.html#class-editorfilesystem-method-update-file) or [scan()](https://docs.godotengine.org/en/4.4/classes/class_editorfilesystem.html#class-editorfilesystem-method-scan).

Using DialogueManager as reference I've seen both `reimport_files()` and `scan()` used. So it is confusing.

addons/storyquest_bootstrap/storyquest_dialogue.gd
manuq: For reference you were right @wjt , it was a matter of using `DirAccess.dir_exists_absolute`. It was another assert that was failing.

addons/storyquest_bootstrap/plugin.gd
manuq: This is what worked in first place for external resources. I still need to try using load() and manipulating the scene as an object, which would be my preference too.

addons/storyquest_bootstrap/plugin.gd
manuq: Yes, I'll give it another chance before removing the draft mark.

manuq:
This is basically working:

[Grabación de pantalla desde 2025-07-08 12-27-10.webm](https://github.com/user-attachments/assets/38437688-d86a-4200-bcc8-859dc6f257e4)

The remaining work is:
- Check if we can edit the copied scenes as objects, rather than as raw text.
- Check if importing the new scenes to the filesystem (to obtain their UIDs) can be improved.

manuq:
Hmm I need to see why the licensing information check is failing. The output doesn't tell anything: https://github.com/endlessm/threadbare/actions/runs/16147551077/job/45570681742?pr=909

manuq:
> Hmm I need to see why the licensing information check is failing. The output doesn't tell anything: https://github.com/endlessm/threadbare/actions/runs/16147551077/job/45570681742?pr=909

Sorted out. It was the `tscn` file inside the new addon.

addons/storyquest_bootstrap/plugin.gd
manuq: Neat, I was able to treat resources as such, not as text files in the WIP appended commit. It still needs some work but is working. I blame myself for having used what the Dialogue Manager does in the tool "Create copy of dialogue example baloon" as reference in the first place.

manuq: I now updated the appended commit and it sets the outro as next scene of the intro. Although I need to rebase because it is changing the previous commits.

manuq:
Rebased and ready for review.

> * Check if we can edit the copied scenes as objects, rather than as raw text.

Done.

> * Check if importing the new scenes to the filesystem (to obtain their UIDs) can be improved.

I dedicated some time to this but couldn't find a better way. I commented in the script what I think should work considering the documentation of `EditorFileSystem` above what works for me in practice. I may open a discussion upstream.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,174 @@
...
+const NEW_STORYQUEST_DIALOG = preload(
+	"res://addons/storyquest_bootstrap/new_storyquest_dialog.tscn"
+)
+
+const TOOL_MENU_LABEL := "Create StoryQuest from template..."
+
+const STORYQUESTS_PATH := "res://scenes/quests/story_quests/"
+# Using + instead of String.path_join() here because it errors with:
+# Assigned value for constant "TEMPLATE_PATH" isn't a constant expression.
+const TEMPLATE_PATH := STORYQUESTS_PATH + "template/"
```
wjt, line 15: Ouch!

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,174 @@
...
+
+func _close_dialog() -> void:
+	_new_storyquest_dialog.queue_free()
+	_new_storyquest_dialog = null
+
+
+func validate_title(title: String) -> PackedStringArray:
+	var errors: PackedStringArray
+	if title.length() < MIN_TITLE_LENGTH:
+		errors.append("⚠ The title length must be bigger than %d." % MIN_TITLE_LENGTH)
```
wjt:
```suggestion
		errors.append("⚠ The title must be at least %d letters long." % MIN_TITLE_LENGTH)
```

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,174 @@
...
+	return errors
+
+
+func validate_filename(filename: String) -> PackedStringArray:
+	var errors: PackedStringArray
+	if DirAccess.dir_exists_absolute(STORYQUESTS_PATH.path_join(filename)):
+		errors.append(
+			"⚠ The StoryQuest folder %s already exists." % STORYQUESTS_PATH.path_join(filename)
+		)
+	return errors
```
wjt: I think we also want to forbid leading `.` characters and `/` characters – or perhaps strip them out when converting the quest name to a directory name.

If you call your quest `../../../../../../../../hello` you can create files outside the project directory.

If you call your quest `a//////a//a/////////` then the `make_dir_absolute` call fails.

wjt: Oh, so does `foo/bar` because intermediate directories are not created.

`String.validate_filename()` strips out characters that are illegal in filenames on some platforms (i.e. Windows) and also strips out slashes. Maybe that would do the job. But actually I would be tempted to go further and replace all characters outside `[a-z0-9_]` with underscores.

wjt: I pushed a patch to address this.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -0,0 +1,176 @@
...
+	assert(error == OK)
+
+	var intro_scene_path: String
+
+	for i in template_info:
+		var directory: String = template_info[i]["components_path"]
+		error = DirAccess.make_dir_recursive_absolute(
+			storyquest_path.path_join(directory.replacen("template_", ""))
+		)
+		assert(error == OK)
```
wjt, line 95: These bare asserts make me a bit sad. I tried using an LLM to rewrite them all, but its output plus `gdformat` was horrible:

```patch
        for i in template_info:
                var directory: String = template_info[i]["components_path"]
-               error = DirAccess.make_dir_recursive_absolute(
-                       storyquest_path.path_join(directory.replacen("template_", ""))
+               var components_dir_path := storyquest_path.path_join(directory.replacen("template_", ""))
+               error = DirAccess.make_dir_recursive_absolute(components_dir_path)
+               assert(
+                       error == OK,
+                       (
+                               "Failed to create components directory %s: %s"
+                               % [components_dir_path, error_string(error)]
+                       )
                )
-               assert(error == OK)
```

wjt: I might rewrite them by hand at some point.

addons/storyquest_bootstrap/plugin.gd
manuq: > I think we also want to forbid leading `.` characters and `/` characters – or perhaps strip them out when converting the quest name to a directory name.
> 
> If you call your quest `../../../../../../../../hello` you can create files outside the project directory.

Ouch, I haven't thought about it. Not very safe.

> If you call your quest `a//////a//a/////////` then the `make_dir_absolute` call fails.

:+1: good catch.

addons/storyquest_bootstrap/plugin.gd
manuq: Maybe instead of assertions it can be a conditional that logs the error and cancel the execution if any ot these errors happen?

wjt (commented):
If you're OK with my patch (and in particular the mapping of e.g. `ñ` in title to `_` in filename) then I think this is good to go

PR #912, opened 2025-06-30
Fix attack and shot

AngheloMP10:


scenes/quests/story_quests/inka/inka_player_components/inka_player/player.tscn3113862180.tmp
```diff
@@ -0,0 +1,176 @@
+[gd_scene load_steps=25 format=3 uid="uid://b8opbag1j4gpj"]
```
wjt, line 1: This and the other few files ending `.tmp` are temporary files which should not be committed (and can be deleted).

PR #913, opened 2025-07-01
Consolidate theme resources

wjt:
For historical reasons, we have ended up with two theme resources. One, in shared_components, is used by HUD input hints and the "Talk"/interact label. The other, which is under the "dialogue" scene because it started out as being used only for the dialogue balloons, is used by menus and other panels in the game such as the credits screen.

Merge the two together. Take the smaller number of styles from shared_components/theme.tres, and add them to
dialogue/components/theme.tres. Adjust all scenes that use the former to use the latter. Then rename the latter over the former.

Fixes https://github.com/endlessm/threadbare/issues/381


scenes/ui_elements/shared_components/theme.tres
```diff
@@ -1,23 +1,196 @@
...
+FlatButton/styles/hover = SubResource("StyleBoxTexture_q6hjb")
+FlatButton/styles/hover_pressed = SubResource("StyleBoxTexture_3vcfk")
+FlatButton/styles/normal = SubResource("StyleBoxTexture_bbim3")
+FlatButton/styles/pressed = SubResource("StyleBoxTexture_3vcfk")
+FlatNextButton/base_type = &"FlatButton"
+FlatNextButton/styles/normal = SubResource("StyleBoxTexture_qndbo")
+GridContainer/constants/h_separation = 32
+GridContainer/constants/v_separation = 32
 HintLabel/base_type = &"Label"
+HintLabel/colors/font_color = Color(1, 1, 1, 1)
```
wjt, line 178: This is the only change that isn't just "merge the two files". `HintLabel` inherits from `Label`, which overrides `font_color`.

manuq (approved):
I tested the branch and didn't notice any regressions. Great!

PR #933, opened 2025-07-09
Grappling hook prototype

manuq:


manuq:
First little game mechanic: use it to cross non-walkable tiles.

[Grabación de pantalla desde 2025-07-11 17-45-02.webm](https://github.com/user-attachments/assets/c201af22-6286-4adf-8430-746603334080)

manuq:
> I found it a bit hard to aim when the pins weren't in a straight line. I wonder whether making the mouse input relative rather than absolute might help. I kept trying to point at each pin but actually only the angle matters.

You mean this problem?

hooking-directions

There is a difference between the mouse angle and the actual propulsion angle. First is between the player character and the mouse, used for testing collision between a raycast and an area. The second is between the player character and an anchor point that things that can be hooked have.

> We did some experimentation and on my laptop I can't move the mouse using the touchpad while holding down W A S D, presumably due to palm rejection (preventing accidental mouse input while typing). If I use the arrow keys the touchpad works - but I can't do that because I'm right-handed and I can't control the touchpad accurately with my left hand.
> 
> We expect many people playing/contributing to Threadbare to only have access to a laptop with touchpad (rather than having an external mouse or gamepad) and we want the game to be playable for them so it may be worth exploring directing the thread based on the last direction the player moved. Levels will need to be designed accordingly to not require grappling at angles that aren't one of the 8 directions.

This is my most important take from your playtest. Our target audience will probably have laptops without an external mouse. So next I'll focus on making this work with keyboard. Thanks!

manuq:
Finally ready for review!

manuq:
> I got stuck here:
> 
> image
> The thread is going in both directions between the needles; but I can't figure out which direction to walk in to cancel the thread, and I can't seem to throw it anywhere further.

Great feedback!

manuq:
> When I use a joypad I tend to use the analogue stick rather than the D-pad. It's not currently possible to aim with the analogue stick though. I didn't look too closely at why not but it's hitting this branch:
> 
> ```gdscript
> 	if (
> 		_event.is_action_released(&"ui_left")
> 		or _event.is_action_released(&"ui_right")
> 		or _event.is_action_released(&"ui_up")
> 		or _event.is_action_released(&"ui_down")
> 	):
> 		return
> ```
> 
> I fear we may need to special-case each of keyboard/dpad, analogue stick, mouse.

Yeah without this, as the comment in the code says:

```
        # Otherwise, if for example left and down are pressed to aim in diagonal and both are released,
        # there is always one that is released first so the aim direction ends up being either left or
        # down, not left AND down.
```

I wonder if there is a better way to do it, instead of special-casing the keyboard/dpad. Maybe with a timer to check if the two input keys were immediately released?

manuq:
> but I think this would still feel fine if the action only repeated if there is a target, so in this case I would throw the thread 4 times and then it would stop until I release and press space, at which point it would be thrown once, unsuccessfully, until I release and press it again, etc.

Great feedback, I will try to make it work like this!

manuq:
> This is **really** satisfying!

Thanks, on the level design I made it optional at this point, but is mandatory in the next sublevel (which I think you didn't reach because of getting stuck in https://github.com/endlessm/threadbare/pull/933#issuecomment-3285044924).

manuq:
> > but I think this would still feel fine if the action only repeated if there is a target, so in this case I would throw the thread 4 times and then it would stop until I release and press space, at which point it would be thrown once, unsuccessfully, until I release and press it again, etc.
> 
> Great feedback, I will try to make it work like this!

Addressed in a fixup commit https://github.com/endlessm/threadbare/pull/933/commits/f47ebab4e0dd04c7a81738ba5036a6c0eb2ce1cd

I also:

- Removed a script that was a leftover https://github.com/endlessm/threadbare/pull/933/commits/7d92b7ac2900c3a4003f86e9579e5f7dc1f5db0d
- Fixed Y sort of some elements in the needles sublevel https://github.com/endlessm/threadbare/pull/933/commits/e047e3069d08ab4a5d20549f2d5a83813700080d
- Fixed the bug described in https://github.com/endlessm/threadbare/pull/933#issuecomment-3283011821 in fixuep https://github.com/endlessm/threadbare/pull/933/commits/c7000a4165dbdc2868c180e4bc1d397ffc2aa352. But I just realized that it changed the behavior, now the pull is cancelled earlier so the "powerup" sublevel can't be beated anymore. More tomorrow!

scenes/game_elements/characters/player/components/player_hook.gd
```diff
@@ -0,0 +1,314 @@
...
+## [br][br]
+## This is a piece of the grappling hook mechanic.
+## [br][br]
+## It should be in group [b]hook_listener[/b] so functions [method hooked],
+## [method hit_wall] and [method hit_air] are called.
+
+## Emitted when the string is thrown from the primary control.
+signal string_thrown
+
+const NON_WALKABLE_FLOOR_LAYER: int = 10
```
wjt, line 20: This can be a follow-up but we should put all the layers into an enum in `scenes/globals/enums.gd`.

scenes/quests/lore_quests/quest_002/2_grappling_hook/components/switch_mode_area.gd
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends Area2D
+## An area inside which the player is set to mode [member Player.Mode.COZY],
+## and outside which is set to mode [member Player.Mode.HOOKING].
+##
+## This is a hack for collecting a thread in a grappling hook scene,
+## until we get rid of player modes.
```
wjt, line 8: 👍🏻

scenes/game_elements/props/hook_control/components/hook_control.gd
```diff
@@ -0,0 +1,206 @@
...
+			var is_loop := false
+			if hooked_to.hook_control:
+				is_loop = is_instance_valid(hooked_to.hook_control.hooked_to)
+				if not is_loop:
+					hooked_to.hook_control.state = State.AIMING
+					hooked_to.hook_control.initial_hook_angle = _get_hook_angle()
+					if hook_area:
+						hooked_to.hook_control.exclude_areas.append(hook_area)
+				state = State.DISABLED
+			get_tree().call_group(&"hook_listener", &"hooked", hooked_to, is_loop)
```
wjt, line 145: I wouldn't have thought to use `call_group` here. I understand why some kind of broadcast is needed ­– the player logic needs to react not only to its own hookcontrol but the hookcontrol on needles.

I think I would have done this the other way round – have `player_hook.gd` discover all the nodes in a `hook_control` group and connect to `hooked`, `hit_wall`, `hit_air`, `released` signals on each one. I think this is fine though.

Edit: Oh! And now as I read further in, I realise why you've done it this way: it's for the `all_hooked.gd` component.

manuq:
> I lied, one more bit of playtesting feedback (for later): I wonder if the camera should follow the tip of the thread. For instance:
> 
> image

Yeah I thought about it, I tried to design the levels to prevent its implementation. But I think that it will be very welcome if we do longer puzzles. One thought I had is, maybe the player should still be on frame, because you can still walk and must walk in the opposite direction of the first thing connected to cancel. Although now after a fixup I did, throwing to the air also cancels the thread. I also thought that in the future we may add enemies that could harm the player, it would be unfair to let that happen off camera. But as things are now, I think the camera should transition to center the active control (the thing with the red arrow from which you can aim). Thanks for the feedback.

scenes/game_elements/props/button_item/components/button_item.gd
```diff
@@ -0,0 +1,37 @@
...
+@onready var _up_down_animation: AnimationPlayer = %UpDownAnimation
+
+
+func _ready() -> void:
+	# Delay the up and down animation according to the global position,
+	# so multiple buttons in a row form a wave:
+	_up_down_animation.stop()
+	var t: float = (sin(global_position.x) + 1) * 0.5 + (cos(global_position.y) + 1) * 0.5
+	await get_tree().create_timer(t).timeout
+	_up_down_animation.play("up-down")
```
wjt: Great idea for how to make this work naturally! I wonder if there is a variation of `AnimationPlayer.play()` that accepts a time offset to start the animation from, rather than using a timer

wjt: I can't see a way to do it so I suggest you leave this as-is!

scenes/game_elements/characters/player/components/player_hook.gd
```diff
@@ -0,0 +1,314 @@
...
+
+
+func _process_pulling(_delta: float) -> void:
+	# Return if the thing that was being pulled was removed:
+	var ending_area := get_ending_area()
+	if not is_instance_valid(ending_area):
+		stop_pulling()
+		return
+
+	var target := ending_area.owner
```
wjt: I was trying to figure out whether it was possible to use a weight of `0.5` to have the hooked thing move towards the player and the player move towards the hooked thing.

I got caught out here because I assembled it directly in the scene rather than in a scene of its own:

Image

and so `ending_area.owner` is the root node of the `grappling_hook_powerup.tscn` scene, not the CharacterBody2D it is a child of.

I wonder if we should change this to an explicit property on the `hookable_area.gd` that gets set by default to its parent or something?

wjt (approved):
I think this is looking pretty good! I confess that I didn't review every last line of code but I spent some time trying to make some new hookable entities of my own and to figure out how the logic flows.

I have a few notes but we could address them as follow-ups.

scenes/game_elements/props/hook_control/components/hook_control.gd
manuq: Let me try the other way round, because AllHooked also has the array of HookableAreas and each of these areas may  have a HookControl.

scenes/game_elements/props/button_item/components/button_item.gd
manuq: You are right, I should `seek()` instead of adding a waiting https://docs.godotengine.org/en/4.4/classes/class_animationplayer.html#class-animationplayer-method-seek

scenes/game_elements/characters/player/components/player_hook.gd
manuq: Great idea, addressed in an appended commit.

[Grabación de pantalla desde 2025-09-17 15-39-01.webm](https://github.com/user-attachments/assets/9960dff0-c84b-49a6-b176-2066067fa185)

scenes/game_elements/props/button_item/components/button_item.gd
manuq: Addressed and amended the commit. It was much simpler, thank you!

manuq:
I addressed a few more things (huge thanks to @wjt). I will merge and open follow up tickets for the rest.

PR #940, opened 2025-07-14
Add parallax effect for void starfield

### LINKED ISSUE #939: Replace void tile variants with shader
We currently have 5 variants of the void tileset, with different amount of stars:

https://media.githubusercontent.com/media/endlessm/threadbare/refs/heads/main/assets/first_party/tiles/Void_Tiles_Empty.png
https://media.githubusercontent.com/media/endlessm/threadbare/refs/heads/main/assets/first_party/tiles/Void_Tiles_Stars_Level1.png
etc.

and a terrain that combines them all with some level of randomness.

I would like to replace this with a shader that draws the stars dynamically, so that we have the stars fixed in place even as the camera moves (just like the stars in the sky don't appear to move if you look up at the sky and walk, because they are so far away), and make the appearance less repetitive, perhaps with some animation.

wjt:
We currently have a “void” tileset, which is used only in the closing
transition of the lorequest and as scenery in stella_stealth. Actually,
we have 5 tilesets with stars at different brightnesses, and a Void
terrain which combines them.

Conceptually, the stars are far away, so it would be nice to have a
parallax effect where the stars do not move even as the camera moves.

Implement this with a new variant of the void tiles plus a shader
applied to the TileMapLayer. The new tiles keep the existing stitched
edges, but then fill the rest of the space with pure red.

The shader uses red as its chroma key, replacing it with stars based on
a noise texture. The position sampled in the noise texture is not the
original UV for the fragment shader, but is instead based on the
viewport position (calculated in the vertex() stage and passed to the
fragment() stage).

A subtle flickering effect is added, using a sine function of the TIME
built-in shader variable (seconds since the engine started).

The new tiles have the same collision shape as the other 5 voids,
namely fully-opaque non_walkable_floor.

This implementation combines ideas from a variety of sources:

- The idea to use a chroma-key to identify the region to fill with stars
  is due to @manuq. (My original idea was to make the void
  transparent, but this only works if the void is below all other layers.)

- The noise-texture starfield is inspired by a video by FencerDevLog
  <https://www.youtube.com/watch?v=EQv5kzqpYOk>, particularly the use of
  pow(r, 20.0) to compress most values in the noise texture towards
  zero.

- https://github.com/Ymanawat/Chroma-Key-Shader-Godot-4 was an
  interesting reference, though since this is a pixel-perfect texture
  there was no need to use the threshold + blending technique there.

- “Repeated texture overlay for tilemaps” by jess.codes
  <https://godotshaders.com/shader/repeated-texture-overlay-for-tilemaps/>
  showed me how to calculate a position in vertex() and pass it through
  to fragment(), though the calculation here is different because we
  want viewport position not world position. (As I write this I realise
  I could use the same mix() technique and avoid branching in the
  shader.)

I updated the 2 uses of the void that are in the game at present, but left the old void in case there are branches out there in the world that use it.

Fixes #939


scenes/game_elements/props/void/README.md
```diff
@@ -0,0 +1,13 @@
+This is a CanvasItem material that is intended to be applied to a TileMapLayer
+that is using the VoidChromakey terrain (or the constituent tiles). It replaces
+the red key color with a procedurally-generated starfield that does not move as
+the camera moves, implying that the stars are at a great distance from the
+camera. The stars flicker slightly at a low rate.
+
+You must set the TileMapLayer's texture filter to “nearest” to avoid an ugly
+fringe where the key color meets the tile border.
```
manuq, line 8: Good point on adding a README with the explanation, since there is no class or other place to do it :+1:

scenes/quests/lore_quests/quest_001/4_closing_transition/components/story_quest_closing_transition.dialogue
```diff
@@ -7,4 +7,5 @@ StoryWeaver: I've done it! I've recovered the elements of the story.
 StoryWeaver: Now it's time to head back to Fray's End with the Memory, Imagination, and Spirit.
 StoryWeaver: Reweaving this story back into the Loom will help mend the holes in this world.
 do animation_player.play("walk_off_scene")
+do wait(1.0)
```
manuq, line 10: :+1:

manuq (approved):
This is awesome, I know I said it multiple times already. I like how you credited with fantastic detail, I must try hard to do the same. The tilting of the stars using a sine function is subtle and makes the effect even better!

scenes/game_elements/props/void/components/void_chromakey.gdshader
```diff
@@ -0,0 +1,72 @@
...
+ * SPDX-FileCopyrightText: The Threadbare Authors
+ * SPDX-License-Identifier: MPL-2.0
+ */
+shader_type canvas_item;
+
+// TODO: Can we get this from SCREEN_PIXEL_SIZE?
+/**
+ * Viewport size.
+ */
+uniform vec2 resolution = vec2(1920.0, 1080.0);
```
manuq, line 12: The way I did it in the past was by passing it through a gdscript. Not ideal.

PR #942, opened 2025-07-15
Show project version on main menu

### LINKED ISSUE #810: Set game version & show in-game
We currently don't set a version in `project.godot`. During export we should set it to the output of `git describe --tags` (`--tags` is necessary because we have so far only created unannotated tags), and then display it in-game somewhere.

wjt:
Fixes #810


wjt:
Hm, this doesn't work in GitHub Actions because it doesn't clone the full history of the branch. Let's see.

wjt:
image

For a release this would instead look like `v0.1.3`.

manuq (approved):
Awesome, nothing to note. I trust you in the election of the third party addon.

PR #943, opened 2025-07-15
Use consistent color for InteractArea CollisionShape2Ds

wjt:
Previously, these were a mix of yellowish colors, cyan for checkpoints, and the default color.

Consistently use the same shade of yellow for all interact areas, as well as the player's interact zone. This makes it a bit easier to distinguish the interaction area from the collision area that prevents walking through objects.


manuq (approved):
Good call, +1 for consistency!

PR #952, opened 2025-07-17
Support modulating trees

wjt:
On entry to `fragment()`, `COLOR = modulation * texture(TEXTURE, UV)`,
roughly. But the tree shader re-samples `TEXTURE` in `fragment()`,
overwriting `COLOR`. This means that the modulation is lost. In particular
this means you can't make a semi-transparent tree with the modulate
alpha channel.

In vertex(), `COLOR` is just the modulation. Save this in a varying, and
reapply it in `fragment()`.

### DIFF
--- a/scenes/game_elements/props/tree/components/tree.gdshader
+++ b/scenes/game_elements/props/tree/components/tree.gdshader
@@ -12,8 +12,8 @@ uniform float animation_frames : hint_range(1.0, 10.0, 1.0) = 6.0;
 varying vec2 world_pos;
 
 void vertex() {
-   // Transform it into world coordinates
-   world_pos = (MODEL_MATRIX * vec4(1.0, 1.0, 0.0, 1.0)).xy;
+	// Transform it into world coordinates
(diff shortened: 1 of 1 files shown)


scenes/game_elements/props/tree/components/tree.gdshader
```diff
@@ -10,10 +10,13 @@ uniform float max_wind_intensity : hint_range(0.0, 10.0, 0.1) = 2.0;
...
 varying vec2 world_pos;
+varying vec4 modulate;
 
 void vertex() {
-   // Transform it into world coordinates
-   world_pos = (MODEL_MATRIX * vec4(1.0, 1.0, 0.0, 1.0)).xy;
+	// Transform it into world coordinates
+	world_pos = (MODEL_MATRIX * vec4(1.0, 1.0, 0.0, 1.0)).xy;
+
+	modulate = COLOR;
```
manuq, line 19: Really? Nice!

manuq (approved):

PR #954, opened 2025-07-18
Consistently indent shaders with tabs

wjt:
This is the style recommended by
https://docs.godotengine.org/en/stable/tutorials/shaders/shaders_style_guide.html
and is the default for the built-in editor.

I also removed an unused shader.


manuq (approved):
Ah I understand #955 now.

Issue #944, opened 2025-07-15
Web share should have a social card


cassidyjames:
When sharing the playable URL on the web, there's no social/preview card data. You can test this by sharing to e.g. Mastodon, Twitter, or Facebook or by using the Share Preview app on Flathub/Endless OS.

<img width="646" height="684" alt="Screenshot" src="https://github.com/user-attachments/assets/6a3f4f25-acf4-4324-938c-db3ac8489425" />

We should provide social card info in the web build so we get something a bit nicer. :)

wjt:
The two ways to accomplish this are:

1. The heavy-handed approach: replace the entire HTML "shell" with a custom version https://docs.godotengine.org/en/4.4/tutorials/platform/web/customizing_html5_shell.html
2. The lighter touch: if we can put everything that is needed into the `<head>` tag, then we can use the "Head Include" field in the web export.

<img width="1246" height="763" alt="Image" src="https://github.com/user-attachments/assets/9280b06f-e9b3-4dbb-8057-7c81b7e7ebe6" />

I think it's the latter: we need a host of `<meta>` tags in the `<head>`?

cassidyjames:
Yep, it's all in the `<head>`. Neat, I didn't realize you could do that on the export side; that approach makes the most sense to me.

We'll need to provide a:
- title
- description
- fully-qualified URL to an image

And then we should make sure we provide:
- Meta title & description
- Twitter tags (title, image, card, description)
- OpenGraph (title, description, image)

It looks like the social preview image set in the repo settings isn't accessible at a known URL like avatars are (e.g. https://github.com/endlessm.png). It's accessible via the GraphQL API, but that seems… excessive. We could upload a copy of the image to the `docs/` folder and link to its raw URL, like we do for the screenshot (I like this option).

wjt:
> Twitter: Image "https://media.githubusercontent.com/media/endlessm/threadbare/refs/heads/main/icon.png" did not meet the requirements: Image is too tiny (128×128px), minimum dimensions are 144×144px.

Sigh. Echoes of #369. Can we tackle all these together:

- Make a list of all icon sizes we need for Flathub, Steam, Itch.io, and share tags
- Make pixel-perfect icons for each :)

Anyway here's what I was testing. I'll stop here:

```
 export_presets.cfg | 10 +++++++++-
 1 file changed, 9 insertions(+), 1 deletion(-)

diff --git a/export_presets.cfg b/export_presets.cfg
[...]

PR #958, opened 2025-07-21
HTML export: Add social card metadata

wjt:
The image URL corresponds to the repository image for https://github.com/endlessm/threadbare. It will need to be manually updated when that image changes (perhaps by storing the image in the repo and adding a CI check that the export file is updated if the image has changed since the referenced commit) but this will do for now.

Fixes https://github.com/endlessm/threadbare/issues/944

### DIFF
--- a/export_presets.cfg
+++ b/export_presets.cfg
@@ -28,7 +28,20 @@ vram_texture_compression/for_desktop=true
 vram_texture_compression/for_mobile=false
 html/export_icon=true
 html/custom_html_shell=""
-html/head_include=""
+html/head_include="<meta property=\"og:title\" content=\"Threadbare\" />
+<meta property=\"og:type\" content=\"website\" />
+<meta property=\"og:description\" content=\"Rebuild an unraveling world\" />
+<meta property=\"og:image\" content=\"https://repository-images.githubusercontent.com/941971441/4633bb09-4b7e-477c-a83b-369b29f23c39\" />
+<meta property=\"og:image:type\" content=\"image/png\" />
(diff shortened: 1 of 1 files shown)


export_presets.cfg
```diff
@@ -28,7 +28,20 @@ vram_texture_compression/for_desktop=true
 vram_texture_compression/for_mobile=false
 html/export_icon=true
 html/custom_html_shell=""
-html/head_include=""
+html/head_include="<meta property=\"og:title\" content=\"Threadbare\" />
+<meta property=\"og:type\" content=\"website\" />
+<meta property=\"og:description\" content=\"Rebuild an unraveling world\" />
+<meta property=\"og:image\" content=\"https://repository-images.githubusercontent.com/941971441/4633bb09-4b7e-477c-a83b-369b29f23c39\" />
```
manuq, line 34: I checked if the image can be embeded in the HTML directly using base64 data, but a quick search tells me that is not going to work for the previews.

export_presets.cfg
```diff
@@ -28,7 +28,20 @@ vram_texture_compression/for_desktop=true
...
+<meta property=\"og:image:type\" content=\"image/png\" />
+<meta property=\"og:image:width\" content=\"1280\" />
+<meta property=\"og:image:height\" content=\"640\" />
+<meta property=\"og:image:alt\" content=\"A cartoon character carrying a sewing needle & button and wearing a thimble for a hat looks towards an elderly cartoon man with a long white beard, wooden staff, and book.\" />
```
manuq, line 38: :+1:

Issue #953, opened 2025-07-17
Player: Add optional animation for running


manuq:
Fallback to the existing behavior, which is to play the walk animation and double the speed.

This enhacement comes from https://github.com/UTPGameLab4-0-Grupo6-Equipo1/threadbare where the learners changed the player script to play a different animation when running. A recent change suggests that it should be easy to add a fallback.

PR #961, opened 2025-07-21
Player: Add optional run animation

manuq:
If the SpriteFrames resource has an animation named "run", play a matching "run" animation in the AnimationPlayer. It assumes that there are 6 frames being played in sequence (like in the walk animation) and it has an audio track for the running steps, playing the same audio FX than the walk animation, but more often (at frames 0, 2 and 4).

If there is no "run" animation, the behavior stays the same as before: play the walk animation at double the speed.

Note: As it happened in commit 48e35f24 the walk animation subresource is moved again, so the diff is bigger than it should.

Fixes https://github.com/endlessm/threadbare/issues/953

### DIFF
--- a/scenes/game_elements/characters/player/components/animation_player.gd
+++ b/scenes/game_elements/characters/player/components/animation_player.gd
@@ -34,6 +34,8 @@ func _get_repel_animation() -> StringName:
 func _process_walk_idle(_delta: float) -> void:
 	if player.velocity.is_zero_approx():
 		play(&"idle")
+	elif player_sprite.sprite_frames.has_animation(&"run") and player.is_running():
+		play(&"run")
 	else:
 		play(&"walk")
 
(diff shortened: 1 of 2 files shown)


manuq:
Confirmed that this works for Sueños nocturnos in my rebase here: https://github.com/UTPGameLab4-0-Grupo6-Equipo1/threadbare/pull/87 . Although the branch name is "nocturnos-rebase-main" I actually ended up rebasing with this branch which contains the optional run animation. So ready for review.

wjt:
> Note: As it happened in commit 48e35f2 the walk animation subresource is moved again, so the diff is bigger than it should.

It's annoying when this happens... I've been known to manually reorder the resource file to make the diff smaller but that's probably taking my commitment to small diffs too far. I wish Godot was better at this.

wjt (approved):
Ideally we would also:

1. Have at least one example of the `run` animation in the tree - but this will be solved in the short term when “Sueños nocturnos” is merged, and perhaps in the longer term by adding a run animation to StoryWeaver?
2. Warn if `run` is provided with the wrong number of frames - I'll make a follow-up PR for this.

Issue #950, opened 2025-07-16
Add talk behavior


manuq:
Start implementing the proposal in https://github.com/endlessm/threadbare/discussions/774

The problem of using inheritance like we currently do is that it is inflexible. For example, the current checkpoint does trigger a dialogue (ie. talks). But since it doesn't inherit from the Talker class, the logic has been duplicated. The same happens with the Loom.

Having a TalkBehavior node that can be added to any game entity, would simplify things and make creating NPCs easier.

PR #965, opened 2025-07-22
Add Talk Behavior node

manuq:
This node wraps few logic, but it is a reusable component that lets developers easily trigger dialogues when the player interacts with an area, with only setup and no scripting. Hopefully it will help learners create their own talkers.

Fixes https://github.com/endlessm/threadbare/issues/950

### DIFF
--- a/scenes/game_elements/characters/npcs/elder/components/elder.gd
+++ b/scenes/game_elements/characters/npcs/elder/components/elder.gd
@@ -2,9 +2,11 @@
 # SPDX-License-Identifier: MPL-2.0
 @tool
 class_name Elder
-extends Talker
+extends NPC
 
 const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
+
+@export var dialogue: DialogueResource = preload("uid://cc3paugq4mma4")
 
 ## Directory of quests that this NPC offers to the player during interactions.
 @export_dir var quest_directory: String
@@ -27,12 +29,19 @@ var chosen_quest: Quest
 var _dialogue_balloon: CanvasLayer
 var _storybook: Storybook
 
+@onready var interact_area: InteractArea = %InteractArea
+@onready var talk_behavior: TalkBehavior = %TalkBehavior
 @onready var _book_sound: AudioStreamPlayer2D = %BookSound
 @onready var _storybook_layer: CanvasLayer = %StorybookLayer
 
 
 func _ready() -> void:
 	super._ready()
+	if Engine.is_editor_hint():
+		return
+	talk_behavior.dialogue = dialogue
+	talk_behavior.before_dialogue = _before_dialogue
+	interact_area.interaction_ended.connect(_on_interaction_ended)
 	animated_sprite_2d.connect("frame_changed", _on_frame_changed)
 
 	if quest_directory:
@@ -52,16 +61,6 @@ func _get_configuration_warnings() -> PackedStringArray:
 	return warnings
 
 
-# Override this talker method so we can vary the dialogue title based on
-# whether the loom offering is possible.
-func _on_interaction_started(player: Player, _from_right: bool) -> void:
-	var title: String = ""
-	if eternal_loom and eternal_loom.is_item_offering_possible():
-		title = "go_to_loom"
-	DialogueManager.dialogue_ended.connect(_on_dialogue_ended, CONNECT_ONE_SHOT)
-	_dialogue_balloon = DialogueManager.show_dialogue_balloon(dialogue, title, [self, player])
-
-
 ## Show a storybook to the player, and wait for them to select a story or close the book.
 func show_storybook() -> void:
 	if not _storybook:
@@ -78,11 +77,14 @@ func show_storybook() -> void:
 	_storybook_layer.remove_child(_storybook)
 
 
-func _on_dialogue_ended(dialogue_resource: DialogueResource) -> void:
-	super(dialogue_resource)
+func _before_dialogue() -> void:
+	if eternal_loom and eternal_loom.is_item_offering_possible():
+		talk_behavior.title = "go_to_loom"
 
+
+func _on_interaction_ended() -> void:
 	if chosen_quest:
-		%InteractArea.disabled = true
+		interact_area.disabled = true
 		GameState.start_quest(chosen_quest)
 		SceneSwitcher.change_to_file_with_transition(
 			chosen_quest.first_scene, ^"", Transition.Effect.FADE, Transition.Effect.FADE
--- a/scenes/game_elements/characters/npcs/talker/components/talker.gd
+++ b/scenes/game_elements/characters/npcs/talker/components/talker.gd
@@ -6,7 +6,6 @@ extends NPC
 
 const DEFAULT_DIALOGUE: DialogueResource = preload("uid://cc3paugq4mma4")
 
-@export var npc_name: String
 @export var dialogue: DialogueResource = DEFAULT_DIALOGUE
 
 var _previous_look_at_side: Enums.LookAtSide = Enums.LookAtSide.UNSPECIFIED
@@ -19,8 +18,6 @@ func _ready() -> void:
 	if Engine.is_editor_hint():
 		return
 	interact_area.interaction_started.connect(_on_interaction_started)
-	if npc_name:
-		interact_area.action = "Talk to %s" % npc_name
 
 
 func _on_interaction_started(player: Player, from_right: bool) -> void:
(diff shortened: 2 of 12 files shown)


scenes/game_logic/talk_behavior.gd
```diff
@@ -0,0 +1,49 @@
...
+## Display a dialogue when interacted.
+##
+## It displays the [member dialogue] passing the parent node and the player as the extra game
+## state.
+## If [member title] is set, it will display the dialogue at that title.[br][br]
+## When the dialogue ends, it finishes the interaction by calling [method
+## InteractArea.end_interaction].[br][br]
+## Optionally, an awaitable function can be passed to [member before_dialogue] if
+## something needs to happen before the [member dialogue] is displayed. This can be used,
+## for example, to dinamically set the [member title] of the dialogue considering the
```
wjt: dynamically

scenes/game_logic/talk_behavior.gd
```diff
@@ -0,0 +1,49 @@
...
+
+
+func _ready() -> void:
+	interact_area.interaction_started.connect(_on_interaction_started)
+
+
+func _on_interaction_started(player: Player, _from_right: bool) -> void:
+	if before_dialogue:
+		await before_dialogue.call()
+	DialogueManager.dialogue_ended.connect(_on_dialogue_ended, CONNECT_ONE_SHOT)
```
wjt: You could instead:

```gdscript
	DialogueManager.show_dialogue_balloon(dialogue, title, [get_parent(), player])
    interact_area.end_interaction()
```

rather than needing a separate callback method. Up to you!

scenes/game_logic/talk_behavior.gd
```diff
@@ -0,0 +1,49 @@
...
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+	if not interact_area:
+		warnings.append("Interact Area property must be set.")
+	return warnings
+
+
+func _ready() -> void:
+	interact_area.interaction_started.connect(_on_interaction_started)
```
wjt, line 42: I guess `InteractArea` is a `@tool` script? But I wonder if you should guard this with a check for running in the editor (because you don't need to connect to the signal in that case; and because this would fail if `InteractArea` stopped being a `@tool`); it's probably fine either way.

scenes/game_logic/talk_behavior.gd
```diff
@@ -0,0 +1,49 @@
...
+## When the dialogue ends, it finishes the interaction by calling [method
+## InteractArea.end_interaction].[br][br]
+## Optionally, an awaitable function can be passed to [member before_dialogue] if
+## something needs to happen before the [member dialogue] is displayed. This can be used,
+## for example, to dinamically set the [member title] of the dialogue considering the
+## game state at the moment the interaction happens.
+
+@export var dialogue: DialogueResource = preload("uid://cc3paugq4mma4")
+@export var title: String = ""
+@export var before_dialogue: Callable
```
wjt, line 22: OOI how do you set a callable in the inspector? I guess I can find out!

scenes/game_elements/characters/npcs/elder/components/elder.gd
```diff
@@ -2,10 +2,13 @@
 # SPDX-License-Identifier: MPL-2.0
 @tool
 class_name Elder
-extends Talker
+extends NPC
 
 const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
 
+@export var npc_name: String
```
wjt: Should we move this to the `NPC` instead?

scenes/game_elements/characters/npcs/elder/components/elder.gd
```diff
@@ -2,10 +2,13 @@
 # SPDX-License-Identifier: MPL-2.0
 @tool
 class_name Elder
-extends Talker
+extends NPC
```
wjt, line 5: The other thing that `Talker` does is set the interact hint on the area to `"Talk to $npc_name"`. I think we are losing that in this change.

scenes/game_elements/characters/npcs/talker/components/talker.gd
```diff
@@ -3,6 +3,7 @@
 @tool
 class_name Talker
 extends NPC
+## @deprecated: Instead of instantiating this class, use a [TalkBehavior] node instead.
```
wjt, line 6: Huh, I didn't realise this was the right place to put the `@deprecated` annotation!

wjt (changes requested):
Nice improvement - even if it's not a big decrease in code size I think it's good to make this change.

A couple of minor suggestions:

scenes/game_elements/characters/npcs/elder/components/elder.gd
wjt: Before:

image

After:

image

manuq: Oh right! Bad regression. First off, yes I agree that `npc_name` should be moved to the NPC and I though about it while doing this PR. But since that's an exported property, I worry that I'll break any potential StoryQuest being submitted. But maybe we just do the breaking change and see?

I guess I can either:
- In Elder, duplicate `interact_area.action = "Talk to %s" % npc_name`.
- In TalkBehavior, set `interact_area.action = "Talk to %s" % npc_name` if `get_parent() is NPC` or if `get_parent().get_property_list()` has a Dictionary `{"name": "npc_name", "type": TYPE_STRING}` (the latter duck-typing is too much IMO).

wjt: I don't think it will break any potential storyquests because if you look at the scene file, the serialisation format just has `npc_name = "Firekeeper"` - it doesn't care which class the property comes from. So you have:

- Instance of (subclass of) Talker: no change, `npc_name` still works (you can check this with the firekeeper in Stella
- Instance of (subclass of) NPC: no change, they now have a `npc_name` property which is unset

The only problematic case is if someone has subclassed `NPC` and also added an `npc_name` property. I think that is very unlikely.

scenes/game_logic/talk_behavior.gd
manuq: I prefer the `await`, thanks for suggesting it. I was unsure for the `CONNECT_ONE_SHOT` added in 1edfb44e, but reading the nice commit message `await` will work.

scenes/game_logic/talk_behavior.gd
manuq: Good point, yes I'll add the guard because TalkBehavior is a tool, and it doesn't need to connect to the InteractArea signal while in the editor.

scenes/game_logic/talk_behavior.gd
manuq: Unfortunately there is now way. Only through scripting.

scenes/game_elements/characters/npcs/elder/components/elder.gd
manuq: @wjt I appended fixup commits for your requests, thank you! For now I decided to duplicate `"Talk to %s" % npc_name` in the Elder script, while we decide if we move `npc_name` to the NPC class.

Another option is introduce a new `Character` or `GameEntity` component with a `character_name` (since `name` is the node's name) and maybe also the `look_at_side`. Then both players and NPCs can use this new component.

manuq: > I don't think it will break any potential storyquests because if you look at the scene file, the serialisation format just has `npc_name = "Firekeeper"` - it doesn't care which class the property comes from. So you have:
> 
> * Instance of (subclass of) Talker: no change, `npc_name` still works (you can check this with the firekeeper in Stella
> * Instance of (subclass of) NPC: no change, they now have a `npc_name` property which is unset
> 
> The only problematic case is if someone has subclassed `NPC` and also added an `npc_name` property. I think that is very unlikely.

Oh good point! I will proceed then. Unless the Character component idea is better for you.

wjt: > Another option is introduce a new `Character` or `GameEntity` component with a `character_name` (since `name` is the node's name) and maybe also the `look_at_side`. Then both players and NPCs can use this new component.

Interesting point. I think moving `npc_name` up to `NPC` is worth doing even if we later make that (larger) change.

manuq: Ready for review again. I moved the `npc_name` where it belongs and updated the commit messages. As a follow up I would like to document NPC and InteractArea.

PR #966, opened 2025-07-23
Play from Fray's End when running from the Editor

manuq:
Change the project main scene to be a new loader scene. The only thing that this scene does is to switch to the Splash scene if the game is not running in the editor, and to Fray's End if the game is running in the editor.

Fray's End is the actual main scene of the game. For development, it is better to skip the splash and intro and land directly in Fray's End.

Playing from the Splash scene is still possible by opening that scene in the Editor and clicking the "Run Current Scene" button.

### DIFF
--- a/scenes/loader.tscn
+++ b/scenes/loader.tscn
@@ -1,0 +1,15 @@
+[gd_scene load_steps=2 format=3 uid="uid://blduub8yuhb2h"]
+
+[sub_resource type="GDScript" id="GDScript_uc6ls"]
+script/source = "extends Node
+
+const SPLASH = preload(\"uid://huuo8mnwsphv\")
+const FRAYS_END = preload(\"uid://cufkthb25mpxy\")
+
+func _ready() -> void:
+	var initial_scene := FRAYS_END if OS.has_feature(\"editor\") else SPLASH
+	get_tree().call_deferred(\"change_scene_to_packed\", initial_scene)
+"
+
+[node name="Loader" type="Node"]
+script = SubResource("GDScript_uc6ls")
--- a/project.godot
+++ b/project.godot
@@ -11,7 +11,7 @@ config_version=5
 [application]
 
 config/name="Threadbare"
-run/main_scene="uid://huuo8mnwsphv"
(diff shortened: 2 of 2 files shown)


scenes/loader.tscn
```diff
@@ -0,0 +1,15 @@
+[gd_scene load_steps=2 format=3 uid="uid://blduub8yuhb2h"]
+
+[sub_resource type="GDScript" id="GDScript_uc6ls"]
+script/source = "extends Node
+
+const SPLASH = preload(\"uid://huuo8mnwsphv\")
+const FRAYS_END = preload(\"uid://cufkthb25mpxy\")
+
+func _ready() -> void:
+	var initial_scene := FRAYS_END if OS.has_feature(\"editor\") else SPLASH
```
manuq: Should this check for the "editor_runtime" feature tag instead? https://docs.godotengine.org/en/4.4/tutorials/export/feature_tags.html

wjt:
I pushed another patch which is my best attempt to also open Fray's End when editing the project and the empty Loader scene loads by default.

I'm not sure this is worth the complexity though. When testing the game in the editor I rarely play from the start with F5 anyway.

manuq:
> Oops, I didn't actually post my other comment:
> 
> I think that opening a blank scene when you first open the game in the editor is worse than loading the logo.

Indeed, I don't like it.

> 
> Which problem are you trying to solve (or both)?
> 
> * The default scene when editing the game is the splash screen, which most people will never edit;
> * When testing the game with `F5` you have to go through the whole logo screen/menu/intro rigmarole to get to the game.

The latter. I am so used to run from the main scene that I don't want to go through the logo and intro. I think it's better to land on Fray's End.

manuq:
This merge broke the web build with "SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.":

image

manuq:
Related engine bug: https://github.com/godotengine/godot/issues/91713

manuq:
The merge above fixed the web build. But then I noticed that the new loader is occluding the debug feature in the scene_switcher autoload, to open scenes from the provided URL hash.

PR #968, opened 2025-07-24
Talk behavior followup

manuq:
Follow up of https://github.com/endlessm/threadbare/issues/950

### DIFF
--- a/scenes/game_logic/talk_behavior.gd
+++ b/scenes/game_logic/talk_behavior.gd
@@ -43,7 +43,7 @@ func _ready() -> void:
 	interact_area.interaction_started.connect(_on_interaction_started)
 
 	if get_parent() is NPC:
-		var npc = get_parent() as NPC
+		var npc := get_parent() as NPC
 		if npc.npc_name:
 			interact_area.action = "Talk to %s" % npc.npc_name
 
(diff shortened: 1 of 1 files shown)


scenes/game_logic/talk_behavior.gd
```diff
@@ -43,7 +43,7 @@ func _ready() -> void:
 	interact_area.interaction_started.connect(_on_interaction_started)
 
 	if get_parent() is NPC:
-		var npc = get_parent() as NPC
+		var npc := get_parent() as NPC
 		if npc.npc_name:
 			interact_area.action = "Talk to %s" % npc.npc_name
```
wjt, line 48: Another way to spell this is:

```gdscript
	var parent := get_parent()
	if parent is NPC:
		if parent.npc_name:
			interact_area.action = "Talk to %s" % parent.npc_name
```

In the happy branch of a `if foo is Bar` block, then `foo`'s type is narrowed to `Bar`.

It would save writing `NPC` twice but is just the same number of lines of code. 🤷🏻

wjt: Or at least, in the code above you don't get any type warnings (either at runtime or in the editor), and you can ctrl-click the `npc_name` reference and jump to the correct documentation.

Annoyingly if you do this:

```
	var parent := get_parent()
	if parent is NPC and parent.npc_name:
		interact_area.action = "Talk to %s" % parent.npc_name
```

then ctrl-click doesn't work. But you still don't get any warnings! 😲

PR #969, opened 2025-07-24
InteractArea: Warn if it doesn't belong to its collision layer

manuq:
The InteractArea instantiated scene or node is supposed to be in a particular collision layer. So add a warning if not. Override the _set() virtual method to update the warnings when the collision layers change.

### DIFF
--- a/scenes/game_elements/props/interact_area/components/interact_area.gd
+++ b/scenes/game_elements/props/interact_area/components/interact_area.gd
@@ -36,6 +36,21 @@ func get_global_interact_label_position() -> Vector2:
 	return to_global(interact_label_position)
 
 
+func _get_configuration_warnings() -> PackedStringArray:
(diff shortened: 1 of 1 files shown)


scenes/game_elements/props/interact_area/components/interact_area.gd
```diff
@@ -36,6 +36,21 @@ func get_global_interact_label_position() -> Vector2:
...
+		warnings.append(
+			"Consider enabling collision with the interactable layer: %d." % INTERACTABLE_LAYER
+		)
+	return warnings
+
+
+func _set(property: StringName, _value: Variant) -> bool:
+	if property == "collision_layer":
+		update_configuration_warnings()
+	return false
```
wjt, line 51: Interesting approach!

wjt (approved):

PR #972, opened 2025-07-24
Use fabric water rock

### LINKED ISSUE #970: Use fabric water rock asset
We have this nice fabric rock-in-water asset in the tree:

Animation frames of a rock in water

And it has a "struck" variant in case we ever do a water version of the sequence puzzle! (Maybe it could be hit by throwing something?)

Animation frames of a fabric rock in water that has been struck by something

But we have not used it in the game yet.

It would be nice to do so!

wjt:
This adds a decoration scene for the fabric water rock, reusing a script that was previously only used for bushes. It places a few rocks in Fray's End; and also adjusts the stealth level to use this scene for its (non-fabric) animated water rocks.

Fixes #970


scenes/game_elements/props/decoration/water_rock/components/water_rock_fabric.tres
```diff
@@ -0,0 +1,94 @@
...
+"texture": SubResource("AtlasTexture_kq6hu")
+}, {
+"duration": 1.0,
+"texture": SubResource("AtlasTexture_c01r6")
+}, {
+"duration": 1.0,
+"texture": SubResource("AtlasTexture_x3akt")
+}],
+"loop": true,
+"name": &"default",
```
manuq, line 78: I wonder if we can come up with a convention for spriteframes animation names for props, like we have "idle", "walk", etc for characters. So they are exchangable. I think this works to start.

manuq (approved):
Excellent!

PR #973, opened 2025-07-24
Stealth: Remove stray shadow under bridge

wjt:
The bridges in this level are raised up 1 level of cliff, so the shadow on the water should be in the row below, not on the same row as the bridge.

One of the bridges had both a correct shadow and an incorrect one. Remove the incorrect one.

Before:

image

After:

Screenshot from 2025-07-24 17-26-07


manuq (approved):
Good catch!

PR #975, opened 2025-07-30
Add character walk behaviors

manuq:
This is the result of the exploration in https://github.com/endlessm/threadbare/pull/777 for discussion https://github.com/endlessm/threadbare/discussions/774 .

These nodes are pure logic, intended to provide an easy way of creating
characters that move around the world with pure setup and no scripting. A more
advanced use is to have a script that can switch between these simple behaviors
to create more complex behaviors.

They are:
- BounceWalkBehavior: Make the character move and bounce on walls.
- ErraticWalkBehavior: Make the character walk around erratically.
- FollowWalkBehavior: Make the character follow a target.
- PathWalkBehavior: Make the character walk along a path.
- InputWalkBehavior: Control the character with input actions to walk and run.

They only take care of controlling the movement of the parent CharacterBody2D
node by setting the velocity and calling move_and_slide. For aesthetics, add a
CharacterSpriteBehavior. This is based in current player_sprite.gd and improved.
In the future, we may also want to support a behavior for animations done with
AnimationPlayer because they can include sound effects that match the visuals.

The walk behavior nodes have exported properties to configure them. There is
quite some repeated code in those. We may consider a single shared resourse
exported.

The walk behavior nodes try to emit as many signals as possible for the
different situations, like when the character turns around when walking on a
path or when the character gets stuck while moving. A script could connect to
these signals to make more interesting behaviors.

The "is stuck" logic is inspired by the one in guard_movement.gd, but changed
to use:
- The boolean returned by move_and_slide() that tells if a collision happened.
- The CharacterBody2D.get_real_velocity()
- The CharacterBody2D.is_on_wall()

For the latter, the character must be properly configured for a top-down game
with motion_mode set to MotionMode.MOTION_MODE_FLOATING.

Resolves https://github.com/endlessm/threadbare/issues/1105

---
  - https://app.asana.com/0/0/1211209074317270

### DIFF
--- a/scenes/game_logic/bounce_walk_behavior.gd
+++ b/scenes/game_logic/bounce_walk_behavior.gd
@@ -1,0 +1,69 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name BounceWalkBehavior
+extends Node2D
+## @experimental
+##
+## Make the character move and bounce on walls.
+
+## Emitted when [member direction] is updated.
+signal direction_changed
+
+## The character walking speed.
+@export_range(10, 1000, 10, "or_greater", "suffix:m/s") var walk_speed: float = 300.0
+
+## The speed to consider that the character is stuck.
+## If less than [member walk_speed], the character may slide on walls instead of bouncing.
+@export_range(0, 1000, 10, "or_greater", "suffix:m/s") var stuck_speed: float = 300.0
+
+## If set, ignore the [member initial_angle] and pick a random one instead.
+@export var pick_random_initial_angle: bool = false:
+	set(value):
+		pick_random_initial_angle = value
+		notify_property_list_changed()
+
+## Angle of the initial direction.
+@export_range(0, 360, 1, "radians_as_degrees") var initial_angle: float = PI / 4.0
+
+## The current direction as a unit Vector2.
+var direction: Vector2
+
+## The controlled character.
+@onready var character: CharacterBody2D = get_parent()
+
+
+func _validate_property(property: Dictionary) -> void:
+	if property.name == "initial_angle" and pick_random_initial_angle:
+		property.usage |= PROPERTY_USAGE_READ_ONLY
+
+
+func _update_direction() -> void:
+	if not direction:
+		direction = Vector2.from_angle(
+			randf_range(0, TAU) if pick_random_initial_angle else initial_angle
+		)
+		direction_changed.emit()
+	elif character.is_on_wall():
+		direction = direction.bounce(character.get_wall_normal())
+		direction_changed.emit()
+
+
+func _ready() -> void:
+	if Engine.is_editor_hint():
+		set_physics_process(false)
+		return
+
+	_update_direction()
+
+
+func _physics_process(_delta: float) -> void:
+	if not direction:
+		_update_direction()
+
+	character.velocity = direction * walk_speed
+	var collided := character.move_and_slide()
+
+	if collided and character.is_on_wall():
+		if character.get_real_velocity().length_squared() <= stuck_speed * stuck_speed:
+			_update_direction()
--- a/scenes/game_logic/character_sprite_behavior.gd
+++ b/scenes/game_logic/character_sprite_behavior.gd
@@ -1,0 +1,74 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name CharacterSpriteBehavior
+extends Node2D
+## Flip horizontally and/or play animations in [member sprite] according
+## to the velocity of [member character].
+
+## The [member CharacterBody2D.velocity] is used to change the [member sprite].
+@export var character: CharacterBody2D
+
+## Whether to play the sprite animations or not. If not, the only thing that will happen is that
+## the sprite will be flipped horizontally according to the velocity of [member character].
+## Use this when using more advanced animation through an AnimationPlayer node.
+@export var play_animations: bool = true
+
+var _is_character_running: bool = false
+
+## The controlled sprite.
+@onready var sprite: AnimatedSprite2D = get_parent()
+
+
+func _ready() -> void:
+	if Engine.is_editor_hint():
+		set_process(false)
+		return
+
+
+func _process(delta: float) -> void:
+	if not character:
+		return
+
+	if play_animations:
+		_process_animations(delta)
+
+	if not is_zero_approx(character.velocity.x):
+		sprite.flip_h = character.velocity.x < 0
+
+
+func _process_animations(_delta: float) -> void:
+	if character.velocity.is_zero_approx():
+		sprite.speed_scale = 1.0
+		sprite.play(&"idle")
+	else:
+		if _is_character_running:
+			if sprite.sprite_frames.has_animation(&"run"):
+				sprite.speed_scale = 1.0
+				sprite.play(&"run")
+			else:
+				sprite.speed_scale = 2.0
+				sprite.play(&"walk")
+		else:
+			sprite.speed_scale = 1.0
+			sprite.play(&"walk")
+
+
+func _notification(what: int) -> void:
+	match what:
+		NOTIFICATION_EDITOR_PRE_SAVE:
+			# Since this is a tool script that plays the animations in the
+			# editor, reset the frame progress before saving the scene.
+			sprite.frame_progress = 0
+
+
+## Force the sprite to flip to a [enum Enums.LookAtSide] direction.
+func look_at_side(side: Enums.LookAtSide) -> void:
+	if side == 0:
+		return
+	sprite.flip_h = side < 0
+
+
+## You can connect this callback to a [member InputWalkBehavior.running_changed] signal.
+func on_running_changed(is_running: bool) -> void:
+	_is_character_running = is_running
--- a/scenes/game_logic/erratic_walk_behavior.gd
+++ b/scenes/game_logic/erratic_walk_behavior.gd
@@ -1,0 +1,79 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name ErraticWalkBehavior
+extends Node2D
+## @experimental
+##
+## Make the character walk around erratically.
+
+## Emitted when [member character] got stuck while walking.
+signal got_stuck
+
+## Emitted when [member direction] is updated.
+signal direction_changed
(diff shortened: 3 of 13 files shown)


scenes/game_logic/bounce_walk_behavior.gd
```diff
@@ -0,0 +1,71 @@
...
+@tool
+class_name BounceWalkBehavior
+extends Node2D
+## Make the character move and bounce on walls.
+
+## Emitted when [member direction] is updated.
+signal direction_changed
+
+## The character walking speed.
+@export_range(10, 100000, 10, "or_greater", "suffix:m/s") var walk_speed: float = 300.0
```
wjt: 100,000 seems way too high for the upper limit. 1000 seems fast enough, and it makes the slider more useful: if the range is 10 to 100,000 the slider is basically at the far left for every useful value.

scenes/game_logic/bounce_walk_behavior.gd
```diff
@@ -0,0 +1,71 @@
...
+
+## The speed to consider that the character is stuck.
+## If less than [member walk_speed], the character may slide on walls instead of emitting
+## the [signal got_stuck] signal.
+## If closer to zero, the character may not ever emit the [signal got_stuck] signal.
```
wjt: I this is copied from ErraticWalkBehavior - there is no `got_stuck` signal here. The actual behaviour is that if the speed is below this, the player bounces rather than sliding along the wall.

scenes/game_logic/bounce_walk_behavior.gd
```diff
@@ -0,0 +1,71 @@
...
+@export_range(0, 100000, 10, "or_greater", "suffix:m/s") var stuck_speed: float = 300.0
```
wjt: If you were feeling really fancy, you could dynamically adjust the upper bound of this property in _validate_property to be `walk_speed` - I don't think the character can ever move faster than that?

scenes/game_logic/erratic_walk_behavior.gd
```diff
@@ -0,0 +1,73 @@
...
+
+	character.velocity = direction * walk_speed
+	var collided := character.move_and_slide()
+	if collided and character.is_on_wall():
+		if character.get_real_velocity().length_squared() <= stuck_speed * stuck_speed:
+			got_stuck.emit()
+			_update_direction()
+			distance = 0.0
+	else:
+		distance += walk_speed * _delta
```
wjt: Rename the parameter to `delta` - the prefix means "unused".

scenes/game_logic/erratic_walk_behavior.gd
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name ErraticWalkBehavior
+extends Node2D
+## Make the character walk around erratically.
+##
+## TODO:
+## - Emit signals and hook with ThrowingEnemy
```
wjt: Yes I think this would be interesting to see.

Some of the other examples of components that I have seen use a different pattern to these. Here your pattern is that each behaviour implements `_process()` (and I guess if we were to introduce `PlayerInputBehaviour` that one would implement `_unhandled_input()` too) and then you change the process mode so that only one behaviour is active at a time. Other examples instead make the behaviours do nothing by default, with methods that must be called directly.

The advantage of doing it this way is that you can add a walk behaviour to a node with no coding required. But as soon as you want to switch between behaviours, you still need to write code to disable and enable different components. I suppose it's no harder than tracking the active component and calling it directly!

scenes/game_logic/erratic_walk_behavior.gd
```diff
@@ -0,0 +1,73 @@
...
+## If less than [member walk_speed], the character may slide on walls instead of emitting
+## the [signal got_stuck] signal.
+## If closer to zero, the character may not ever emit the [signal got_stuck] signal.
+@export_range(0, 100000, 10, "or_greater", "suffix:m/s") var stuck_speed: float = 100.0
+
+## The turn direction will be randomly picked between this and [member turn_angle_end].
+@export_range(0, 180, 1, "radians_as_degrees") var turn_angle_left: float = PI / 2.0
+
+## The turn direction will be randomly picked between [member turn_angle_start] and this.
+@export_range(0, 180, 1, "radians_as_degrees") var turn_angle_right: float = PI / 2.0
```
wjt: The docstrings refer to turn_angle_end and _start, while the properties are _left and _right. I think I gently prefer left and right.

(I would suggest we use _anticlockwise and _clockwise but not only are these really long words, I remembered that US English (which CONTRIBUTING says is the correct dialect) says _counterclockwise not anti. dextrorotatory and levorotatory? 😁 )

scenes/game_logic/follow_walk_behavior.gd
```diff
@@ -0,0 +1,67 @@
...
+
+## The current distance travelled since last direction update.
+var distance: float = 0
+
+## The controlled character.
+@onready var character: CharacterBody2D = get_parent()
+
+
+func _update_direction() -> void:
+	direction = character.global_position.direction_to(target.global_position)
```
wjt, line 76: The motion is a bit abrupt with these sharp turns. Maybe we could adjust the angle gradually - I found rotate_toward() which is like move_toward() but for angles.

But this would be a refinement!

scenes/game_logic/follow_walk_behavior.gd
```diff
@@ -0,0 +1,67 @@
...
+
+	_update_direction()
+
+
+func _process(_delta: float) -> void:
+	if Engine.is_editor_hint():
+		return
+
+	if not direction:
+		_update_direction()
```
wjt, line 89: I think there's a missing check for reaching the target (modulo `target_reached_distance`), keeping a boolean for whether the target is currently in range, and emitting `target_reached` when going from "out of range" to "in range" (but probably with some hysteresis, i.e. if I'm walking at almost exactly the follower's speed, I wouldn't want the signal to be repeatedly emitted every few frames when I dip in and out of its target range).

wjt (commented):
This looks really promising!

scenes/game_logic/bounce_walk_behavior.gd
manuq: Good point! This was copied from current player. So I will change that one too. And we have the "or_greater" so it's not a limitation.

scenes/game_logic/bounce_walk_behavior.gd
manuq: Since all these behavior nodes share some exports, I was thinking in exporting a resource called CharacterBehaviorOptions with the shared ones, what do you think? In that case I can comment that `got_stuck` is not used for the BounceWalkBehavior. Hmm thinking aloud, then I won't be able to reference `[signal got_stuck]` from these comments.

scenes/game_logic/bounce_walk_behavior.gd
manuq: This is compared with `CharacterBody2D.get_real_velocity()` which I'm not sure if will ever be bigger than walk speed. So maybe yes, I could add a _validate_property.

scenes/game_logic/erratic_walk_behavior.gd
manuq: By the way I'm switching to use `_physics_process` !

scenes/game_logic/erratic_walk_behavior.gd
manuq: Ideally I'd add a visual debugging showing what are these angles about in the editor on top of the character.

scenes/game_logic/follow_walk_behavior.gd
manuq: Oh I need to try rotate_toward() for the grappling hook as replacement of lerp_angle().

scenes/game_logic/follow_walk_behavior.gd
manuq: Very good point, I changed it to a `target_reached_changed` signal that will emit true or false when the target becomes in range and when it becomes out of range.

(Sorry I haven't seen this comment before but it looks like we are aligned!)

manuq: Hmm I think the same problem could be in the `got_stuck` signal. I tried it with behaviors that do something immediately like turn around when stuck, but some others may be emitting the signal repeatedly in every physics loop.

scenes/game_logic/character_sprite_behavior.gd
```diff
@@ -0,0 +1,74 @@
...
+## Force the sprite to flip to a [enum Enums.LookAtSide] direction.
+func look_at_side(side: Enums.LookAtSide) -> void:
+	if side == 0:
+		return
+	sprite.flip_h = side < 0
+
+
+## You can connect this callback to a [member InputWalkBehavior.running_changed] signal.
+func on_running_changed(is_running: bool) -> void:
+	_is_character_running = is_running
```
wjt, line 74: I was going to say that you should have called this `set_is_running` or something but actually I think this name is more descriptive of how it should be used.

scenes/game_logic/follow_walk_behavior.gd
```diff
@@ -0,0 +1,106 @@
...
+
+## The current distance travelled since last direction update.
+var distance: float = 0
+
+## True if the character has reached the target.
+var is_target_reached: bool:
+	set = _set_is_target_reached
+
+## The controlled character.
+@onready var character: CharacterBody2D = get_parent()
```
wjt, line 53: I don't think it needs to be done in this PR, but I think we could catch a class of user errors if we updated this property (and the equivalents in other behaviors; perhaps via a `base_behavior.gd`?) on `_enter_tree()` and `_exit_tree()`, which I think are called if you move a node around the hierarchy in the editor; and added a configuration warning if the parent has the wrong type.

scenes/game_logic/path_walk_behavior.gd
```diff
@@ -0,0 +1,174 @@
...
+signal got_stuck
+
+## Emitted when turning around.
+signal direction_changed
+
+## Emitted when a "pointy" part of the path is reached.
+## This could be used to wait standing for a bit in these points.
+## If the path is not closed, both endings are considered pointy so this signal will also
+## emit in them.
+signal pointy_path_reached
```
wjt, line 28: I love this signal name!

scenes/game_logic/follow_walk_behavior.gd
manuq: I agree that this is important! So I hurried a follow up: https://github.com/endlessm/threadbare/pull/1144

Issue #976, opened 2025-07-30
Development version fails to load with parse error


manuq:
### Description

See https://github.com/endlessm/threadbare/pull/966#issuecomment-3132421006

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #977, opened 2025-07-30
Fix EditorInterface parse error bug

manuq:
Fix https://github.com/endlessm/threadbare/issues/976

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -406,11 +406,10 @@ func _notification(what: int) -> void:
 			_reset()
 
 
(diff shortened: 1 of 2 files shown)


cassidyjames (approved):
I don't 100% understand the fix here, but:
1. I trust you @manuq :D
2. I can confirm this branch is playable on the web
3. I can confirm this branch still opens in Fray's End when running from the editor

manuq:
> I don't 100% understand the fix here, but:
> 
> 1. I trust you @manuq :D
> 2. I can confirm this branch is playable on the web
> 3. I can confirm this branch still opens in Fray's End when running from the editor

You understood everything then, thanks!

wjt:
Oops, my bad. Thanks for fixing this!

Issue #725, opened 2025-06-03
Font scaling is odd


manuq:
The font characters "dance" because of the bad scaling or aliasing:

https://github.com/user-attachments/assets/b42f4332-ab75-4dd5-be93-45608476171f

manuq:
Things I tried so far, without luck:
- Resetting all `texture_filter = linear` in Fray's End and in the `fixed_size_label.tscn` scene.
- Changing antialiasing in project settings from Grayscale to LCD Subpixel.
- Changing the import settings of the m6x11plus.ttf using the Advanced settings and then reimporting.

I see the issue even if in `fixed_size_label.tscn` scene I reset the node to use the default theme. So I think part (or all) the problem is due to the way we make the label float and follow an object in a scene, with fixed scale regardless of the camera zoom.

manuq:
Today I've found that the most important thing to fix here is to prevent the label being repositioned in its own canvas layer. Every frame it is trying to catch up with the anchor point of the thing being interacted, because of the translation from world coordinates to screen coordinates. I opened a PR that has a proposal for fixing it.

wjt:
I wonder if we can take a different approach here:

1. Have the hint be in the game layer, not a CanvasLayer atop it
2. When about to display the hint: 
    - [get the current viewport](https://docs.godotengine.org/en/4.4/classes/class_node.html#class-node-method-get-viewport),
    - [get its Camera2D](https://docs.godotengine.org/en/4.4/classes/class_viewport.html#class-viewport-method-get-camera-2d),
    - scale the label as the inverse of the camera's current scale

wjt:
I wonder if one reason that the label is playing catch-up is that the Camera2D repositions itself in its `_process` callback (or rather the C++ equivalent), and our code also repositions in `_process`. But the Camera2D is a later child of Player than the `PlayerInteraction`...

wjt:
We removed the floating interact label ⇒ this is resolved!

PR #978, opened 2025-07-30
Remove label dizzying effect

manuq:
There is a dizzying effect in the labels when something can be interacted. This is because the labels are in a layer on top and they are positioned in screen (viewport) coordinates, by translating the anchor point of the object being interacted from world coordinates. This is so their size is independient of the camera zoom.

That translation from world coordinates to screen coordinates is always catching up with the previous frames, producing a dizzying effect and blurry text while the camera is moving.

The proposal here is to avoid repositioning the label every frame. Do it at a rate that depends on how far is the new position, thus how important it is to do so. Also allow some offset between the anchor point and the current position. And finally, also fade in and out the label while repositioning.

Playable at https://endlessm.github.io/threadbare/branches/endlessm/remove-label-dizzying-effect/#world_map/frays_end

Before:

[Grabación de pantalla desde 2025-07-30 14-12-06.webm](https://github.com/user-attachments/assets/ef88f119-8ef1-4f01-b5f5-586ba088b2ee)

After:

[Grabación de pantalla desde 2025-07-30 14-05-08.webm](https://github.com/user-attachments/assets/d7bc4c3b-33a1-4459-80d0-c028026bd41f)

Fix https://github.com/endlessm/threadbare/issues/725

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -181,4 +181,3 @@ sokoban_skip={
 
 renderer/rendering_method="gl_compatibility"
 renderer/rendering_method.mobile="gl_compatibility"
-2d/snap/snap_2d_transforms_to_pixel=true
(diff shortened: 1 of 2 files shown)


manuq:
Two weeks have passed and I'm not actively working on it. And also I don't think that this is an improvement, it makes the effect weird in a different way.

PR #985, opened 2025-08-04
Update to Godot Dialogue Manager 3.7.1

wjt:
Previously, we had 3.6.1 plus a patch that I sent upstream, which was included in the 3.6.2 release. So this covers the following changes:

- https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v3.7.1
- https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v3.7.0
- https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v3.6.3
- https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v3.6.2

Some minor improvements and fixes, many of which are to the in-editor interface rather than runtime-related.


addons/dialogue_manager/l10n/en.mo
wjt, line 1: I think this was added unintentionally upstream. I proposed https://github.com/nathanhoad/godot_dialogue_manager/pull/952 to delete it.

manuq:
Awesome, is great to have your patch merged upstream.

Issue #979, opened 2025-07-30
Regression in web builds loading from  URL hash


manuq:
### Description

The PR https://github.com/endlessm/threadbare/pull/966 introduced a loader as the main scene. For web builds, the new loader now switching to the splash scene always, occluding the debug feature in the `scene_switcher` autoload that opens the scene from the provided URL hash.

Acceptance criteria: A link like https://endlessm.github.io/threadbare/branches/endlessm/main/#world_map/frays_end should open Fray's End again.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #986, opened 2025-08-04
Revert loader script

manuq:
Resolves https://github.com/endlessm/threadbare/issues/979


manuq:
@wjt I used the playable link and confirmed that it brings back the hash URL debugging: https://endlessm.github.io/threadbare/branches/endlessm/revert-loader-script/#world_map/frays_end

Issue #989, opened 2025-08-04
Remove snap transforms to pixel constrain?


manuq:
### Description

Related to https://github.com/endlessm/threadbare/issues/638

This is another thing I found while working on the upcoming style guide. The documentation for this setting says that this is for low-resolution pixel art games and that it comes at the cost of less smooth movement. Specially when camera smoothing is enabled (our case):

> If true, [CanvasItem](https://docs.godotengine.org/en/4.4/classes/class_canvasitem.html#class-canvasitem) nodes will internally snap to full pixels. Useful for low-resolution pixel art games. Their position can still be sub-pixel, but the decimals will not have effect as the position is rounded. This can lead to a crisper appearance at the cost of less smooth movement, especially when [Camera2D](https://docs.godotengine.org/en/4.4/classes/class_camera2d.html#class-camera2d) smoothing is enabled.

This was added 6a53c4aae without being a clear fix for anything in particular.

Here is a video explaining why is better to not have this constrain in high resolution pixel art games: https://youtu.be/-hSRLmk3nLo?t=88 . The less smooth camera movement is showcased.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #990, opened 2025-08-04
Project: Remove snap transforms to pixel

manuq:
Revert commit 6a53c4aae . This is not a pixel-perfect or low-resolution pixel art game. So it is better if pixels aren't constrained for smoother movement.

Resolves https://github.com/endlessm/threadbare/issues/989

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -181,4 +181,3 @@ sokoban_skip={
 
 renderer/rendering_method="gl_compatibility"
 renderer/rendering_method.mobile="gl_compatibility"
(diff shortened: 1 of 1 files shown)


wjt (approved):
I have to say that I don't notice much difference between https://endlessm.github.io/threadbare/#world_map/frays_end (before this change) and https://endlessm.github.io/threadbare/branches/endlessm/remove-snap-transforms-to-pixel/#world_map/frays_end (after this change). OK!

PR #999, opened 2025-08-07
AreaFiller: Add test scene

wjt:
This demonstrates how to create:

1. A boundary forest, using a StaticBody2D;
2. An Area2D with individual trees that the player can walk through;
3. A flowerbed.

The signs are meant to give hints for how the level designer can use this component.

I also fixed an issue with the rock prop that I noticed while making this scene.


wjt:
Playable in https://endlessm.github.io/threadbare/branches/endlessm/area-filler-example/#game_elements/props/area_filler/components/area_filler_test though it's not really about playing this one!

manuq (approved):
Excellent, adding test scenes like these is great playable documentation.

PR #1011, opened 2025-08-08
Add cat fellow

manuq:
Based on a beloved real cat.

Add the sprite and instantiate it as NPC in Fray's End. The idle animation is a single static frame for now, the tail should be animated.


wjt:
This conflicts with my changes in #1008. I'll fix it up.

Issue #1062, opened 2025-08-11
Change Flatpak app ID from org.endlessos.*


wjt:
Our organisation is now Endless Access, and we have a new domain name (actually more than one), so having the app ID be `org.endlessos.threadbare` would be non-ideal.

We should change it to an ID within one of our current domains, keeping in mind that we will need to be able to put a file at `/.well-known/` on the domain we choose.

PR #1065, opened 2025-08-11
Update Flatpak app ID to org.endlessaccess.threadbare

wjt:
Our organisation has been renamed!

https://endlessaccess.org/ redirects to
https://access.endlessstudios.com/ so another option was com.endlessstudios.access.threadbare, but that's much uglier, and our email addresses will be @endlessaccess.org, so let's use that one.

Update the manifest, metainfo, desktop file and CI pipeline accordingly.

Fixes https://github.com/endlessm/threadbare/issues/1062

### DIFF
--- a/linux/metainfo.xml
+++ b/linux/metainfo.xml
@@ -11,21 +11,21 @@ SPDX-License-Identifier: CC-BY-4.0
     <color type="primary" scheme_preference="dark">#435b7d</color>
   </branding>
   <developer_name translatable="no">Endless Access</developer_name>
-  <developer id="org.endlessos">
+  <developer id="org.endlessaccess">
     <name translatable="no">Endless Access</name>
   </developer>
   <description>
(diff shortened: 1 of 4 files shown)


linux/metainfo.xml
```diff
@@ -11,21 +11,21 @@ SPDX-License-Identifier: CC-BY-4.0
     <color type="primary" scheme_preference="dark">#435b7d</color>
   </branding>
   <developer_name translatable="no">Endless Access</developer_name>
-  <developer id="org.endlessos">
+  <developer id="org.endlessaccess">
```
wjt, line 14: This is the only piece I'm not sure about. Like it or not, we have other apps published with the `org.endlessos` developer ID, so we may want to keep this as `org.endlessos` to group them together.

(This is one of my least favourite "improvements" in AppStream 1.0.)

wjt: No, it turns out that we never did update any of those apps to have `<developer id="org.endlessos">...` so this is fine.

PR #1066, opened 2025-08-11
Refresh AppStream metainfo

### LINKED ISSUE #1064: Refresh AppStream metainfo
I noticed that the screenshot link is now incorrect (since Git LFS was enabled, I think!). There may be other things to review too.

wjt:
Fixes #1064

### DIFF
--- a/linux/metainfo.xml
+++ b/linux/metainfo.xml
@@ -45,5 +45,6 @@ SPDX-License-Identifier: CC-BY-4.0
   <recommends>
     <control>keyboard</control>
     <control>gamepad</control>
+    <display_length side="shortest" compare="ge">1080</display_length>
   </recommends>
 </component>
--- a/linux/metainfo.xml
+++ b/linux/metainfo.xml
@@ -32,7 +32,7 @@ SPDX-License-Identifier: CC-BY-4.0
   <url type="contribute">https://github.com/endlessm/threadbare/contribute</url>
   <screenshots>
     <screenshot type="default">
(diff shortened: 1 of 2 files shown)


linux/metainfo.xml
```diff
@@ -25,22 +25,36 @@ SPDX-License-Identifier: CC-BY-4.0
...
+      <url type="details">https://github.com/endlessm/threadbare/releases/tag/v0.1.1</url>
+    </release>
+    <release version="0.1" date="2025-05-23">
+      <url type="details">https://github.com/endlessm/threadbare/releases/tag/v0.1</url>
     </release>
   </releases>
   <recommends>
     <control>keyboard</control>
     <control>gamepad</control>
+    <display_length side="shortest" compare="ge">1080</display_length>
```
wjt, line 55: Oh, actually, since:

- https://github.com/endlessm/threadbare/pull/1009

We do actually have a required minimum size. I'll refresh this to add that.

linux/metainfo.xml
```diff
@@ -25,22 +25,38 @@ SPDX-License-Identifier: CC-BY-4.0
...
+      <image>https://media.githubusercontent.com/media/endlessm/threadbare/67b2cad522c26b56abbfa334c2553a87d9b54af0/docs/screenshot.png</image>
     </screenshot>
   </screenshots>
   <releases>
-    <release version="0.0.0" date="2025-04-28">
-      <description>
-        <p>Pre-release</p>
-      </description>
+    <release version="0.1.3" date="2025-08-11">
+      <url type="details">https://github.com/endlessm/threadbare/releases/tag/v0.1.3</url>
```
wjt, line 40: This is assuming that 0.1.3 releases today. But we can always change the date. (And really the date is not too critical. What matters is the order of the dates!)

PR #1069, opened 2025-08-11
TalkBehavior node: Fix for non NPCs

manuq:
The node is supposed to work for any parent node. And if the parent node is an NPC (i.e. it inherits from the NPC class),then improve the label to use the `npc_name`. But currently there is a bug because npc.npc_name is being referenced even if `get_parent() as NPC` returns null.


manuq:
Issue found while making a rock talk for the v0.1.3 release notes.

scenes/game_logic/talk_behavior.gd
```diff
@@ -44,7 +44,7 @@ func _ready() -> void:
 
 	if get_parent() is NPC:
 		var npc := get_parent() as NPC
-		if npc.npc_name:
+		if npc:
 			interact_area.action = "Talk to %s" % npc.npc_name
```
wjt:
```suggestion
var npc := get_parent() as NPC
if npc and npc.npc_name:
    interact_area.action = "Talk to %s" % npc.npc_name
```

?

wjt:
oh-no-rock

manuq:
@wjt you are right, it was working before and this is not blocking the release. Another error confused me, the dialogue file was referencing `npc_name`.

wjt (approved):
Lgtm!

manuq:
@wjt sorry I missed adding the co-authored-by in the 1st commit since you wrote it.

Issue #1010, opened 2025-08-08
Template: Add tileset template


manuq:
The current template folder doesn't have a tileset. Add one with tiles for a floor variant and an elevation variant.

PR #1080, opened 2025-08-13
Template: Add tileset

manuq:
A simple tileset with terrains for floor, elevation and walls. Based on the style guide.

This is how it looks:

image

I didn't want to touch the current template levels for now, I added a separate test scene. But maybe is better to use it in the levels? 

I spent a while with the elevation. The corners are tricky in this simplified 16 tile set. Specially without color contrast, I wanted the template to be gray. It's so gray and boring that makes me want to replace it. I guess that's a good sign.

Resolves https://github.com/endlessm/threadbare/issues/1010


wjt (approved):
Yes, I think the fact that it's so dull works in its favour!

PR #1082, opened 2025-08-13
Reorder fields of project.godot

wjt:
Commit b99c0e81f87434e6c3b5a13020c76f16c4b9fadf added config/tags to this file; whenever I save the project, Godot reorders the file.

Reorder it to be the way that it is saved.


manuq (approved):
Yes, odd, is happening to me as well...

Issue #1083, opened 2025-08-13
Playtest all quests & file aspect ratio bugs


wjt:
We recently changed the game from having a fixed-size 1920×1080 (16:9) viewport, to expanding horizontally or vertically if the aspect ratio is different to 16:9:

- https://github.com/endlessm/threadbare/pull/992

As a result, when the game is played in a window which has a different aspect ratio to 16:9 there are many levels where you can see further than the level designer originally intended, into parts of the scene which were not intended to be visible. One example of such an issue is:

- https://github.com/endlessm/threadbare/issues/998

There are also issues with other UI elements, such as the control hints shown when the player is on the bottom island of Fray's End:

- https://github.com/endlessm/threadbare/issues/997

Using the development branch of the game in the Godot editor, play through all quests (there are at least 7: 2 from the LoreQuest Elder and 5 from the StoryQuest Elder), adjusting the window size as you go to be either wider or taller than 16:9. Find more problems like the ones above, and for each problem you find, [check if it has already been reported](https://github.com/endlessm/threadbare/issues?q=is%3Aissue%20state%3Aopen%20type%3ABug), and if not, [file a new issue for each problem](https://github.com/endlessm/threadbare/issues/new?template=bug.yml).

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Management & Production: Quality Assurance & Playtesting | Problem Solving | 50 |
| Management & Production: Quality Assurance & Playtesting | Creativity & Design Thinking | 50 |
| Management & Production: Quality Assurance & Playtesting | Technical Literacy | 75 |

russell-cj:
I Plan to work on this

russell-cj:
Hello @manuq , hello @wjt ,

From my side, I’ve already reviewed all the cases related to the aspect ratio issues, and as far as I can see, those should be **all the affected levels**.

Therefore, I believe this issue can now be considered complete.  
I remain attentive to any feedback or confirmation from you both.

PR #1084, opened 2025-08-14
Add scene to debug aspect ratio issues

manuq:
This scene can play a level inside a subviewport, and show semi-transparent red bars when the viewport is too tall or too wide. So bughunters of ratio issues know where is the limit of their search, by ignoring anything that is far away in the red area.

Add Fray's End to the subviewport, but this is supposed to be replaced by the scene to debug.

Note that playing a level through a subviewport has limitations. For instance I wasn't able to use the story book and select a quest in Fray's End.

The minimum and maximum aspect ratios in this commit are rather arbitrary.

This is how it looks (I disabled the minimum window size restriction to take these pics):

Captura desde 2025-08-14 13-39-25

Captura desde 2025-08-14 13-39-13

### DIFF
--- a/scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.gd
+++ b/scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.gd
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends CanvasLayer
+
+
+func _ready() -> void:
+	if not ProjectSettings.get_setting("threadbare/debugging/debug_aspect_ratio"):
+		queue_free()
--- a/scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.tscn
+++ b/scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.tscn
@@ -1,8 +1,10 @@
-[gd_scene load_steps=2 format=3 uid="uid://c5oi15v411wc6"]
+[gd_scene load_steps=3 format=3 uid="uid://c5oi15v411wc6"]
 
 [ext_resource type="Script" uid="uid://brldj3icxiy5y" path="res://scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger_rect.gd" id="1_bf0cn"]
+[ext_resource type="Script" uid="uid://cnv4dqfa0h7cm" path="res://scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.gd" id="1_mi05q"]
 
 [node name="AspectRatioDebugger" type="CanvasLayer"]
+script = ExtResource("1_mi05q")
 
 [node name="OutOfBoundsBars" type="Control" parent="."]
 custom_minimum_size = Vector2(1920, 1080)
--- a/scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.gd.uid
+++ b/scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.gd.uid
@@ -1,0 +1,1 @@
+uid://cnv4dqfa0h7cm


manuq:
@wjt what do you think? Can somethink like this help bug hunters in https://github.com/endlessm/threadbare/issues/1083 ?

manuq:
> In principle I think this is a good idea. But I wonder whether we could include it in one of the existing global scenes, or a new one, enabled & disabled in settings? That way changing between scenes of the game would work?

I did so, thanks for the hint. This removes all the oddities.

With this I started an addon for storing custom settings in Project Settings. I did it this way after reading [ProjectSettings.save()](https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-method-save) that says "Note: This method is intended to be used by editor plugins, as modified ProjectSettings can't be loaded back in the running app.". But maybe I went to far and becoming the current `scenes/globals/settings/settings.gd` a `@tool` would be enough?

scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.gd
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends CanvasLayer
+
+
+func _ready() -> void:
+	if not ProjectSettings.get_setting("threadbare/debugging/debug_aspect_ratio"):
+		queue_free()
```
manuq, line 8: The alternative to add this global unconditionally and then free it in the default case is to add the node programatically from another global like Settings.

wjt:
Force-pushed to fix `reuse lint`.

scenes/globals/settings/settings.gd
```diff
@@ -21,6 +21,9 @@ const VIDEO_WINDOW_MODE_KEY := "Window Mode"
 ## than the base resolution.
 const MINIMUM_WINDOW_SIZE := Vector2i(1280, 720)
 
+const MINIMUM_ASPECT_RATIO := 1.5
+const MAXIMUM_ASPECT_RATIO := 2.2
```
wjt: How did you pick this aspect ratio range?

I looked at https://store.steampowered.com/hwsurvey/ and there is one 5:4 (1.25) ratio: 1280 x 1024 (0.26%). I'm actually surprised it's so high. Maybe we can ignore this ratio but it doesn't look completely unreasonable to me when I test it:

Image

There are also a bunch of ultrawides (21:9 or 32:9) but I think we can ignore them.

scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger_rect.gd
```diff
@@ -0,0 +1,26 @@
...
+extends ColorRect
+## A ColorRect that shows vertical or horizontal bars when the aspect ratio
+## is too wide or too tall.
+##
+## The project sets the aspect ratio to expand, but there aren't minimum
+## or maximum limits. This is a visual clue for dressing the scenes.
+
+
+func _draw():
+	var aspect = get_size().aspect()
```
wjt: I spotted a simplification: `get_size()` can be replaced with `size` property access.

scenes/globals/settings/settings.gd
manuq: > How did you pick this aspect ratio range?

Arbitrary! From the commit message "The minimum and maximum aspect ratios in this commit are rather arbitrary." I'm glad that you found the Steam hardware survey.

> 
> I looked at https://store.steampowered.com/hwsurvey/ and there is one 5:4 (1.25) ratio: 1280 x 1024 (0.26%). I'm actually surprised it's so high. Maybe we can ignore this ratio but it doesn't look completely unreasonable to me when I test it:

Yep that doesn't look bad at all!

> There are also a bunch of ultrawides (21:9 or 32:9) but I think we can ignore them.

:+1:

scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger_rect.gd
manuq: Oh nice!

scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.tscn
```diff
@@ -6,12 +6,12 @@
...
-[node name="ColorRect" type="ColorRect" parent="."]
+[node name="OutOfBoundsBars" type="Control" parent="."]
 custom_minimum_size = Vector2(1920, 1080)
+layout_mode = 3
 anchors_preset = 15
 anchor_right = 1.0
 anchor_bottom = 1.0
 grow_horizontal = 2
 grow_vertical = 2
-color = Color(1, 1, 1, 0)
```
wjt: I don't think it needs to be a ColorRect; so I changed it to a bare Control.

scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger_rect.gd
```diff
@@ -1,26 +1,25 @@
...
 	# Too tall, add vertical bars:
 	elif aspect < Settings.MINIMUM_ASPECT_RATIO:
-		var bar_width := ((get_size().y * Settings.MINIMUM_ASPECT_RATIO) - get_size().x) / 2.0
-		draw_rect(Rect2(0, 0, get_size().x, bar_width), Color(Color.RED, 0.4))
-		draw_rect(
-			Rect2(0, get_size().y - bar_width, get_size().x, bar_width), Color(Color.RED, 0.4)
-		)
+		var bar_width := ((size.y * Settings.MINIMUM_ASPECT_RATIO) - size.x) / 2.0
+		draw_rect(Rect2(0, 0, size.x, bar_width), BAR_COLOR)
+		draw_rect(Rect2(0, size.y - bar_width, size.x, bar_width), BAR_COLOR)
```
wjt: By extracting BAR_COLOR and using `size` the ugly line wrapping is avoided!

scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.tscn
manuq: Oh really? Great to know that Control is enough for custom drawing. Of course I see now that Control inherits CanvasItem which defines `_draw`.

scenes/globals/settings/settings.gd
wjt: Oh yes, you did say. I didn't read.

I suggest we change this to 1.25.

manuq: Agreed!

PR #1085, opened 2025-08-15
Add new Void Runner level

wjt:
This is a prototype of a new mechanic, where the player has to outrun an
expanding void in order to get from A to B. If the vortex that is
spreading the void engulfs the player, they unravel; as the vortex
travels across the map, it spreads the void, hiding any props beneath.
It's somewhat inspired by the levels in Crash Bandicoot where a rock is
rolling towards the player.

The vortex moves at the same speed as the player, but speeds up to the
player's running speed if they get too far ahead. Because the player can
run continuously, without consuming stamina, the vortex has to move at
the running speed when the player is running; but I wanted to make it
possible to beat the level without running. It is possible, though it's
a bit easier if you hold down the run button throughout.

Most of the scene starts off covered up by the void. When the player
reaches their goal and interacts with an NPC, the void is pushed back,
revealing new areas of the environment.

This is currently the only challenge of a new LoreQuest. Since all
quests must grant the player three threads, interacting with the monk
grants the player 2 of the 3 threads. The intention is that we'll add
two more challenges to this quest: the second introducing another new
mechanic, and the third combining the two.

All art, dialogue, animations are placeholders, though I couldn't resist
dressing the scene quite thoroughly.

The quest is named and described without using the letter ‘e’ in homage
to La Disparition by Georges Perec, translated into English as A Void,
which I have never read.


wjt:
https://endlessm.github.io/threadbare/branches/endlessm/void-runner/#quests/lore_quests/quest_002/1_void_runner/void_runner

scenes/quests/lore_quests/quest_002/1_void_runner/components/monk.gd
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends NPC
+## @experimental
```
manuq, line 5: 👍

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
```diff
@@ -0,0 +1,171 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name TileMapCover  # TODO: better naem
```
manuq:
```suggestion
class_name TileMapCover  # TODO: better name
```

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
```diff
@@ -0,0 +1,171 @@
...
+	_unconsumed_nodes.erase(coord)
+	# TODO: store previous modulate.a to restore in uncover_all()
+	if immediate:
+		for node: Node2D in nodes:
+			node.modulate.a = 0.0
+	else:
+		var tween := create_tween().set_parallel(true)
+		for node: Node2D in nodes:
+			tween.tween_property(node, "modulate:a", 0.0, _DURATION).set_ease(Tween.EASE_OUT)
+		await tween.finished
```
manuq, line 150: I wonder if after tweening the modulate alpha to zero, we should set the node visible to false as well. The outcome is the same, but for performance is better to set visible to false. Nodes that modulate to alpha zero are still rendered.

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
```diff
@@ -0,0 +1,171 @@
...
+
+	for node: Node in nodes:
+		# TODO: also store original process mode?
+		node.process_mode = Node.PROCESS_MODE_DISABLED
+	_consumed_nodes[coord] = nodes
```
manuq:
```suggestion
	for node: Node2D in nodes:
		# TODO: also store original process mode?
		node.process_mode = Node.PROCESS_MODE_DISABLED
		# TODO: store previous visible value to restore in uncover_all()
		node.visible = false
```

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
```diff
@@ -0,0 +1,171 @@
...
+	tween.set_trans(Tween.TRANS_CUBIC).set_ease(Tween.EASE_IN_OUT)
+	tween.set_parallel(true)
+	tween.tween_property(self, "modulate:a", 0.0, duration / 2)
+	for nodes: Array in _consumed_nodes.values():
+		for node: Node2D in nodes:
+			tween.tween_property(node, "modulate:a", 1.0, duration)
+			node.process_mode = Node.PROCESS_MODE_INHERIT
+
+	await tween.finished
+	clear()
```
manuq, line 172: Nice, this reminded me that this inherits from TileMapLayer :D

scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
```diff
@@ -0,0 +1,104 @@
...
+const VOID_TERRAIN: int = 9
+const NEIGHBORS := [
+	# TileSet.CELL_NEIGHBOR_BOTTOM_RIGHT_CORNER,
+	TileSet.CELL_NEIGHBOR_BOTTOM_SIDE,
+	# TileSet.CELL_NEIGHBOR_BOTTOM_LEFT_CORNER,
+	TileSet.CELL_NEIGHBOR_LEFT_SIDE,
+	# TileSet.CELL_NEIGHBOR_TOP_LEFT_CORNER,
+	TileSet.CELL_NEIGHBOR_TOP_SIDE,
+	# TileSet.CELL_NEIGHBOR_TOP_RIGHT_CORNER,
+	TileSet.CELL_NEIGHBOR_RIGHT_SIDE,
```
manuq: Let's remove the commented lines here.

scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
```diff
@@ -0,0 +1,104 @@
...
+	var current_agent_position: Vector2 = global_position
+	var next_path_position: Vector2 = navigation_agent.get_next_path_position()
+	var running := navigation_agent.distance_to_target() > (64 * 3)
+	var speed := run_speed if running else walk_speed
+	# TODO: smoothly change between running & walking speed?
+	velocity = current_agent_position.direction_to(next_path_position) * speed
+	move_and_slide()
+
+
+func _process(_delta: float) -> void:
```
manuq, line 70: Thanks for starting this decouple between `_physics_process` and `_process`!

wjt:
> The running level feels great. I only managed to beat the level by running, which I think is expected. Maybe making the character always running is better for this case.

The enemy movement is set up so that it's possible to beat the level without running. (I have tested this! Though it's very hard, one slight error and you lose.) That's why it matches the player's walking pace when within a certain range, and matches the player's running pace when outside it. The problem I was trying to solve is: if the monster moves at a constant speed, then either the level is impossible without running, or the level is way too easy if you run.

Some tuning may be needed here! Perhaps by making the path more winding we can give the enemy more of a chance to catch up – it can float across the void – or perhaps it's OK if it's too easy if you run, because that gives us more opportunity to put extra collectible items into the level in future, or just give the player a little more breathing room.

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
wjt: Ah yes, I remember typoing this but at the time I expected to come up with a better name before merging...!

wjt: Maybe this name is OK. The obvious other choice was "Void" but I wanted to make it generic enough that a learner could make a level with e.g. a slime mold.

manuq: Yes TileMapCover is fine, I would just remove the TODO comment.

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
wjt: Good point!

scenes/quests/lore_quests/quest_002/1_void_runner/components/tile_map_cover.gd
wjt: Done, and also restored it when restoring the world (that was a head-scratcher for a few minutes!)

PR #1086, opened 2025-08-15
SceneSwitcher: Respond to manual hash changes

wjt:
Previously, if you manually edited the URL hash in the browser address bar and pressed Enter, nothing would happen. This made it harder to manually construct the URL for a scene: you received no feedback as to whether you constructed the URL correctly.

We can monitor the window's hashchange event to be notified when the URL hash changes. However, this event is also emitted when the hash is changed by this code, so we need to take care not to enter an infinite loop in that case.

Store the URL that we most recently set. If we observe the URL changing to something different, then this must be user-initiated; switch to that scene.

### DIFF
--- a/scenes/globals/scene_switcher/scene_switcher.gd
+++ b/scenes/globals/scene_switcher/scene_switcher.gd
@@ -11,6 +11,13 @@ const _SCENE_SUFFIX = ".tscn"
 ## Proxy object for the 'window' DOM object, or null if not running on the web
 var _window: JavaScriptObject
 
+## Proxy object for the [method _on_hash_changed] callback, or null if not running on the web
+var _on_hash_changed_ref: JavaScriptObject
+
+## The last URL that was set by [method _set_hash], if running on the web.
+## If we observe the URL changing to something different, the user has edited the URL manually.
+var _current_url: String
+
 ## Matches the expected absolute path for a scene, with a capture group
 ## representing a more human-readable substring.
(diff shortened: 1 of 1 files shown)


wjt:
Test this by opening this link in your browser:

https://endlessm.github.io/threadbare/branches/endlessm/scene-switcher-respond-to-manual-hash-changes/

Then click the address bar, delete any `#foo/bar`, type/paste `#world_map/frays_end` or `#uid://d1o1hx0x80d5k` at the end of the URL, and press Enter.

scenes/globals/scene_switcher/scene_switcher.gd
```diff
@@ -69,9 +78,17 @@ func _set_hash(resource_path: String) -> void:
 		# Replace the current URL rather than simply updating window.location to
 		# avoid creating misleading history entries that don't work if you press
 		# the browser's back button.
+		_current_url = url.href
 		_window.location.replace(url.href)
 
 
+func _hashchange_cb(args: Array) -> void:
```
manuq: I know that "cb" stands for "callback" as it comes from `JavaScriptBridge.create_callback(...)`. I wonder if that will be obvious. Can we make it explicit? Either `_hashchange_callback` or `_on_hash_changed` ?

wjt: Good point. I renamed it to `_on_hash_changed`.

Issue #1089, opened 2025-08-18
Discord merge notification action failed


wjt:
### Description

When #1087 was merged, the action which is supposed to post a notification to a Discord channel failed:

https://github.com/endlessm/threadbare/actions/runs/17043897279

```
Env var DISCORD_WEBHOOK is not defined. Maybe try to set it if you are running the script manually.
```

If I had to guess, I would guess that the difference is that the PR is from a fork, and the necessary secrets are not supplied to the action as a result, because the `pull_request:` trigger runs in the context of the source branch. By contrast `pull_request_target:` runs in the context of the target branch.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

wjt:
As seen in https://github.com/endlessm/threadbare/actions/runs/17154171924/job/48666996819, merging a PR from a fork runs the action with the secret.

But nothing is seen in threadbare-merges on Discord, not for PRs from forks nor for PRs from us.

That's consistent with this bug report against the action we're using:

- https://github.com/Ilshidur/action-discord/issues/82

PR #1093, opened 2025-08-19
ci: Run Discord merge hook in target branch context

wjt:
The pull_request trigger runs the workflow in the context of the source branch, not the target branch. This means that, if the pull request is from a fork, the workflow does not have access to the target repository's secrets.

This is for security reasons: otherwise, a malicious contributor could change the pull_request-triggered workflow to exfiltrate the secrets. However, in this case we really need the secret to be present, or else the hook cannot work.

Run this workflow from the pull_request_target trigger, which runs in the context of the target ref. Having read the dire warnings in the [pull_request_target documentation][0], I read the source code of the action we're using and confirmed that it does what it says it does and will not download untrusted external code or steal our secrets.

[0]: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

I believe this will fix https://github.com/endlessm/threadbare/issues/1089 but it's impossible to test without first merging this change, then testing what happens when we merge a PR from a fork and a PR from this repo.

### DIFF
--- a/.github/workflows/discord.yml
+++ b/.github/workflows/discord.yml
@@ -3,7 +3,7 @@
 name: Discord Notification
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - closed
     branches:


manuq (approved):
Nice, let's see how next pull request from a fork looks like with the Discord hook.

PR #1106, opened 2025-08-21
scene_switcher: Allow relative paths without res:// prefix

wjt:
Previously, Fray's End could be reached from any of the following URL hashes:

- `#world_map/frays_end`, which is what the game sets
- `#res://scenes/world_map/frays_end.tscn`, the full canonical path to the scene, which is perhaps the easiest to copy in the editor's FileSystem view
- `#world_map/frays_end.tscn`, which I don't imagine anyone will ever use
- `#res://scenes/world_map/frays_end`, ditto
- `#uid://cufkthb25mpxy`, which I use but I don't think anyone else does

However `scenes/world_map/frays_end` and
`scenes/world_map/frays_end.tscn` were not previously accepted: either you had to include the full `res://scenes/` prefix, or omit it all. But `scenes/world_map/frays_end.tscn` is the easiest thing to copy from the command-line or from GitHub.

Accept paths that begin with `scenes/`, prefixing just `res://` in that case.

### DIFF
--- a/scenes/globals/scene_switcher/scene_switcher.gd
+++ b/scenes/globals/scene_switcher/scene_switcher.gd
@@ -2,10 +2,14 @@
 # SPDX-License-Identifier: MPL-2.0
 extends Node
 
-## Prefix to strip from scene path when setting URL hash
-const _SCENE_PREFIX = "res://scenes/"
+## Prefixes to try adding to non-absolute path in URL hash, which may have been stripped to make it
+## more human-readable.
+const _SCENE_PREFIXES = [
+	"res://",
+	"res://scenes/",
+]
 
-## Suffix to strip from scene path when setting URL hash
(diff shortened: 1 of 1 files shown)


scenes/globals/scene_switcher/scene_switcher.gd
```diff
@@ -2,10 +2,14 @@
...
 extends Node
 
-## Prefix to strip from scene path when setting URL hash
-const _SCENE_PREFIX = "res://scenes/"
-
-## Suffix to strip from scene path when setting URL hash
+## Prefixes to try adding to non-absolute path in URL hash, which may have been stripped to make it
+## more human-readable.
+const _SCENE_PREFIXES = [
+	"res://",
```
manuq, line 8: Yes, much easier to copy and paste!

manuq (approved):

Issue #671, opened 2025-05-20
Add sound effect to the "hushrooms"


manuq:
There are mushrooms in Fray's End that dissapear when the player gets near, and appear when the player gets far. The idea is to add a "hush" sound to them. It could be two sounds, one for hiding and one for revealing. Either record the sound yourself using eg. Audacity, or find a good match with a free license (for example in freesounds.org).

PlayMatters:
Posted the Hush sound effect as .mp3 in Discord.

ValePS81:
Me gustaría trabajar en esto.

PR #1111, opened 2025-08-21
Add sound effect to mushrooms

ValePS81:
#671

https://valeps81.github.io/threadbare/


scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
```diff
@@ -4,10 +4,12 @@
 extends Node2D
 
 @onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
-
+@onready var hide_player: AudioStreamPlayer2D = $AudioStreamPlayer2D_Hide
+@onready var reveal_player: AudioStreamPlayer2D = $AudioStreamPlayer2D_Reveal
```
wjt: The linter is failing because it wants two blank lines before `func _hide()`.


```suggestion
@onready var hide_player: AudioStreamPlayer2D = $AudioStreamPlayer2D_Hide
@onready var reveal_player: AudioStreamPlayer2D = $AudioStreamPlayer2D_Reveal


```

scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.tscn
```diff
@@ -1,9 +1,11 @@
-[gd_scene load_steps=21 format=3 uid="uid://bbsc3wssndtfe"]
+[gd_scene load_steps=23 format=3 uid="uid://bbsc3wssndtfe"]
 
 [ext_resource type="Script" uid="uid://dnb4ys8yvi1pj" path="res://scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd" id="1_v208a"]
 [ext_resource type="Texture2D" uid="uid://5s56lmeleerc" path="res://scenes/game_elements/props/decoration/hiding_mushroom/components/Mushroom_Hiding.png" id="2_u5x8h"]
 [ext_resource type="Texture2D" uid="uid://caajyl18koevp" path="res://scenes/game_elements/props/decoration/hiding_mushroom/components/Mushroom_Idle.png" id="3_gyonr"]
 [ext_resource type="Texture2D" uid="uid://b3jrnsvk46nji" path="res://scenes/game_elements/props/decoration/hiding_mushroom/components/Mushroom_Revealing.png" id="4_nlg6p"]
+[ext_resource type="AudioStream" uid="uid://dgwytqo5nvymw" path="res://assets/first_party/sounds/mushroom/Hu.mp3" id="5_gyonr"]
```
wjt: I like this filename!

scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.tscn
```diff
@@ -140,5 +142,13 @@ collision_layer = 4
 [node name="CollisionShape2D" type="CollisionShape2D" parent="PlayerDetector"]
 shape = SubResource("CircleShape2D_v208a")
 
+[node name="AudioStreamPlayer2D_Hide" type="AudioStreamPlayer2D" parent="."]
+stream = ExtResource("5_gyonr")
+volume_db = 2.0
+
+[node name="AudioStreamPlayer2D_Reveal" type="AudioStreamPlayer2D" parent="."]
```
wjt: I don't think you need the type of the node in its name. I think you could rename these to `HideSound` and `RevealSound` (and update the code accordingly).

scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.tscn
```diff
@@ -140,5 +142,13 @@ collision_layer = 4
 [node name="CollisionShape2D" type="CollisionShape2D" parent="PlayerDetector"]
 shape = SubResource("CircleShape2D_v208a")
 
+[node name="AudioStreamPlayer2D_Hide" type="AudioStreamPlayer2D" parent="."]
+stream = ExtResource("5_gyonr")
+volume_db = 2.0
```
wjt: Testing your change, I can see that what you're trying to adjust for is that the "hu" sound is much much quieter than the "sh" sound. I don't think this 3 dB adjustment (+2 here, -1 for the "sh") is quite enough. In my testing when I increase this to `+8 dB` they sound about right next to each other.

If you prefer, another option would be to edit the sound effects in a tool like Audacity and use the "Normalize" tool to adjust each one so their peak volume is the same as each other. That way you wouldn't need to adjust the level here.

scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.tscn
```diff
@@ -140,5 +142,13 @@ collision_layer = 4
 [node name="CollisionShape2D" type="CollisionShape2D" parent="PlayerDetector"]
 shape = SubResource("CircleShape2D_v208a")
 
+[node name="AudioStreamPlayer2D_Hide" type="AudioStreamPlayer2D" parent="."]
```
wjt: Both sound effects should have the "Bus" property set to "SFX" (an abbreviation of "sound effects"). Currently they are assigned to the "Master" bus, which means that the in-game volume control for sound effects doesn't affect their volume.

To test if this is set correctly, run the game, press Escape or Pause on your keyboard or Start/Menu on a controller, go to the options, and adjust the Sound Effects and Music volumes both to zero. Then walk past some mushrooms - they should be silenced.

Issue #993, opened 2025-08-06
Player can collide with invisible thread


manuq:
### Description

Notably, the collectible thread at round 3 of ink combat is in front of the player initial position:

https://github.com/user-attachments/assets/2a289cfe-753d-435b-99e5-17fd7c995fea

### XP Summary (total 225)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |
### Checklist

- [x] I've reviewed and agree to follow the code of conduct

wjt:
One way to fix this would be to add and remove the `player` layer from the `CollisionObject2D`'s `collision_layer` mask based on the CollectibleItem's `revealed` property, similar to how the `Sprite2D`'s `visible` property is changed.

JoaquinPando:
Me gustaria trabajar en esto

JoaquinPando:
Hi mentors, here is the planning sheet for the issue. I'll be sending the pull request shortly:
https://docs.google.com/document/d/1NmgYR8kpb-GmgsOzUrpKsgK4uWyPP8sVqr1VEqlyHfs/edit?usp=sharing

PR #1118, opened 2025-08-22
Collectible: Sync physical collision with 'revealed' state

JoaquinPando:
The collectible's StaticBody2D remained active while the item was hidden, incorrectly blocking player movement. This change modifies the _update_based_on_revealed function to also toggle the 'disabled' property of the physical CollisionShape2D. This ensures the physical collision is only active when the item is visible and interactable.

Thanks a lot for the feedback, Will — it really helped me! I hope I got it right this time.

Fixes https://github.com/endlessm/threadbare/issues/993

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -46,7 +46,6 @@ editor/translations/update_pot_files_automatically=false
 
 window/size/viewport_width=1920
 window/size/viewport_height=1080
-window/size/mode=4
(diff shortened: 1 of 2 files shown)


manuq:
Thanks a lot @JoaquinPando! I have squashed the commits and merged, with a minor edit of the pull request title and description.

Issue #998, opened 2025-08-07
Camera bounds in music_puzzle extend past the dressed area


wjt:
### Description

In the lorequest's [music puzzle](https://endlessm.github.io/threadbare/branches/endlessm/main/#quests/lore_quests/quest_001/1_music_puzzle/music_puzzle), if you make the viewport really tall or wide you can see beyond the dressed area of the scene:

<img width="946" height="1420" alt="Image" src="https://github.com/user-attachments/assets/06a9ba21-e834-4cf3-b8ea-9ee680ed9d2c" />

<img width="1654" height="563" alt="Image" src="https://github.com/user-attachments/assets/50b6c462-3a85-49d5-b3bc-34f190f03ae8" />

This could probably be fixed for most aspect ratios by adjusting the `Camera2D` node in this scene, changing its left, top, and right limits. If necessary, a few more trees could be added. (You can use the `AreaFiller` node to fill in an area with trees - see `scenes/game_elements/props/area_filler/components/area_filler_test.tscn` for an example.)

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |
### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #1126, opened 2025-08-26
Limit camera in music puzzle

DonSeja06:
Adjust the Camera2D node in this scene updating the limits of the left, top, and right values.

Fixes https://github.com/endlessm/threadbare/issues/998


manuq (approved):
Excellent, this works very well. Next time, please use a better title for the pull request. For example "Limit camera in music puzzle". The pull request description is good but:
- It should use imperative mood. So instead of "I did this..." you should write "Do this...".
- The last line should be "Fix https://github.com/endlessm/threadbare/issues/998"

Check https://github.com/endlessm/threadbare/blob/main/docs/CONTRIBUTING.md#git-commit-messages-and-pull-request-descriptions

Since it's almost there I'm going to edit and merge this time!

PR #1133, opened 2025-09-01
Change game base resolution to 720p

manuq:
Downscale the viewport size by two thirds, from 1080p to 720p. Because this is
the resolution of handheld consoles like Steam Deck or Nintendo Switch.
    
Also: change the window stretch mode from Viewport to Canvas Items. This is not
a pixel-perfect game. Otherwise by just downscaling the viewport size, the
background shows shakiness while the player (and thus the camera) moves.
    
Change the zoom of Camera2D nodes in scenes depending on their content.

Resolves https://github.com/endlessm/threadbare/issues/638

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -45,9 +45,9 @@ editor/translations/update_pot_files_automatically=false
 
 [display]
 
-window/size/viewport_width=1920
-window/size/viewport_height=1080
-window/stretch/mode="viewport"
+window/size/viewport_width=1280
+window/size/viewport_height=720
+window/stretch/mode="canvas_items"
 window/stretch/aspect="expand"
 
 [editor_plugins]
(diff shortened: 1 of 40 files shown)


manuq:
A couple things spotted:
- The title scene overflows because the Threadbare image is too big and the text too.
- The Storybook overflows too, same thing.
- The text in dialogues should be smaller now. Same as the interaction label.

wjt:
I think we should land this and follow up with:

1. Adjusting the text size in the theme (which should fix most of those issues)
2. Fixing the title screen

wjt (approved):
LGTM. I pushed one simple fix to the aspect ratio debugger. The font & theme adjustments will be a larger task so let's defer that to a follow-up task. I'll land this and open some issues for the things you identified.

Issue #1138, opened 2025-09-02
Fix title screen for 720p base resolution


wjt:
<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/ef51cf48-075a-45b3-8bdc-7cdb4f33855e" />

Beside adjusting the UI theme (#1136), the logo has a custom minimum width of 1200. It could be that just reducing that proportionately is enough to make the title screen fit.

PR #1141, opened 2025-09-02
Main Menu: Fit 720p resolution

wjt:
The main menu had some hardcoded dimensions for the previous base resolution, 1920×1080, which meant that it did not fit 1280×720:

- The logo had a custom minimum width of 1200px, and a 128px margin to its left; 1328px > 1280px.
- The button box had 128px margins to its left and right.

Remove these. To my eye the result is OK, though it will look better with a smaller button font.

Fixes https://github.com/endlessm/threadbare/issues/1138


manuq (approved):
Yes removing these overrides and custom size the title screen is back to be usable!

PR #1144, opened 2025-09-02
Walk behaviors followup

manuq:


scenes/game_logic/base_character_behavior.gd
```diff
@@ -0,0 +1,27 @@
...
+
+func _exit_tree() -> void:
+	update_configuration_warnings()
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+	if get_parent() is not CharacterBody2D:
+		warnings.append("Parent is not a CharacterBody2D.")
+	return warnings
```
wjt: I believe you actually need to update `character` in these functions too. A node only becomes ready once, even if it is subsequently moved in the tree.


```suggestion
var character: CharacterBody2D


func _enter_tree() -> void:
	character = get_parent() as CharacterBody2D
	update_configuration_warnings()


func _exit_tree() -> void:
	character = null
	update_configuration_warnings()


func _get_configuration_warnings() -> PackedStringArray:
	var warnings: PackedStringArray
	if not character:
		warnings.append("Parent is not a CharacterBody2D.")
	return warnings
```

According to https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#casting, if I'm reading it correctly, `get_parent() as CharacterBody2D` should return `null` if the parent is not a `CharacterBody2D`. But I haven't tested this.

wjt (changes requested):

scenes/game_logic/base_character_behavior.gd
manuq: Good point. I saw you did it like this in the `area_filler.gd`. My line of thought was: the warning is supposed to just work in the editor, then in the game the parent has to be a CharacterBody2D and it's fine if it raises an error when not. But your way will cover the cases when the node is moved at runtime using `reparent()` or `remove_child()`, `add_child()`.

Now I hope I'm not overthinking but, what about?
- We export the `character` property. So it could be customized to something different than the parent node.
- We set it to the parent node when it enters the tree, only if it's unset (null) and the parent is of the matching class.
- We warn if the property isn't set.

This has the advantage that if you move the behavior node around in the scene tree, and `character` was previously set to be the parent, it is remembered (the editor automatically updates the node paths).

I did it in a fixup commit because I'm unsure. What do you think?

wjt: Interesting idea! I'll look…

scenes/game_elements/characters/components/gym_area_test.tscn
```diff
@@ -8,16 +8,16 @@
 [ext_resource type="PackedScene" uid="uid://bbsc3wssndtfe" path="res://scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.tscn" id="6_l45sp"]
 [ext_resource type="PackedScene" uid="uid://dv4f232y8w8dv" path="res://scenes/game_elements/props/decoration/water_rock/water_rock.tscn" id="7_1aydi"]
 [ext_resource type="SpriteFrames" uid="uid://b3r84ksew5djp" path="res://scenes/game_elements/characters/enemies/throwing_enemy/components/ink_drinker_frames_purple.tres" id="9_6nfeg"]
-[ext_resource type="Script" uid="uid://cx0sk4uhvnii0" path="res://scenes/game_logic/erratic_walk_behavior.gd" id="10_k43ew"]
```
wjt: ugh, why does godot do this sometimes...

scenes/game_logic/base_character_behavior.gd
manuq: > Interesting idea! I'll look…

Keep in mind that I need to update the message of the commit fixed up. Either this or apply your suggestion!

scenes/game_elements/characters/components/gym_area_test.tscn
manuq: > ugh, why does godot do this sometimes...

Yes first I partially added the changes in this scene with `git add --patch`. But I think Godot does it because the references `metadata/_custom_type_script = "uid://cx0sk4uhvnii0"` have been removed in my fixup. So I added it all.

manuq: No I am wrong. I moved a node in the scene, saved the scene, and the UIDs are back...

Issue #1151, opened 2025-09-04
Use InputWalkBehavior for player


wjt:
commit 6af3a371f5f9c2654593c4bd65f61a5906b78876 added an `InputWalkBehavior` node which is based on the walking logic in `player.gd`. We should switch the player to use this behavior and remove the duplicated code.

manuq:
> Similar to [#1122](https://github.com/endlessm/threadbare/issues/1122), I wonder if we should factor these out into 1 or more resources. For instance we could have a CharacterWalkSpeeds resource which contains the 4 from `input_walk_behavior.gd`; the 3 extra properties would be ignored on the other walking behaviours for now, but you could imagine making the follow_walk_behavior switch between walking and running based on how far away the target is.

This is a great idea! The player (and the input walk behavior) also have "stopping_step" and "moving_step" to smooth the transition between moving and stopped, and I think the other walk behaviors would benefit from them.

wjt:
The next question here is: how do we handle playing the right animation on the sprite when there are more states than idle/walk/run? https://github.com/endlessm/threadbare/discussions/774#discussioncomment-14401610

PR #1153, opened 2025-09-04
Factor out a "speeds" resource for walk behaviors

wjt:
Previously, each behavior defined a `walk_speed` property, with the input
behavior also defining `run_speed` and parameters controlling acceleration and
deceleration. It seems likely that we will in due course want the
non-input-controlled walk behaviors to have these other parameters, to give them
more natural movement.

Define a new CharacterSpeeds resource that holds the 4 parameters from
InputWalkBehavior. Adjust all walk behaviors to have a `speeds` property.

The property defaults to `null`; it is set to an instance with the default
values on ready in that case, but only when running in-game. This is because we
do not want to persist the default values in each scene where this node is used,
only if they have been changed; and there is no way in GDScript to define a
constant instance of CharacterSpeeds. (If we defined a
default_character_speeds.tres and set this as the default, then it would be very
easy to accidentally edit the global defaults.)

See discussion at
#1151 (comment).

I also fixed some type warnings, and grouped the walk behaviors together in a directory of their own.


scenes/game_logic/walk_behaviors/bounce_walk_behavior.gd
```diff
@@ -10,8 +10,9 @@ extends BaseCharacterBehavior
 ## Emitted when [member direction] is updated.
 signal direction_changed
 
-## The character walking speed.
-@export_range(10, 1000, 10, "or_greater", "suffix:m/s") var walk_speed: float = 300.0
+## Parameters controlling the speed at which this character walks. If unset, the default values of
+## [CharacterSpeeds] are used.
+@export var speeds: CharacterSpeeds
 
 ## The speed to consider that the character is stuck.
```
wjt: I wonder about also including this property in the resource. It's not used in the InputWalkBehavior but it is present in all the others; and you could imagine it being useful to change your player-character's animation in the case when you're walking straight into a wall?

manuq: Yes I think it makes sense to have the stuck speed in `CharacterSpeeds`.

manuq (approved):
This is awesome! Yes please also add the stuck speed in the new resource.

scenes/game_logic/walk_behaviors/character_speeds.gd
```diff
@@ -0,0 +1,41 @@
...
+## How fast does the player transition from walking/running to stopped.
+## A low value will make the character look as slipping on ice.
+## A high value will stop the character immediately.
+@export_range(10, 4000, 10, "or_greater", "suffix:m/s²") var stopping_step: float = 1500.0
+
+## How fast does the player transition from stopped to walking/running.
+@export_range(10, 4000, 10, "or_greater", "suffix:m/s²") var moving_step: float = 4000.0
+
+## The speed below which the character is considered stuck.
+## [br][br]
```
manuq, line 28: I like the `[br][br]` in the empty line, 👍

scenes/game_logic/walk_behaviors/erratic_walk_behavior.gd
```diff
@@ -57,20 +55,23 @@ func _ready() -> void:
...
 func _physics_process(delta: float) -> void:
 	if not direction:
 		_update_direction()
 
-	character.velocity = character.velocity.lerp(direction * walk_speed, direction_weight)
+	character.velocity = character.velocity.lerp(direction * speeds.walk_speed, direction_weight)
 	var collided := character.move_and_slide()
 	if collided and character.is_on_wall():
-		if character.get_real_velocity().length_squared() <= stuck_speed * stuck_speed:
+		if speeds.is_stuck(character):
```
manuq, line 69: 👍

PR #1162, opened 2025-09-09
Change font to Jersey

### LINKED ISSUE #201: Replace font for the game

wjt:
This is a pixel font designed for knitting. As well as looking good and fitting
the game thematically, it has better glyph coverage than our current font, and
has a greater variety of sizes available.

Fixes #201


wjt:
| Before | After |
| ------ | ----- |
| ? | image |

PR #1168, opened 2025-09-10
Update theme for 720p resolution

### LINKED ISSUE #1136: Adjust theme font & box sizes for 720p base resolution
Since changing the base resolution to 720p #1133, the main menu, dialogue balloons, and other UI elements are way too big.

Most of these are controlled by the shared theme at `res://scenes/ui_elements/components/theme.tres`.

I suggest that for starters we halve the default font size, from 36px to 18px. The font we are using is [m6x11plus](https://managore.itch.io/m6x11) which should be used at sizes 18, 36, 54, …. The current uses of 36px should become 18px. The current much larger fonts used in some places could be changed to 36?

We'll need to adjust the sizing of the various PanelContainers so that the text remains vertically centred.

It could be that we'll need to switch to a different font (#201) to make this look good!

### LINKED ISSUE #1142: Credits screen doesn't fit 720p
### Description

Image

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |

wjt:
This adjusts the font sizes used in the game (building on #1162 which gives us more font sizes to use), and adjusts the margins, separation, and sizes of various elements to better fit the UI in the reduced resolution.

All screenshots taken fullscreen on a 3840x2160 monitor for fairness: this common hidpi display resolution is precisely 200% larger than 1080p and 300% larger than 720p, so all screenshots have integer scaling.

| 0.1.3 release (1080p) | main (720p) | this branch |
| -- | -- | -- |
| image | image | image |
| image | image | image |

Fixes #1136
Fixes #1142 

---
  - https://app.asana.com/0/0/1211211871317294

### DIFF
--- a/scenes/ui_elements/dialogue/balloon.tscn
+++ b/scenes/ui_elements/dialogue/balloon.tscn
@@ -20,11 +20,11 @@ theme = ExtResource("2_pfnde")
 
 [node name="PanelContainer" type="PanelContainer" parent="Balloon"]
 layout_mode = 0
-offset_right = 384.0
-offset_bottom = 384.0
+offset_right = 480.0
+offset_bottom = 359.0
 
 [node name="VBoxContainer" type="VBoxContainer" parent="Balloon/PanelContainer"]
-custom_minimum_size = Vector2(384, 128)
+custom_minimum_size = Vector2(352, 128)
 layout_mode = 2
 
 [node name="CharacterPanel" type="PanelContainer" parent="Balloon/PanelContainer/VBoxContainer"]
@@ -57,6 +57,7 @@ text = "next >"
 
 [node name="ResponsesMenu" type="VBoxContainer" parent="Balloon/PanelContainer/VBoxContainer" node_paths=PackedStringArray("response_template")]
 unique_name_in_owner = true
+visible = false
 layout_mode = 2
 size_flags_vertical = 8
 theme_override_constants/separation = 2


scenes/ui_elements/components/theme.tres
```diff
@@ -118,9 +119,9 @@ axis_stretch_horizontal = 2
 axis_stretch_vertical = 2
 
 [sub_resource type="StyleBoxTexture" id="StyleBoxTexture_mu1ca"]
-content_margin_left = 96.0
+content_margin_left = 64.0
 content_margin_top = 64.0
-content_margin_right = 96.0
+content_margin_right = 64.0
```
manuq: I agree that there is less space to spare in 720p! But at the same time I find a bit odd that the balloon is sparing space on the left of the texture showing the background behind instead of having the solid paper background for better margin and readability. The text is too close to the border. Ideally we should expand the texture to the left:

Image

I gave it a try, I thought it was impossible without editing the asset in an external program because the texture comes in 64x64 tiles. But there is an "Expand Margin" option. Setting left to 16 px looks much better to me. And right too to keep things centered. Before:

Image

After:

Image

Now this also enlarged the texture to the right, we can compensate by reducing the width in the custom minimum size by 16 px:

Image

I think this looks better. Here is a commit for reference, in case you agree: https://github.com/endlessm/threadbare/commit/036157a6dbba97037c8bfd8593b468507809af56

I checked the credits and the intro dialogues and didn't saw anything overflow because of the 16 px horizontal reduction. What do you think?

wjt: I think this is much better! I didn't know this was supported. I'll look at your patch.

wjt: I amended it to apply the same to the top and bottom margins; and to adjust the default size of the balloon so that it does not need to resize unless there are more than 3 lines of dialogue, or 2 or more responses.

manuq: > I amended it to apply the same to the top and bottom margins; and to adjust the default size of the balloon so that it does not need to resize unless there are more than 3 lines of dialogue, or 2 or more responses.

I was about to suggest doing the same in the vertical axis, you read my mind!

Issue #771, opened 2025-06-12
Record which quests have been played to completion


wjt:
Currently there is no record of which quests the player have completed. We would like to store this information, so that in future we can show it in the storybook (and potentially change Fray's End based on how many quests the player has completed).

This information should be stored in the [game state](https://github.com/endlessm/threadbare/blob/main/scenes/globals/game_state/game_state.gd), alongside the "quest currently being played" information, and should be cleared if you choose "New Game" on the main menu.

We need to store which specific quests have been played, not just how many; and we do not need to record whether a quest has been played once versus more than once.

(Presenting this information in the storybook is not part of this task.)

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

manuq:
> > [@AlexanderV24](https://github.com/AlexanderV24)Hola, ¿cómo va todo? ¿Necesitas ayuda con la tarea? ¿Entiendes lo que hay que hacer?
> 
> Hi, I have a question. How would I know when a quest is considered completed? In which event would it be? What I had in mind is that when the 3 items are collected, it gets marked as completed. Would that idea be okay?

Maybe it should be after the 3 items are sent to the loom in the Fray's End scene? Because that's when another quest can be started.

PR #1176, opened 2025-09-11
Record which quests have been played to completion

AlexanderV24:
Add a new field to the game state which holds an array of the paths of completed quests. Update it when the player takes threads to the loom. Clear it when the state is cleared.

Resolves https://github.com/endlessm/threadbare/issues/771

---

I tested in a new scene simulating the collection and delivery of the objects. I deleted that and didn't add it to the changes because it was to save time during testing.

https://docs.google.com/document/d/1wQst5g8896AjaVjZsPAC9E2uumBQm0mpBo9vJWxlDOA/edit?tab=t.0

### DIFF
--- a/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
+++ b/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
@@ -36,6 +36,7 @@ func _ready() -> void:
 		GameState.set_incorporating_threads(false)
 
 
+
 func _on_interaction_ended() -> void:
 	if _have_threads:
 		# Hide interact label during scene transition
@@ -48,6 +49,9 @@ func _on_interaction_ended() -> void:
 func on_offering_succeeded() -> void:
 	loom_offering_animation_player.play(&"loom_offering")
 	await loom_offering_animation_player.animation_finished
+	#Mark mission completed
+	var quest_path: String = GameState._state.get_value(GameState.QUEST_SECTION, GameState.QUEST_PATH_KEY, "")
+	GameState.add_completed_mission(quest_path)
 	GameState.clear_inventory()
 
 
--- a/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
+++ b/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
@@ -36,7 +36,6 @@ func _ready() -> void:
 		GameState.set_incorporating_threads(false)
 
 
-
 func _on_interaction_ended() -> void:
 	if _have_threads:
 		# Hide interact label during scene transition
@@ -49,10 +48,8 @@ func _on_interaction_ended() -> void:
 func on_offering_succeeded() -> void:
 	loom_offering_animation_player.play(&"loom_offering")
 	await loom_offering_animation_player.animation_finished
-	#Mark mission completed
-	var quest_path: String = GameState._state.get_value(GameState.QUEST_SECTION, GameState.QUEST_PATH_KEY, "")
-	GameState.add_completed_mission(quest_path)
 	GameState.clear_inventory()
+	GameState.mark_quest_completed()
 
 
 func is_item_offering_possible() -> bool:
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -167,4 +167,4 @@ func _save() -> void:
 		return
 	var err := _state.save(GAME_STATE_PATH)
 	if err != OK:
-		push_error("Failed to save settings to %s: %s" % [GAME_STATE_PATH, err])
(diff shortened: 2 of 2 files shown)


scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -48,6 +49,9 @@ func _on_interaction_ended() -> void:
 func on_offering_succeeded() -> void:
 	loom_offering_animation_player.play(&"loom_offering")
 	await loom_offering_animation_player.animation_finished
+	#Mark mission completed
+	var quest_path: String = GameState._state.get_value(GameState.QUEST_SECTION, GameState.QUEST_PATH_KEY, "")
+	GameState.add_completed_mission(quest_path)
```
wjt: This method does not exist in `game_state.gd`. Are you sure that you have committed all of your changes?

AlexanderV24: I'm checking, and yes, indeed, when I made the pull request, the last thing I did in game_state.gd wasn't saved. Do I need to make another pull request? Or what should I do?

manuq: @AlexanderV24 no need for a different pull request! Just `git add` the missing file and append a commit to this one.

scenes/globals/game_state/game_state.gd
```diff
@@ -21,6 +21,7 @@ const QUEST_SECTION := "quest"
 const QUEST_PATH_KEY := "resource_path"
 const QUEST_CURRENTSCENE_KEY := "current_scene"
 const QUEST_SPAWNPOINT_KEY := "current_spawn_point"
+const QUEST_COMPLETED_KEY := "completed_missions"
```
wjt: We call them "quests", not "missions". Please use "quests" throughout.

```suggestion
const QUEST_COMPLETED_KEY := "completed_quests"
```

scenes/globals/game_state/game_state.gd
```diff
@@ -34,6 +35,7 @@ const TRANSIENT_SCENES := [
 ## can be added to the loom.
 @export var inventory: Array[InventoryItem] = []
 @export var current_spawn_point: NodePath
+@export var completed_missions: Array[String] = []
```
wjt:
```suggestion
@export var completed_quests: Array[String] = []
```

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
wjt: The `_` prefix on the `_state` variable is a convention that means that the variable is private to the class that defines it. https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#functions-and-variables

In this case `_state` is private to `game_state.gd`, so accessing it directly from `eternal_loom.gd` is not good style.

I think you can simplify this API. You don't actually need the quest path here, so you can change this to:

```suggestion
	GameState.mark_quest_completed()
```

and adjust `game_state.gd` as follows.

scenes/globals/game_state/game_state.gd
```diff
@@ -87,6 +89,13 @@ func set_current_spawn_point(spawn_point: NodePath = ^"") -> void:
 	current_spawn_point = spawn_point
 	_state.set_value(QUEST_SECTION, QUEST_SPAWNPOINT_KEY, current_spawn_point)
 	_save()
+	
+
+func add_completed_mission(mission_name: String) -> void:
+	if mission_name not in completed_missions:
+		completed_missions.append(mission_name)
+		_state.set_value(QUEST_SECTION, QUEST_COMPLETED_KEY, completed_missions)
+		_save()
```
wjt:
```suggestion
func mark_quest_completed() -> void:
	var current_quest := _state.get_value(QUEST_SECTION, QUEST_PATH_KEY, null)
	if current_quest and current_quest not in completed_quests:
		completed_quests.append(current_quest)
		_state.set_value(QUEST_SECTION, QUEST_COMPLETED_KEY, completed_quests)
		_save()
```

scenes/globals/game_state/game_state.gd
```diff
@@ -167,4 +178,4 @@ func _save() -> void:
 		return
 	var err := _state.save(GAME_STATE_PATH)
 	if err != OK:
-		push_error("Failed to save settings to %s: %s" % [GAME_STATE_PATH, err])
+		push_error("Failed to save settings to %s : %s" % [GAME_STATE_PATH, err])
```
wjt: I think this was OK before:

```suggestion
		push_error("Failed to save settings to %s: %s" % [GAME_STATE_PATH, err])
```

wjt (approved):
I applied my suggestions.

PR #1177, opened 2025-09-11
Fix: dynamic on-screen control hints

### LINKED ISSUE #403: On-screen control hints assume keyboard

nico-noodle:
Previously, on-screen control hints always displayed keyboard icons, even when the player was using a gamepad. This confused testing, since players assumed only the keyboard was supported. Gamepad input was already functional, but hints did not update dynamically.

This change integrates the godot_input_helper addon and refactors input handling to centralize device detection. On-screen hint scenes now listen to InputHelper.device_changed and update automatically to show the correct icons for the last active device (keyboard, PlayStation, Xbox, Switch, etc.). The mapping API is used so that each action consistently displays the proper button symbol.

As part of this work:
- shaker.gd was updated to remove redundant _last_controller_id logic.
- input_key.gd was simplified to handle only gamepad input.
- Two new scripts were added:
   - keyboard_input.gd for handling keyboard hints.
   - interact_input.gd for managing the interaction button across both keyboard and gamepad.
- Hint scene scripts were extended to react to device changes and swap textures dynamically.
- New assets for the main device families were added and wired into the mapping system.

Success criteria
- Control hints reflect the last active input device (keyboard or gamepad).
- Switching between devices updates hints immediately.
- Input handling is centralized through InputHelper, with no duplication.
- No regressions in normal gameplay or input handling.

Fixes #403

### DIFF
--- a/scenes/world_map/frays_end.tscn
+++ b/scenes/world_map/frays_end.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=72 format=4 uid="uid://cufkthb25mpxy"]
+[gd_scene load_steps=74 format=4 uid="uid://cufkthb25mpxy"]
 
 [ext_resource type="TileSet" uid="uid://07fq3rspk8ia" path="res://scenes/tileset.tres" id="1_4evgk"]
 [ext_resource type="Script" uid="uid://8div00rgvsds" path="res://scenes/world_map/components/frays_end.gd" id="1_4gse6"]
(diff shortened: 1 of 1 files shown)


scenes/game_elements/fx/shaker/components/shaker.gd
```diff
@@ -67,8 +67,13 @@ func _ready() -> void:
...
 	noise.seed = randi()
 	started.emit()
-	if _last_controller_id >= 0:
-		Input.start_joy_vibration(_last_controller_id, 0.5, 0.5, time)
+	# Previously: we used _last_controller_id to trigger vibration.
+	# Now we rely on InputHelper.device_index, which is -1 if no controller is active.
+	# if _last_controller_id >= 0:
+	#     Input.start_joy_vibration(_last_controller_id, 0.5, 0.5, time)
+	if Engine.has_singleton("InputHelper") and InputHelper.device_index >= 0:
+		Input.start_joy_vibration(InputHelper.device_index, 0.5, 0.5, time)
```
manuq: `var _last_controller_id: int = -1` is still defined at the top of the script, please remove it.

The explanation in the comment is good! It seems better to go into the pull request description, because it describes how it worked before and why it changed. The commented code in the last 2 lines is definitely not needed, because we use Git for revisioning and the old implementation will be remembered in the Git history. Please:
- Copy and paste the "Previously ... Now ..." lines into the pull request description.
- Remove all 4 commented lines.

scenes/game_elements/props/hint/input_key/input_key.gd
```diff
@@ -3,10 +3,124 @@
...
+@export var steam_pressed_texture: Texture2D
+
+var current_device: String = ""
+var is_keyboard_mode: bool = true
 
 func _physics_process(_delta: float) -> void:
-	if Input.is_action_pressed(action_name):
-		modulate = Color.GRAY
+	if is_keyboard_mode:
+		return  # El nodo de mando no hace nada en modo teclado
```
manuq: We prefer the comments to be in a separate line. And it should be in English:

```suggestion
		# The command node doesn't do anything in keyboard mode:
		return
```

I left a suggestion with a literal translation, but I'm not sure what did you mean with "El nodo de mando".

PR #1178, opened 2025-09-12
Factor NavigationFollowWalkBehavior out of void_spreading_enemy.gd

### LINKED ISSUE #1175: Make void-spreading enemy use a WalkBehavior
Currently `void_spreading_enemy.gd` contains both movement logic (following another node using a NavigationAgent2D) and world-erasing logic.

I would like to try to move the movement/navigation to our nascent library of WalkBehaviors. I imagine it would enter the “stuck” state when the target is unreachable.

wjt:
Previously, void_spreading_enemy.gd contained movement logic somewhat
entwined with the other logic for this entity. But the movement logic
was relatively generic – follow a target using a NavigationAgent2D – and
you could imagine a different flavour of this entity that moves in a
different way, perhaps following a path.

Create NavigationFollowWalkBehavior based on the logic in
void_spreading_enemy.gd, with similar functionality to
FollowWalkBehavior but using a NavigationAgent2D rather than
straight-line pathfinding.

As with the previous implementation, if the distance to the target is
above a configurable threshold, the character runs; otherwise, they
walk. Like InputWalkBehavior, a running_changed signal is emitted, which
can be connected to CharacterSpriteBehavior to control animations,
though this is not used in the void runner scene.

Signals are emitted when navigation ends (either indicating that the
target has been reached, or that it cannot be reached), though this is
also not used in the void runner scene.

I also added an example of this behavior to the gym.

Fixes #1175

### DIFF
--- a/scenes/game_logic/walk_behaviors/navigation_follow_walk_behavior.gd
+++ b/scenes/game_logic/walk_behaviors/navigation_follow_walk_behavior.gd
@@ -1,0 +1,161 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name NavigationFollowWalkBehavior
+extends BaseCharacterBehavior
+## @experimental
+##
+## Make the character follow a target, using a [NavigationAgent2D] to
+## route towards it.
+
+## Emitted when the character starts or stops running.
+signal running_changed(is_running: bool)
+
+## Emitted when [member target] becomes reached or not.
+signal target_reached_changed(is_reached: bool)
+
+## Emitted when [member is_target_reachable] changes.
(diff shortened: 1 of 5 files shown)


scenes/game_logic/walk_behaviors/navigation_follow_walk_behavior.gd
```diff
@@ -0,0 +1,161 @@
...
+## route towards it.
+
+## Emitted when the character starts or stops running.
+signal running_changed(is_running: bool)
+
+## Emitted when [member target] becomes reached or not.
+signal target_reached_changed(is_reached: bool)
+
+## Emitted when [member is_target_reachable] changes.
+signal target_unreachable_changed(is_reachable: bool)
```
manuq, line 18: Nice!

PR #1180, opened 2025-09-12
CharacterSpriteBehavior: Handle left- and front-facing sprites

wjt:
Previously, this behavior assumed that the animations it controls face right.

The majority of sprites used in the game do face right, but there are some exceptions: for example, the musician faces forward, and the knitwitch faces left.

Of course, neither of these sprites have a walk animation; and if we did want the KnitWitch to move, we could invert the asset. But you could imagine having a sprite that can move but should not be flipped based on its direction (perhaps a lightning storm).

Add a new property to CharacterSpriteBehavior, indicating which way the sprite faces. Use this property to not flip the sprite if it faces UNSPECIFIED; and invert the current flip_h logic if the resting direction is LEFT rather than RIGHT. Set the default to RIGHT, matching the convention for characters in the game which can move.

### DIFF
--- a/scenes/game_logic/character_sprite_behavior.gd
+++ b/scenes/game_logic/character_sprite_behavior.gd
@@ -12,7 +12,8 @@ extends Node2D
 @export var character: CharacterBody2D
 
 ## Whether to play the sprite animations or not. If not, the only thing that will happen is that
-## the sprite will be flipped horizontally according to the velocity of [member character].
+## the sprite will be flipped horizontally according to the velocity of [member character] and
+## the value of [member default_facing].
 ## Use this when using more advanced animation through an AnimationPlayer node.
 @export var play_animations: bool = true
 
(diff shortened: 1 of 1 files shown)


manuq:
> I wrote this because I initially used the KnitWitch as the placeholder sprite for the new behavior in #1178, and the KnitWitch faces left.
> 
> I switched to an InkDrinker when I realised that the KnitWitch doesn't have a walk animation, which caused a stream of warnings. But the patch is written, so here it is. We could decide not to use this and instead declare that all sprites that can walk should be drawn facing right, not left or forwards.

Yeah I think it will be less confusing if we use a convention: all our sprites should face to the right of the screen if they can be flipped. In the case of the musician and the KnitWitch I think we don't flip them. I have documented this convention in the visual style guide and at one point I asked PixelFrog to flip the StoryVores that were delivered.

PR #1184, opened 2025-09-14
AreaFiller: Support undo/redo

wjt:
It has often annoyed me that the "Refill" action cannot be undone with Ctrl-Z.

Integrate with EditorUndoRedoManager to make this possible.

In addition, change how the child nodes are instantiated so they are persisted to the scene in just the same way as if they were instantiated by hand, fixing an issue where loading and saving the scene later, without changes, would change the `.tscn`.

### DIFF
--- a/scenes/game_elements/props/area_filler/area_filler.gd
+++ b/scenes/game_elements/props/area_filler/area_filler.gd
@@ -49,6 +49,12 @@ func _exit_tree() -> void:
 	update_configuration_warnings()
 
 
+func _ready() -> void:
+	if not Engine.is_editor_hint():
+		self.queue_free()
+		return
+
+
 func _get_configuration_warnings() -> PackedStringArray:
 	var warnings: PackedStringArray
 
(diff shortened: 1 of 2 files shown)


scenes/game_elements/props/area_filler/area_filler.gd
```diff
@@ -100,7 +100,8 @@ func _generate_points() -> PackedVector2Array:
 
 
 func _prepare_child(pos: Vector2) -> Node2D:
-	var child: Node2D = scenes.pick_random().instantiate()
+	var scene: PackedScene = scenes.pick_random()
+	var child: Node2D = scene.instantiate(PackedScene.GenEditState.GEN_EDIT_STATE_INSTANCE)
```
wjt, line 104: I wrote https://willthompson.co.uk/notes/instancing-scenes-in-tool-script/ so I can remember this next time I need it.

I learnt how to do this from https://github.com/godotengine/godot-docs/issues/8801

manuq: This is news to me! Thanks for the blog post.

manuq (approved):
Excellent!

PR #1187, opened 2025-09-15
Add custom mouse pointer

wjt:
This uses an asset included in Tiny Swords.

It only overrides the default cursor shape: I believe we don't use any other
shapes in the game.


project.godot
```diff
@@ -49,6 +49,9 @@ window/size/viewport_width=1280
 window/size/viewport_height=720
 window/stretch/mode="canvas_items"
 window/stretch/aspect="expand"
+mouse_cursor/custom_image="uid://i20s4be8qvfl"
+mouse_cursor/custom_image_hotspot=Vector2(22, 17)
```
wjt: The visible cursor in this image is smaller than 32×32, but the texture is 64×64 with a lot of transparent padding. As [documented here](https://docs.godotengine.org/en/stable/classes/class_input.html#class-input-method-set-custom-mouse-cursor), on the web platform:

> Cursor images larger than 32×32 will also only be displayed if the mouse cursor image is entirely located within the page for [security reasons](https://chromestatus.com/feature/5825971391299584).

and this includes the transparent portion of the texture.

So if we are going to do this, we should add a copy of the texture which is ≤ 32×32 by removing the transparent padding.

wjt: Revised the PR to add a 32×32px copy of the image

Issue #376, opened 2025-04-29
Add “click to play” for web builds


wjt:
Due to web platform limitations, web pages cannot typically play audio until you have interacted with the page. (An interaction is a click or keystroke.) In particular this means that the sewing-machine sound doesn't reliably play during the splash-screen animation, and (when we add it) that the theme tune will not reliably play until the user presses a key or clicks something.

(If you're wondering why YouTube videos autoplay without needing a click, see https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Autoplay#autoplay_availability and the linked Chrome policy.)

It is possible to customise the HTML "shell" that the web build of a game uses. The documentation [gives](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Autoplay#autoplay_availability) adding a click-to-play button in order to trigger fullscreen as a reason to do this: similarly, fullscreen can only be triggered with a user interaction on the web.

The customization needed is roughly to replace `startGame()` with `init().then(() => { show a "click to play" button, await a click }).then(() => engine.start())`. Here are the [JavaScript API docs for the web engine](https://docs.godotengine.org/en/latest/tutorials/platform/web/html5_shell_classref.html).

wjt:
To go into a little more detail about how I would implement this feature:

First, copy the default HTML shell from Godot 4.4 https://github.com/godotengine/godot/blob/master/misc/dist/html/full-size.html into Threadbare's repo. I would put it into a new `web/` folder. Commit this.

Now, the section that needs to be customised is this bit:

https://github.com/godotengine/godot/blob/4.4/misc/dist/html/full-size.html#L200-L210

`engine.startGame` is a wrapper around `engine.init()` + `engine.start()`. The API documentation is at https://docs.godotengine.org/en/latest/tutorials/platform/web/html5_shell_classref.html but you may prefer to read its source code:

https://github.com/godotengine/godot/blob/4.4/platform/web/js/engine/engine.js#L196-L210

Replace the call to `engine.startGame` with the calls it makes internally, to `init()`, `preloadFile()`, then `start()`.

Now, change the HTML page to have a "click to play" button that is only displayed once `init()` and `preloadFile()` have finished. Change the code to only call `start()` when this button is pressed.

PR #1195, opened 2025-09-16
Initial workaround for autoplay restrictions in main scene (needs debugging)

daff2:
Integrated a "Click to start" button directly into the main scene to comply with browser autoplay restrictions. Adjusted the logic so that game start depends on user interaction. However, the scene currently only starts when pressing the ESC key; the button click event does not yet reliably trigger. The solution is partially implemented and needs further debugging to make the button a reliable trigger.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,15 @@ repos:
   rev: v5.0.0
   hooks:
   - id: check-added-large-files
-    args:
-      - --enforce-all
+	args:
+	  - --enforce-all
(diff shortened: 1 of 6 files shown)


wjt (changes requested):
I don't think this is the right approach. The "Click to play" should not be in the Godot game, but in the HTML + JavaScript web page that loads it. I outlined this approach in https://github.com/endlessm/threadbare/issues/376 and have now written some more detail in https://github.com/endlessm/threadbare/issues/376#issuecomment-3299381734

Issue #696, opened 2025-05-22
Make SequencePuzzle automatically detect SequencePuzzleSteps


wjt:
Currently SequencePuzzle has a `steps` property that is a list of (NodePaths of) SequencePuzzleStep nodes. Steps that aren't set here are ignored by the SequencePuzzle.

But by contrast, it detect any descendent node that is a SequencePuzzleObject:

https://github.com/endlessm/threadbare/blob/20132e0e0586638082839b711bf58cee8af7018e/scenes/game_logic/sequence_puzzle.gd#L53-L60

It would be perhaps less error-prone to do the same for SequencePuzzleStep.

This is a soft-breaking change in the sense that, in the edge case where someone has the SequencePuzzleStep nodes arranged in a different order to the `steps` property, then switching to using the scene hierarchy order would reorder the puzzle. That seems relatively unlikely to cause problems though.

### XP Summary (total 250)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

WhoD-S:
Buenas, me gustaria trabajar en este proyecto.

manuq:
@WhoD-S hi, how is this going? Do you need help with the task? Do you understand what needs to be done?

Issue #1163, opened 2025-09-09
godot-git-describe addon edits project.godot while running game in editor


wjt:
### Description

In #942 I added an addon which makes the output of `git describe` available to the game at runtime. What I hadn't quite realised is that it works by editing `project.godot` at the start of the export process, and reverting it at the end; and similarly, when you run the game in the editor, it edits `project.godot`, and reverts its changes when the launched game is closed.

This means that if you run `git commit -a` while the game is still running, you will accidentally commit the current revision's `git describe` output and commit hash to Git, which is not intended. I have seen this in a few pull requests, and caught myself almost doing this.

This is pretty annoying and error-prone. It would be nice to not do this.

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 150 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

PR #1196, opened 2025-09-16
Feat: Auto-detect SequencePuzzle steps

WhoD-S:
Previously, the SequencePuzzle's `steps` property had to be set in the editor to a list of `SequencePuzzleStep` nodes. This turned out to be error-prone: it was easy to add a new `SequencePuzzleStep` node but forget to add it to the `steps` property.

Instead, use a similar approach to how `SequencePuzzleObject` nodes are discovered: find all `SequencePuzzleStep` nodes that are children (or grandchildren, etc.) of the `SequencePuzzle` node. For backwards-compatibility, only do this if the `steps` property is empty.

Resolves: https://github.com/endlessm/threadbare/issues/696

### DIFF
--- a/scenes/game_logic/sequence_puzzle.gd
+++ b/scenes/game_logic/sequence_puzzle.gd
@@ -15,6 +15,9 @@ signal step_solved(step_index: int)
 ## The order in which the player must interact with objects to solve each step of the puzzle.
 ## If this is empty, [SequencePuzzleStep] nodes that are children (or grandchildren, etc.) of
 ## this node will be used, in depth-first order.
+##
+## @deprecated: Don't set this property directly: rely on the SequencePuzzleStep nodes being found
+##              automatically.
 @export var steps: Array[SequencePuzzleStep]
 
 ## If enabled, the [SequencePuzzleHintSign] for the current step of the puzzle


project.godot
```diff
@@ -8,13 +8,18 @@
...
+git_describe/commit_hash="e374f8b6e94f36948a2016c172b57fa68737c4db"
+
 [application]
 
 config/name="Threadbare"
 config/tags=PackedStringArray("2d", "endless", "game", "tilemap", "top_down")
 run/main_scene="uid://huuo8mnwsphv"
 config/features=PackedStringArray("4.4", "GL Compatibility")
 config/icon="uid://bc70phmq55dkf"
+config/git_describe="e374f8b"
```
wjt: These changes are due committing while the game is running – I filed #1163 about a way we could avoid this happening. For now, quit Godot, then use the following command to revert these changes:

```bash
git restore --source=HEAD~ project.godot
git add project.godot
git commit -m 'Revert unintended changes to project.godot'
```

wjt: `HEAD~` means "the commit before the current one".

scenes/game_logic/sequence_puzzle.gd
```diff
@@ -41,6 +41,12 @@ var _position: int = 0
 
 func _ready() -> void:
 	_find_objects()
+	
+	# If the steps array is empty, automatically discover the puzzle steps from the scene children.
+	if steps.is_empty():
+		for child in get_children():
+			if child is SequencePuzzleStep:
+				steps.append(child)
```
wjt: This only looks at the immediate children of the `SequencePuzzle` node. But look at the scene tree for the lorequest music puzzle:

Image

The steps are grandchildren, not children, of the `SequencePuzzle` node.

I would do something like this. Define the following recursive helper function:

```gdscript
func _find_steps(node: Node) -> void:
	if node is SequencePuzzleStep:
		steps.append(node)
	for child in get_children():
		_find_steps(child)
```

Then call it from `_ready`:


```suggestion
	# If the steps array is empty, find all steps that are within this node
	if steps.is_empty():
		_find_steps(self)
```

wjt:
I'll apply the changes I suggested.

PR #1201, opened 2025-09-16
Import HTML shell from export template

daff2:
This adds the HTML shell from the Godot 4.4.1 export template, without modification (except to add a trailing newline).

In future, this will be customised to require a click or keypress to start the game.

Helps #376


wjt (approved):
Thanks. I added the necessary project setting for this to be used by the web export, and the necessary license file to satisfy our automated checks.

PR #1165, opened 2025-09-09
Audio: Implement 18-note handbell scale for puzzles

JoaquinPando:
This PR implements the full 1.5-octave Nepalese handbell sound font. It replaces the old placeholder notes in the puzzles with this new, complete scale.

This provides a consistent sound and allows for greater melodic flexibility for future levels.

Fixes #1104

PR #1204, opened 2025-09-17
Remove unused xylophone samples

wjt:
Remove unused xylophone samples

These were replaced in every scene by commit
0a552f40763a5e8ced52ebaef4eb64b931dadb34, and are now unused.


wjt:
This is a follow-up to #1165 which (correctly) did not remove the old samples – there are several open pull requests for work-in-progress StoryQuests that use the old samples so I am glad that we left them in the repository. That's why this PR is in draft state.

But at some point it would be good to remove them. @JoaquinPando If you like, you could leave review on this PR and see if it looks correct to you or if I missed anything!

Issue #784, opened 2025-06-13
Add signal when collectible is collected(?)


wjt:
I wanted to design a level which had more than one collectible; so rather than having each collectible in a separate scene, you would move through the level, collecting one collectible at a time.

I wanted to make it so that it would be impossible to leave the first area of the level without collecting its collectible, by placing a door in the map and opening it when the collectible is collected.

But there is no signal emitted by collectible when it is collected.

----

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 150 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

Slayer2505:
@wjt @manuq 
I do believe a simple solution is posible to maintain the simplicity of the project, if you'd likie, I can look into possible solution that addresses everything mentioned earlier, including saving the player's position.

wjt:
If you want to investigate an improved state-saving solution which:

- Addresses #1254 
- Supports saving other level state like in https://github.com/endlessm/threadbare/issues/1208

then that's great, but I think we should keep the simple model that the collectible finishes the level.

PR #1210, opened 2025-09-17
Fix duplicate collectible issue with unique IDs

Slayer2505:
Este PR corrige un error que permitía recolectar el mismo elemento dos veces después de continuar el juego. 

Cambios realizados:

- Se agregó un identificador único para cada objeto coleccionable, permitiendo que el sistema de guardado rastree qué elementos ya fueron recopilados y evitando duplicados.
- Se creó una lista de IDs de elementos recopilados que se carga correctamente al iniciar o continuar el juego, garantizando que los elementos no reaparezcan físicamente si ya están en el inventario.
- Se modificó la lógica de interacción con el jugador para que, si un objeto ya fue recolectado, la transición de diálogos y escenas se ejecute automáticamente sin necesidad de acercarse al objeto.
- Se ajustó la persistencia del estado del juego para que funcione correctamente con la lista de IDs y se eviten errores al cargar datos guardados.

Resultado: El error de duplicación se corrige, el inventario mantiene consistencia, y la experiencia del jugador se mantiene fluida al continuar el juego.

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -3,34 +3,23 @@
 @tool
 class_name CollectibleItem extends Node2D
 
-## Overworld collectible that can be interacted with. When a player interacts
-## with it, an [InventoryItem] is added to the [Inventory]
-
-## Wether the collectible can be seen or collected. This allows the collectible
-## to be placed in the scene even when some condition has to be met for it to
-## appear.
 @export var revealed: bool = true:
 	set(new_value):
 		revealed = new_value
 		_update_based_on_revealed()
 
-## If provided, switch to this scene after collecting and possibly displaying a dialogue.
 @export_file("*.tscn") var next_scene: String
-
-## [InventoryItem] provided by this collectible when interacted with.
 @export var item: InventoryItem:
 	set = _set_item
 
 @export_category("Dialogue")
-
-## If provided, this dialogue will be displayed after the player collects this item.
 @export var collected_dialogue: DialogueResource:
 	set(new_value):
 		collected_dialogue = new_value
 		notify_property_list_changed()
 
-## The dialogue title from where [member collected_dialogue] will start.
 @export var dialogue_title: StringName = ""
+@export var unique_id: String = ""
 
 @onready var interact_area: InteractArea = $InteractArea
 @onready var animation_player: AnimationPlayer = $AnimationPlayer
@@ -38,31 +27,24 @@ class_name CollectibleItem extends Node2D
 @onready var appear_sound: AudioStreamPlayer = %AppearSound
 @onready var physical_collider: CollisionShape2D = $StaticBody2D/CollisionShape2D
 
-
 func _validate_property(property: Dictionary) -> void:
 	match property.name:
 		"dialogue_title":
 			if not collected_dialogue:
 				property.usage |= PROPERTY_USAGE_READ_ONLY
 
-
 func _get_configuration_warnings() -> PackedStringArray:
 	if not item:
 		return ["item property must be set"]
 	return []
 
-
 func _set_item(new_value: InventoryItem) -> void:
 	item = new_value
-
 	if sprite_2d:
 		sprite_2d.texture = item.get_world_texture() if item else null
-
 	if interact_area:
 		interact_area.action = "Collect " + item.type_name() if item else "Collect"
-
 	update_configuration_warnings()
-
 
 func _ready() -> void:
 	_set_item(item)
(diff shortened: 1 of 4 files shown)


scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -3,67 +3,49 @@
...
-
-## If provided, this dialogue will be displayed after the player collects this item.
 @export var collected_dialogue: DialogueResource:
 	set(new_value):
 		collected_dialogue = new_value
 		notify_property_list_changed()
 
-## The dialogue title from where [member collected_dialogue] will start.
 @export var dialogue_title: StringName = ""
+@export var unique_id: String = ""
```
manuq, line 22: Please identify each collectible in the current scene by their node path, instead of adding a new identifier.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -74,35 +56,42 @@ func _ready() -> void:
 
 	interact_area.interaction_started.connect(self._on_interacted)
 
+	# Si ya estaba recogido → lanzar flujo directamente
+	if unique_id != "" and GameState.has_collected(unique_id):
+		await _call_after_collected(null)
```
manuq, line 61: So if I understand correctly, this is supposed to happen if the item has `collected_dialogue` set:

1. Player collects the item.
2. Dialogue starts.
3. Player quits the game.
4. Player resumes the game.
5. Dialogue starts.

And that happens right away. I don't think that's right. Imagine the case of a scene with 2 revealed items, both with dialogue set (and no `next_scene` set). Player grabs both, then quits (after reading both dialogues). When resuming the game, both dialogues will try to start at the same time!

By design, progress within a scene is not supposed to be persisted. For example we don't save the player position in the scene. That is the bug depicted in https://github.com/endlessm/threadbare/issues/784#issuecomment-3266999783

So I like the idea of keeping track of which items have been collected in the current scene through an identifier like their node path. But instead of using that information for triggering their effects (display a dialogue, go to the next scene), I would use that information to reset them (remove them from the game state), if they happen to match a node in the current scene.

manuq: A better option, although harder to implement, is to not persist the items collected in the game state until the scene changes, but trigger their effects (display dialogue, show it in the HUD).

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -3,67 +3,49 @@
 @tool
 class_name CollectibleItem extends Node2D
 
-## Overworld collectible that can be interacted with. When a player interacts
-## with it, an [InventoryItem] is added to the [Inventory]
-
-## Wether the collectible can be seen or collected. This allows the collectible
-## to be placed in the scene even when some condition has to be met for it to
-## appear.
```
wjt, line 11: Why does this PR delete all the documentation comments?

scenes/game_elements/props/collectible_item/components/collectible_item.gd
wjt: We already use the fact that the path to a node in a scene is a unique identifier elsewhere in the game: that's how we save and restore the last Checkpoint (spawn point) that the player reached.

----

Another approach would be that, rather than each `CollectibleItem` instance defining its own `InventoryItem` as we do now, we would define `memory.tres`, `imagination.tres`, `spirit.tres` resources and reference them each time. These resources would have UIDs.

We'd have to figure out how that would interact with https://github.com/endlessm/threadbare/issues/1234 where the request is that a story might include two different Imagination threads.

I don't think we should introduce our own new identifiers.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
wjt: Yes, I think it's not correct to re-display the dialogue when respawning in the scene. Here's a simple scene I made on top of this branch:

Image

You can't get past the green thread without collecting it. (That works around the lack of a good way to trigger a change in the level, e.g. opening a door – which was the original subject of #784.)

Collecting the thread triggers a brief dialogue:

Image

Now I allow myself to be caught:

Image

When I respawn, the dialogue triggers (during the fade-in) but the green thread is still in the scene:

Image

When the dialogue finishes, the thread disappears:

Image

Slayer2505:
@wjt @manuq 
With your permissiion, i can continue exploring more solutions that cover the points mentioned earlier, aiming to keep things simple to maintain the overall simplicity of the project. I'll be waiting for your response, althought i'll also keep exploring different solutions in the meantime. Thank you

PR #1211, opened 2025-09-17
Fix: prevent intro dialogue from replaying on respawn in stealth challenges

### LINKED ISSUE #949: Intro dialogue triggers after respawning in stealth challenges
### Description

When you start playing the [Stella stealth level](https://endlessm.github.io/threadbare/branches/endlessm/main/#quests/story_quests/stella/1_stella_stealth/stella_stealth) (for example) a sentence of narration is shown:

Image

But if you are caught prior to reaching a checkpoint, when you respawn you see the same introduction again.

The same is true of the template, and of all StoryQuests I've seen so far (because it's in the template).

Our Cinematic node is being used to trigger this opening dialog (and similar dialogue in each scene of the storyquest template and Stella). It was originally intended only for use in non-interactive cutscenes so there's no provision for recording that it has been played.

I think that:

- When you first enter the Stella stealth level, the dialogue should play
- If you die before reaching a checkpoint, the dialogue should not play
- If you quit and then continue before reaching a checkpoint: I could be convinced either way TBH

### Checklist

JSobero:
This PR fixes #949 by ensuring the intro dialogue only plays once at the start of the level, and no longer replays after respawning.

### DIFF
--- a/scenes/ui_elements/cinematic/cinematic.gd
+++ b/scenes/ui_elements/cinematic/cinematic.gd
@@ -21,16 +21,15 @@ extends Node2D
 
 
 func _ready() -> void:
-	DialogueManager.show_dialogue_balloon(dialogue, "", [self])
-	await DialogueManager.dialogue_ended
+	if not GameState.intro_dialogue_shown:
+		DialogueManager.show_dialogue_balloon(dialogue, "", [self])
+		await DialogueManager.dialogue_ended
+		GameState.intro_dialogue_shown = true
 
 	if next_scene:
-		(
-			SceneSwitcher
-			. change_to_file_with_transition(
-				next_scene,
-				spawn_point_path,
-				Transition.Effect.FADE,
-				Transition.Effect.FADE,
-			)
+		SceneSwitcher.change_to_file_with_transition(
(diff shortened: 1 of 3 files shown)


scenes/ui_elements/cinematic/cinematic.gd
```diff
@@ -21,16 +21,15 @@ extends Node2D
 
 
 func _ready() -> void:
-	DialogueManager.show_dialogue_balloon(dialogue, "", [self])
-	await DialogueManager.dialogue_ended
+	if not GameState.intro_dialogue_shown:
+		DialogueManager.show_dialogue_balloon(dialogue, "", [self])
+		await DialogueManager.dialogue_ended
+		GameState.intro_dialogue_shown = true
```
wjt, line 27: There is more than one level that shows intro dialogue. So in this implementation, after you play one such level, no more intro dialogue will be played until you quit and restart the game.

Something would need to clear this flag when switching to a new level.

wjt: I pushed a change to clear this flag when the scene changes.

PR #1217, opened 2025-09-18
Issue template: prevent unwanted breaks in descriptions

cassidyjames:
Super minor nit, but I noticed the top form description breaking from the newline:

> image

We don't need the literal line breaks in the Markdown string, so replace the literal string `|` with a folded string `>`.


wjt (approved):
Good catch!

Issue #171, opened 2025-04-11
Add woollen/patchwork staircase assets


LauraFoglia:
The last row of this spritesheet is from this asset pack: https://pixelfrog-assets.itch.io/tiny-swords

The rest of the rows were modified to match the wooly/patchwork style of the game.

We need new tiles to replace the staircase ones.

![Image](https://github.com/user-attachments/assets/251eeead-72ac-407b-829e-ec5b72356075)

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Art: Aesthetics & Visual Design | Problem Solving | 50 |
| Art: Aesthetics & Visual Design | Creativity & Design Thinking | 75 |
| Art: Aesthetics & Visual Design | Technical Literacy | 75 |

wjt:
We no longer have a staircase in the world map. (Stephen is adding one to some other level, but it's a staircase to nowhere at the edge of the scene: just a prop.)

So I think this is less important now. Retitling.

PR #1221, opened 2025-09-19
Add wooden staircase tiles

wjt:
Add wooden staircase tiles

Previously, the staircase tiles in Cliff_Tiles were inherited from Tiny
Swords.

Replace them with very nice new wooden stairs, drawn by 159adrian on
https://github.com/endlessm/threadbare/issues/171. I modified the PNG on
that ticket to move the stairs to exactly the same region, so that no
changes are needed to scenes that already use the old stairs.

The only such scenes are in StoryQuests: Stella and El Juguete Perdido.
I think the wooden stairs look better in both; but if someone wanted to
use the previous stone stairs, they could use
assets/third_party/tiny-swords/Terrain/Ground/Tilemap_Elevation.png in
their own TileSet.

Co-authored-by: 159adrian <213216604+159adrian@users.noreply.github.com>

### DIFF
--- a/assets/first_party/tiles/Cliff_Tiles.png
+++ b/assets/first_party/tiles/Cliff_Tiles.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5e18fa6cab18c7edc9a1a038d01279b224e470e2cc1e0c00821f4a088e4228c6
-size 37198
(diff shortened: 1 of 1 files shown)


159adrian:
> @159adrian would you like to be credited as "159adrian" or by a different name? If so, what name would you to be credited as?

I don't have any problem with that name. I used to write that name since I was a kid(6-7 years old)

Issue #1198, opened 2025-09-16
Update to Godot 4.5


wjt:
Godot 4.5 has been released. I think we should probably update Threadbare to use it relatively soon because new contributors are likely to download the latest Godot version available from the Godot website at any given time, and we have a new cohort of learners starting in early October.

I also believe that 4.5 will resolve an annoying bug where tooltips for properties defined in `.gd` files do not show reliably, which limits the usefulness of our lovingly-written docstrings!

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

PR #1226, opened 2025-09-22
Update project to Godot 4.5

wjt:
Update version in project.godot. Update CI pipeline and Flatpak build to match.

Reimport all assets. The new .import properties are all the new defaults. I
updated all files by deleting .godot and reimporting the project.

Fixes https://github.com/endlessm/threadbare/issues/1198


wjt:
Flatpak pipeline depends on:

- https://github.com/flathub/org.godotengine.godot.BaseApp/pull/52

wjt:
Now that the 4.5 base app has hit Flathub, that part also builds & runs successfully.

@manuq I won't bother updating this to fix conflicts when #1243 is still outstanding and will cause more conflicts, but what do you think about updating all `.import` files like this versus leaving them as they are and updating them as they happen to be modified? I think we should do it all at once because otherwise people will accidentally commit these (automatically-made) changes together with other stuff.

wjt:
Opening all scenes in a directory is not so hard to do manually:

image

Then you can save all open scenes with one action:

image

manuq:
> Opening all scenes in a directory is not so hard to do manually:

Oh thanks, I didn't know that!

wjt:
Testing Flatpak build again in https://github.com/endlessm/threadbare/actions/runs/18091876356...

manuq:
> @manuq I refreshed this again – does it look OK?

Yes, I'm sorry that you had to redo the imports several times.

manuq (approved):

PR #1230, opened 2025-09-23
pre-commit: Disable style checks for StoryQuests

### LINKED ISSUE #1227: Relax some linter checks for StoryQuests
We have configured the various linters (licensing, formatting, code style) to not run as GitHub Actions in forks, but when people come to submit their work to us (or if they  set up `pre-commit` locally) the checks run.

As seen on https://github.com/endlessm/threadbare/pull/946 / https://github.com/endlessm/threadbare/actions/runs/17930321063/job/51020963668?pr=946 it is common for a large number of checks to fail at this point.

We could consider relaxing some of these checks to, e.g., not apply `gdlint` to `.gd` files in StoryQuests. The reason I added linting to the project as a whole is that this game doubles as a learning tool and I think the code in the project should be consistent and high-quality. Of course linting doesn't guarantee high-quality code but in my experience it helps (and avoids discussions about how to format code). But code in a StoryQuest (rather than in the main project) may be less likely to be treated as an example by other contributors in future.

Before doing that, though, I would like to see whether we can improve the checks to make it easy to comply.

wjt:
We enforce coding style and formatting checks on the project as a whole
because we want the reusable code in this project to be exemplary, and
we believe that consistent style and formatting helps to maintain code
quality – and to avoid discussions about style and formatting!

There is less need to apply the same checks to scripts, shaders, and dialogue
files in StoryQuests: these are only used within that specific StoryQuest,
rather than being intended for other quests or components to reuse or extend.
The exceptions is the template quest, and Stella: we do want to maintain
consistent style there.

Exclude all storyquests, except the template and stella, from GDScript style
checks and editorconfig formatting checks. Keep other checks enabled, such as
for large files that should be stored in Git LFS.

This is currently mostly hypothetical – none of the StoryQuests in the tree have
any scripts or shaders, just dialogue files – but there are other quests being
developed which do include custom scripts and shaders.

In addition, treat `.piskel` files like `.svg` files: don't apply end-of-line checks to them, even though they are text files, because they are not intended to be human-edited.

Resolves #1227

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,13 +24,26 @@ repos:
   rev: v3.4.0
   hooks:
   - id: editorconfig-checker
+    exclude: |
+      (?x)^(
+        scenes/quests/story_quests/(?!stella/|template/)
+      )
 
 - repo: https://github.com/Scony/godot-gdscript-toolkit
   rev: 4.3.4
   hooks:
   - id: gdformat
+    exclude: |
+      (?x)^(
+        scenes/quests/story_quests/(?!stella/|template/)
+      )
   - id: gdlint
-    exclude: script_templates/
+    exclude: |
+      (?x)^(
+        script_templates/|
+        scenes/quests/story_quests/(?!stella/|template/)
+      )
+
 exclude: |
   (?x)^(
     addons/dialogue_manager/(.*)|


.pre-commit-config.yaml
```diff
@@ -14,9 +14,9 @@ repos:
   - id: check-illegal-windows-names
   - id: check-yaml
   - id: end-of-file-fixer
-    exclude: \.svg$
+    exclude_types: [image]
   - id: trailing-whitespace
-    exclude: \.svg$
+    exclude_types: [image]
```
manuq: The https://pypi.org/project/identify/ was a good reference, thanks.

PR #1239, opened 2025-09-24
Tweak void runner difficulty

### LINKED ISSUE #1237: Adjust void-runner difficulty
Several playtesters have said that the current void-runner level is too hard. e.g.:

- https://github.com/endlessm/threadbare/discussions/1173

I don't want to spend too much time finessing the current level design – I expect it to change as we connect it up better to the rest of the quest – but I think there are some things we could tweak:

- Enemy walking speed: currently it's the same as the character's walking speed; maybe this could be lower
- Enemy running distance: currently it runs if it is further than 192px (3 tile-widths) away; maybe this could be further
- Tree positions: some of them are obnoxiously placed.

wjt:
This reduces the difficulty of the level a little, hopefully making it challenging but fair rather than seeming unfair.

Resolves #1237

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
@@ -25,7 +25,12 @@ const NEIGHBORS := [
 	TileSet.CELL_NEIGHBOR_RIGHT_SIDE,
 ]
 
+const IDLE_EMIT_DISTANCE := sqrt(2 * (64.0 ** 2))
+
 @export var void_layer: TileMapCover
+
+@export var idle_patrol_path: Path2D:
+	set = _set_idle_patrol_path
 
 var player: Player:
 	set = _set_player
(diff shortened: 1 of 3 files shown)


scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
```diff
@@ -25,36 +25,58 @@ const NEIGHBORS := [
...
+	if not is_node_ready():
+		return
+
+	match state:
+		State.IDLE:
+			path_walk_behavior.process_mode = Node.PROCESS_MODE_INHERIT
+			follow_walk_behavior.process_mode = Node.PROCESS_MODE_DISABLED
+		State.CHASING:
+			path_walk_behavior.process_mode = Node.PROCESS_MODE_DISABLED
+			follow_walk_behavior.process_mode = Node.PROCESS_MODE_INHERIT
```
wjt, line 72: It was really nice to be able to reuse the path-following behavior and only write a small amount of code to switch between them when the state changes.

manuq: Nice, in the future we could have a helper function to simplify the mapping of states to behaviors, setting process_mode accordingly. But the match is readable to me.

PR #1240, opened 2025-09-24
Split Elders into two scenes

wjt:
Previously, we had one elder.tscn, which we instantiated twice. It inherited
from npc.gd and so had a sprite_frames property, which we set as appropriate on
each elder.

But the two elders have different shapes: they hold their staffs in different
hands. So the collision shapes have to be different. We handled this with custom
code that matches the name of the SpriteFrames resource and flips the collision
shapes.

I think it is clearer to have two copies of the scene, one for each elder, and
then have the collision shapes in each scene match the sprite. If we later add
more elders, they might have shapes which are not the same or mirror images of
these two: in that case, it will be easier to add a new scene for each one.

Rename the existing elder.tscn to lore_quest_elder.tscn, setting the correct
dialogue there. Duplicate it as story_quest_elder.tscn, and adjust the collision
shapes, SpriteFrames, dialogue, etc. as the script used to do. (In fact I did
this with the script.)

Then adjust the script to inherit directly from CharacterBody2D rather than
npc.gd. Remove the top-level dialogue property: this is set directly on the
TalkBehavior now. Remove the code to flip the collision shapes. Remove the
top-level idle_sound_stream property: this was the same on each elder and now
could be adjusted directly on the stream in each scene.

Update frays_end.tscn to instance one of each scene.

Set the quest_directory property in each scene, rather than on the instances in
frays_end.tscn.

Set the InteractBehavior's action property directly: this is the only thing we
lose by not inheriting from npc.gd.

### DIFF
--- a/scenes/game_elements/characters/npcs/elder/story_quest_elder.tscn
+++ b/scenes/game_elements/characters/npcs/elder/story_quest_elder.tscn
@@ -30,6 +30,7 @@ position = Vector2(0, -40)
 sprite_frames = ExtResource("3_rah8h")
 animation = &"idle"
 autoplay = "idle"
+frame = 5
 
 [node name="BodyShape" type="CollisionShape2D" parent="."]
 position = Vector2(-1, 0)


scenes/game_elements/characters/npcs/elder/story_quest_elder.tscn
```diff
@@ -30,6 +30,7 @@ position = Vector2(0, -40)
 sprite_frames = ExtResource("3_rah8h")
 animation = &"idle"
 autoplay = "idle"
+frame = 5
```
manuq, line 33: :100:

Issue #1225, opened 2025-09-22
Add AnimationPlayer version of CharacterSpriteBehavior


manuq:
In c59059c5 we added CharacterSpriteBehavior for the visual representation of characters that move using a walk behavior. That would be good enough for simple characters. But for more complex ones, we need an AnimationPlayer version of it. Examples:
- Sound effects. Simple characters could trigger a sound when a signal of walk behaviors is emitted. But complex characters may need a sound effect syncronized precisely with a specific frame of the frame-by-frame animation. So adding an audio track to an AnimationPlayer along with a track for the AnimatedSprite2D frames is needed.
- Start/stop particles emition.
- Animating any other properties of the character. Like the "disabled" property of a collision shape, or its position, or the shape itself. This is how the repel action is implemented.

PR #1242, opened 2025-09-24
Add AnimationPlayer version of CharacterSpriteBehavior

manuq:
Resolve https://github.com/endlessm/threadbare/issues/1225

### DIFF
--- a/scenes/game_logic/character_animation_player_behavior.gd
+++ b/scenes/game_logic/character_animation_player_behavior.gd
@@ -1,0 +1,69 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name CharacterAnimationPlayerBehavior
+extends Node2D
+## @experimental
+##
+## Play animations in [member animation_player] according
+## to the velocity of [member character].
+## [br][br]
+## For creating simple characters, consider using [CharacterSpriteBehavior] instead.
+
+## The [member CharacterBody2D.velocity] is used to change the [member animation_player].
+## [br][br]
+## [b]Note:[/b] If the grandparent node is a CharacterBody2D and this isn't set,
+## the grandparent node will be automatically assigned to this variable.
+@export var character: CharacterBody2D
+
+## The controlled animation player.
+## [br][br]
+## [b]Note:[/b] If the parent node is an AnimationPlayer and this isn't set,
+## the parent node will be automatically assigned to this variable.
+@export var animation_player: AnimationPlayer:
+	set = _set_animation_player
+
+var _is_character_running: bool = false
+
+
+func _enter_tree() -> void:
+	if not animation_player and get_parent() is AnimationPlayer:
+		animation_player = get_parent()
+	if not character and get_parent() and get_parent().get_parent() is CharacterBody2D:
+		character = get_parent().get_parent()
+
+
+func _set_animation_player(new_animation_player: AnimationPlayer) -> void:
+	animation_player = new_animation_player
+	update_configuration_warnings()
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+	if animation_player is not AnimationPlayer:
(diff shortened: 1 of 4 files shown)


scenes/game_logic/character_animation_player_behavior.gd
```diff
@@ -0,0 +1,69 @@
...
+	if not character:
+		return
+
+	if character.velocity.is_zero_approx():
+		animation_player.play(&"idle")
+	else:
+		if _is_character_running:
+			animation_player.play(&"run")
+		else:
+			animation_player.play(&"walk")
```
manuq, line 64: Initially I did similar than the sprite behavior and played the walk animation at double speed if there was no "run" animation. But since this is intended for more advanced control over the character visuals, I decided to remove it.

scenes/game_logic/character_player_behavior.gd
```diff
@@ -0,0 +1,69 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name CharacterPlayerBehavior
```
wjt: The only thing I'm not keen on here is the name. "Player" is kind of synonymous with "Player-Character" but I think here you're using it to mean "[Animation]Player". CharacterAnimationPlayerBehavior?

But also: will this behaviour be used in practice? If your player-character has actions other than idle/walk/run, then more complicated logic will be needed. And the player does in practice have other actions, such as being defeated.

manuq: > The only thing I'm not keen on here is the name. "Player" is kind of synonymous with "Player-Character" but I think here you're using it to mean "[Animation]Player". CharacterAnimationPlayerBehavior?

Yeah, CharacterAnimationPlayerBehavior was actually my first attempt, then I found it too long. But I didn't realize that "Player" is confusing. It is! The first thing we all think is of the player character.

> But also: will this behaviour be used in practice? If your player-character has actions other than idle/walk/run, then more complicated logic will be needed. And the player does in practice have other actions, such as being defeated.

Fair. Today I had a similar situation with CharacterSpriteBehavior and changed it's process mode to play an animation other than idle/walk/run: https://github.com/endlessm/threadbare/blob/c679323e749d7a4d4ebd5a3e0ac5e6bb7d8a5f77/scenes/demo.tscn#L69-L70

Otherwise that "alerted" animation is being overriden. So I guess that's how this and CharacterSpriteBehavior would work in characters that use these behaviors but also expand them. What do you think?

scenes/game_logic/character_animation_player_behavior.gd
wjt: Doubling the walk animation speed looks and sounds better, though. :(

Try opening the gym, holding down right-arrow but rapidly pressing and releasing the shift key. It looks like this:

https://github.com/user-attachments/assets/3313e694-b0fb-4424-a3e7-7c968b06a717

and you can't hear it in the screen recording but the sound effect also sounds bad because it gets doubled up each time you cross the walk/run speed threshold.

I think it's the same cause in both cases: each time we switch between the walk & run animations, we are starting again from the start of the animation.

With these particular animations it might work to start the animation not from the beginning but from the same relative offset. The walk animation is 0.6s long, the run animation is 0.3s. So if we are playing `walk` and need to switch to `run`, we can calculate the progress ratio as `animation_player.current_animation_position / animation_player.current_animation_length` and restore that in the other animation. But if we did that it would be simpler to just play the walk animation at double speed.

I don't know what the right way is to solve this.

scenes/game_logic/character_player_behavior.gd
wjt: Maybe… or perhaps use a different behaviour that copy-pastas this code? We can experiment.

scenes/game_logic/character_animation_player_behavior.gd
manuq: Good point, it sounds very bad! Sorry that I didn't notice this before, but yes let's solve it as follow up.

Issue #1213, opened 2025-09-18
Clearly flag template files as not to be edited


wjt:
A request from our learning team:

## User Story

- **When** I open the project and see the template content in Godot,
- **I want to** clearly understand that these files are only for making copies,
- **So I can** avoid accidentally editing original template scenes, resources, or assets.

## Requirements

1. **Rename template folder** to something short but clear, e.g. `!!NO_EDIT`.
    - Names should remain brief enough that the rest of the filename (e.g. player_components) is still visible in the file tree.
2. **Visual distinction**
    - Recolor the folder(s) red (or another “warning” color) in Godot’s file system if supported.
    - If recoloring isn’t feasible, rely on naming conventions (!! prefix).
    - **Scope**: Apply these conventions to **all contents inside the template directory**, including:
        - Subfolders (intro, outro, minigames, etc.)
        - Scene files
        - Scripts
        - Assets (PNGs, audio, etc.)
        - Resource files (.tres, .res, sprite frames, tile sets, etc.)
    - Goal: No file within the template folder should appear “safe to edit.”

wjt:
Here is how the FileSystem view would look after implementing the following 2 requests:

> Rename template folder to something short but clear, e.g. !!NO_EDIT.
> Recolor the folder(s) red (or another “warning” color) in Godot’s file system if supported.

<img width="484" height="726" alt="Image" src="https://github.com/user-attachments/assets/d2136cbf-fa3f-4832-ad69-994e85eea109" />

For the record the folder already had a “warning” color, yellow:

<img width="484" height="726" alt="Image" src="https://github.com/user-attachments/assets/9c3333e1-30bf-4dbf-b07d-70ae0c1509df" />

but placing `!` at the start of the name does move the folder to the top of the list.

@jgbourque @hydrolet If I understand the request correctly, this is sufficient to satisfy it?

I think this might help people who primarily navigate their project using the FileSystem view. (From memory, that includes you, @hydrolet.) I don't think it will help people who e.g. navigate from a scene to a dialogue by clicking on the Cinematic node's Dialogue property in the Inspector (from memory, that's what you do, @jgbourque):

<img width="531" height="384" alt="Image" src="https://github.com/user-attachments/assets/7c9ea83b-0b81-4334-8c66-1e414ad1d235" />

[...]

PR #1243, opened 2025-09-24
Rename template to NO_EDIT

manuq:
This will hopefully help contributors with accidental changes.

animated_texture_rect.gd: Use UID path to preload the template sprite
frames resource file.
    
Adjust the following files to the renamed directory:
- .pre-commit-config.yaml
- addons/storyquest_bootstrap/plugin.gd
- project.godot

Resolve https://github.com/endlessm/threadbare/issues/1213

---
  - https://app.asana.com/0/0/1211209074317243


scenes/quests/story_quests/NO_EDIT/0_template_intro/template_intro_components/template_intro_image.png.import
```diff
@@ -3,28 +3,34 @@
...
+source_file="res://scenes/quests/story_quests/NO_EDIT/0_template_intro/template_intro_components/template_intro_image.png"
+dest_files=["res://.godot/imported/template_intro_image.png-b2234a1131462d8ba4b3d04377c946a8.ctex"]
 
 [params]
 
 compress/mode=0
 compress/high_quality=false
 compress/lossy_quality=0.7
+compress/uastc_level=0
+compress/rdo_quality_loss=0.0
```
wjt: Did you write this PR with Godot 4.5? These properties are added by that version. When I try to edit the project in 4.4 it removes them again.

manuq: > Did you write this PR with Godot 4.5? These properties are added by that version. When I try to edit the project in 4.4 it removes them again.

Ugh yes sorry, I fixed it now.

wjt (commented):
I still get this when I open the project after switching to this branch:

Image

> Unable to load addon script from path: 'res://addons/storyquest_bootstrap/plugin.gd'. This might be due to a code error in that script.
Disabling the addon at 'res://addons/storyquest_bootstrap/plugin.cfg' to prevent further errors.

In the Output tab I see:

> ERROR: res://addons/storyquest_bootstrap/plugin.gd:166 - Parse Error: Could not resolve external class member "STORY_QUEST_TEMPLATE".
> ERROR: modules/gdscript/gdscript.cpp:3022 - Failed to load script "res://addons/storyquest_bootstrap/plugin.gd" with error "Parse error".

And indeed when I save, `project.godot` is modified to disable the plugin.

I'll see if I can fix this

wjt:
I edited this branch to add 2 patches before the rename:

- Fix the plugin load problem
- Change AnimatedTextureRect to default to null SpriteFrames, just like TextureRect

manuq:
> @hydrolet clarified on Slack that it is option 2 that she wants. I would tend to agree that there's no point in just renaming the folder: if you only have 1 scene or dialogue file open, its path is not shown in the tab, only the filename. So I think we should do the more thorough renaming as in 2.

OK! Let me work on it.

manuq:
@wjt I added 2 fixup commits. Could you review? In `addons/storyquest_bootstrap/plugin.gd` I factored out the prefix to a constant: `const TEMPLATE_PREFIX := "NO_EDIT"` in case we ever change it, it can be done in a single place.

addons/storyquest_bootstrap/plugin.gd
```diff
@@ -10,24 +10,37 @@ const NEW_STORYQUEST_DIALOG = preload(
...
 const MIN_TITLE_LENGTH := 4
 
 var template_info := {
 	"intro":
 	{
-		"scene_path": "/0_template_intro/template_intro.tscn",
-		"components_path": "/0_template_intro/template_intro_components/",
-		"dialogue_path": "/0_template_intro/template_intro_components/template_intro.dialogue",
-		"image_path": "0_template_intro/template_intro_components/template_intro_image.png",
+		"scene_path": "/0_{prefix}_intro/{prefix}_intro.tscn".format({"prefix": TEMPLATE_PREFIX}),
```
wjt, line 23: I personally think that the repeated `.format({"prefix": TEMPLATE_PREFIX})` is much less legible than repeating `NO_EDIT`, even at the cost having to search-and-replace every line if we rename it again. But let's go with this for now.

wjt (approved):
OK, let's see if this works!

PR #1245, opened 2025-09-25
Fade out when quitting

wjt:
Previously, choosing “Exit Game” on the main menu abruptly terminated the game.

Instead, play the normal fade-out transition, which also fades out the music,
then quit.


manuq (approved):
Nice touch, I like it. And is using the existing Transitions autoload.

Issue #1216, opened 2025-09-18
Import errors on first open


wjt:
### Description

When you first open a fresh clone of Threadbare in Godot, you get the following errors in the Output:

>  ERROR: BUG: File queued for import, but can't be imported, importer for type 'dialogue_manager' not found.
>  ERROR: Failed loading resource: res://scenes/ui_elements/cinematic/cinematic_placeholder.dialogue. Make sure resources have been imported by opening the project in the editor at least once.
>  ERROR: res://scenes/ui_elements/cinematic/cinematic.gd:9 - Parse Error: Could not preload resource file "uid://b7ad8nar1hmfs".
>  ERROR: modules/gdscript/gdscript.cpp:3022 - Failed to load script "res://scenes/ui_elements/cinematic/cinematic.gd" with error "Parse error".
>  ERROR: BUG: File queued for import, but can't be imported, importer for type 'dialogue_manager' not found.
>  ERROR: Failed loading resource: res://scenes/menus/intro/components/intro.dialogue. Make sure resources have been imported by opening the project in the editor at least once.

I think what's happening is:

1. The default scene is `res://scenes/menus/splash/splash.tscn`; this is opened on startup
2. It uses `splash.gd` which `preload()`s `res://scenes/menus/title/title_screen.tscn` as a PackedScene
3. `title_screen.tscn` references `res://scenes/menus/intro/intro.tscn` in a `PackedScene` property
4. `intro.tscn` uses `cinematic.gd`
5. `cinematic.gd` `preload()`s a `DialogueResource`
6. But at this point the Godot Dialogue Manager addon has not yet loaded
7. Boom

I think we could fix this by breaking the links in 2 or 3 to not use `PackedScene`.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

wjt:
Actually it's not so simple: after replacing those two `PackedScene` properties with strings that happen to be the `uid://` paths to scenes, I still see the same 5 errors.

Removing the `preload()` of a dialogue file in `cinematic.gd` cuts it down to just these two:

>  ERROR: BUG: File queued for import, but can't be imported, importer for type 'dialogue_manager' not found.
>  ERROR: Failed loading resource: res://scenes/menus/intro/components/intro.dialogue. Make sure resources have been imported by opening the project in the editor at least once.

PR #1249, opened 2025-09-25
Add an “Abandon Quest” option to the pause menu

### LINKED ISSUE #432: Add a way to abandon a quest

wjt:
This is only shown if the player is currently on a quest, including the
Sokoban puzzles. Update the button visibility each time the pause menu
is shown.

If chosen, the inventory is cleared, as is the incorporating_threads
flag that is set while inside the Eternal Loom; then Fray's End is
loaded. If Fray's End happens to already be loaded, that's OK: it will
be reloaded.

Resolves #432

This also fixes some finer points of the game state, necessary to make this work correctly.

### DIFF
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -93,6 +93,10 @@ func set_current_spawn_point(spawn_point: NodePath = ^"") -> void:
 	_save()
 
 
+func is_on_quest() -> bool:
+	return _state.has_section_key(QUEST_SECTION, QUEST_PATH_KEY)
+
+
 ## Marks the current quest (if any) as completed.
 func mark_quest_completed() -> void:
 	var quest_name: String = _state.get_value(QUEST_SECTION, QUEST_PATH_KEY, "")
@@ -118,6 +122,12 @@ func add_collected_item(item: InventoryItem) -> void:
 	collected_items_changed.emit(items_collected())
 	_update_inventory_state()
 	_save()
+
+
+func abandon_quest() -> void:
+	set_incorporating_threads(false)
+	_state.erase_section_key(QUEST_SECTION, QUEST_PATH_KEY)
+	clear_inventory()
 
 
 ## Remove all [InventoryItem] from the [member inventory].
--- a/scenes/globals/pause/pause_overlay.gd
+++ b/scenes/globals/pause/pause_overlay.gd
@@ -3,10 +3,12 @@
 extends CanvasLayer
 
 const TITLE_SCENE: PackedScene = preload("uid://stdqc6ttomff")
+const FRAYS_END := "uid://cufkthb25mpxy"
 
 @onready var pause_menu: Control = %PauseMenu
 @onready var resume_button: Button = %ResumeButton
 @onready var options: Control = %Options
+@onready var abandon_quest_button: Button = %AbandonQuestButton
 
 
 func _ready() -> void:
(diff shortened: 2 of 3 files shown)


scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -34,6 +34,7 @@ func _ready() -> void:
 		)
 		await DialogueManager.dialogue_ended
 		GameState.set_incorporating_threads(false)
+		GameState.mark_quest_completed()
```
manuq, line 37: :+1:

scenes/globals/game_state/game_state.gd
```diff
@@ -96,9 +96,11 @@ func set_current_spawn_point(spawn_point: NodePath = ^"") -> void:
...
 func mark_quest_completed() -> void:
 	var quest_name: String = _state.get_value(QUEST_SECTION, QUEST_PATH_KEY, "")
-	if quest_name and quest_name not in completed_quests:
-		completed_quests.append(quest_name)
-		_state.set_value(GLOBAL_SECTION, COMPLETED_QUESTS_KEY, completed_quests)
+	if quest_name:
+		if quest_name not in completed_quests:
+			completed_quests.append(quest_name)
+			_state.set_value(GLOBAL_SECTION, COMPLETED_QUESTS_KEY, completed_quests)
+		_state.erase_section_key(QUEST_SECTION, QUEST_PATH_KEY)
```
manuq, line 107: :+1:

scenes/globals/pause/pause_overlay.gd
```diff
@@ -25,10 +27,19 @@ func toggle_pause() -> void:
...
 		pause_menu.show()
 		resume_button.grab_focus()
 
 
+func _on_abandon_quest_pressed() -> void:
+	toggle_pause()
+	GameState.abandon_quest()
+	SceneSwitcher.change_to_file_with_transition(
+		FRAYS_END, ^"", Transition.Effect.FADE, Transition.Effect.FADE
+	)
```
manuq, line 40: :+1:

scenes/globals/pause/pause_overlay.gd
```diff
@@ -3,10 +3,12 @@
 extends CanvasLayer
 
 const TITLE_SCENE: PackedScene = preload("uid://stdqc6ttomff")
+const FRAYS_END := "uid://cufkthb25mpxy"
```
manuq, line 6: At this point I wonder if Fray's End UID should be in a global/singleton/autoload.

manuq (approved):
Excellent, this will make demo'ing the game faster!

scenes/globals/pause/pause_overlay.gd
wjt: The line above this also explains something I couldn't understand when fixing #1216: even after I removed the `PackedScene` reference between the splash screen and the title screen, `Cinematic` was still getting loaded. I think it was by this!

Issue #1117, opened 2025-08-22
Fix grey background beneath grass on top island of Fray's End


wjt:
There is no clifftop tile underneath the grass at the bottom edge of the top island of Fray's End:

<img width="1401" height="360" alt="Image" src="https://github.com/user-attachments/assets/03dd2736-dba5-4d33-bf27-0ab7d39df804" />

As a result, if you look closely you can see the grey default background colour visible at the edge of the grass:

<img width="640" height="360" alt="Image" src="https://github.com/user-attachments/assets/54053b72-8f3d-46cc-ba54-f1fba2e0b958" />

Unfortunately it is not quite as simple as painting a row of stone tiles into the `TileMapLayers/Stone_Level1` node because this is actually on top of the grass. Instead a lower level of stone (earlier in the scene tree) needs to be added; the cliff that are conceptually below the grass should be moved there; and then the missing tiles can be painted in.

PR #1251, opened 2025-09-26
Fray's End: Add cliff tiles below grass

wjt:
Fray's End: Add cliff tiles below grass

Previously, there was no cliff below the bottom border of the top
island, so if you looked closely you could see the grey default
background colour peeking through.

The cliffs in the row of tiles below the edge of the grass were
previously part of Stone_Level1, which is above Grass_Level1.

Add a new, lower layer, Stone_Level0. Move the cliffs that are
conceptually below the grass from Stone_Level1 to Stone_Level0. Add
clifftop below the edge of the grass.

Resolves https://github.com/endlessm/threadbare/issues/1117


wjt:
| Before | After |
| ------ | ----- |
| image | image |

Issue #1197, opened 2025-09-16
"ObjectDB instances leaked at exit" warning


wjt:
### Description

Steps to reproduce:

1. Launch the game from a terminal window, or some other place where you can see error/warning messages from the engine
2. Play the game a little
3. Quit the game

I see the following:

```
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
     at: cleanup (core/object/object.cpp:2378)
```

I don't know if this is a significant problem but I tend to believe that warnings should be treated as important until proven otherwise.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Optimization & Performance | Problem Solving | 150 |
| Engineering: Optimization & Performance | Creativity & Design Thinking | 75 |
| Engineering: Optimization & Performance | Technical Literacy | 150 |

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

wjt:
If I do the following:

- Export game for Linux
- Run exported game with `./build/threadbare.x86_64 --verbose`
- At the title screen, choose Exit Game

Then I see the following:

```
ERROR: 1 RID allocations of type 'N5GLES37TextureE' were leaked at exit.
ERROR: Texture with GL ID of 16: leaked 5460 bytes.
   at: ~Utilities (drivers/gles3/storage/utilities.cpp:77)
ERROR: Parameter "RenderingServer::get_singleton()" is null.
   at: ~CompressedTexture2D (scene/resources/compressed_texture.cpp:464)
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
     at: cleanup (core/object/object.cpp:2378)
Leaked instance: GDScriptFunctionState:9223372096078349844
Hint: Leaked instances typically happen when nodes are removed from the scene tree (with `remove_child()`) but not freed (with `free()` or `queue_free()`).
Orphan StringName: quit (static: 0, total: 1)
Orphan StringName: finished (static: 0, total: 1)
Orphan StringName: Music (static: 0, total: 1)
Orphan StringName: _introduce_scene (static: 0, total: 1)
StringName: 4 unclaimed string names at exit.
```

`quit` is the name of the method being called at exit. `_introduce_scene` and `finished` are the names of a method and signal to be called and emitted by `do_transition` - since commit 3176904157ca4ed3b298c7984c7b723670d15ae6 we quit halfway through a fade-out.

`Music` is, I would guess, the name of one of the audio buses:

```console
[...]

PR #1253, opened 2025-09-26
Fix ObjectDB & StringName leaks on exit

wjt:
Fix ObjectDB & StringName leaks on exit

Previously, quitting the game from the main menu would show (under
`--verbose`):

    WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
        at: cleanup (core/object/object.cpp:2378)
    Leaked instance: GDScriptFunctionState:9223372096078349844
    Hint: Leaked instances typically happen when nodes are removed from the scene tree (with `remove_child()`) but not freed (with `free()` or `queue_free()`).
    Orphan StringName: quit (static: 0, total: 1)
    Orphan StringName: finished (static: 0, total: 1)
    Orphan StringName: Music (static: 0, total: 1)
    Orphan StringName: _introduce_scene (static: 0, total: 1)
    StringName: 4 unclaimed string names at exit.

`quit` is the name of the method that quits the game; `finished` and
`_introduce_scene` are members of `transitions.gd` which are used at the
end of `do_transition`. Since commit 3176904 we quit midway through
`do_transition`.

Add a new method, `do_out_transition`, which only does the fade-out and
leaves the black cover layer visible. When quitting the game, await this
method, then call quit(), deferred, at the end of the frame.

This fixes the leaks above. (There are still an error about a leaked
texture, which appears to be unrelated.)

Resolves https://github.com/endlessm/threadbare/issues/1197

### DIFF
--- a/scenes/globals/scene_switcher/transitions/transitions.gd
+++ b/scenes/globals/scene_switcher/transitions/transitions.gd
@@ -95,6 +95,18 @@ func do_transition(
 	finished.emit()
 
 
+## Transition out, leaving the screen blanked at the end of the transition.
+## [br][br]
+## This is only intended to fade out when quitting the game. Unlike [method do_transition], this
+## does not emit [signal started] or [signal finished]; use [code]await[/code] to wait for the
+## fade-out to finish.
+func do_out_transition(
+	out_transition: Transition.Effect = Transition.Effect.FADE,
+):
+	visible = true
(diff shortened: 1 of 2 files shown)


manuq (approved):
Great finding. Yes adding a new function to just do the first part of the transition seems simpler than adding complexity to the existing one.

PR #1257, opened 2025-09-27
Grappling hook: Fancier camera

manuq:
Add a camera behavior node (a node that can control a camera) to keep a target on frame, within a safe region. When the target is off frame, the camera will pan to contain it. The safe region is intended to not move the camera unnecessarily. 

[Grabación de pantalla desde 2025-09-26 22-09-53.webm](https://github.com/user-attachments/assets/92718eab-850a-4b70-a7b8-94dcf11479a7)

This sketch tries to explain the safe region:

on-off-frame

The camera region is a 720x1280 (16:9) rectangle that fits the viewport. The dashed area outside shows the canvas if the viewport aspect ratio doesn't match the camera aspect ratio. Inside the camera region, there is a "safe region" rectangle, defined with left/top/right/bottom margins going from zero (all margin) to one (no margin at all). The recording above uses 0.5 for all margins, so the safe region is 1/4 rectangle of the camera rectangle. Setting zero to all margins there is no safe region, so the camera moves even when (for instance) the grappling hook grabs a button item.


scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_needles.tscn
```diff
@@ -72,6 +73,13 @@ limit_bottom = 1728
 position_smoothing_enabled = true
 editor_draw_limits = true
 
+[node name="FrameCameraBehavior" type="Node2D" parent="OnTheGround/Player/Camera2D" node_paths=PackedStringArray("camera", "main_target", "frame_target")]
+unique_name_in_owner = true
+script = ExtResource("15_dwcg4")
+camera = NodePath("..")
+main_target = NodePath("../..")
+frame_target = NodePath("../../PlayerHook/HookEnding")
```
wjt, line 81: How did you set this in the editor? It's a child of the `player.tscn` scene so I can't find a way to do it, except perhaps enabling editable children on the player temporarily?

wjt (approved):
Playing this, it feels good!

But it does feel a little like you are reimplementing the drag margin logic that the camera already has. Another way to do this might be:

1. Add a second Camera2D as child of HookEnding in the player scene
2. Configure this with the desired drag margins; don't enable it.
3. When the player begins hooking:
    - Set the hook camera to `enabled = true`
    - Call `make_current()` on it
4. When the player ends hooking:
    - Set `enabled = false` which should restore the next enabled camera in the tree – the one attached to the player

As I type this out – maybe it's more complicated than what you have implemented :)

scenes/quests/lore_quests/quest_002/2_grappling_hook/components/grappling_hook_needles.gd
```diff
@@ -5,9 +5,13 @@ extends Node2D
 
 @onready var door: Node2D = %Door
 @onready var award_buttons: Node2D = %AwardButtons
+@onready var player: Player = %Player
+@onready var frame_camera_behavior: FrameCameraBehavior = %FrameCameraBehavior
 
 
 func _ready() -> void:
+	frame_camera_behavior.frame_target = player.get_node("PlayerHook/HookEnding")
```
wjt, line 13: Why is this done in code here, but in the scene file in `grappling_hook_needles.tscn`?

scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_needles.tscn
manuq: Hmm yes I used editable children while developing, then turned it off.  Good finding, it shouldn't be part of the changes.

scenes/quests/lore_quests/quest_002/2_grappling_hook/components/grappling_hook_needles.gd
manuq: The code should do it.

manuq:
> Playing this, it feels good!
> 
> But it does feel a little like you are reimplementing the drag margin logic that the camera already has. Another way to do this might be:
> 
> 1. Add a second Camera2D as child of HookEnding in the player scene
> 2. Configure this with the desired drag margins; don't enable it.
> 3. When the player begins hooking:
>    
>    * Set the hook camera to `enabled = true`
>    * Call `make_current()` on it
> 4. When the player ends hooking:
>    
>    * Set `enabled = false` which should restore the next enabled camera in the tree – the one attached to the player
> 
> As I type this out – maybe it's more complicated than what you have implemented :)

Having a second camera is an interesting option. Let me experiment with it!

PR #1258, opened 2025-09-27
Add new crops to Fray's End

### LINKED ISSUE #677: Replace scaled bushes with crops
In Fray's End there are scaled bushes as placeholders for crops. Create an asset for the crops and replace them:

Image

Be inspired by the game's lore when creating the crops. They could be made of fabric or have a crochet pattern, for example.

shiroidK:
Resolves #677


scenes/quests/story_quests/template/3_template_sequence_puzzle/template_combat.tscn
wjt, line 1: I don't think you meant to move this file?

scenes/ui_elements/cinematic/balloon.tscn
wjt, line 1: Same with this one – I don't think you meant to move it.

wjt (approved):
Nice improvement!

One suggestion for the future: you have drawn these crops at 64×64px, but then you have scaled them down – by different amounts in each axis. For example, the barley is scaled to 0.4 on the X axis and 0.6 on the Y axis.

They would look even better in the game if they were not scaled but drawn at the intended size. And if something must be scaled, prefer to scale both axes by the same amount, and by a power of 2 (for example, 0.5×0.5 or 0.25×0.25).

PR #1259, opened 2025-09-29
Add tomato, carrot, barley prop scenes

wjt:
Previously, these were placed in Fray's End by instancing bush.tscn,
then setting its sprite_frames and scale.

Make a standalone scene for each prop, so that the sprite and scale can be set in one place.

Set the scale factor to be the same in both axes, and a power of 2: this
looks better and reduces jitter when moving the camera.

Replace all instances in Fray's End with these scenes. Group them under
new parent nodes.

Set a bush that was accidentally turned into a carrot back into a bush.

Factor out the script that randomises the start frame of an AnimatedSprite2D from the 2 different parent script that did this.

### DIFF
--- a/scenes/game_elements/props/decoration/random_frame_animated_sprite.gd
+++ b/scenes/game_elements/props/decoration/random_frame_animated_sprite.gd
@@ -1,0 +1,18 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends AnimatedSprite2D
+## Starts a decoration's animation from a random frame.
+##
+## The frame offset is not persisted to the owning scene.
+
+
+func _ready() -> void:
+	var frames_length: int = sprite_frames.get_frame_count(animation)
+	frame = randi_range(0, frames_length)
+
+
+func _notification(what: int) -> void:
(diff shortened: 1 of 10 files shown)


scenes/game_elements/props/decoration/random_frame_animated_sprite.gd
```diff
@@ -0,0 +1,18 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends AnimatedSprite2D
```
manuq, line 4: I wonder if this could be a behavior node (using composition) rather than a subclass of AnimatedSprite2D.

manuq (approved):
Excellent! Thanks for factoring out the random frame that we want in multiple scenes.

scenes/game_elements/props/decoration/random_frame_animated_sprite.gd
wjt: I thought about that. I did it this way because in none of the relevant scenes did the AnimatedSprite2D node have an attached script so I thought this was simpler. Maybe we can change it in future if it becomes a problem?

Issue #673, opened 2025-05-20
Design & implement non-placeholder storybook


wjt:
Currently the storybook looks like this:

![Image](https://github.com/user-attachments/assets/bf329666-be3f-4c0d-98de-602a0b84ecc6)

This is of course very much a placeholder. Conceptually it should look more like a book.

wjt:
I think it would actually look better to have one quest per page (i.e. 1-2 quests visible at a time), with arrows at the sides to turn the pages. People who know my taste in games will not be surprised to know that what I have in mind is the journal from Spelunky:

![Image](https://github.com/user-attachments/assets/99273294-8d5e-4e2d-848d-9a8c3bb45d57)

![Image](https://github.com/user-attachments/assets/5640e3d9-caa0-4924-9719-f8c0c303b7df)

I imagine that the book would fall open to the first 2-page spread that contains a quest that the player has not yet played. (This does raise the question of how we sort StoryQuests, and how we let the player search them, but that's also an issue with the current placeholder. Maybe there is a scrollable table of contents at the start of the book with a search feature.)

wjt:
Talking to my daughter this morning, she pointed out that the world is made of fabric and so the storybook should be too. I think she's right!

Maybe it can be a fabric swatch book like this:

![Books of samples of tartan fabric](https://i.pinimg.com/originals/ed/78/b2/ed78b23a0c89b43b302150aa08692c03.jpg)

wjt:
We currently don't have any way to stop a quest mid-way, do another quest, then come back to the first quest and pick up where we left off.

panconqueso04:
Quiero trabajar en esto:D

DonSeja06:
I Plan to work on this

Gatorrante:
Me gustaría trabajar en esto también!

panconqueso04:
Done, we’ve already been in touch and are coordinating.

Gatorrante:
Preliminar design of the book 

<img width="2800" height="1900" alt="Image" src="https://github.com/user-attachments/assets/a8253601-1e62-4bd9-8191-376ae9d9c483" />

PR #1263, opened 2025-09-29
Add non-placeholder storybook

manuq:
Replace the placeholder book UI with an animated storybook that matches the intended design and aesthetics.

Add a spritesheet with left/right page-turn animations. It contains pixel art textures inspired by fabric.

UI Redesign:

- Split the book into two pages: quest list on the left, quest details on the right.

- Add a ScrollContainer for quest descriptions to preserve the book’s layout.

- Add left/right arrow buttons for page navigation.

Animations & UX:

- UI elements temporarily hide while the page-turn animation plays, showing only the book.

- Restore UI focus after the turn animation is finished.

Fix https://github.com/endlessm/threadbare/issues/673


scenes/menus/storybook/components/storybook.gd
```diff
@@ -26,14 +26,22 @@ const QUEST_RESOURCE_NAME := "quest.tres"
...
+@onready var animated_book: AnimatedSprite2D = %AnimatedSprite2D
+@onready var ui_container: Control = %StoryBookContent
+@onready var left_button: Button = %Left_Button
+@onready var right_button: Button = %Right_Button
 
 
 func _enumerate_quests() -> Array[Quest]:
 	var has_template: bool = false
-	var quests: Array[Quest]
+	var quests: Array[Quest] = []
```
manuq, line 43: @wjt do you think that this needs the same as 16eeb2dd ? `var quests: Array[Quest] = [] as Array[Quest]` ?

scenes/menus/storybook/components/storybook.gd
```diff
@@ -68,24 +79,101 @@ func _ready() -> void:
...
+	if previous_button:
+		previous_button.focus_neighbor_bottom = back_button.get_path()
+		back_button.focus_neighbor_top = previous_button.get_path()
 
+	left_button.pressed.connect(_on_left_button_pressed)
+	right_button.pressed.connect(_on_right_button_pressed)
 	reset_focus()
 
 
+func _change_focus_to(index: int) -> void:
```
manuq: This could be renamed to `_switch_to_page(page_index: int)`.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -26,14 +26,22 @@ const QUEST_RESOURCE_NAME := "quest.tres"
 ## have a [code]quest.tres[/code] file within.
 @export_dir var quest_directory: String = "res://scenes/quests/story_quests"
 
+var _last_focused_button: Button = null
+var _last_focus_index: int = -1
```
manuq: This is the index of the current page. I think I'll rename it to `_current_page_index`.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -68,24 +79,101 @@ func _ready() -> void:
...
+	right_button.pressed.connect(_on_right_button_pressed)
 	reset_focus()
 
 
+func _change_focus_to(index: int) -> void:
+	if _navigation_locked:
+		return
+
+	if not quest_list:
+		return
```
manuq: `quest_list` is an `@onready` so this won't ever happen. I'll remove it.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -68,24 +79,101 @@ func _ready() -> void:
...
+	_navigation_locked = true
+
+	var button: Button = quest_list.get_child(index)
+	if not is_instance_valid(button):
+		_navigation_locked = false
+		return
+
+	var quest: Quest = button.get_meta("quest") if button.has_meta("quest") else null
+
+	var old_index := _last_focus_index
```
manuq: This third variable `old_index` is not necessary if we update the index at the bottom of the function.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -26,14 +26,22 @@ const QUEST_RESOURCE_NAME := "quest.tres"
 ## have a [code]quest.tres[/code] file within.
 @export_dir var quest_directory: String = "res://scenes/quests/story_quests"
 
+var _last_focused_button: Button = null
```
manuq: This variable is not needed, as the button can be obtained from the current page index: `quest_list.get_child(_current_page_index)`.

scenes/menus/storybook/components/storybook.gd
wjt: No, it doesn't. In this context Godot infers the type of the right-hand-side to be `Array[Quest]`.

The issue in 16eeb2ddae4e0b977169f02702bd8dc5609b204b was that `[]` was a `Variant` parameter. Here's the signature:

```gdscript
func get_value(section: String, key: String, default: Variant = null) -> Variant: ...
```

You might hope it would be something like this, but it's not:

```gdscript
func get_value<T>(section: String, key: String, default: T = null) -> T: ...
```

Here's a simpler example:

```gdscript
func _ready() -> void:
	var foo: Array[String] = []
	foo.append("hello")
	
	var bar: Variant = []
	foo = bar
```

The final line is a type error:

> Trying to assign an array of type "Array" to a variable of type "Array[String]".

but the initialisation of `foo` is fine.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -51,12 +58,15 @@ func _enumerate_quests() -> Array[Quest]:
...
 func _ready() -> void:
+	animated_book.animation_finished.connect(_on_animation_finished)
+
 	var previous_button: Button = null
 	for quest in _enumerate_quests():
 		var button := Button.new()
 		button.text = quest.get_title()
 		button.theme_type_variation = "FlatButton"
 		quest_list.add_child(button)
+		button.set_meta("quest", quest)
```
wjt, line 69: I've never seen `Object.set_meta` before!

scenes/menus/storybook/components/storybook.gd
manuq: Excellent, thanks for clarifying.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -68,24 +78,97 @@ func _ready() -> void:
...
+		button.grab_focus()
+
+
+func _on_left_button_pressed() -> void:
+	if _navigation_locked:
+		return
+	var target := _current_page_index - 1
+	if target < 0:
+		target = max(0, quest_list.get_child_count() - 1)
+	_switch_to_page(target)
```
wjt, line 146: Doesn't need to be changed, but:

```suggestion
	_switch_to_page(posmod(_current_page_index - 1, quest_list.get_child_count()))
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -68,24 +78,97 @@ func _ready() -> void:
...
+
+
+func _on_right_button_pressed() -> void:
+	if _navigation_locked:
+		return
+	var target := _current_page_index + 1
+	if target >= quest_list.get_child_count():
+		target = 0
+	_switch_to_page(target)
```
wjt, line 155: Even simpler when the first argument is known not to be negative:

```suggestion
	_switch_to_page(_current_page_index + 1 % quest_list.get_child_count())
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -68,24 +78,97 @@ func _ready() -> void:
...
+		return
+	if page_index < 0 or page_index >= total:
+		return
+	if page_index == _current_page_index:
+		return
+
+	_navigation_locked = true
+
+	var button: Button = quest_list.get_child(page_index)
+	if not is_instance_valid(button):
```
wjt, line 105: In general I'm a bit suspicious of `is_instance_valid`: if false, in my experience it normally means there is some bug elsewhere in the code, because it shouldn't be holding onto references to freed instances.

But I don't think this needs to block merging this change.

wjt (approved):
Very nice!

scenes/menus/storybook/components/storybook.gd
manuq: I haven't used it much, It is wild. Metadata can be anything. There is a `+ Add Metadata` button in the Inspector at the bottom of any node or resource you are inspecting.

scenes/menus/storybook/components/storybook.gd
manuq: Much better!

PR #1267, opened 2025-09-30
Intro: Play dialogue at normal speed

wjt:
The dialogue in the intro is set to play at half speed. I added this way back in March when I added this dialogue #46 but there's no explanation there for why I made it half speed.
Play it at the normal speed.

Also, simplify how we wait for the fade-in to finish before starting the intro dialogue.


manuq (approved):
Nice to also use a signal for the transition waiting.

Issue #1265, opened 2025-09-30
Guard: Integrate with character behavior


manuq:
Refactor the guard, integrating the character behavior node(s).

The guard moves along a path while patrolling, so the PathWalkBehavior node can be used for that state. Also when chasing the player there is a "leave breadcrumbs" logic (registering the points walked) And returning from chasing the player, there is a "collecting breadcrumbs" logic. So maybe PathWalkBehavior can also be used when returning, by dynamically building a path using straight lines between the breadcrumb points.

PR #1268, opened 2025-09-30
Guard: Integrate with character behavior

manuq:
Resolve https://github.com/endlessm/threadbare/issues/1265


scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -437,35 +431,6 @@ func edit_patrol_path() -> void:
...
-		var debug_color: Color = Color.RED
-		var line_width: float = 5.0
-		var point_size: float = 15.
-
-		if patrol_path and patrol_path.curve:
-			var curve: Curve2D = patrol_path.curve
-			## Draw the patrol path segments
-			if curve.point_count > 1:
-				for point_idx in curve.point_count - 1:
-					draw_line(
```
wjt: I think it's good to remove this because it interacts badly with the debug option to have the guard move in the editor: the path is drawn relative to the guard's current position not original position.

wjt (commented):
I did not look at the WIP patch but the preparatory patches look good to me

manuq:
While doing this I found that the calculations of the path walk behavior were wrong for the edge cases, and for scaled or rotated paths. So I have split that into a separate pull request.

wjt:
I fixed the bug, it was easy enough and needed fixing anyway!

In any case I will close this PR - we don't need to keep including this in every new web build - but leave the branch.

PR #1289, opened 2025-10-03
pre-commit: Recognise piskel files via exclude_types

wjt:
My upstream pull request was merged:

- https://github.com/pre-commit/identify/pull/541

The maintainer chose to edit my contribution to not add the `image` tag
to piskel files:

> the `image` tag was an idea that didn't really pan out so I'm just
> going to remove it here

Of course that's their prerogative as maintainer!

Unfortunately this means that my plan from commit
e92fd6d85a60328d0767550d0149bcdee84eb858, to exclude all image files
with `exclude_types: [image]`, is not going to work. Instead we have to
add the `piskel` type explicitly to the list, and also any further
text-based image file formats that we find in future.

Add `piskel` to the exclude list. Replace the `image` type with `svg`:
`svg` is the only format with that type, and since that tag will not be
applied to any new formats, it's clearer to be explicit.

Remove the regex branches that match piskel files by name.

If you have `pre-commit` set up locally, this change will cause the hook
to fail:

    =====> Type tag 'piskel' is not recognized.  Try upgrading identify and pre-commit?

If you installed `pre-commit` with `uv tool install` you can run:

    uv tool upgrade pre-commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,17 +16,15 @@ repos:
   - id: end-of-file-fixer
     exclude: |
       (?x)(
-        \.piskel$|
         ^scenes/quests/story_quests/(?!stella/|NO_EDIT/)
       )
-    exclude_types: [image]
+    exclude_types: [svg, piskel]
   - id: trailing-whitespace
     exclude: |
       (?x)(
-        \.piskel$|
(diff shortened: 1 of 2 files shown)


manuq (approved):
Thanks for the instructions for updating the pre-commit hook.

PR #1292, opened 2025-10-03
shjourney: Replace xylophone sound effects

wjt:
In the main game, we have new, better-sounding sequence puzzle sounds. And unlike the xylophone samples that were used before, we have a full scale and a half, including all semitones.

Replace the xylophone sounds in the dragon puzzle with handbell sounds. Change the scale from C major (C D E F G A) to C minor (C D D♯=E♭ F G G♯=A♭): this matches the background music in this level, which is in C minor.


wjt:
@xXLeopoldoXx and team: please can you review this change?

xXLeopoldoXx:
> @xXLeopoldoXx and team: please can you review this change?

Of course @wjt , we will review it and adapt it

wjt:
No need to adapt: if it looks good, give it an approving review!

The other reason for this change is that in #1204 I plan to remove the xylophone noises

PR #1294, opened 2025-10-03
Replace godot_git_describe with minimal export-only plugin

wjt:
godot_git_describe works by changing settings in project.godot to the output of `git describe`. One of its interesting features is that it doesn't only do this when the project is exported: it also hooks the debugger and changes the setting when the project is run in the editor.

This is useful behaviour because it makes it much easier to test UI that uses these properties from the editor.

But it turns out to be annoying that project.godot is modified while the game is running. People often commit all changes in the tree, with `git commit -a` or similar, and this ends up accidentally including the version for whatever Git revision they are on.

Replace that plugin with a from-scratch implementation of the same general idea, but only registering an EditorExportPlugin rather than also hooking the debugger. This implementation is much simpler: it just runs a single, hard-coded command, and we assume that `git` can be invoked by that name. I'm not particularly concerned whether this works on non-Linux platforms because it's only really important in our CI.

I imagined a more complicated implementation where you define a resource:

    class_name GitDescribe
    extends Resource

    @export var version: String

    func get_version() -> String:
        if version:
            return version

        return `git describe --tags`

then make the export plugin customise it:

    func _customize_resource(resource: Resource, path: String) -> Resource:
        if resource is GitDescribe:
            resource = resource.duplicate()
            resource.version = resource.get_version()
            return resource

But this seems more complicated for not much upside compared to a simpler implementation.

Fixes https://github.com/endlessm/threadbare/issues/1163


addons/threadbare_git_describe/export.gd
```diff
@@ -0,0 +1,30 @@
...
+func _export_begin(
+	_features: PackedStringArray, _is_debug: bool, _path: String, _flags: int
+) -> void:
+	var output: Array[String] = []
+	var ret := OS.execute("git", ["describe", "--tags"], output)
+	if ret != 0:
+		printerr("git describe --tags failed: %d" % ret)
+	else:
+		var version := output[0].strip_edges()
+		set_version(version)
```
manuq, line 26: Looks good!

PR #1295, opened 2025-10-06
Re-enable threadbare_project_settings plugin

wjt:
This was accidentally disabled in commit
783701b70678719f77ae2b15415a459a5b06ceb6, probably by me when I resolved
merge conflicts in this file.

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -59,7 +59,7 @@ window/size/mode.web=0
 
 [editor_plugins]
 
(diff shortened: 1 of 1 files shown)


wjt:
It's really annoying that the list of enabled plugins is stored in this format that is guaranteed to conflict if parallel changes are made…

manuq (approved):

PR #1297, opened 2025-10-06
ci: Use treeless clone for export workflow

wjt:
ci: Use treeless clone for export workflow

When building & exporting the game, we need the full history of commits so that
`git describe` works, but we do not need the contents of any historical commits.
It is wasteful to download it all, only to discard it as soon as the clone is
finished.

Use a treeless clone, which fetches the full commit history, but only downloads
trees (and blobs, i.e. files) when needed.

https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -31,9 +31,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           lfs: true
-          # TODO: can we use --filter to avoid fetching the trees & blobs for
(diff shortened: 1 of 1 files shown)


wjt:
Doesn't work:

```
  /usr/bin/git lfs fetch origin refs/remotes/pull/1297/merge
  Fetching reference refs/remotes/pull/1297/merge
  Could not scan for Git LFS files
```

OK, never mind!

PR #1299, opened 2025-10-06
Reindent REUSE.toml; move Flatpak manifest

wjt:
Having these two files at the top level of the repo means it's easy to accidentally edit them in Godot. And in both cases, opening & saving the file would reindent it with tabs rather than spaces.

Fix this in 2 different ways:

- Reindent `REUSE.toml`
- Move the Flatpak manifest

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -117,7 +117,7 @@ jobs:
       - uses: flatpak/flatpak-github-actions/flatpak-builder@v6
         with:
           bundle: threadbare.flatpak
-          manifest-path: org.endlessaccess.threadbare.yml
+          manifest-path: linux/org.endlessaccess.threadbare.yml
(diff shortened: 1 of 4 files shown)


wjt:
I cancelled the build jobs because I'm retriggering with the Flatpak build enabled:

https://github.com/endlessm/threadbare/actions/runs/18289009956

Oh hey, actually, we can make the default conditional on the paths that are modified!

.github/workflows/export.yml
```diff
@@ -99,7 +108,7 @@ jobs:
   flatpak:
     name: Build Flatpak
     needs: build
-    if: ${{ inputs.build_flatpak || github.event_name == 'release' }}
+    if: ${{ inputs.build_flatpak || github.event_name == 'release' || needs.build.outputs.linux-changed == 'true' }}
```
manuq, line 111: 👍

manuq (approved):

PR #1300, opened 2025-10-06
Replace dialogue SFX with “wah wah” vocalisation

wjt:
This isn't perfect - I think it's a bit too silly, and it seems wrong for narration without a speaker. But this is more what I imagine it might sound like.

There are actually a few pretty good gibberish vocalisations on Freesound!


manuq:
> e.g. https://endlessm.github.io/threadbare/branches/endlessm/dialogue-wah-wah/#res://scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn

Funny!

PR #1302, opened 2025-10-07
Fix path walk behavior

manuq:
These are a number of fixes found while doing https://github.com/endlessm/threadbare/pull/1268


scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -437,35 +431,6 @@ func edit_patrol_path() -> void:
 		editor_interface.edit_node.call_deferred(patrol_path)
 
 
-func _draw() -> void:
-	## Only draw the patrol path debug line if we are in the editor
-	## and the guard node is selected.
```
wjt, line 442: I think this change makes sense. I looked into whether one can change the color drawn in the editor for a Path2D in a different way. The answer is "kind of". Path3D has a `debug_custom_color` but Path2D doesn't: https://github.com/godotengine/godot-proposals/issues/12108#issuecomment-2766572610. The best available workaround is using self_modulate but that doesn't really work because it modulates the default green colour.

If `Path2D.debug_custom_color` existed then we could readd a similar behaviour by adjusting that colour when the PathFollowBehavior is selected in the editor, and resetting it when deselected.

scenes/game_logic/walk_behaviors/path_walk_behavior.gd
```diff
@@ -144,23 +144,39 @@ func _physics_process(delta: float) -> void:
 				direction *= -1
 
 
+func _is_curve_smooth(point_in: Vector2, point_out: Vector2) -> bool:
```
wjt, line 182: **PathWalkBehavior: Fix check for pointy in closed path**

This commit looks good but its message ends with:

```
If the path is closed, then the check for and the out control point of
the
```

which looks unfinished.

scenes/game_logic/base_character_behavior.gd
```diff
@@ -1,8 +1,7 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 @tool
-# TODO: Prepend "abstract" when switching to Godot 4.5:
-class_name BaseCharacterBehavior
+@abstract class_name BaseCharacterBehavior
```
wjt: The only reason to hold off on this is that currently https://github.com/inkandswitch/patchwork-godot-plugin/ is based on Godot 4.4; currently one can test it with Threadbare because we don't actually use any 4.5-specific features, but this would change that.

scenes/game_elements/characters/enemies/guard/components/guard.gd
manuq: True, this is a diff I have in my `project.godot` right now:
```
shapes/paths/geometry_color=Color(0.72156864, 0.003921569, 0.9098039, 0.7764706)
```
I had to change it away from the default color for debugging to red / purple and add more opacity, because the floors in the game are usually green / yellow.

scenes/game_logic/walk_behaviors/path_walk_behavior.gd
manuq: Oh good catch! That was a left over, I forgot to remove the line. I merged that explanation into the first paragraph "Add a helper function _is_curve_smooth() which normally receives the in and out control points of the same vertex, but for the special case receives the control points of the two endings."

scenes/game_logic/base_character_behavior.gd
manuq: It is a very important reason, thanks for letting me know.

scenes/game_elements/characters/enemies/guard/components/guard.gd
wjt: Interesting - maybe we should commit that?

PR #1307, opened 2025-10-07
Remove fallback to attack_01 for repel action

### LINKED ISSUE #1305: repel_deprecated fallback doesn't work
### Description

In commit 96876c9f0e950f6bb8761272c214d91234232f84, we changed `player.tscn` to use the `attack_02` animation rather than `attack_01`. A fallback path was added to detect the case where the `attack_02` animation is missing and continue to use `attack_01` instead.

Unfortunately this doesn't work. The `repel` animation doesn't only animate the player: it also drives the AirStream and AirStreamCollision nodes, which are what actually does the repelling.

This breaks the Stella and Sueños Nocturnos combat scenes.

We should probably fix those in-tree quests by simply renaming `attack_01` to `attack_02` in the SpriteFrames. But having done that, we have 2 choices:

1. Delete the fallback;
2. Fix it.

### XP Summary (total 225)
| Subdomain | Skill | XP |
| - | - | - |

### Checklist

wjt:
As described in #1305, this fallback does not actually work correctly.

Update all SpriteFrames in the game to implement attack_02 where necessary.

Then remove the incomplete fallback path.

Resolves #1305


wjt:
That explains it! Thanks for doing the additional detective work. I was confused because I was sure the fallback path had worked before.

manuq:
> That explains it! Thanks for doing the additional detective work. I was confused because I was sure the fallback path had worked before.

This was probably due to an interactive rebase I did at one point. My bad!

PR #1308, opened 2025-10-07
Add ThrowProjectileBehavior

manuq:
A behavior node to throw (spawn or instantiate) projectile scenes. The projectile scenes can be any Node2D, but if it is a RigidBody2D then an initial impulse will be applied.

This is intended as an easy way to make a character throw projectiles.

[Grabación de pantalla desde 2025-10-07 17-04-57.webm](https://github.com/user-attachments/assets/21e53621-b402-4032-beb2-5f4410c65c89)

The existing throwing enemy is inflexible: the target is hardcoded to be the player character and its logic to the "fill barrels" game. This behavior node can also be treated as a generic spawner to instantiate other game entities.

### DIFF
--- a/scenes/game_logic/throw_projectile_behavior.gd
+++ b/scenes/game_logic/throw_projectile_behavior.gd
@@ -18,7 +18,8 @@ enum AddMode {
 @export var projectile_scenes: Array[PackedScene]
 
 ## The target to aim the throw.
-@export var target: Node2D
+@export var target: Node2D:
+	set = _set_target
 
 ## The projectile will be instantiated at this distance from this node,
 ## in the direction of the [member target].
@@ -28,18 +29,54 @@ enum AddMode {
 @export_range(10., 1000., 5., "or_greater", "or_less", "suffix:m/s") var impulse: float = 30.0
 
 ## Where to add the projectiles.
-@export var add_mode: AddMode
+@export var add_mode: AddMode:
+	set = _set_add_mode
 
 ## Projectiles may be added as children of this node depending on [member add_mode].
 ## [br][br]
 ## [b]Note:[/b] If there is a grandparent node and this isn't set,
 ## the grandparent node will be automatically assigned to this variable.
-@export var reference: Node2D
+@export var reference: Node2D:
+	set = _set_reference
+
+
+func _set_target(new_value: Node2D) -> void:
+	target = new_value
+	update_configuration_warnings()
+
+
+func _set_add_mode(new_value: AddMode) -> void:
+	add_mode = new_value
+	update_configuration_warnings()
+	notify_property_list_changed()
+
+
+func _set_reference(new_value: Node2D) -> void:
+	reference = new_value
+	update_configuration_warnings()
 
 
 func _enter_tree() -> void:
 	if not reference and get_parent() and get_parent().get_parent():
 		reference = get_parent().get_parent()
+
+
+func _validate_property(property: Dictionary) -> void:
+	match property["name"]:
+		"reference":
+			if add_mode == AddMode.CHILD_OF_ROOT:
+				property.usage |= PROPERTY_USAGE_READ_ONLY
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+	if add_mode == AddMode.CHILD_OF_REFERENCE and not reference:
+		warnings.append("Reference node must be set if Add Mode is Child of Reference")
+	if not projectile_scenes or projectile_scenes.is_empty():
+		warnings.append("At least one projectile scene must be set")
+	if not target:
+		warnings.append("Target must be set")
+	return warnings
 
 
 ## Instantiate the projectile and add it to the scene.


scenes/game_logic/throw_projectile_behavior.gd
```diff
@@ -0,0 +1,63 @@
...
+## depending on [member add_mode].
+## [br][br]
+## [b]Note:[/b] If there is a grandparent node and this isn't set,
+## the grandparent node will be automatically assigned to this variable.
+@export var reference: Node2D
+
+
+func _enter_tree() -> void:
+	if not reference and get_parent() and get_parent().get_parent():
+		reference = get_parent().get_parent()
```
manuq, line 61: I realize that for this to work in the editor it has to be a `@tool`.

scenes/game_logic/throw_projectile_behavior.gd
```diff
@@ -0,0 +1,64 @@
...
+## Behavior to throw projectiles.
+
+## Controls where the projectile instances will be added.
+enum AddMode {
+	## Add projectiles as child of the reference node.
+	CHILD_OF_REFERENCE,
+	## Add projectiles as siblings of the reference node.
+	SIBLING_OF_REFERENCE,
+	## Add projectiles to the current scene root.
+	CHILD_OF_ROOT,
```
wjt, line 15: Perhaps I'm misunderstanding the "reference" concept. Would the `reference` node typically be the CharacterBody2D, or its parent?

The reason I ask is:

- If it's typically the CharacterBody2D, why would you ever want a projectile to be a child of the character?
- If it's the CharacterBody2D's parent, why would you want SIBLING_OF_REFERENCE?

scenes/game_logic/throw_projectile_behavior.gd
```diff
@@ -0,0 +1,64 @@
...
+	if not reference and get_parent() and get_parent().get_parent():
+		reference = get_parent().get_parent()
+
+
+## Instantiate the projectile and add it to the scene.
+## [br][br]
+## Hint: You can connect a [signal Timer.timeout] signal to this method
+## for periodic throws.
+func throw() -> void:
+	var projectile := projectile_scene.instantiate()
```
wjt: In the current `throwing_enemy.gd`, instantiating a projectile copies a whole load of properties onto the new projectile. c.f. #1309.

I think this approach of simply instantiating a reference scene is a better approach but I wonder how you envisage varying certain projectile properties:

- For the LoreQuest fill game, we want to spawn projectiles that match the *current* barrels in the scene.
- For Stella, the original design was that only the starfish would match the boat, while the other fish would not match the boat. This proved impossible to implement with the provided components so we had to just allow any fish to fill the boat.

You could imagine doing this by changing the `projectile_scene` property to `Array[PackedScene]` and sampling randomly from it at each throw. Then:

- For Stella, the four invisible throwers would each choose randomly from the same 4 scenes, one of which is the starfish (and the barrel is configured to only accept that scene)
- For the LoreQuest combat scenes… perhaps all the enemies have the same `Array[PackedScene]` (same instance, not copies), and then something else in the scene keeps it up to date as barrels are filled?

scenes/game_logic/throw_projectile_behavior.gd
manuq: Ouch sorry that I forgot to answer here!

> Perhaps I'm misunderstanding the "reference" concept. Would the `reference` node typically be the CharacterBody2D, or its parent?

I was trying to add flexibility so "reference" could be the character (for sibling) or its parent (for add as child).

> The reason I ask is:
> 
> * If it's typically the CharacterBody2D, why would you ever want a projectile to be a child of the character?
> * If it's the CharacterBody2D's parent, why would you want SIBLING_OF_REFERENCE?

Yeah it's true that
- Add as sibling does only make sense if the reference is the character.
- Add as child does not make sense if the reference is the character.

After a week, I now thik that only 2 options are enough:
- "child of root": The existing way for the ink combat. Which spawns all projectiles on top of everything.
- "as child of node" (better name for calling this node "reference"?): Mainly for spawning the projectiles in a Y-sorted node.

scenes/game_logic/throw_projectile_behavior.gd
manuq: > In the current `throwing_enemy.gd`, instantiating a projectile copies a whole load of properties onto the new projectile. c.f. #1309.
> 
> I think this approach of simply instantiating a reference scene is a better approach but I wonder how you envisage varying certain projectile properties:
> 
> * For the LoreQuest fill game, we want to spawn projectiles that match the _current_ barrels in the scene.
> * For Stella, the original design was that only the starfish would match the boat, while the other fish would not match the boat. This proved impossible to implement with the provided components so we had to just allow any fish to fill the boat.
> 
> You could imagine doing this by changing the `projectile_scene` property to `Array[PackedScene]` and sampling randomly from it at each throw. Then:
> 
> * For Stella, the four invisible throwers would each choose randomly from the same 4 scenes, one of which is the starfish (and the barrel is configured to only accept that scene)
> * For the LoreQuest combat scenes… perhaps all the enemies have the same `Array[PackedScene]` (same instance, not copies), and then something else in the scene keeps it up to date as barrels are filled?

This is a great idea! Let's change the `projectile_scene` property to an `Array[PackedScene]`, and sample randomly from it at each throw.

scenes/game_logic/throw_projectile_behavior.gd
```diff
@@ -0,0 +1,60 @@
...
+@export_range(0., 100., 1., "or_greater", "suffix:m") var distance: float = 100.0
+
+## If the projectile is a [RigidBody2D], apply this impulse when throwing.
+@export_range(10., 1000., 5., "or_greater", "or_less", "suffix:m/s") var impulse: float = 30.0
+
+## Where to add the projectiles.
+@export var add_mode: AddMode
+
+## Projectiles may be added as children or siblings of this node
+## depending on [member add_mode].
```
wjt: I think "siblings" here is no longer true.

scenes/game_logic/throw_projectile_behavior.gd
```diff
@@ -0,0 +1,60 @@
...
+
+## The projectile will be instantiated at this distance from this node,
+## in the direction of the [member target].
+@export_range(0., 100., 1., "or_greater", "suffix:m") var distance: float = 100.0
+
+## If the projectile is a [RigidBody2D], apply this impulse when throwing.
+@export_range(10., 1000., 5., "or_greater", "or_less", "suffix:m/s") var impulse: float = 30.0
+
+## Where to add the projectiles.
+@export var add_mode: AddMode
```
wjt: I am not sure but shouldn't the default be to make the projectiles children of the scene root?

scenes/game_logic/throw_projectile_behavior.gd
```diff
@@ -0,0 +1,60 @@
...
+
+## Projectiles may be added as children or siblings of this node
+## depending on [member add_mode].
+## [br][br]
+## [b]Note:[/b] If there is a grandparent node and this isn't set,
+## the grandparent node will be automatically assigned to this variable.
+@export var reference: Node2D
```
wjt: Perhaps use a `_validate_property` hook to hide this property, or make it read-only, if `add_mode` is set to `CHILD_OF_ROOT`?

scenes/game_logic/throw_projectile_behavior.gd
manuq: Good catch!

scenes/game_logic/throw_projectile_behavior.gd
wjt: Actually on closer inspection I think this default behaviour makes sense. The root node may not have y-sorting enabled, but the enemy's parent node probably will.

scenes/game_logic/throw_projectile_behavior.gd
wjt: Pushed a patch to do this

PR #1310, opened 2025-10-08
Fix errors in Sueños Nocturnos tilesets

### LINKED ISSUE #1306: Tile-outside-texture errors from Sueños Nocturnos tilesets
I see the following when opening various scenes, or when exporting the project:

```
ERROR: Cannot create tile. The tile is outside the texture or tiles are already present in the space the tile would cover.
   at: create_tile (scene/resources/2d/tile_set.cpp:4986)
ERROR: The TileSetAtlasSource atlas has no tile at (2, 3).
   at: has_alternative_tile (scene/resources/2d/tile_set.cpp:5438)
ERROR: TileSetAtlasSource has no tile at (2, 3).
   at: create_alternative_tile (scene/resources/2d/tile_set.cpp:5386)
ERROR: Cannot create tile. The tile is outside the texture or tiles are already present in the space the tile would cover.
   at: create_tile (scene/resources/2d/tile_set.cpp:4986)
ERROR: The TileSetAtlasSource atlas has no tile at (3, 3).
   at: has_alternative_tile (scene/resources/2d/tile_set.cpp:5438)
ERROR: TileSetAtlasSource has no tile at (3, 3).
   at: create_alternative_tile (scene/resources/2d/tile_set.cpp:5386)

[repeated for many many pages]
```

As discussed at https://github.com/godotengine/godot/issues/109170 there is not enough information in the error message to identify the file at fault, let alone which atlas it is.

wjt:
Loading any of the Sueños Nocturnos scenes, or exporting the project, prints errors like the following, hundreds of times:

    ERROR: Cannot create tile. The tile is outside the texture or tiles are already present in the space the tile would cover.
    ERROR: The TileSetAtlasSource atlas has no tile at (3, 2).
    ERROR: TileSetAtlasSource has no tile at (3, 2).

These are caused by three different variations on the same problem:

1. TileSetAtlasSource with no texture assigned
2. TileSetAtlasSource with an empty texture assigned
3. TileSetAtlasSource with tiles drawn outside the area of the assigned texture

Fix each of these classes of problems. I also found an unused TileSet which I deleted as step 0.

Resolves #1306


wjt:
(I haven't tested the scenes that use these tilesets very thoroughly but, at a high level, they can't have used these tiles before because the tiles are missing.)

manuq (approved):

Issue #444, opened 2025-05-02
Loading screen for shader compilation


JuanFdS:
Right now, the first time an ink blob is shot, the game freezes for an instant. This is probably due to shader precompilation. If we add more particles effects like these, the game will have stutter each time they are shown for the first time.

If we want to add more effects like these, we'd need a loading screen at the beginning of the game that displays those effects for a frame so they don't cause stutter while playing.

### XP Summary (total 300)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Optimization & Performance | Problem Solving | 75 |
| Engineering: Optimization & Performance | Creativity & Design Thinking | 75 |
| Engineering: Optimization & Performance | Technical Literacy | 150 |

wjt:
Unfortunately Shader Baker is not supported in our case. https://docs.godotengine.org/en/stable/tutorials/performance/pipeline_compilations.html#shader-baker documents that only Vulkan and D3D12 are supported. We use the Compatibility (GL) renderer, which is the only one supported on the web platform.

But looking at https://docs.godotengine.org/en/stable/tutorials/performance/pipeline_compilations.html#pipeline-precompilation-instancing there is another approach: add a disabled, hidden instance of the ink blob projectile to the first scene that uses it. This causes Godot 4.4+ to precompile it.t

PR #1311, opened 2025-10-08
export_presets: Update with Godot 4.5 options

wjt:
This is just the result of exporting the project in Godot 4.5.

### DIFF
--- a/export_presets.cfg
+++ b/export_presets.cfg
@@ -54,6 +54,8 @@ progressive_web_app/icon_144x144=""
 progressive_web_app/icon_180x180=""
 progressive_web_app/icon_512x512=""
 progressive_web_app/background_color=Color(0, 0, 0, 1)
+threads/emscripten_pool_size=8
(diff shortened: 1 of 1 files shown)


export_presets.cfg
```diff
@@ -83,6 +85,7 @@ debug/export_console_wrapper=1
 binary_format/embed_pck=false
 texture_format/s3tc_bptc=true
 texture_format/etc2_astc=false
+shader_baker/enabled=false
```
manuq, line 88: I wonder if enabling shader baker would help with the freeze the first time an ink blob appears https://github.com/endlessm/threadbare/issues/444 . Although there are tradeoffs https://godotengine.org/article/dev-snapshot-godot-4-5-dev-5/#shader-baker

wjt: Interesting!

But this line is for the Linux export preset. I checked the settings and the option isn't offered for the web target. I wonder why not…

PR #1312, opened 2025-10-08
CODEOWNERS: Mark VERSO team as joint owners of their quest

wjt:


.github/CODEOWNERS
```diff
@@ -46,3 +46,4 @@ docs/issue_template.md    @endlessm/community
 # StoryQuest maintainers
 scenes/quests/story_quests/stella @endlessm/games @endlessm/learning
 scenes/quests/story_quests/shjourney @endlessm/games @endlessm/suenos-nocturnos
+scenes/quests/story_quests/verso     @endlessm/games @endlessm/verso
```
manuq, line 49: Should we also add their individual names to the AUTHORS file?

manuq (approved):

.github/CODEOWNERS
wjt: I pushed a patch to add them to AUTHORS.

wjt:
I'm bypassing the rules because "Build for web" will never succeed; #1314 is merged so if I refreshed the branch this would build successfully but that's a waste of electrons.

PR #1321, opened 2025-10-09
Guard: Fix flipping the sprite horizontally

manuq:
In 57b0a2e5 I removed more than necessary:

```
if not is_zero_approx(velocity.x):
    animated_sprite_2d.flip_h = velocity.x < 0
```

Bring it back, but now using CharacterSpriteBehavior and CharacterAnimationPlayerBehavior nodes.

When the guard is alerted, disable the node controlling the AnimationPlayer for walking / being idle, so it can be disabled to play the alerted animation.

Inside the scene file, the position of the idle animation subresource changes for no reason.

### DIFF
--- a/scenes/game_elements/characters/enemies/guard/components/guard.gd
+++ b/scenes/game_elements/characters/enemies/guard/components/guard.gd
@@ -98,6 +98,13 @@ var _player: Player
 @onready var sight_ray_cast: RayCast2D = %SightRayCast
 ## Control to hold debug info that can be toggled on or off.
 @onready var debug_info: Label = %DebugInfo
+
+## Reference to the node controlling the AnimationPlayer for walking / being idle,
+## so it can be disabled to play the alerted animation.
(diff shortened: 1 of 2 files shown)


scenes/game_elements/characters/enemies/guard/components/guard.gd
```diff
@@ -269,6 +276,7 @@ func _set_state(new_state: State) -> void:
 			if not _alert_sound.playing:
 				_alert_sound.play()
 		State.ALERTED:
+			character_animation_player_behavior.process_mode = Node.PROCESS_MODE_DISABLED
```
wjt, line 279: Don't you need to re-enable the behavior on moving to a different state?

manuq: Nope because "alerted" is the final state. There is no way for the guard to go from alerted to another state.

PR #1329, opened 2025-10-13
Nora: Reset combat & stealth camera zoom to 1×

wjt:
Previously these were set to 1.4. This causes ugly flickering on 1px
horizontal/vertical lines as the camera pans around. I think it's a
leftover from the game being at 1080p resolution.

Reset the zoom to 1×, which also makes a little more of the level
visible (which I think makes it fairer).

The sequence puzzle is set to 0.7×; this has the same problem but is not
so easy to fix because if you reset it to 1× then you can't see all of
the objects when interacting with the book.


wjt:
I'm merging this without review because it's such a simple change. Post-merge review welcome of course!

Issue #1143, opened 2025-09-02
Collectible threads HUD is too large


wjt:
### Description

Since the switch to 720p as the base resolution, the HUD in the top-right corner of quests is ridiculously big, e.g.:

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/7d722016-6e90-4a21-ab63-92a740fb07e7" />

Happily the underlying thread textures, e.g. `res://assets/first_party/collectibles/memory.png`, are 32×32.

The banner is bigger but that should be a solvable problem!

To fix this problem, we should adjust `res://scenes/ui_elements/story_quest_progress/story_quest_progress.tscn` and its constituent scenes to be smaller. Note that there is a "collected" animation that may need to be adjusted. (Currently the sizes seem to be set absolutely rather than relying on the textures' natural sizes.)

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 50 |
| Game Design: Game Feel & UX | Technical Literacy | 75 |

PR #1342, opened 2025-10-13
HUD: Reduce its size

manuq:
Set a minimum size of 32x32 to items (which are  TextureRects). This is the same size as the actual PNG asset.

Adjust the stylebox texture override. Set texture margins to 64 (it was 65 which is wrong). Use expand margins and content margins to reduce the transparent zone at the top and right of the texture.

Before / after:

Captura desde 2025-10-13 15-21-10

Captura desde 2025-10-13 15-28-04

Captura desde 2025-10-13 15-50-01

(The one in the middle is an alternative solution, with the thread textures at double size, 64x64)

Fix https://github.com/endlessm/threadbare/issues/1143

### DIFF
--- a/scenes/ui_elements/story_quest_progress/story_quest_progress.tscn
+++ b/scenes/ui_elements/story_quest_progress/story_quest_progress.tscn
@@ -5,11 +5,15 @@
 [ext_resource type="PackedScene" uid="uid://1mjm4atk2j6e" path="res://scenes/ui_elements/story_quest_progress/components/item_slot/item_slot.tscn" id="3_mwl7j"]
 
 [sub_resource type="StyleBoxTexture" id="StyleBoxTexture_hqnrr"]
+content_margin_top = 32.0
+content_margin_right = 48.0
 texture = ExtResource("1_cikk2")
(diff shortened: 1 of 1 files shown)


scenes/ui_elements/story_quest_progress/story_quest_progress.tscn
```diff
@@ -5,33 +5,45 @@
 [ext_resource type="PackedScene" uid="uid://1mjm4atk2j6e" path="res://scenes/ui_elements/story_quest_progress/components/item_slot/item_slot.tscn" id="3_mwl7j"]
 
 [sub_resource type="StyleBoxTexture" id="StyleBoxTexture_hqnrr"]
+content_margin_top = 32.0
+content_margin_right = 48.0
 texture = ExtResource("1_cikk2")
-texture_margin_left = 65.0
-texture_margin_top = 65.0
-texture_margin_right = 65.0
-texture_margin_bottom = 65.0
```
manuq, line 12: These 65s were wrong! The 192x192 nine patch texture is meant to be split in 64x64 patches!

PR #1344, opened 2025-10-14
storyquest_bootstrap: Copy template recursively

### LINKED ISSUE #1214: Extend template instantiation tool
A request from our learning team, building on https://github.com/endlessm/threadbare/pull/909:

## User Story

- **When I** copy a template folder (intro, outro, minigames),
- **I want to** automatically duplicate all included files, subfolders, and referenced resources (e.g., `.tres` dialogue files, sprite frames, tile sets), with scenes relinked to these new copies,
- **So I can** safely edit my own StoryQuest without breaking templates or the base game.

## Requirements

Full duplication:

- Copy the entire folder structure (intro, outro, 3 minigames)
- As more mini-games are added, create the option to duplicate mini-games as a standalone process.
- Include all nested subfolders and files.

Resource duplication:

- Duplicate all resources referenced by those scenes: .tres files, sprite frames, tile sets, etc.

Resource relinking:

- Ensure the new scenes point to their duplicated resources, not originals.
- No references in the learner’s copy should remain tied to the template folder or base game files.

Naming & organization:

- Place copies under a new learner-specific folder.
- (Optional) support renaming all instances of “template” with a chosen StoryQuest name (not just removing the word “template” from the duplicated files).

Robustness

- Learners should never need to manually rename or relink files.
- The copied folder should be ready to open and edit immediately.

wjt:
Previously this plugin only copied the opening and closing cutscenes, and relinked them. We would like the plugin to copy the whole NO_EDIT template, including the 3 challenges.

The previous approach was to hard-code which elements of the two scenes need to be copied and relinked. This works fine for these simple scenes but to handle the complex interactive scenes we need a more general approach.

Define a copy() function which copies a resource, delegating to per-type special cases to handle each type of resource that we need to handle: Quest, PackedScene, CompressedTexture2D, DialogueResource, and SpriteFrames. For scenes, check every property of every node in the scene which could be either a Resource or string UID, and recursively call copy() if it points to a resource in the NO_EDIT template.

Since this is a graph, not a tree, maintain a cache of source UID to copied resource, and consult this before copying any given resource.

Kick off the process by copying quest.tres, which is the root of the graph.

Resolves #1214

----

In addition, I fixed a format string error in quest.gd; and added a crude progress indicator during the copy since it now takes longer.

---
  - https://app.asana.com/0/0/1211587602398303

### DIFF
--- a/addons/storyquest_bootstrap/new_storyquest_dialog.gd
+++ b/addons/storyquest_bootstrap/new_storyquest_dialog.gd
@@ -4,6 +4,7 @@
 extends Window
 
 signal create_storyquest(title: String, description: String, filename: String)
+signal cancel
 
 @export var storyquests_path: String
 @export var validate_title: Callable
@@ -19,6 +20,7 @@ var _filename: String
 @onready var errors_label: RichTextLabel = %ErrorsLabel
 @onready var folder_res_label: Label = %FolderResLabel
 @onready var description_edit: TextEdit = %DescriptionEdit
+@onready var progress_bar: ProgressBar = %ProgressBar
 
 
 func _ready() -> void:
@@ -32,7 +34,17 @@ func _ready() -> void:
 
 
 func _on_create_button_pressed() -> void:
+	title_edit.editable = false
+	description_edit.editable = false
+	create_button.disabled = true
+	progress_bar.visible = true
+	progress_bar.indeterminate = true
 	create_storyquest.emit(_title, _description, _filename)
+
+
+func _on_close_requested() -> void:
+	if not progress_bar.visible:
+		cancel.emit()
 
 
 func _make_filename(title: String) -> String:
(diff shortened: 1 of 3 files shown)


scenes/menus/storybook/components/quest.gd
```diff
@@ -58,7 +58,7 @@ func _validate_property(property: Dictionary) -> void:
 
 
 func _to_string() -> String:
-	return '<Quest %s: "%s">' % [resource_path]
+	return '<Quest %s: "%s">' % [resource_path, title]
```
manuq, line 61: Ouch!

addons/storyquest_bootstrap/copier.gd
```diff
@@ -0,0 +1,210 @@
...
+	await reimport(copy_path)
+	return load(copy_path)
+
+
+func copy(uid: String, resource: Resource) -> Resource:
+	assert(uid.begins_with("uid://"), "'%s' is not a uid" % uid)
+	assert(
+		resource.resource_path.begins_with(TEMPLATE_PATH),
+		"'%s' is not a storyquest resource" % resource
+	)
```
manuq, line 157: Nice, this prevents the recursive copy from jumping outside of the template directory.

Issue #1350, opened 2025-10-14
Grappling hook: Keep momentum


manuq:
If the player is standing and use the grappling hook on a pin, the gained velocity moves the player far:

https://github.com/user-attachments/assets/c166c503-7131-4f63-a348-14948f9a488c

But if the player is walking or even running, the gained velocity is overriden by the walk/run velocity:

https://github.com/user-attachments/assets/e3a1d258-8651-41a0-a81f-db74591b32cc

This is unacceptable if we want to combine the grappling hook with the Void runner.

PR #1352, opened 2025-10-14
Player: Repurpose stopping and moving steps

manuq:
These exported attributes control how fast the player velocity reaches the desired velocity as determined by the input controls.

Previously, when the player released the input actions (the input vector was zero), stopping_step was used. And otherwise moving_step was used.

Now use stopping_step if the desired velocity is smaller than the actual velocity. And use moving_step otherwise.

Do the same in InputWalkBehavior. And adopt the format of this one in player.gd.

In the video, I'm using the keyboard when going from left to right, and a controller when going from right to left:

[input.webm](https://github.com/user-attachments/assets/3be1841b-732f-412f-8c75-225ebe252d20)

Fix https://github.com/endlessm/threadbare/issues/1350


wjt (approved):
This feels better. You could imagine renaming these to "acceleration" and "deceleration" - that's what the new definition means I think - but they're just names and I think the docstring is fine.

PR #1353, opened 2025-10-14
Void particles: Make visibility rect big enough

manuq:
Since the Void enemy has a trail of particles, the particles may dissapear abruptly when the enemy goes off screen. Making the visibility rect big enough ensures that it doesn't happen.

Before:

[particles-before.webm](https://github.com/user-attachments/assets/7345dcc8-66f4-4811-aa59-fbb264d2d591)

After:

[particles-after.webm](https://github.com/user-attachments/assets/c0651957-966f-4f15-a5a3-7e3e97179938)

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_particles.tscn
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_particles.tscn
@@ -44,4 +44,5 @@ amount = 80
 texture = ExtResource("1_hohl7")
 one_shot = true
 explosiveness = 1.0
+visibility_rect = Rect2(-1000, -1000, 2000, 2000)
 process_material = SubResource("ParticleProcessMaterial_rrhr1")


scenes/quests/lore_quests/quest_002/1_void_runner/components/void_particles.tscn
```diff
@@ -44,4 +44,5 @@ amount = 80
 texture = ExtResource("1_hohl7")
 one_shot = true
 explosiveness = 1.0
+visibility_rect = Rect2(-1000, -1000, 2000, 2000)
```
manuq, line 47: Honestly, I made this big enough but I don't really know what would be a sensible value.

wjt: I have never seen this property before!

I guess we can calculate how big it needs to be:

- The enemy's max speed is (say) 500 px/s
- The player's max speed is also 500 px/s
- So the max relative speed between the player node and the particles node is 1000px/s
- The lifetime of the particles is 1 second - so the maximum distance that can be closed between the player and the particles over their lifetime is 1000px

So I think you're right, it needs to be 1000px on each side of the node. If the Particles node is 1000px outside the viewport at time T, it can be *just* in the viewport at T + 1s.

What I don't understand about this property - and actually the whole particle node - is that in the inspector the box is 2000px tall but only 400px wide:

image

but similarly the circle representing the emission area (I think) is weirdly an ellipse.

manuq: Nice calculation. I found that the weird ellipse is because of the frame-by-frame animation, which is using a horizontal spritesheet in this case:

[Grabación de pantalla desde 2025-10-15 07-05-41.webm](https://github.com/user-attachments/assets/91f6841e-dba2-4d93-9293-2eab06d90180)

I'll file a bug upstream.

manuq: Filed as https://github.com/godotengine/godot/issues/111677

Issue #1355, opened 2025-10-14
Grappling hook: Improve controls


manuq:
Feedback from playtest https://github.com/endlessm/threadbare/discussions/1327

- Mouse + keyboard: mouse left click "fires" the string (mouse move already aims).
- Joypad: right trigger (R2) "fires" the string, right stick aims.

Issue #1303, opened 2025-10-07
Add dedicated up/down/left/right/interact/repel actions


wjt:
Currently we (ab)use the built-in `ui_up`, `ui_down`, `ui_left`, `ui_right` actions for in-game movement, and `ui_accept` for interacting and for repelling projectile during combat. These are meant to be used by Godot's `Control` elements, as used in menus, not for in-game movement. We should:

1. Add `up`, `down`, `left`, and `right` actions to the game config, with the same keyboard/controller bindings as we have configured for the corresponding `ui_` action.
2. Add a `repel` action, with the same bindings as `ui_accept`.
3. Add an `interact` action, again with the same bindings.
4. Find all references to the `ui_*` actions in the game (excluding the `addons/` folder) and replace them with these new actions as appropriate.

You may find this command useful to search the project:

```console
$ git grep -E 'ui_(up|down|left|right|accept)' ':!addons'
```

(This uses an [extended regular expression](https://git-scm.com/docs/git-grep#Documentation/git-grep.txt--E) and a fancy [pathspec](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-pathspec).)

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

PR #1356, opened 2025-10-15
Refactor input actions

manuq:
godot.project: Remove all customizations to the built-in actions (ui_accept, ui_cancel, ui_up, ui_down, ui_left, ui_right).

Add specific actions:

- move_up, move_down, move_left, move_right for movement.

- aim_up, aim_down, aim_left, aim_right for aiming with the grappling hook.

- interact, repel, throw.

Do the mapping as before, with the following changes on top:

- throw: Left mouse button, joypad right trigger (Sony R2, Xbox RT)

- aim: Use the right stick for joypads.

- dialogue_skip: Use it also for quitting the Storybook.

Also update throw_hint.tscn with the joypad right trigger textures of different consoles.

Also add an event check for keyboard in hook_control.tscn for the keyboard special case.

Note: The grappling_hook_start.tscn is still using the movement_input_hints.tscn which is now incorrect for joypad (it should be the right stick). This is left intentionally out because none of the existing scripts for input hints (gamepad_input.gd or interact_input.gd) can be used as-is.

Fix https://github.com/endlessm/threadbare/issues/1303

Fix https://github.com/endlessm/threadbare/issues/1355


manuq:
While testing this with a joypad, I am no longer able to navigate the UI menus. it looks like we do need to overload `ui_accept` with a joypad button. Or am I missing something?

project.godot
```diff
@@ -83,52 +83,6 @@ hook_listener=""
...
-]
-}
-ui_down={
-"deadzone": 0.5,
-"events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":4194322,"physical_keycode":0,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)
-, Object(InputEventJoypadButton,"resource_local_to_scene":false,"resource_name":"","device":0,"button_index":12,"pressure":0.0,"pressed":false,"script":null)
-, Object(InputEventJoypadMotion,"resource_local_to_scene":false,"resource_name":"","device":0,"axis":1,"axis_value":1.0,"script":null)
-, Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":83,"key_label":0,"unicode":115,"location":0,"echo":false,"script":null)
-]
-}
```
wjt, line 131: Yes, I think we need to keep the joypad bindings for the ui_* actions to be able to navigate menus.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -156,7 +156,7 @@ func _on_right_button_pressed() -> void:
 
 
 func _input(event: InputEvent) -> void:
-	if event.is_action_pressed("ui_cancel"):
```
wjt, line 159: I think ui_cancel is the correct action here. (The storybook is a fancy menu.)

scenes/ui_elements/input_hints/throw_hint.tscn
```diff
@@ -1,12 +1,13 @@
...
 [ext_resource type="Script" uid="uid://cbj0406q05dly" path="res://scenes/game_elements/props/hint/input_key/interact_input.gd" id="3_g0g4i"]
 [ext_resource type="Texture2D" uid="uid://p8kty3uxifxu" path="res://assets/third_party/inputs/keyboard-and-mouse/Dark/Space_Key_Dark.png" id="4_81p17"]
-[ext_resource type="Texture2D" uid="uid://dm4a8fn88sjf5" path="res://assets/third_party/inputs/PS5/PS5_Cross.png" id="5_vu6nt"]
-[ext_resource type="Texture2D" uid="uid://bc2ktlkypid78" path="res://assets/third_party/inputs/Switch/Switch_A.png" id="6_trb8c"]
-[ext_resource type="Texture2D" uid="uid://dul7b6elfhcm0" path="res://assets/third_party/inputs/Steam Deck/SteamDeck_A.png" id="7_3bs2i"]
+[ext_resource type="Texture2D" uid="uid://8vjuvwg6uvkq" path="res://assets/third_party/inputs/Xbox Series/XboxSeriesX_RT.png" id="5_cgyew"]
+[ext_resource type="Texture2D" uid="uid://bersmjts8ia2e" path="res://assets/third_party/inputs/PS5/PS5_R2.png" id="6_g0g4i"]
+[ext_resource type="Texture2D" uid="uid://22x1c6271exs" path="res://assets/third_party/inputs/Switch/Switch_RT.png" id="7_81p17"]
```
wjt, line 9: I believe you have a Switch so I'll take your word for it that this is the right label for the button! :)

scenes/menus/storybook/components/storybook.gd
manuq: Ah good catch.

project.godot
manuq: Excellent, I'll bring them back.

scenes/ui_elements/input_hints/throw_hint.tscn
manuq: > I believe you have a Switch so I'll take your word for it that this is the right label for the button! :)

That's the one I'm guessing. I'll connect the joycons via Bluetooth to confirm.

project.godot
manuq: @wjt I bring back joypad button 0 (Xbox A) for `ui_accept`. And with that I'm able to navigate the menus as before. Is there anything else in the ui_* actions I need to keep?

wjt: I had mapped `B` to `ui_cancel` and it can be used to go back on some menus

wjt: I amended your patch to:

1. Bind `B` to `ui_cancel`
2. Set the `ui_accept` and `ui_cancel` bindings to only apply to device 0. I think this is weird but it matches the built-in `ui_{up,down,left,right}` bindings, and it's what we had before.

wjt (approved):

project.godot
manuq: Excellent!

PR #1357, opened 2025-10-15
Prefix quest files (not dirs) with quest name

wjt:
Previously, bootstrapping a quest named "dennis" would copy `scenes/quests/story_quests/NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth_components/NO_EDIT_stealth.dialogue` to `scenes/quests/story_quests/dennis/1_stealth/stealth_components/stealth.dialogue`.

This doesn't match what we have been teaching students to do by hand, which is to name
`scenes/quests/story_quests/dennis/1_dennis_stealth/dennis_stealth_components/dennis_stealth.dialogue`.

Compromise on including the quest name in the filename, to make it always visible when the filename is shown, but not in the intermediate folders:
`scenes/quests/story_quests/dennis/1_stealth/stealth_components/dennis_stealth.dialogue`.

This does have the disadvantage that if a learner names their quest “It was the best of time, it was the worst of times” then all the filenames will begin with `it_was_the_best_of_times_it_was_the_worst_of_times`, and if they want to change this they have to rename ~66 files, whereas with the previous behaviour, they would only have to rename the quest's root folder of their quest.

### DIFF
--- a/addons/storyquest_bootstrap/new_storyquest_dialog.tscn
+++ b/addons/storyquest_bootstrap/new_storyquest_dialog.tscn
@@ -6,6 +6,7 @@
 bg_color = Color(0, 0, 0, 1)
 
 [node name="StoryQuestDialogue" type="Window"]
+oversampling_override = 1.0
 title = "Create StoryQuest"
 initial_position = 1
 size = Vector2i(600, 400)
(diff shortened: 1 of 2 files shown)


wjt:
@jgbourque but for now - shall we change the behaviour in the way implemented here?

wjt:
I chose the limit of 20 characters because `el_juguete_perdido` is 18 characters and that seems OK.

addons/storyquest_bootstrap/new_storyquest_dialog.tscn
```diff
@@ -6,6 +6,7 @@
 bg_color = Color(0, 0, 0, 1)
 
 [node name="StoryQuestDialogue" type="Window"]
+oversampling_override = 1.0
```
manuq, line 9: Is this intentional?

manuq (approved):
Excellent, I'm looking forward for an ice cream dreams story quest!

jgbourque:
This is great thanks @wjt

addons/storyquest_bootstrap/new_storyquest_dialog.tscn
wjt: No - but Godot really really wants to set this whenever I open and save the scene.

PR #1367, opened 2025-10-16
Void spreading enemy: Defeat player while patrolling

manuq:
And also make it possible to follow (but not caught) other innocent creatures of Threadbare (or any Node2D).

Remove the "is Player" restriction when starting the chase.

The enemy can now defeat the player regardless of the current state or who is following.

This is a safe change for the existing Void runner level because the player can't collide with the enemy while patrolling in the existing path.


wjt (approved):
Nice!

PR #1370, opened 2025-10-17
Void spreading enemy: Add defeat method

manuq:
When called, wait for all particles to finish and then queue free.

This can be a callback of an Area2D body_entered signal.

This allows a scene to have with multiple Void enemies, each one chasing the player after the other. The player has to run (or walk!) from the Area2D that triggers the chasing to the Area2D that defeats the enemy.

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
@@ -11,6 +11,8 @@ enum State {
 	CHASING,
 	## The void has engulfed the player
 	CAUGHT,
+	## The void has been defeated
+	DEFEATED,
 }
 
 const VOID_PARTICLES = preload(
@@ -40,6 +42,7 @@ var state := State.IDLE:
 
 var _last_position: Vector2
 var _distance_since_emit: float = 0.0
+var _live_particles: int = 0
 
 @onready var path_walk_behavior: PathWalkBehavior = %PathWalkBehavior
 @onready var follow_walk_behavior: NavigationFollowWalkBehavior = %NavigationFollowWalkBehavior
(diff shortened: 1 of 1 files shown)


scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.gd
```diff
@@ -85,7 +90,18 @@ func start(detected_node: Node2D) -> void:
 		state = State.CHASING
 
 
+func defeat() -> void:
+	state = State.DEFEATED
+	for particles: GPUParticles2D in get_children().filter(
+		func(c: Node) -> bool: return c is GPUParticles2D
+	):
+		await particles.finished
```
wjt: I think this might work in practice but in theory it won't work as you expect. Suppose you have 2 GPUParticles2D child nodes, A and B. Suppose they are in the scene tree in that order. But suppose B emits finished before A. This loop will first await A.finished, and then await B.finished which will never come.

In practice I think it works because particles are will emit `finished` in the order that they appear in the scene tree.

I think a better approach would be to do this in `_emit_particles`:

```gdscript
func _emit_particle() -> void:
    _live_particles += 1

    # current body

    _live_particles -= 1
    if state == DEFEATED and _live_particles == 0:
        queue_free()
```

And then in this function, replace this loop with:

```gdscript
if _live_particles == 0:
    queue_free()
# else wait for `_emit_particles` to free this node after all particles are finished.
```

manuq: You have a very good point, I was in doubt and since it worked in practice, I moved on. While doing this I checked an addon: https://godotengine.org/asset-library/asset/3592

wjt: Nice find - I was wondering how hard it would be to write that the other day!

Issue #1349, opened 2025-10-14
Create void runner + grappling level


wjt:
We believe that combining the mechanics from #1174 and #1212 will be fun. Let's make a level combining the two mechanics as the final round of the Void quest.

PR #1371, opened 2025-10-17
Complete the skeleton of the lore second quest

manuq:
With a placeholder, although already playable, 3rd level that combines the Void runner with the grappling hook.

Helps https://github.com/endlessm/threadbare/issues/1349


wjt:
Good idea! I think that would also help with the problem on the very play of the first void level, where no-one quite sees it coming until it eats them.

Issue #1376, opened 2025-10-17
Grappling hook: Add Void enemy lurking around


manuq:
From playtest: For continuity, show the Void enemy in the 2nd level of the 2nd lore quest. Ideas:

- Add it to the first sublevel. Make it split the player from the Spirit thread that is so close, shaping the tiles.
- Add it to the powerup sublevel, going back and forth the big gap that can only be crossed after adquiring the longer thread. Make it patrol slowly, so this is the first introduction of Void enemies that have to be dodged by timing the jump.

PR #1377, opened 2025-10-17
Grappling hook: Add Void enemy lurking around

manuq:
Resolves https://github.com/endlessm/threadbare/issues/1376


wjt (approved):
Nice idea!

PR #1380, opened 2025-10-20
Add stealth/sequence puzzle combo level

### LINKED ISSUE #1363: Sequence puzzle + stealth combo level
@jgbourque and I have been collaborating on combining the sequence puzzle with the stealth mechanic - so you have to avoid guards while dashing between rocks to kick and solve the puzzle.

I think this could form the first round of a third quest.

We're developing it with [Patchwork](https://github.com/inkandswitch/patchwork-godot-plugin) but when it reaches some level of done-ness we can add it to the Git repo. This will be a nice test of the cross-VCS contribution flow.

wjt:
This combines two existing mechanics to form a new flavour of puzzle. Like in
the music puzzle, you have to tap some rocks in a series of sequences to ignite
3 fires and reveal a collectible thread; but you have to do so while avoiding
detection by two Storyvores, like in the stealth level.

I added a new placeholder quest that just contains this level. I think we could
place other levels without a home in here, such as implementations of some of
the nice ink combat designs that were submitted.

Pleasingly, no code changes were needed: the two mechanics work quite well
together. I think this combination works around one common complaint about the
music "puzzle": that there is no puzzle, because you just interact with each
altar and then copy the sequence displayed. Here the skill is in dodging the
guards as you do so. Much more fun.

 There are some papercuts though:

- When you interact with a sign to see a demo of the sequence, you can't move
  until the demo is complete; but the guards keep moving and can catch you.

- I think it would be possible to run across the bridge towards the collectible
  with a guard chasing you, and have the guard defeat you while the collectible
  animation is playing. We made the bridge pretty long to try to avoid this!

- The layout means that when you are standing by the third altar, you can't see
  all 4 rocks. I worked around this by setting the sequence on each altar to
  only use the rocks you can see.

Resolves #1363


scenes/quests/lore_quests/quest_003/1_stealth_sequence_combo/stealth_sequence_combo.tscn
```diff
@@ -0,0 +1,290 @@
...
+}
+point_count = 4
+
+[sub_resource type="Resource" id="Resource_idy4y"]
+script = ExtResource("9_1l5nh")
+metadata/_custom_type_script = "uid://bgmwplmj3bfls"
+
+[node name="StealthSequenceCombo" type="Node2D"]
+y_sort_enabled = true
+metadata/patchwork_id = "e0d6d619601849f5be54868c6b970eb3"
```
wjt, line 48: These IDs come from the Patchwork plugin, where we initially prototyped this level.

I am deliberately leaving them here because I want to see if there are any ill-effects of doing so.

scenes/quests/lore_quests/quest_003/quest.tres
```diff
@@ -0,0 +1,10 @@
+[gd_resource type="Resource" script_class="Quest" load_steps=2 format=3 uid="uid://dx8ew8remlcps"]
+
+[ext_resource type="Script" uid="uid://dts1hwdy3phin" path="res://scenes/menus/storybook/components/quest.gd" id="1_onpv7"]
+
+[resource]
+script = ExtResource("1_onpv7")
+title = "Placeholder"
+description = "A quest to hold scenes that don't yet have a permanent home."
```
manuq, line 8: 👍

Issue #686, opened 2025-05-21
Sequence puzzle wobble hints trigger incorrectly


wjt:
The sequence/musical-rocks puzzle has a hint mechanism where the next object you need to interact with animates/wobbles silently every so often.

The conditions under which this should happen are not really documented but (from memory and looking at the code) the idea was that it would only start once you have asked the musician for a hint about the current step a certain number of times:

https://github.com/endlessm/threadbare/blob/cfb224f8fb19f44486d28e5e38c42451495ec8d8/scenes/game_logic/sequence_puzzle.gd#L22

https://github.com/endlessm/threadbare/blob/cfb224f8fb19f44486d28e5e38c42451495ec8d8/scenes/game_logic/sequence_puzzle.gd#L96-L102

However a few lines later, hint timer is unconditionally started whenever you make a correct interaction:

https://github.com/endlessm/threadbare/blob/cfb224f8fb19f44486d28e5e38c42451495ec8d8/scenes/game_logic/sequence_puzzle.gd#L105

There is also nothing that stops the hint timer firing while the sequence is being demonstrated. It is very confusing if you're getting a demo of the sequence and two rocks wobble at once.

We should specify what the intended behaviour of these wobble hints is, document it, and implement it. We should also provide a way to disable these entirely. (I actually have a feeling that we can just remove the whole mechanic at this point because we made the musician just directly tell the player what to do, and the hint signs/altars demonstrate the sequence directly.)

### XP Summary (total 250)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

wjt:
Personally I think it would be simplest to remove the wobble hints given how much easier we made the puzzle. Maybe we could later look at a way to make it a separate component that could be added as a child of the `SequencePuzzle`.

wjt:
@HunterMk12 are you still working on this? If you're not, we can return it to the backlog. (You're always welcome to come back and work on it again later of course!)

PR #1394, opened 2025-10-21
Musician dialogue: Remove call to stop_hints()

manuq:
Which was removed in 6dde5c9c

Helps https://github.com/endlessm/threadbare/issues/686


wjt (approved):
Ouch, thanks!

Issue #1395, opened 2025-10-21
Player can get stuck in trees


manuq:
### Description

From a playtest (thanks 12 years old nephew!) the player can get stuck in one tree of the current Void + grappling hook level. This is very easy to reproduce, just walk towards the tree:

https://github.com/user-attachments/assets/6eb5faa5-6953-45d4-9a02-7f04fb021b58

https://github.com/user-attachments/assets/1179fa67-66cd-471b-826f-379ad067828b

We saw this before with barrels in the ink combat levels, and the workaround back then was to relocate the barrels. This time I found that changing the CharacterBody2d [safe_marging](https://docs.godotengine.org/en/4.5/classes/class_characterbody2d.html#class-characterbody2d-property-safe-margin) in player affects this for good. So it might be a better solution.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #1396, opened 2025-10-21
Player: Add one pixel of safe margin

manuq:
This seems enough for preventing the player getting stuck in props like trees.

Fix https://github.com/endlessm/threadbare/issues/1395

### DIFF
--- a/scenes/game_elements/characters/player/player.tscn
+++ b/scenes/game_elements/characters/player/player.tscn
@@ -507,6 +507,7 @@ _data = {
 collision_mask = 531
 motion_mode = 1
 wall_min_slide_angle = 0.017453292
+safe_margin = 1.0
 script = ExtResource("1_g2els")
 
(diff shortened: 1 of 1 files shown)


scenes/game_elements/characters/player/player.tscn
```diff
@@ -507,6 +507,7 @@ _data = {
 collision_mask = 531
 motion_mode = 1
 wall_min_slide_angle = 0.017453292
+safe_margin = 1.0
```
manuq, line 510: The default is: 0.08

wjt:
With this change I was able to get stuck on the monk's island in https://endlessm.github.io/threadbare/branches/endlessm/fix-player-stuck/#quests/lore_quests/quest_002/1_void_runner/void_runner:

image

but that's not a regression, I could get stuck there before. I don't think you're suggesting that this change is a silver bullet that will solve all such problems; I think you're right that it should help in some cases. Let's try it.

PR #1399, opened 2025-10-21
Prototype: More void + grappling hook

manuq:
Here I prototyped how it would look to use needles to go back the way that the Void has devoured.

### DIFF
--- a/scenes/quests/lore_quests/quest_002/3_void_grappling/components/return_needles.gd
+++ b/scenes/quests/lore_quests/quest_002/3_void_grappling/components/return_needles.gd
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends Node2D
+
+
+func _ready() -> void:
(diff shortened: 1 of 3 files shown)


manuq:
> I think the special thread-following camera logic isn't working quite right here, I think it is not considering that the player is moving as it bounces between the needles.

Yes, I noticed this too and forgot to comment about it.

> It's also a little frustrating to line up each shot. Should we be even kinder to the player and give each needle a default direction that is the correct direction (or the most obvious correct direction if there are 2 options, making the player work harder to go a fancier route)?

I was thinking in defaulting to the direction that is closer to the one the thread coming has. What do you think?

manuq:
> > I was thinking in defaulting to the direction that is closer to the one the thread coming has. What do you think?
> 
> Worth a try!

Actually this is how it works today with keyboard:

[recording.webm](https://github.com/user-attachments/assets/fe80031f-d3cb-4fde-857f-af719352b72c)

With joypad or mouse is difficult because the direction gets updated in the immediate next physics process loop to the direction that the input action is.

PR #1402, opened 2025-10-22
Void runner: Allow enemy to consume starting starting island

### LINKED ISSUE #1398: Void Level 1: Shadow Removes Path Inside Safe Zone
### Description

**Description**
While testing the first Void level in Threadbeard (the one with the Runner mechanic and shadow chase), I discovered that the shadow can remove part of the safe zone at the start of the level.

After moving slightly forward to trigger the shadow and then going back toward the starting area (with the blue house and the armored NPC), the path behind the player starts disappearing. This causes the player to become trapped, unable to move forward or complete the level.

**Where it happened**
Location: Start of the first Void level
Action: I advanced enough to trigger the shadow, then moved back to test how far it would chase me.
Result: The shadow removed the ground in the starting area, locking the player inside.

**Steps to reproduce**
1. Start the level “Void – 1.”
3. Move forward until the shadow begins to chase you.
5. Immediately move back toward the starting area (blue house).
7. Observe how the shadow deletes tiles behind the player, including part of the safe zone.
9. The player becomes stuck with no way to progress.

**Expected behavior**

The shadow should not destroy tiles within the safe zone, and the player should be able to return there safely.

**Actual behavior**

The shadow continues to delete tiles even inside the starting area, creating a softlock that prevents further progress.

**Attachments**

Image

https://github.com/user-attachments/assets/bcdd1ac9-1e50-4c26-91ff-ce055449d06d

Image

### Checklist

wjt:
Previously the enemy, once triggered, was not able to follow you back
onto the starting islands. So if you triggered the enemy then went left
back to the starting island, you would be stuck: the enemy would not
follow (and eat) you, and there's nothing else you can do.

Expand the navigation polygon to include those two islands. Adjust the
Void layer to add the entities on those islands to the nodes that can be
consumed. Unnest the forest from the Node2D that groups the other props
on the island to make this work.

Resolves #1398


wjt:
@AlealarcViva Could you take a look at [the test build of this branch](https://endlessm.github.io/threadbare/branches/endlessm/void-runner-consume-starting-island/#uid://bm4ewr8p48x0i) (link here goes straight to this scene) and see what you think?

Issue #1338, opened 2025-10-13
Update game logo & icon


wjt:
We have a new logo. Various variants in https://drive.google.com/drive/folders/1rzexrHyJnXd_yTtuyGmBFMvwWZ3BRjfd

I can think of at least 3 things we should change:

- [x] The Midjourney-generated logo on the title screen
- [ ] The loading screen (currently shows the Godot logo, we should show ours)
- [x] The Flatpak icon

Annoyingly I had only just finished making our current icon comply with Flathub requirements #1158 and I predict we'll get the same complaints about the new one.

PR #1407, opened 2025-10-22
Use Threadbare logo in bootsplash

wjt:
This is the 512px variant with no text and a needle through the spool.

Configure the splash screen not to scale the logo, and (in case we later do want
to scale it) to use nearest-neighbour interpolation rather than linear
interpolation.

Resolves https://github.com/endlessm/threadbare/issues/1338


manuq:
Love it!
> We can also change the background colour of the splash screen. I think the neutral grey is OK though.

Yes, somehow the gray and the blue progress bar (the defaults) go well.

manuq (approved):

PR #1409, opened 2025-10-22
Duplicate and rename template and all files

Fernort:
Se esta agregando en la story quest la carpeta aventuras_de_un_capa


scenes/quests/story_quests/aventuras_de_un_capa/quest.tres
```diff
@@ -0,0 +1,10 @@
+[gd_resource type="Resource" script_class="Quest" load_steps=2 format=3 uid="uid://b1r57leuevlty"]
+
+[ext_resource type="Script" uid="uid://dts1hwdy3phin" path="res://scenes/menus/storybook/components/quest.gd" id="1_8wsae"]
+
+[resource]
+script = ExtResource("1_8wsae")
+title = "StoryQuestContradiccion"
```
manuq, line 7: The folder is named `aventuras_de_un_capa` but the title here is "StoryQuestContradiccion". And the prefix used is also `story_quest_contradi` (except for the tileset which is using `aventurasdeuncapa` as prefix). Please make it consistent. Try to not use "story quest" in the prefix or in the title.

manuq (changes requested):

Issue #1372, opened 2025-10-17
Void enemy: Add clues when it starts chasing


manuq:
Add both a visual clue and an audio clue when the enemy starts chasing the player.

As placeholders, we could reuse the guard's exclamation sign and the alerted sound.

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

wjt:
I wonder - should it also have the slight pause before it starts chasing?

manuq:
> I wonder - should it also have the slight pause before it starts chasing?

Yes, a slight pause may probably prepare the player for the chasing. I also thought about a 3, 2, 1, go! counter like in an actual race, but may be too off theme.

PR #1422, opened 2025-10-23
Add clues when Void enemy starts chasing

manuq:
Also update the asset of the exclamation sign which was oversized for the game base resolution.

Resolve https://github.com/endlessm/threadbare/issues/1372


wjt (approved):
Great - I think this will help.

PR #1423, opened 2025-10-23
Rama mathias

Jake2420:
Cambios por el grupo1


JuaniCargnel:
Hi @wjt! Sorry about that — this was a mistake from the students, they never canceled the PR! They just forgot to change it so that the PR would go to the org main branch.

Regarding the “Revert Update,” that was me — I did it when we had to update the ORGs. I had made a PR to update them, but then the option to disable protection came up, so I did the revert to make all the ORGs consistent... it didn’t work i guess, haha. Could that cause problems in the future? Everything should be fine now, but I might have messed something up — sorry about that.

wjt:
> Also, I just want to clarify again that this PR we’re talking in should be rejected (or closed) since it was just a mistake.

OK!

PR #1433, opened 2025-10-24
Pull Request Confirmacion 

andree-git:
Comprobar si funciona el Pull Request

### DIFF
--- a/scenes/quests/story_quests/laberinto_del_fauno/0_intro/intro_components/laberinto_del_fauno_intro.dialogue
+++ b/scenes/quests/story_quests/laberinto_del_fauno/0_intro/intro_components/laberinto_del_fauno_intro.dialogue
@@ -1,14 +1,29 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 
+# Diálogo: Ofelia llega al cuartel militar.
 ~ start
-do animation_player.play(&"walk_on")
-do animation_player.animation_finished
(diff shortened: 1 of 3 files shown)


addons/dialogue_manager/components/editor_property/editor_property_control.tscn
```diff
@@ -1,9 +1,13 @@
-[gd_scene load_steps=4 format=3 uid="uid://ycn6uaj7dsrh"]
+[gd_scene load_steps=5 format=3 uid="uid://ycn6uaj7dsrh"]
```
wjt, line 1: I guess you didn't intend to change all these files in `addons/dialogue_manager`. You should revert them:

```console
$ git restore --source=6c579a4a257abdc5f23b5ac09a0a98da88c191b5 addons/dialogue_manager
$ git add addons/dialogue_manager
$ git commit -m 'Revert unintended changes to Dialogue Manager addon'
```

scenes/quests/story_quests/mystoryquest_fauno/quest.tres
wjt, line 1: You don't need `mystoryquest` in the folder name for the quest. It would be better to use `fauno` for the folder name and all filenames if that is the theme of your quest.

addons/dialogue_manager/components/editor_property/editor_property_control.tscn
wjt: This still needs to be addressed before we can merge this!

Issue #1410, opened 2025-10-23
Void quest: Add initial narrative


manuq:
The first level (Void runner) already has an initial narrative, but the other 2 don't. It looks like the recurring caracter will be the monk, which at the end of level 1 tells to the StoryWeaver: "If they can bring a story to the town, they may be able to push back the Void and open the path once more". Then the monk will give a grappling hook to the StoryWeaver at the start of level 2. And then at the end of level 3 the StoryWeaver reaches the town, and shares the story from the monk. The Void is pushed back, opening up that previous path that couldn’t be used (leading to the very beginning). They acquire the third thread and can return to the Eternal Loom.

Note: this is prone to change but will give a better idea of how the quest looks with cinematics and dialogues.

PR #1434, opened 2025-10-24
Grappling hook level: Add intro cinematic

manuq:
The monk tells the StoryWeaver where to go, but suddenly the Void erases the path. The monk gives her grappling hook (tool name is placeholder) to the StoryWeaver so they can navigate in the place.

Note: The start and end grappling hook levels will have to be redesigned because the thread is below the dialogue.

Helps https://github.com/endlessm/threadbare/issues/1410

### DIFF
--- a/scenes/quests/lore_quests/quest_002/2_grappling_hook/intro_start.dialogue
+++ b/scenes/quests/lore_quests/quest_002/2_grappling_hook/intro_start.dialogue
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+Matilda: Go take the thread, it will teleport you to the town.
+do animation_player.play(&"eat_floor")
+do animation_player.animation_finished
+Matilda: Hmmm.. the Void is blooming fast here.
+Matilda: You’d only be able to reach the town using a [b]grappling hook[/b]. That's how I managed to survive. Take mine.
+=> END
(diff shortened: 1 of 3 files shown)


manuq:
> Play this branch at https://endlessm.github.io/threadbare/branches/endlessm/more-void-initial-narrative.
> 

Actually here https://endlessm.github.io/threadbare/branches/endlessm/more-void-initial-narrative/#quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_start

I think I'll redesign the level because the dialogue occluding the thread being mentioned is bad.

scenes/quests/lore_quests/quest_002/2_grappling_hook/intro_start.dialogue
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+Matilda: Go take the thread, it will teleport you to the town.
+do animation_player.play(&"eat_floor")
+do animation_player.animation_finished
+Matilda: Hmmm.. the Void is blooming fast here.
+Matilda: You’d only be able to reach the town using a [b]grappling hook[/b]. That's how I managed to survive. Take mine.
```
wjt, line 8: Nice idea to embolden the text! However it doesn't work on the web platform:

Image

presumably because we don't have a bold-weight font.

Maybe we can (later) change this to use colour instead?

PR #1436, opened 2025-10-24
Void runner: improve tree collisions

Omar071201:
Corrected an issue where the player's progression was blocked due to unexpected collisions with certain tree assets.
The update adjusts the bounding boxes and collision meshes for the problematic tree models, ensuring accurate interaction while maintaining their visual appearance.
This fix restores proper player movement through the environment, eliminating the progression-blocking scenario reported by the community.

Fixes #1424

image

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=37 format=4 uid="uid://bm4ewr8p48x0i"]
+[gd_scene load_steps=36 format=4 uid="uid://bm4ewr8p48x0i"]
 
 [ext_resource type="PackedScene" uid="uid://2rbpl811wlv1" path="res://scenes/game_elements/props/background_music/background_music.tscn" id="1_g2rto"]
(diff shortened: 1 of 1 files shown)


scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
```diff
@@ -832,6 +835,13 @@ position = Vector2(3505.48, 1740.11)
 scale = Vector2(0.937681, 1.0068)
 sprite_frames = ExtResource("21_xqo3t")
 
+[node name="StaticBody2D" type="StaticBody2D" parent="OnTheGround/SanctuaryWoods"]
+position = Vector2(3551, 1666)
+
+[node name="CollisionShape2D" type="CollisionShape2D" parent="OnTheGround/SanctuaryWoods/StaticBody2D"]
+position = Vector2(5.5, -61.5)
+shape = SubResource("RectangleShape2D_g2rto")
```
manuq: Have you considered using a CollisionPolygon2D shape instead? I think it will work better, because you can draw an irregular limit with it.

Omar071201: Oh great, I'm going to pull it when I get back from university.

wjt: The polygon is better, great!

Omar071201:
Yes, I've switched now, thanks for the suggestion.

PR #1445, opened 2025-10-27
tileset: Add missing "walls" collision polygon

wjt:
tileset: Add missing "walls" collision polygon

This is the cliff-front tile of the 1-tile-wide column of the
Cliff_Mines_Tiles atlas. Previously it had no collision polygon. Add a
full-tile polygon, matching similar cliff-fronts.


manuq (commented):
This one, right? Good catch.

Image

PR #1447, opened 2025-10-27
Adjust tile map layers

wjt:
This is another set of changes to make every TileMapLayer in the game only use one of a particular group of tiles (e.g. "stones", "grass/dirt/etc", "water"). With this done, it will be possible to split up the TileSet (#1336).

It includes a tweak to El Juguete Perdido that I missed in #1416, and adjusts four other scenes in the same way.

PR #1448, opened 2025-10-28
Split big TileSet into 6 smaller TileSets & update all scenes

### LINKED ISSUE #1336: Split tileset by kind of tile
As discussed in https://github.com/endlessm/threadbare/discussions/1209, we currently have one big tileset, `scenes/tileset.tres`, which contains all the tiles used in the base game.

Let's split this by "type" of layer:

- water
- foam (or maybe this should live with water?)
- cliff faces + clifftops
- grass + faded grass + sand + dirt - all surfaces that can sit on top of rock and each other
- bridges + bridge-shadows
- void

This will make it harder to accidentally place tiles in the wrong layer: you will only have the appropriate tiles for that layer in the toolbox. I think it will also make it more obvious that you don't need to edit one big tileset to add your own tiles: you can create new tilesets.

wjt:
Currently we have one big TileSet which contains 14 atlases, covering every tile we use in the base game. This makes it easy to accidentally mix (for example) grass, cliffs and water on the same layer, which can later make it really hard to figure out why there's a patch of grass that you can't walk across.

Split the current tileset into 6 smaller tilesets:

- bridges
- elevation (cliffs & stairs)
- exterior floors (grass, mud, sand)
- foam
- water
- chromakey void

(Don't include the now-unused static void tiles.) These 6 TileSets preserve the same atlas IDs so they are drop-in replacements when a TileMapLayer only references atlases from one such TileSet.

Then update every TileMapLayer in every scene in the repository to use the appropriate smaller tileset. (I have previously fixed every TileMapLayer that mixed different kinds of tile so that this can be done.)

Leave the old, now-unused tileset in the tree for now. We can remove it in a few months once any extant branches that reference it have been merged and those scenes updated.

Fixes #1336 

----

That's the what; here's the how.

First, I manually duplicated the tileset 6 times and removed the unwanted terrains, patterns, and atlas sources from each duplicate. It would probably have been possible to automate this but it was easy enough to do it by hand. I committed this to a branch called `split-tileset`, which will become relevant later.

Next, I wrote a script that searched the project for scenes using the old TileSet. This script checks each layer in such scenes to see if one of the new smaller TileSets can be swapped in. If so, it does so; if not, it prints a message about mixed tiles. (This is how I found scenes to adjust in #1416 and https://github.com/endlessm/threadbare/pull/1447.) This script is included in the PR so that we can re-run it later as branches using the unified tileset are later merged.

Now, I wanted to find a way to preserve the `git blame` history for the TileSet across the split. I remembered reading about a way to do this on [The Old New Thing](https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904) six years ago – I've always wanted to try this trick. And it turns out it works!

I went back to the `main` branch and made a new `split-tileset-2` branch. Then, for each new tileset, I made a branch with commits:

1. **Rename** the old tileset to the new tileset's name.
2. **Edit the copy** to change its UID and remove all unwanted layers (by restoring its contents from the `split-tileset` branch I made earlier.

I merged all these branches back to `split-tileset-2`. Finally I appended a commit to restore the original file.

https://gist.github.com/wjt/5b5608084fa9cb6e47a816a2addd8f90

This process means `git blame tiles/foam.tres` shows the commits that changed each line, not that I created the file today (except the lines that I actually changed).

Unfortunately `git log` (even with `--follow`) is a little weird for the old & new tilesets, but I think it's OK.

But it does make for a kind of complicated commit graph!

```
* dd29f453 (HEAD, origin/split-tileset-2, split-tileset-2) storyquest bootstrap: Copy all (split) TileSets
* ca90b052 Update all scenes to use split TileSets
* 72d812d9 Add editor script to update TileSets in all scenes
* 50f78170 Restore scenes/tileset.tres
*---------.   171d6e15 Split scenes/tileset.tres into separate tilesets
|\ \ \ \ \ \
| | | | | | * a230e82e Remove unrelated tiles from tiles/foam.tres
| | | | | | * cd6e5cea Move scenes/tileset.tres to tiles/foam.tres
| |_|_|_|_|/
|/| | | | |
| | | | | * 16cbc587 Remove unrelated tiles from tiles/water.tres
| | | | | * f9ecdf74 Move scenes/tileset.tres to tiles/water.tres
| |_|_|_|/
|/| | | |
| | | | * 19fe9f35 Remove unrelated tiles from tiles/void_chromakey.tres
| | | | * d26a6e61 Move scenes/tileset.tres to tiles/void_chromakey.tres
| |_|_|/
|/| | |
| | | * 303fbad8 Remove unrelated tiles from tiles/bridges.tres
| | | * 3b2da532 Move scenes/tileset.tres to tiles/bridges.tres
| |_|/
|/| |
| | * 1f034594 Remove unrelated tiles from tiles/elevation.tres
| | * 859857af Move scenes/tileset.tres to tiles/elevation.tres
| |/
|/|
| * 4e5df825 Remove unrelated tiles from tiles/exterior_floors.tres
| * 504f1981 Move scenes/tileset.tres to tiles/exterior_floors.tres
|/
*   d3a3f2b1 (origin/main, origin/HEAD, main) Merge pull request #1456 from endlessm/wjt/player-add-defeat-method
```

I am happy with the result but I won't be offended if others think that it would be better to just have simple commits for each copy, accepting that `git blame` will be wrong for the split tilesets.

### DIFF
--- a/addons/storyquest_bootstrap/copier.gd
+++ b/addons/storyquest_bootstrap/copier.gd
@@ -7,6 +7,7 @@ const STORYQUESTS_PATH := "res://scenes/quests/story_quests/"
 const TEMPLATE_PREFIX := "NO_EDIT"
 const TEMPLATE_PATH := STORYQUESTS_PATH + TEMPLATE_PREFIX + "/"
 const QUEST_FILENAME := "quest.tres"
+const TILES_PATH := "res://tiles/"
 
 ## Map from UID to already-copied resource
 var orig_uid_to_copy: Dictionary[String, Resource]
@@ -235,7 +236,24 @@ func copy_uid(uid: String) -> String:
 	return copied_uid
 
 
+func copy_tilesets() -> void:
+	var dir := DirAccess.open(TILES_PATH)
+	assert(dir, error_string(DirAccess.get_open_error()))
+
+	dir.list_dir_begin()
+	var file := dir.get_next()
+	while file != "":
+		if not dir.current_is_dir() and file.ends_with(".tres"):
+			var tileset := ResourceLoader.load(TILES_PATH.path_join(file))
+			if tileset is TileSet:
+				await copy_resource(tileset)
+		file = dir.get_next()
+	dir.list_dir_end()
+
+
 func create_storyquest() -> void:
+	copy_tilesets()
+
 	var quest: Quest = load(TEMPLATE_PATH.path_join(QUEST_FILENAME))
 	await copy_resource(quest)
 	EditorInterface.save_all_scenes()


wjt:
@manuq I consider this ready for review but I would like to get Justin's feedback on behalf of the learning team before merging it. (Happily the branch is mostly generated by a computer so I can easily refresh it if we don't merge it immediately.)

wjt:
> The change does mean they will have to copy more files outside of what the automation tool does for them ( they will be given two but no more),

We could change the duplicator to duplicate all 6?

wjt:
Discussing this with Justin: yes, duplicating all 6 is desired, not just the 2 that are used. I think that's fine because, since they're split up, it'll be very easy to check "show owners" and see which ones are unused.

wjt:
Please don't squash-merge this PR by accident!

manuq:
> The CI failure will be resolved by #1458

I re-ran the CI after merging this and is still failing, hmm.

Issue #1348, opened 2025-10-14
Duplicate TileSet for NO_EDIT template


wjt:
Currently the NO_EDIT template scenes use `scenes/tileset.tres`. It's really common for people making a StoryQuest to want to add their own tiles, and the "obvious" thing to do is to edit the tileset that's already in use.

@jgbourque has asked for this to change, so that the tileset(s) used in the NO_EDIT quest are duplicated, like the SpriteFrames etc.

I think this makes sense. I would hope that splitting up the tileset (#1336) would help to make it clearer that you don't need one big tileset - you can have lots of little ones - but still, currently the easy thing to do is the "wrong" thing, so I think it's worth a try splitting it. I would prefer to first break apart the big tileset though, so that there is less duplication.

PR #1449, opened 2025-10-28
storyquest_bootstrap: Copy TileSets

wjt:
Previously, the StoryQuest bootstrapping tool did not duplicate the TileSet(s)
    used by the NO_EDIT template levels, because the TileSet resource is outside the
    NO_EDIT directory. It's very tempting to just edit the TileSet that a layer uses
    in-place, without stopping to duplicate it (or make a new TileSet). This is a
    common reason why past StoryQuest submissions have needed revision before being
    accepted.

When traversing the NO_EDIT template, consider any TileSet resource as something
    to be duplicated, even if it is not in the NO_EDIT folder. Duplicate any such
    tilesets from outside the NO_EDIT folder to a tiles/ subfolder of the new quest,
    and prefix their names with "questname_" like other files.

Resolves https://github.com/endlessm/threadbare/issues/1348

---
  - https://app.asana.com/0/0/1211719193201604

### DIFF
--- a/addons/storyquest_bootstrap/copier.gd
+++ b/addons/storyquest_bootstrap/copier.gd
@@ -42,7 +42,7 @@ func is_template_uid(uid: String) -> bool:
 
 
 func is_template_resource(resource: Resource) -> bool:
-	return resource.resource_path.begins_with(TEMPLATE_PATH)
+	return resource.resource_path.begins_with(TEMPLATE_PATH) or resource is TileSet
 
 
 func reimport(_path: String) -> void:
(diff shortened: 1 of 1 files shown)


addons/storyquest_bootstrap/copier.gd
```diff
@@ -187,6 +207,8 @@ func copy(uid: String, resource: Resource) -> Resource:
 		copied = await copy_as_file(resource, copy_path)
 	elif resource is SpriteFrames:
 		copied = await copy_sprite_frames(resource, copy_path)
+	elif resource is TileSet:
+		copied = await copy_tile_set(resource, copy_path)
```
manuq, line 211: 👍

addons/storyquest_bootstrap/copier.gd
```diff
@@ -156,20 +169,27 @@ func copy_as_file(resource: Resource, copy_path: String) -> Resource:
...
+		subdir = subpath.get_base_dir()
+		filename = subpath.get_file()
+
+		subdir = subdir.replace(TEMPLATE_PREFIX + "_", "")
+		filename = filename.replace(TEMPLATE_PREFIX, quest_name)
+	elif resource is TileSet:
+		subdir = "tiles"
+		filename = quest_name + "_" + resource.resource_path.get_file()
+	else:
+		assert(false, "'%s' is not a storyquest or tileset resource" % resource)
```
manuq, line 190: 👍

manuq (approved):

PR #1452, opened 2025-10-28
storyquest_bootstrap: Forbid "story" and "quest" words

manuq:
In title and folder. The rationale for the folder is that the quests are already inside a story_quests/ folder. The rationale for the title is to prevent dull titles like "My Story Quest" or similar.

### DIFF
--- a/addons/storyquest_bootstrap/new_storyquest_dialog.gd
+++ b/addons/storyquest_bootstrap/new_storyquest_dialog.gd
@@ -99,8 +99,12 @@ func _revalidate() -> void:
 			errors.append("⚠ ‘%s’ is not allowed in the folder name" % match.get_string(0))
 	if _filename.to_lower() != _filename:
 		errors.append("⚠ Folder name must be lower-case")
+	if "story" in _filename and "quest" in _filename:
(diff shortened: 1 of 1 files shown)


addons/storyquest_bootstrap/new_storyquest_dialog.gd
```diff
@@ -99,8 +99,12 @@ func _revalidate() -> void:
 			errors.append("⚠ ‘%s’ is not allowed in the folder name" % match.get_string(0))
 	if _filename.to_lower() != _filename:
 		errors.append("⚠ Folder name must be lower-case")
+	if "story" in _filename and "quest" in _filename:
```
manuq, line 102: I should have mentioned that this is "story" and "quest" both at the same time. We could be more strict and disallow including "story" or "quest" but maybe that's too much.

manuq:
> I think this is worth doing, but I would also suggest to @jgbourque @Stregoica777 that the relevant guides are tweaked to tell people that, although the text or demo video might use `mystoryquest` in the name, learners should not call their quest `mystoryquest` or `mystoryquest foo`.

Absolutely, that is much more important!

PR #1454, opened 2025-10-28
Grappling hook level: Redesign start and end scenes

manuq:
So the magical thread is not occluded by the dialogue. And to give it
more narrative sense:
    
- From the Void runner level, there is an ellipsis.
    
- The StoryWeaver is about to grab a magical thread and be teleported.
    
- But the Void ruins their plans. Has to take an alternative path.
    
- At the end of the level, the StoryWeaver comes from the other
  direction.

Also add a dialogue at the start of the Void + grappling level.

Helps https://github.com/endlessm/threadbare/issues/1410


wjt:
image

Is this pin-in-the-void intentional?

wjt:
That's in the start level. Here's the "same" thing in the "end" level:

image

.oO( can we make these actually be the same scene, by storing some state in `GameState`? one for another day perhaps )

So I think you meant to move that pin up a little? Or is the enemy eating one tile too many perhaps?

manuq:
@wjt thanks for spotting the difference! That island was 1 row taller. It is indeed hard to keep these 2 scenes in sync. The "only" game state needed would be the grappling hook "has more string" and I had that in my prototype. But I think we agreed to leave that out for now to avoid adding more state.

Issue #1026, opened 2025-08-11
Design a new InkDrinker level


saspiers:
Play through the InkDrinker game e by selecting the Musician Lore Quest via the Mage on the left side of the loom. 

Create and attach a document to this issue with a design for a new variation on the InkDrinker game. There should be three sections that increase in difficulty. New designs can be variations on existing the existing puzzle or add new elements to the environment/mechanics of the game. 

Ideally the design includes: 

- Written description of the new elements introduced into the game/puzzle
- Sketches explaining how the game/puzzle works 
- Steps for completing the puzzle/game
- a list of tasks needed to complete the creation of this puzzle (ex. art assets, audio assets, game functionality) 

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 75 |
| Game Design: Core Gameplay & Mechanics | Creativity & Design Thinking | 75 |
| Game Design: Core Gameplay & Mechanics | Technical Literacy | 50 |

Russell-CJ:
”I Plan to work on this”

lesy-carolina:
We present the thematic planning sheet. [#1026](https://docs.google.com/document/d/1Pnyrn5i7sRamFlFKHiBJVQZTYoCSG26Mffnvg4PVKiw/edit?usp=sharing)

russell-cj:
Hi @manuq, @saspiers @manuq @lesy-carolina , here’s the link to the [Notion](https://www.notion.so/Issue-1026-256b5b0e697780b6952efd886041d2e9), which contains all the notes, links, and progress on this issue.
It’s not finished yet — I’m sharing it so you can leave any feedback on what you think should be changed or improved.

manuq:
@russell-cj nice! I like the idea of the fragile jars in https://www.notion.so/Ideas-de-niveles-261b5b0e697780a891caef6d5ff30646 ! Keep in mind that the delivered design should be in English. I can help you with that.

russell-cj:
@manuq, Yes, the **PDF contains everything**, but in a more visual/static way.  

All the **links** are also included inside the PDF.  

If you’d like to interact with the resources in more detail, you can check them in **Figma** or **Notion**. There you’ll be able to adjust video playback speed, leave comments, or edit the designs directly.

manuq:
Thank you, added to the artwork/game design contributions page: https://github.com/endlessm/threadbare/discussions/1129

PR #1455, opened 2025-10-29
Implement InkDrinker Level 4 Design

### LINKED ISSUE #1383: Implement designed InkDrinker level 4
Implement level 4 as designed in https://github.com/endlessm/threadbare/issues/1026

Add the level to `lore_quests/quest_003` (not as level 4 of the existing quest). Do not implement new mechanics for now, let's track that separately.

### XP Summary (total 300)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Technical Literacy | 150 |

russell-cj:
This PR implements Ink Drinker Level 4 according to the design defined in issue #1026.
The level was added under lore_quests/quest_003 and includes the terrain layout, enemies, interactive objects, and all elements for the combat encounter.

Resolves  #1383

### DIFF
--- a/scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn
+++ b/scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn
@@ -46,7 +46,7 @@ _data = {
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_oj6a4"]
 size = Vector2(101, 49)
 
-[node name="InkCombatRound4" type="Node2D"]
+[node name="InkCombat" type="Node2D"]
 
 [node name="BackgroundMusic" parent="." instance=ExtResource("1_aj2gl")]
 stream = ExtResource("2_h1kdf")


scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn
manuq, line 1: Following the pattern in the other lore quests, the folder of the level should have a number as prefix. This one is the second so prepend`2_`:

```
scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/
```

scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn
```diff
@@ -0,0 +1,207 @@
...
+
+[node name="Door Exit" parent="OnTheGround/LevelExit" instance=ExtResource("2_jwan6")]
+position = Vector2(0, 82)
+play_victory_fanfare_on_open = true
+
+[node name="Teleporter" type="Area2D" parent="OnTheGround/LevelExit"]
+position = Vector2(0, -25)
+collision_layer = 4
+script = ExtResource("6_h1kdf")
+next_scene = "uid://bo6qfusag3fae"
```
manuq, line 125: When I finish the level, I am teleported to the 3rd round of the first lore quest. To keep the quest playable, it should go to Fray's End (`scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn`).

wjt (approved):
Great start! I'll merge this but I have a few suggestions for improvements after testing it a little.

I think you should move the Camera2D node under the Player node, and adjust its bounds to the "dressed" area of the level. (I know the existing combat scenes are not like this, but I think that's a problem we should fix!)

I can walk off the screen here: 

Image

This is because:

1. The camera bounds are set incorrectly
2. You have used the wrong clifftop tile here: you should use the one for the right-hand edge of a cliff, not the left-hand edge. With Debug -> Visible Collision Shapes enabled, and overriding the camera, you can see that the wrong edge has a collision shape:

Image

We should dress a bit more of the scene. Perhaps extend the water and/or cliffs as appropriate on the left and right sides, and extend the ground & paths to the top and bottom (with water either side perhaps). The viewport is no longer locked to 16:9 aspect ratio https://github.com/endlessm/threadbare/issues/1083 so if the window is too wide we see:

Image

Or too tall:

Image

In the project settings you can enable a flag that draws red borders if the aspect ratio is unreasonably wide or tall:

Image

We don't require the levels to look good if the screen is really wide or tall, but of course if it's easy to do (e.g. lots of water) then that's great.

While taking these screenshots I discovered that you can actually finish this level by just standing still!

Image

Not sure if this is intentional :)

scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn
russell-cj: Perfect, thanks for the clarification. I’ll rename the level folder with the prefix `2_` to follow the pattern of the other missions and maintain consistency in the projec.

scenes/quests/lore_quests/quest_003/ink_drinker_levels/ink_combat_round_4.tscn
russell-cj: Thanks for the observation. This is related to what I mentioned in the first comment: I temporarily left the Teleporter pointing to Level 3 because the final flow wasn’t defined yet. Now that I have a clearer idea, I’ll update the next scene so the player reaches Fray’s End correctly.

Issue #1334, opened 2025-10-13
Being caught by two guards at once causes double-respawn


wjt:
### Description

In stealth levels, it's possible to be caught by two guards at once. When this happens, the player respawns twice. In this video you can see that the player warps directly to the checkpoint, but then the level fades out and back in again:

https://github.com/user-attachments/assets/6f89769a-2c86-416f-8a53-5463df68fc5b

This is because in `stealth_game_logic.gd` the following callback runs each time `player_detected` is emitted by a guard:

```
func _on_player_detected(player: Player) -> void:
	player.mode = Player.Mode.DEFEATED
	await get_tree().create_timer(2.0).timeout
	SceneSwitcher.reload_with_transition(Transition.Effect.FADE, Transition.Effect.FADE)
```

The simplest fix would probably be to check in this function if `player.mode` is already `DEFEATED`, and do nothing if so. But there may be a better way!

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

PR #1456, opened 2025-10-29
Player: Add defeat() method

wjt:
Player: Add defeat() method

Previously the code to mark the player as defeated, wait a few seconds,
then reload the current scene was duplicated between two kinds of enemy.
In addition, it was not idempotent: if a second guard caught the player
during the 2-second pause, the scene would reload twice.

Add a defeat() method to the player that does nothing if the mode is
already DEFEATED, and otherwise carries out the same set of steps. Call
this in the two relevant places.

It's a bit of a kludge to have the `falling` parameter but I think this
is still an improvement.

Resolves https://github.com/endlessm/threadbare/issues/1334

### DIFF
--- a/scenes/game_elements/characters/player/components/player.gd
+++ b/scenes/game_elements/characters/player/components/player.gd
@@ -223,3 +223,22 @@ func _set_walk_sound_stream(new_value: AudioStream) -> void:
 	if not is_node_ready():
 		await ready
 	_walk_sound.stream = walk_sound_stream
+
+
+## Sets the player's [member mode] to [constant DEFEATED], if it is
+## not already. Reloads the current scene after a short interval.
+## [br][br]
+## If [param falling] is [code]true[/code], scale the player to zero, as if they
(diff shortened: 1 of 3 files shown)


scenes/game_elements/characters/player/components/player.gd
```diff
@@ -223,3 +223,22 @@ func _set_walk_sound_stream(new_value: AudioStream) -> void:
...
+## are falling into the screen as they unravel.
+func defeat(falling: bool = false) -> void:
+	if mode == Player.Mode.DEFEATED:
+		return
+
+	mode = Player.Mode.DEFEATED
+
+	if falling:
+		var tween := create_tween()
+		tween.tween_property(self, "scale", Vector2.ZERO, 2.0)
```
manuq, line 241: I think that the falling parameter is fine. In the future this could be its own AnimationPlayer animation with both the scale change + custom frame-by-frame animation + sound effect. In which case we'll need a Player.Mode.FALLING mode/state.

Issue #1280, opened 2025-10-02
Switch to input prompt asset pack with unified arrow keys


wjt:
The controller prompt asset pack we are currently using doesn't have an asset that shows all four keyboard arrow keys at once. Instead it has separate assets for each key, 100×100px.

But for joypads, the dpad is a single 100×100px asset, with variations for the different pressed states.

This means our input hints need to handle keyboard directional input separately to joypad directional input; and the switch between the two leaves something to be desired!

I think we should switch to an asset pack that has unified arrow keys, such as https://kenney.nl/assets/input-prompts. (These are 64×64, substantially smaller, which I think looks better too.)

Then we can unify `gamepad_input.gd` and `keyboard_input.gd` into a single `directional_input_hint.gd` script which treats keyboard and joypad-dpad the same way.

### XP Summary (total 475)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 75 |
| Game Design: Game Feel & UX | Technical Literacy | 50 |

PR #1464, opened 2025-10-30
feat: implement unified directional key support using the new input p…

AlealarcViva:
This pull request addresses issue #1280, which focuses on unifying the arrow key input system to ensure consistent and responsive behavior across all UI components.

Details:
- Implemented a new input handling approach for arrow keys within the control scripts.
- Replaced individual arrow key icons (↑, ↓, ←, →) with a unified asset.
- Updated input handling logic to support both keyboard and controller inputs in a shared flow.
- Cleaned up and refactored code to prepare for the new directional_input_hint.gd implementation.

Fix https://github.com/endlessm/threadbare/issues/1280


AlealarcViva:
image
Would this type of button be better?

manuq:
> image Would this type of button be better?

For now I think it's fine! I haven't merged yet because the license check is failing. We have to update this:

https://github.com/endlessm/threadbare/blob/11ee3b1d8f2167b386b7ff9dc58c182f2da7d159/REUSE.toml#L89-L94

Or this:

https://github.com/endlessm/threadbare/blob/11ee3b1d8f2167b386b7ff9dc58c182f2da7d159/REUSE.toml#L47-L50

manuq:
@AlealarcViva I have appended a few fixes:

- The `scenes/world_map/frays_end.tscn` has changed recently and was having a git conflict. It was easy to fix.
- The assets in `assets/third_party/inputs/` are now not only PNG, but also SVG and other files, so I updated the `REUSE.toml` file.
- More importantly, the attribution of those files now correspond to Kenney. So I have updated it too. The license is still the same CC0-1.0.

PR #1474, opened 2025-11-02
Repositions intro concept art 

### LINKED ISSUE #1166: Concept art in intro does not fit viewport
### Description

The intro scene was designed for 1920×1080 resolution. We reduced the game's base resolution but did not update this scene accordingly; as a result most of the sketched concept art is off the right/bottom edge of the viewport:

Image

Image

The art currently has its position set manually. One way to fix this bug would be to update the position & scaling of each Sketch node to fit the screen. Another way would be to replace the Sprite2D nodes with TextureRect nodes, and set their position to shrink right, expand vertically, preserve aspect ratio. This would keep them right-justified if the window is wider than the base aspect ratio, and expand them vertically if the window is taller than the base aspect ratio.

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |

Gatorrante:
This PR replaces the problematic `Sprite2D` nodes (`Sketch01` through `Sketch05`) with `TextureRect` nodes as @wjt suggested. This approach leverages Godot's UI system to handle scaling and positioning dynamically, making it robust to resolution changes.

https://github.com/user-attachments/assets/d4f5df64-9bcc-4729-83cb-af9b8484938a

The following configuration was applied to each `TextureRect` to ensure the art aligns to the right and scales correctly:

* **Layout / Anchors Preset:** Set to `Bottom Right` to anchor the art to the right edge of the parent `Control` node.
* **Expand Mode:** `Ignore Size`
* **Stretch Mode:** `Keep Aspect Centered`

Closes #1166

### DIFF
--- a/scenes/menus/intro/intro.tscn
+++ b/scenes/menus/intro/intro.tscn
@@ -293,31 +293,86 @@ texture = ExtResource("2_a6oma")
 expand_mode = 3
 stretch_mode = 6
 
-[node name="Sketch01" type="Sprite2D" parent="CanvasLayer/Control"]
+[node name="Sketch01" type="TextureRect" parent="CanvasLayer/Control"]
 visible = false
 material = SubResource("CanvasItemMaterial_mqh0g")
-position = Vector2(1379, 540)
+layout_mode = 1
+anchors_preset = 3
+anchor_left = 1.0
+anchor_top = 1.0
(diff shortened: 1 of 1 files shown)


wjt (approved):
This looks great. Nice work, and well explained.

PR #1475, opened 2025-11-02
Copia renombre template

Leonel763:
copie y renombre el template a el_eclipse


scenes/menus/splash/splash.tscn
wjt, line 92: It looks like you accidentally duplicated these nodes of the splash screen. You will need to revert this change in your fork.

PR #1481, opened 2025-11-03
Story quest grupo4

JordanPD03:


scenes/game_elements/characters/player/components/player.gd
```diff
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
+#test Push 03
```
wjt: You'll need to revert this change to the player script before submitting the StoryQuest upstream.

Issue #1478, opened 2025-11-03
Generic controller shows keyboard input hint for directional input


wjt:
### Description

When I use keyboard input, the directional input hints correctly show the layout of the arrow keys:

<img width="408" height="216" alt="Image" src="https://github.com/user-attachments/assets/b0682ca8-0139-4cdf-9471-21f38e2d82db" />

When I use my PlayStation 2 controller, a PlayStation d-pad is shown:

<img width="408" height="216" alt="Image" src="https://github.com/user-attachments/assets/f76abc14-4d2f-48f4-821b-9870e6c95e44" />

But when I use a Stadia controller, the keyboard hint is shown:

<img width="408" height="216" alt="Image" src="https://github.com/user-attachments/assets/369d5f33-4d63-4dc8-bf05-b0a9ae422970" />

I think the problem is that the controller is classed as "generic" and the wrong textures are used for the "controller default texture" and "controller pressed texture". 

<img width="649" height="156" alt="Image" src="https://github.com/user-attachments/assets/06639297-aab8-4a1d-ad0b-7ce9472b3af1" />

I suggest using the XBox or Steam textures for the generic case. (We could even remove the `controller_default_texture` and `controller_pressed_texture` properties and update the `_:` branches of the `match` statements in `scenes/game_elements/props/hint/input_key/directional_input_hint.gd` - this would match how `scenes/game_elements/props/hint/input_key/interact_input.gd`  works.)

### XP Summary (total 175)

| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

AlealarcViva:
Thanks for the heads-up! I’ve checked, and it does seem related to #1280.
I’ll work on a fix and submit a PR shortly.

PR #1482, opened 2025-11-03
Input hint: Fix generic controller showing keyboard directional hint

AlealarcViva:
Generic controllers were incorrectly displaying the keyboard directional input hint instead of a controller one.
This occurred because generic devices were not explicitly handled in the match statement, causing them to use keyboard textures as fallback.

Changes made:

Removed controller_default_texture and controller_pressed_texture properties.

Updated logic in scenes/game_elements/props/hint/input_key/directional_input_hint.gd to align with interact_input.gd.

Added fallback to Xbox-style textures for generic controllers.

Ensured consistent controller hint visuals across all platforms.

Fixes: https://github.com/endlessm/threadbare/issues/1478

### DIFF
--- a/scenes/game_elements/props/hint/input_key/directional_input_hint.gd
+++ b/scenes/game_elements/props/hint/input_key/directional_input_hint.gd
@@ -112,12 +112,13 @@ func _physics_process(_delta: float) -> void:
 				InputHelper.DEVICE_STEAMDECK_CONTROLLER:
 					texture = steam_controller_texture
 				_:
-					# Fallback controller image (prefer Xbox, then Steam)
-					texture = (
-						xbox_controller_texture
-						if xbox_controller_texture
-						else steam_controller_texture
-						)
+					texture = null
+
+			if not texture:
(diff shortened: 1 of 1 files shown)


scenes/game_elements/props/hint/input_key/directional_input_hint.gd
```diff
@@ -100,29 +97,34 @@ func _physics_process(_delta: float) -> void:
 			match current_device:
 				InputHelper.DEVICE_XBOX_CONTROLLER:
 					texture = (
-						xbox_pressed_texture if xbox_pressed_texture else controller_pressed_texture
+						xbox_pressed_texture if xbox_pressed_texture else null
 					)
```
wjt:
```suggestion
					texture = xbox_pressed_texture
```

because `x if x else null`, when `x` is either `null` or a non-`null` reference, means the same as `x`.

Same change on the other branches.

wjt:
image

For future reference, "Fixes:" and the issue link have to be on the same line for this to work.

wjt (approved):

PR #1484, opened 2025-11-03
ci: Cache Git LFS objects

wjt:
We clone the repo from scratch on every build. Previously, we fetched all LFS objects for the current commit each time, too.

GitHub has a monthly quota for free LFS bandwidth - we hit it last week.

But the LFS objects don't change at all between most builds that we run. So use GitHub Actions' built-in cache to cache and restore LFS objects. This technique is described here:
https://github.com/actions/checkout/issues/165#issuecomment-2776048200

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -32,12 +32,29 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
         with:
-          lfs: true
           # TODO: can we use --filter to avoid fetching the trees & blobs for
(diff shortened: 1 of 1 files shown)


manuq (approved):
Let's give it a try! Thanks for referencing the "cache for LFS" snippet.

wjt:
The Flatpak build still checks out all LFS files - a huge waste because it only needs a 4 KB PNG - but I tried and failed several different ways to do the same improvement there. And that job only runs conditionally, so…

PR #1491, opened 2025-11-04
Void: Add closing transition

manuq:
This is a pure cinematic scene. The StoryWeaver reaches the town and reproduces the story. The Void retreats. The people from the town celebrate and gives the last thread to the StoryWeaver.

Go to Fray's End when the cinematic ends.

Void + grappling hook level:
- Remove the thread, now given by dialogue in the closing transition.
- Add a teleporter to go to the new scene.
- Add layers for water and foam.

Fix https://github.com/endlessm/threadbare/issues/1410

To play it you can beat the Void + grapping hook level: https://play.threadbare.game/branches/endlessm/void-outro/#quests/lore_quests/quest_002/3_void_grappling/void_grappling

Or go here directly: https://play.threadbare.game/branches/endlessm/void-outro/#quests/lore_quests/quest_002/4_closing_transition/closing_transition


scenes/quests/lore_quests/quest_002/4_closing_transition/components/closing_cinematic.dialogue
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+do animation_player.play("enter")
+do animation_player.animation_finished
+Oh what do you have there? Is that a story from the monk?
+Tell us! Tell us!
+do Transitions.do_transition(reveal)
```
wjt, line 8: This causes the music to fade out and not fade back in. I wonder why! But I think we can address that later.

wjt (approved):
Nice! I like that the villagers give you the thread.

scenes/quests/lore_quests/quest_002/4_closing_transition/components/closing_cinematic.dialogue
manuq: Oh good catch, I haven't noticed it.

PR #1492, opened 2025-11-04
ci: Don't fetch all LFS objects for Flatpak build

wjt:
Git LFS. But it doesn't need the other tens of megabytes of assets,
which take time to fetch and also count against our monthly Git LFS
bandwidth quota.

It is possible to tell Git LFS to only check out certain large files,
rather than all LFS objects referenced in the working copy. But there is
an easier way: we can use git's sparse-checkout feature to only check
out the linux/ directory (plus the files in .) for this job. Then Git
LFS will only fetch the objects used in that directory.

To make this work we have to change linux/icon-256.png from a symbolic
link to a copy of the icon from assets, but I think that is a price
worth paying.

I also updated the workflow to run the Flatpak job automatically if the workflow itself changes; and to use the 25.08 flatpak-builder image.

### DIFF
--- a/.github/workflows/export.yml
+++ b/.github/workflows/export.yml
@@ -126,7 +126,7 @@ jobs:
   flatpak:
     name: Build Flatpak
     needs: build
-    if: ${{ inputs.build_flatpak || github.event_name == 'release' || needs.build.outputs.linux-changed == 'true' }}
+    #if: ${{ inputs.build_flatpak || github.event_name == 'release' || needs.build.outputs.linux-changed == 'true' }}
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/flathub-infra/flatpak-github-actions:freedesktop-25.08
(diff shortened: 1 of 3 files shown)


.github/workflows/export.yml
```diff
@@ -125,21 +126,25 @@ jobs:
...
     container:
-      image: ghcr.io/flathub-infra/flatpak-github-actions:freedesktop-24.08
+      image: ghcr.io/flathub-infra/flatpak-github-actions:freedesktop-25.08
       options: --privileged
     steps:
       - uses: actions/checkout@v5
         with:
           lfs: true
+          sparse-checkout: |
+            linux
```
manuq, line 139: Nice!

manuq (approved):
Thanks for digging into this issue, I agree that not keeping the icon as a symbolic link is preferred if this is the price.

Issue #491, opened 2025-05-07
Review Sokoban introductory dialogue


wjt:
The [following dialogue](https://github.com/endlessm/threadbare/blob/main/scenes/eternal_loom_sokoban/sokoban.dialogue) plays when you interact with the Eternal Loom and start the first of three Sokoban puzzles:

> It looks like there are some broken threads inside The Eternal Loom!
> You'll need to push each bow onto the matching frayed section to fix it.
> If you get stuck, you can undo a step, or reset entirely!

We may want to adjust this text.

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Narrative & Storytelling | Problem Solving | 50 |
| Game Design: Narrative & Storytelling | Creativity & Design Thinking | 75 |
| Game Design: Narrative & Storytelling | Technical Literacy | 50 |

wjt:
@hydrolet said she will look at this!

wjt:
@saspiers Do you want to take a look at lore-ifying this dialogue?

JuaniCargnel:
Hi @manuq I wanted to ask if it’s possible to assign this issue to me so I can show the students from More: Threadbare the process of resolving and submitting an issue. (I chose this one for convenience, but if there’s another one you’d prefer me to take, that’s totally fine!) 
I think the visual example (I’m going to record a short video) would be helpful for the students, and I suggested it in class — they agreed.
Thank you so much!

PR #1493, opened 2025-11-04
Change the introductory dialogue for the Sokoban puzzle.

JuaniCargnel:
The dialogue now feels friendlier and more expressive, while keeping the explanation of the Sokoban rules and controls.

Fixes https://github.com/endlessm/threadbare/issues/491

### DIFF
--- a/scenes/eternal_loom_sokoban/sokoban.dialogue
+++ b/scenes/eternal_loom_sokoban/sokoban.dialogue
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 ~ start
-Oh dear… it looks like a few threads inside the Eternal Loom have come loose!
+Oh dear... it looks like a few threads inside the Eternal Loom have come loose!
 Without them, the fabric of the world might start to unravel — what a mess that would be!
 Your task is to push each bow onto its matching frayed thread to mend the weave.
 Don’t worry if things get tangled — you can always undo your last step or reset the puzzle entirely.


scenes/eternal_loom_sokoban/sokoban.dialogue
```diff
@@ -1,7 +1,9 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 ~ start
-It looks like there are some broken threads inside The Eternal Loom!
-You'll need to push each bow onto the matching frayed section to fix it.
-If you get stuck, you can undo a step, or reset entirely!
+Oh dear… it looks like a few threads inside the Eternal Loom have come loose!
```
manuq: I also prefer using the "…" unicode character. But the dialogue plays character by character, so having 3 separate period characters looks better:
```suggestion
Oh dear... it looks like a few threads inside the Eternal Loom have come loose!
```

We have three period characters in other dialogues, like:

https://github.com/endlessm/threadbare/blob/31ab539ea2a5473e630af9c293409dd193011791/scenes/quests/lore_quests/quest_001/1_music_puzzle/components/dialogues/musician.dialogue#L14-L18

JuaniCargnel: Great, thanks @manuq for the suggestion! To apply it, do I just press “Commit suggestion” and resolve the conversation? Or is there anything else I should do?

manuq: Yes, you can do that and then we squash and merge.

Issue #1147, opened 2025-09-03
Incorporate broken bridges sprite


wjt:
In [#334](https://github.com/endlessm/threadbare/issues/334#issuecomment-3246666459),  @159adrian updated the bridges sprite sheet to add broken horizontal and vertical segments:

> <img width="192" height="256" alt="Image" src="https://github.com/user-attachments/assets/2c764530-4040-4c1b-8c1b-ac824d445261" /> 

To incorporate this into the game so that it can be used in levels:

- Add the PNG to `res://assets/first_party/tiles`.
- Update the texture in `res://tiles/bridges.tres` source 3 to use this PNG as the texture rather than the Tiny Swords `Bridge_All.png` texture - I believe it's a drop-in upgrade
- Add the new tiles so they can be used

Then we will have to add collisions so that the player cannot cross the broken bit of the bridge. Perhaps confusingly the way that we do this is by having alternatives of the water tile:

- Under passable bridges we place Alternative 1 which has no collision
- Under the existing half-bridge, used in fray's end, we place Alternative 2 which has a collision at the bottom

So for the new tiles we will need new alternative waters with different collision shapes.

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 50 |
| Game Design: Core Gameplay & Mechanics | Creativity & Design Thinking | 50 |
| Game Design: Core Gameplay & Mechanics | Technical Literacy | 75 |

sebastianangulod:
I plan to work on this

wjt:
@sebastianangulod We recently rearranged the tilesets. This tile should be added to `tiles/bridges.tres`; `scenes/tileset.tres` is now unused.

sebastianangulod:
Okay thank you. I will start working on it during this week. Here is the planning sheet as a bit of progress: https://docs.google.com/document/d/1rnFPaNhLRURBxQ5oPerFnZz69H8MklFgqf7eFrBXHNo/edit?usp=sharing

PR #1496, opened 2025-11-05
Incorporate broken bridges sprite

sebastianangulod:
In assets/first_party/tiles, implement the new bridges sprite with the broken bridges
In tiles/bridges.tres, implement broken bridges sprites to the tileset with collisions.
Fixes: #1147

### DIFF
--- a/scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn
+++ b/scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn
@@ -1509,7 +1509,7 @@ curve = SubResource("Curve2D_306df")
 
 [node name="PlayerFollower" type="PathFollow2D" parent="ForestSoundPath"]
 position = Vector2(64, 2088)
-rotation = -1.5493928
+rotation = -1.58865
 script = ExtResource("22_nonj7")
 
 [node name="AudioStreamPlayer2D" type="AudioStreamPlayer2D" parent="ForestSoundPath/PlayerFollower"]
@@ -1527,7 +1527,7 @@ curve = SubResource("Curve2D_l8fgn")
 
 [node name="PlayerFollower" type="PathFollow2D" parent="RiverSoundPath"]
 position = Vector2(-408, 1104)
-rotation = -0.12495447
+rotation = -0.0696546
 script = ExtResource("22_nonj7")
 
 [node name="AudioStreamPlayer2D" type="AudioStreamPlayer2D" parent="RiverSoundPath/PlayerFollower"]


scenes/quests/lore_quests/quest_001/4_closing_transition/closing_transition.tscn
```diff
@@ -185,7 +185,7 @@ tile_map_data = PackedByteArray("AAAQAAIADQAAAAAAAAAQAAMADQAAAAIAAAARAAIADQABAAA
 tile_set = ExtResource("9_u7k2u")
 
 [node name="Bridges" type="TileMapLayer" parent="TileMapLayers"]
-tile_map_data = PackedByteArray("AAAKAAUAAwAAAAAAAAALAAUAAwABAAAAAAAMAAUAAwABAAAAAAANAAUAAwABAAAAAAAOAAUAAwACAAAAAAASAAUAAwAAAAAAAAATAAUAAwABAAAAAAAUAAUAAwABAAAAAAAVAAUAAwABAAAAAAAWAAUAAwACAAAAAAA=")
```
wjt, line 188: Can you take a screenshot of what you have changed here?

wjt: OK it's here:

Screenshot from 2025-11-05 09-32-13

I don't think this really makes sense because StoryWeaver walks across this bridge in the cutscene. I think if you want to use the new tiles, you could put them in `res://scenes/quests/lore_quests/quest_003/1_stealth_sequence_combo/stealth_sequence_combo.tscn`, something like this:

Screenshot from 2025-11-05 09-34-01

But please revert this change to `res://scenes/quests/lore_quests/quest_001/4_closing_transition/closing_transition.tscn`.

sebastianangulod: I was testing the tileset and I though I reverted that change, my bad. Check it again.

tiles/bridges.tres
```diff
@@ -35,6 +35,11 @@ texture_region_size = Vector2i(64, 64)
 0:3/0/terrain = 0
 0:3/0/terrains_peering_bit/top_side = 0
 2:3/0 = 0
+2:2/0 = 0
+2:2/0/physics_layer_0/polygon_0/points = PackedVector2Array(-19.593307, -26.124409, -18.799213, 22.622784, 23.897308, 22.941414, 23.791874, -26.124409, -20.059814, -26.124409)
+1:3/0 = 0
+1:3/0/physics_layer_0/polygon_0/points = PackedVector2Array(-26.530273, -21.450008, 27.402245, -21.029629, 27.402245, 15.294277, -27.094748, 15.805267, -27.094748, -22.578957)
+1:3/0/physics_layer_1/polygon_0/points = PackedVector2Array(-27.523932, -21.925844, -27.523932, 19.593304, 26.590916, 19.126797, 26.590916, -20.99283, -27.523932, -20.99283)
```
wjt: Sorry, I missed this the first time through. This is a collision on the tileset's physics layer 1 which corresponds to "sight occluders".

I don't think that's right. I don't think the bridge occludes sight. I think this line should be removed.

```suggestion
```

PR #1508, opened 2025-11-05
Stealth/sequence combo: Use broken-bridge tile

wjt:
My original idea with the wrecked-bridge tiles on this level was to
imply that, in a former time, it was possible to walk horizontally
between all four islands; but now the player must take the longer
routes, avoiding the guards.

Make this even more tantalising by placing a broken bridge between the
third and fourth islands. This doesn't change the gameplay but I think
it's a nice touch - plus it uses the nice new tile!


wjt:
image

@sebastianangulod What do you think?

Issue #679, opened 2025-05-21
Sequence puzzle object sprite has 0.7 scale factor


wjt:
I hadn't noticed before today, but in 9defdbab361471bd92d371357f990b490c62e5f3 which swapped from a placeholder rock to a fabric rock asset, a scale of (0.7, 0.7) was applied to the AnimatedSprite2D.

This makes it harder to design new assets for sequence puzzle objects, because when used in-game they will be scaled relative to the player.

To address this I would suggest working as follows:

1. Reset the scale factor in `res://scenes/game_elements/props/sequence_puzzle_object/sequence_puzzle_object.tscn` to (1, 1)
2. Find all scenes that reference it:
   
   <img width="667" height="344" alt="Image" src="https://github.com/user-attachments/assets/d9104428-873f-4c0b-8d9d-54730750873b" />
3. In each scene, determine what adjustments need to be made, if any.

I expect that the Stella scene will need the most changes because the six objects are carefully positioned to line up with a turtle asset. Other scenes will probably be OK with a bit of tweaking of positions.

### XP Summary (total 300)

| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 150 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 75 |
| Game Design: Game Feel & UX | Technical Literacy | 75 |

Miguel3098:
Here is my planification sheet: https://docs.google.com/document/d/1GsufMi8qi63vm27SbmSnx29J8NhyLUPwPNJW8RVeAp0/edit?tab=t.0

PR #1518, opened 2025-11-06
Change the sequence puzzle object sprite 

Miguel3098:
Make the required change to the sequence object in the puzzle and verify that it matches in each scene.
Fixes https://github.com/endlessm/threadbare/issues/679

### DIFF
--- a/scenes/game_elements/props/sequence_puzzle_object/sequence_puzzle_object.tscn
+++ b/scenes/game_elements/props/sequence_puzzle_object/sequence_puzzle_object.tscn
@@ -15,7 +15,7 @@ script = ExtResource("1_kw7av")
 
 [node name="AnimatedSprite2D" type="AnimatedSprite2D" parent="."]
 unique_name_in_owner = true
-scale = Vector2(0.7, 0.7)
+scale = Vector2(1.1, 1.1)
 sprite_frames = ExtResource("2_cltuv")
 
 [node name="CollisionShape2D" type="CollisionShape2D" parent="."]


scenes/game_elements/props/sequence_puzzle_object/sequence_puzzle_object.tscn
```diff
@@ -15,7 +15,7 @@ script = ExtResource("1_kw7av")
 
 [node name="AnimatedSprite2D" type="AnimatedSprite2D" parent="."]
 unique_name_in_owner = true
-scale = Vector2(0.7, 0.7)
+scale = Vector2(1.1, 1.1)
```
wjt, line 18: This new scale is wrong. The objective was to have the scale set to 1 (unscaled) so that the sprite appears at exactly its original size.

wjt (changes requested):
Once you have adjusted the sprite scale you'll also need to make adjustments to the surrounding game elements.

With the new scale, the collision shape (the oval) is now much smaller than the visible size of the object:

Image

Image

As a result the player can walk through the rock/shape.

----

Here is `stella_sequence_puzzle.tscn` before this change:

Image

and after this change:

Image

This scene was the reason I noticed this strange scaling & what motivates this change. The objects are meant to be gems on the back of the tortoise. But with the objects' scaling having been changed, they no longer fit. The Tortoise sprite has been scaled, and so have the Gems in this scene. You'll need to adjust these to fit together.

manuq:
@Miguel3098 have you seen the feedback from Will? Let us know when this is ready for review again.

PR #1517, opened 2025-11-06
Input hints:Refactor input hint system to use resource-based textures

AlealarcViva:
Replaces hardcoded input hint textures with a resource-driven system using JoypadButtonTextures resources for each device type. Adds InputHintManager singleton to handle device detection and texture lookup. Updates directional_input_hint.gd and movement_input_hints.tscn to use the new system, simplifying node configuration and making it easier to extend support for new devices. Also adds .gdignore files for unused third-party input prompt assets to reduce import time. 
**Notes**:
This work was done in a new branch updated with the latest main changes to avoid conflicts.

Resolves #1285
Resolves #1499
**Related to:**
#1511

PR #1532, opened 2025-11-10
Input hints: Remove one-off script create_gdignore.sh

AlealarcViva:
The create_gdignore.sh script was a temporary helper used to generate .gdignore files.
It is no longer needed and should not be tracked by Git. Removing it fixes the license
check failure related to this file.

Fixes https://github.com/endlessm/threadbare/issues/1517


AlealarcViva:
@manuq I have made the changes and pushed them as you suggested. Thanks for the guidance

Issue #1185, opened 2025-09-14
Player can be moved after completing Sokoban puzzle


wjt:
### Description

When you finish a Sokoban puzzle, you are taken to the next scene (another puzzle in the sequence, or back to Fray's End). As with all our scene transitions, there is a 1-second fade-out. During this time, it is possible to move the player:

https://github.com/user-attachments/assets/8b39b63e-1e02-4fc2-9e31-3d7016ec19bb

I think it should not be possible to move the player during the transition.

### XP Summary (total 225)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

sebastianangulod:
Here is the planning sheet for this: https://docs.google.com/document/d/1ehqm5GkhfRLtEXfRoM6RmAkKHek51KsUBqEBVL_3z4E/edit?tab=t.0

PR #1553, opened 2025-11-12
Fix player can be moved after completing Sokoban puzzle

sebastianangulod:
In Sokoban Puzzles, fix being able to move the player during the small time of transition to the next scene.

Fixes: #1185


scenes/eternal_loom_sokoban/components/system/rules/rule_engine.gd
```diff
@@ -178,7 +178,7 @@ func _check_goal() -> void:
 
 	if result:
 		goals_reached.emit()
-
+		directional_input.enabled = false
```
manuq, line 181: Ah this does the trick!

manuq (approved):

Issue #1544, opened 2025-11-11
InkDrinker_1: Camera bounds show areas outside the scene when using aspect ratios other than 16:9


russell-cj:
### Problem Description
When playing the **InkDrinker 1** level (*scene: `ink_combat_round_1`*) with an aspect ratio different from **16:9** (for example, *21:9* or *4:3*), the camera bounds allow parts of the scene to be visible that were not meant to be seen.  
This happens when the game window expands horizontally or vertically, exposing areas outside the region "dressed" by the level designer.

### Steps to Reproduce
1. Open the project in the **Godot** editor.  
2. In **Project Settings**, enable **Debug Aspect Ratio**.  
3. Start the level **InkDrinker 1** (*scene: `ink_combat_round_1`*).  
4. Resize the window (for example, make it wider or taller).  
5. Observe that the camera shows undesigned or empty areas outside the intended playable region.

### Expected Result
The camera should remain within the designed playable area and not display regions outside the intended environment.

---

### Suggested Fix
This issue could be resolved by **adding more decoration (e.g., water)** around the edges of the scene and **adjusting the camera limits** so that the view stays centered on the action.  

Currently, the visible tiles are **19 horizontally** and **11 vertically**, so the approximate bounds should be:  
- Right limit: `19 * 64`  
- Bottom limit: `11 * 64`

>[!NOTE]
>- Found while testing the main issue #1083 .
>- Scene name: `ink_combat_round_1`

russell-cj:
<img width="728" height="903" alt="Image" src="https://github.com/user-attachments/assets/3a213439-30ba-4fcb-825f-68a75a0623c6" />

<img width="1238" height="512" alt="Image" src="https://github.com/user-attachments/assets/ff14ebbe-7cbe-4b60-9112-fb2f0674e683" />

PR #1566, opened 2025-11-13
InkDrinker_1: Adjust camera limits and improve visuals to prevent out-of-bounds area

russell-cj:
This *pull request* fixes an issue in the **InkDrinker 1** level (`ink_combat_round_1.tscn`) where the camera showed unintended areas of the scene when using aspect ratios different from 16:9 (e.g., 21:9 or 4:3).

You can test the level directly here:
[Play InkDrinker_1 level](https://play.threadbare.game/branches/endlessm/issue_1544/#quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1)

resolves: #1544


manuq (approved):
Excellent! Is nice that you also adjusted the foam. The exit door at the top goes no-where but I guess it's way out from the safe zone (I havent checked with the red overlay).

PR #1573, opened 2025-11-13
Switch background music when ink combat completed

wjt:
Previously, the intense combat music would continue to play even after
the final round had been completed and the collectible revealed.

Define an AudioStreamInteractive with 2 clips, the combat music and the
bed used in Fray's End. Set this as the background music stream on all 6
ink combat scenes, rather than setting the combat music directly: this
ensures that the music doesn't restart at the beginning between rounds 1
and 2 of combat.

Then, add a method to background_music.gd to switch to a different clip.
Connect to the goal_reached signal and, in addition to playing the
animation that reveals the collectible, also switch the music to the
the "cozy" clip.

This doesn't work quite right when switching between
quest_003/2_ink_drinker_levels/ink_combat_round_6.tscn and
frays_end.tscn. frays_end.tscn uses the less-dramatic bed music directly;
ink_combat_round_6.tscn uses a AudioStreamInteractive. background_music.gd
treats these as different streams and so restarts the music when switching to
frays_end.

But it sounds OK and I think the trade-off is worth it. I think this could be
improved in future by going further down this path: put *all* the background
music into an AudioStreamInteractive, with appropriate fades; or even have our
background music logic build an AudioStreamInteractive dynamically.

### DIFF
--- a/scenes/game_elements/props/background_music/components/background_music.gd
+++ b/scenes/game_elements/props/background_music/components/background_music.gd
@@ -69,3 +69,18 @@ func _exit_tree() -> void:
 	## in the editor, the music will keep playing.
 	if Engine.is_editor_hint():
 		_stop()
+
+
+## If [member stream] is an [AudioStreamInteractive], and music is playing,
+## switch to [param clip_name].
+func switch_to_clip(clip_name: StringName) -> void:
+	if not audio_stream_player.is_playing():
+		push_warning("Not currently playing")
+		return
+
+	if audio_stream_player.stream is not AudioStreamInteractive:
+		push_warning("Switching clips is only supported with AudioStreamInteractive")
+		return
+
+	var playback := audio_stream_player.get_stream_playback() as AudioStreamPlaybackInteractive
+	playback.switch_to_clip_by_name(clip_name)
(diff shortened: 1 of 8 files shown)


scenes/game_elements/props/background_music/components/background_music.gd
```diff
@@ -69,3 +69,18 @@ func _exit_tree() -> void:
...
+
+## If [member stream] is an [AudioStreamInteractive], and music is playing,
+## switch to [param clip_name].
+func switch_to_clip(clip_name: StringName) -> void:
+	if not audio_stream_player.is_playing():
+		push_warning("Not currently playing")
+		return
+
+	if audio_stream_player.stream is not AudioStreamInteractive:
+		push_warning("Switching clips is only supported with AudioStreamInteractive")
```
manuq, line 82: I like the use of `push_warning`s here

scenes/quests/lore_quests/quest_001/2_ink_combat/components/combat_music.tres
```diff
@@ -0,0 +1,21 @@
+[gd_resource type="AudioStreamInteractive" load_steps=3 format=3 uid="uid://dkgtyb0x1i0m2"]
```
manuq, line 1: Welcome new node!

manuq (approved):

wjt:
I added the missing word

Issue #1584, opened 2025-11-14
Input action error in directional_input_hint.gd


manuq:
### Description

This happens since we merged https://github.com/endlessm/threadbare/pull/1517 .

When playing the scene `res://scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_start.tscn` I'm getting this error:

```
E 0:00:08:876   directional_input_hint.gd:77 @ _physics_process(): The InputMap action "move_unpressed" doesn't exist. Did you mean "move_up"?
  <Error C++>   Condition "!InputMap::get_singleton()->has_action(p_action)" is true. Returning: false
  <C++ Fuente>  core/input/input.cpp:376 @ is_action_pressed()
  <Rastreo de Pila>directional_input_hint.gd:77 @ _physics_process()
```

It happens in every physics process, which is very often.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #1585, opened 2025-11-14
Directional input hint: Fix inexisting action name

manuq:
The string "move_unpressed" is used by this script, but is not an actual input action of the game. So checking for
Input.is_action_pressed("move_unpressed") produces an error in every physics process step.

Fix https://github.com/endlessm/threadbare/issues/1584


manuq:
@AlealarcViva I will merge this without your review because it's pretty serious.

Issue #1135, opened 2025-09-02
Improve the musician asset


manuq:
The first Lore quest is called "The Musician's Quest" so this is an important character. The pixel art needs a few improvements to be aligned with [our visual style guidelines](https://github.com/endlessm/threadbare/discussions/1071):

<img width="642" height="490" alt="Image" src="https://github.com/user-attachments/assets/44e117cf-1e17-4de3-946a-779980bfc66f" />

For example:
- It doesn't have a shadow.
- The perspective is a bit odd, is being looked from the side like standing on a platform, not from the top. See the feet.
- It doesn't have a 2 pixels outline.
- The face is using too many shades of skin color, it should be cel shaded with 3 colors.

Also the hair color is full white (#fff) which is not in the guidelines, but not ideal.

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Art: Animation & Effects | Problem Solving | 50 |
| Art: Animation & Effects | Creativity & Design Thinking | 75 |
| Art: Animation & Effects | Technical Literacy | 75 |

Omar071201:
![Image](https://github.com/user-attachments/assets/0b550828-2a33-48a0-8791-f50d412e78ff)

manuq:
@Omar071201 thanks! I think that works great for a special "guitar solo playing" animation. Maybe celebrating when the player solves the puzzle? Or for the outro scene of the musician quest? I would expect to hear the guitar as well when this animation is playing, but we can't do that during the music puzzle or it will be very confusing.

For an idle animation is also a bit too charged: Multiple things overlapping:
- tapping the foot to the rhythm
- hands moving
- music notes coming out in 3 colors
- body going up and down

Idle animations should be subtle, or it will grab the player eye for no reason. I think that if you make the character just hold the guitar (not playing) and leave only two:
- tapping the foot to the rhythm
- body going up and down

It will be more clear, and the tapping will be more readable.

So please keep this animation with a different name like "playing guitar", but also provide an improved "idle" animation.

As per a single frame, from the list in the description you solved:
- Fixing the shading to be 3 variants per color (cel shading)
- Adding a shadow
- Fixing the perspective (at least in the feet)

And missed:
- Fixing the outline to be 2 pixels wide
- Replacing the full white color (#FFFFFF) in hair, eye brows, teeth (and since now is smiling is even more noticeable). A light gray, almost white, would be better. Here I tried #b2bcc8 and #99a6ac as a quick example:

[...]

PR #1590, opened 2025-11-14
New musician sprite

Omar071201:
This PR adds the new idle animation style for the musician, following Threadbare's visual style.

Fixes: #1135


Omar071201:
@manuq I had to create a new one in order to make the necessary changes.

manuq (changes requested):
Thanks @Omar071201 this is already an improvement. I suggest you also make the teeth a tone of gray for the same reasons that the hair was changed. And ideally also close the mouth a bit as it was before. Now it seems too open for only showing the teeth. What do you think? Did you want to give the musician a big smile?

Image

Also please fix these issues with alpha in the shadow,. make it a single plain area:

Image

Here is how it looks in game. To me the head up-down movement seems a bit too frenetic for the calm level. Previously it was more subtle, just moving the leg, like waiting patiently. Now the leg movement is barely noticeable because the head movement is stronger. What do you think?

https://github.com/user-attachments/assets/ae4ebfdb-48b4-4b6c-8180-8a62e7febc43

scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn
manuq, line 1: The change in this scene is unrelated. Please restore it from the main branch! The rest looks good, great asset and animation!

Issue #1511, opened 2025-11-05
Use texture atlases and remove unused input hint assets


wjt:
While looking at why building & publishing the game #1483, I noticed that a significant amount of time seems to be spent importing the assets from `assets/third_party/inputs/kenney_input-prompts_1.4/`. There are `3979` assets here, counted by the number of `.import` files:

```console
$ find assets/third_party/inputs/kenney_input-prompts_1.4/ -name '*.import' | wc -l
3979
```

This is 83% of the total number of assets in the whole project:

```console
$ find -name '*.import' | wc -l
4783
```

I hadn't realised there were quite so many more files in this pack compared to the previous one (which had 310 files - still more than 25% of the total at the time!) and that this would have a significant impact on import & export time.

I suggest that rather than using the individual assets, like `assets/third_party/inputs/kenney_input-prompts_1.4/Steam\ Deck/Default/steamdeck_button_a.png`, let's instead use the atlases that contain all the icons for a given controller, like `assets/third_party/inputs/kenney_input-prompts_1.4/Steam\ Deck/steam-deck_sheet_default.png`. We can use an `AtlasTexture` to reference just a small area from the bigger image. There is an XML file for each of these sheets, e.g. `assets/third_party/inputs/kenney_input-prompts_1.4/Steam\ Deck/steam-deck_sheet_default.xml`, which lists out which buttons are at which coordinates. We can either manually create an AtlasTexture resource file for each of the small number of buttons we need (with reference to the coordinates in that file), or look for a tool or addon that can do this for us. I found https://bitbucket.org/iseesharp83workspace/kenney-spritesheet-importer-godot-plugin in a quick search, it looks like it might work.

Then we should delete all of the files we're not using. Maybe we keep the unified sheet for other controller types, even if we're not showing them right now.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Optimization & Performance | Problem Solving | 150 |
| Engineering: Optimization & Performance | Creativity & Design Thinking | 75 |
| Engineering: Optimization & Performance | Technical Literacy | 150 |

PR #1593, opened 2025-11-14
Input hints: Use texture atlases and remove unused assets

AlealarcViva:
This change replaces the thousands of individual input-hint icons from the Kenney pack with atlas-based textures to reduce import and export time significantly.
The previous implementation imported 3,979 separate images, which represented over 80% of the project's assets. Using atlases drastically reduces the number of imported resources and improves build performance.

I created atlas .tres files for each controller type using an internal tool (the public plugin was not compatible with Godot 3). Only the icons currently used by the game were preserved.
All unused device folders were removed or marked with .gdignore. The remaining PNG sheets and XML metadata are now stored in assets/third_party/inputs/atlas_kenney_input_prompts_1.4/, grouped by controller type.

The atlas icons use the default 64×64 PNG sheets; visually they match the pixel-art style of the game, but the team can evaluate switching to the higher-resolution double-PNG sheets if needed.

Additionally, the erroneous reference to the non-existent move_unpressed action has been removed to prevent runtime errors during physics processing.

**Fixes https://github.com/endlessm/threadbare/issues/1511

Fixes https://github.com/endlessm/threadbare/issues/1584**


AlealarcViva:
I was trying to fix the bug and I found that it was calling the wrong .tres files, so Godot couldn’t find them. I fixed that issue, but it seems the root cause was something else. Still, when I ran it, no errors appeared on my side. 


Uploading prueba 1.mp4…

manuq:
@AlealarcViva is hard to spot in the GitHub view because of the many files changed. But the root of the conflict seems to be that we have squashed and merged PR https://github.com/endlessm/threadbare/pull/1517/ with suggestions applied like [stylistic changes](https://github.com/endlessm/threadbare/pull/1517/commits/0ed1886c23c3aff98d6269c80513c6198306156c) or [removing redundant cache](https://github.com/endlessm/threadbare/pull/1517/commits/a74bac7e9e984e79baeb10b6d612b9b1efbcbc1d). But instead of sync'ing your branch after it was merged, you continued in the same branch. So now are those applied suggestions which are entering conflict.

wjt:
I think this PR was superseded by https://github.com/endlessm/threadbare/pull/1635? If so, can we close this one?

Issue #1604, opened 2025-11-16
NO_EDIT_stealth – Visual issues with non–16:9 aspect ratios


russell-cj:
When testing the `NO_EDIT_stealth.tscn` scene using aspect ratios other than **16:9**, multiple visual inconsistencies appear. Some areas of the environment become exposed, showing empty or unintended regions, and certain objects fail to align correctly depending on the window's aspect ratio. These issues occur when widening or stretching the game window horizontally or vertically.

<img width="1640" height="762" alt="Image" src="https://github.com/user-attachments/assets/5030ee01-09c8-4920-9777-2fb409e14dc6" />

https://github.com/user-attachments/assets/733616be-97a1-4edd-ad8f-735610676ca1

### Reproduction Steps
1. Open the scene:  
   `res://scenes/quests/template_quests/NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth.tscn`
2. Run the scene using aspect ratios other than **16:9** (e.g., 4:3, 21:9, 5:4).
3. Observe any exposed areas, camera boundary issues, or misaligned elements.

### Expected Result
- The camera should remain within the intended playable and decorated area.
- No empty or unintended sections of the scene should be visible at any supported aspect ratio.

### Suggested Solution
- Add additional decoration (e.g., stones, grass, walls, foliage) around the scene borders to cover all camera-reachable areas.
- Adjust camera limits to ensure the framing remains centered on the designed region and does not reveal out-of-bounds content.

### Notes
> [!NOTE]
> - Found while testing the main issue **#1083**.  
> - Scene path: `res://scenes/quests/template_quests/NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth.tscn`

PR #1618, opened 2025-11-17
NO_EDIT_stealth: Fix visual issues on non–16:9 aspect ratios

russell-cj:
This *pull request* fixes an issue in the **NO_EDIT_stealth** scene (`NO_EDIT_stealth.tscn`) where empty or unintended areas became visible when using aspect ratios different from 16:9 (such as 21:9, 4:3, or 5:4).

Extra tiles were added to the `Stone` TileMap layer to cover exposed borders, and the `Camera2D` limits were updated so the camera remains properly centered within the intended playable area across all supported aspect ratios.

[Play NO_EDIT_stealth](https://play.threadbare.game/branches/endlessm/issue_1604/#quests/template_quests/NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth)

resolves: #1604


scenes/quests/template_quests/NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth.tscn
```diff
@@ -62,14 +63,16 @@ position = Vector2(131, 463)
...
+limit_bottom = 1280
 position_smoothing_enabled = true
 editor_draw_limits = true
 
 [node name="EnemyGuards" type="Node2D" parent="."]
 y_sort_enabled = true
 
 [node name="Guard1-GoingBackAndForth" parent="EnemyGuards" node_paths=PackedStringArray("patrol_path") instance=ExtResource("6_fg25g")]
-position = Vector2(526, 470)
+position = Vector2(525.9997, 470.00003)
```
manuq, line 75: I notice that Godot editor started doing this since we last upgraded: aproximating an integer with a floating point. I wonder why. Maybe is worth checking if there is a bug upstream about this, and open one if not.

PR #1626, opened 2025-11-17
Add Singularidad StoryQuest

PrismenoDrako:


scenes/tileset.tres
manuq, line 1: This is something I overlooked when reviewing https://github.com/endlessm/threadbare/pull/1538. This is the tileset used for the whole game. I'm sorry but you can't modify it. Instead, you should:
- Duplicate this file to your own StoryQuest folder. For example to `singularity_tileset.tres`
- Change the TileMapLayers in your scenes to use your own copy.
- Restore this `scenes/tileset.tres` file to how it was before.

Issue #1606, opened 2025-11-16
NO_EDIT_sequence_puzzle.tscn – Visual issues with non–16:9 aspect ratios


russell-cj:
While testing `NO_EDIT_sequence_puzzle.tscn` at aspect ratios other than 16:9, parts of the puzzle area or surrounding scene become visible beyond the intended play region. This may reveal undeveloped edges.

<img width="1911" height="843" alt="Image" src="https://github.com/user-attachments/assets/9cb6ba25-feb7-4627-9877-d269f95b75c1" />

### Reproduction Steps
1. Open the scene:  
   `res://scenes/quests/template_quests/NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sequence_puzzle.tscn`
2. Run the scene with aspect ratios other than **16:9** (e.g., 4:3, 21:9).
3. Resize the window (wider and taller) and observe the puzzle area, camera bounds, and UI hints.

### Expected Result
- The sequence puzzle area and its surroundings should remain visually complete and centered, with no out-of-bounds regions shown .

### Suggested Solution
- Add or extend decorative tiles/elements around puzzle edges to hide potential gaps.
- Adjust camera clipping/limits so the viewport cannot reveal outside regions.

### Notes
> [!NOTE]
> - Found while testing the main issue **#1083**.  
> - Scene path: `res://scenes/quests/template_quests/NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sequence_puzzle.tscn`

PR #1629, opened 2025-11-18
NO_EDIT_sequence_puzzle: Fix visual issues on non–16:9 aspect ratios

russell-cj:
This *pull request* fixes an issue in the **NO_EDIT_sequence_puzzle** scene (`NO_EDIT_sequence_puzzle.tscn`) where out-of-bounds or undeveloped regions became visible when using aspect ratios different from 16:9 (such as 21:9 or 4:3).

A new `Void` TileMap layer was added to fill the outer edges of the scene, the `Camera2D` limits were updated to keep the puzzle centered, and the `Grass` and `Sand` tiles were slightly adjusted to match the new background. These changes ensure the entire puzzle area stays visually contained across all supported aspect ratios.

[Play NO_EDIT_sequence_puzzle](https://play.threadbare.game/branches/endlessm/issue_1606/#quests/template_quests/NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sequence_puzzle)

resolves: #1606


russell-cj:
Hi @manuq , @wjt , @jgbourque , @Stregoica777 
I created this PR taking into account the suggestions provided by the Learning team. It is mainly meant to give you a visual representation of how I interpreted the idea. As suggested, I only added a single node. I did modify the Camera2D node, but if the team decides that the camera should not be changed, I can restore it to its original state without any problem.

manuq:
@russell-cj thanks!

@jgbourque can you please confirm that this is what we want? And if so, do we want the same for the intro and outro? I ask because [a different solution](https://play.threadbare.game/branches/endlessm/issue_1606/#quests/template_quests/NO_EDIT/0_NO_EDIT_intro/NO_EDIT_intro) has landed for the intro (enlarging the background).

image

image

jgbourque:
Simply enlarging the background or I believe, as @russell-cj has done with the Stealth template: adding additional stone to expand that walled in look and cover the gray missing tiles is good I think.

But in the case of the Sequence puzzle, it might make more sense to just copy the intro.tscn solution of enlarging the background. That way the path the player walks in the cut scene has the appearance of continuing off screen instead of looking like it just leads to the void and ends abruptly on either side.

manuq:
> Hi @manuq , @jgbourque
> 
> I’ve just pushed a new commit to this PR.

Thanks! Before merging, I added a fixup commit to extend a bit the camera limits and to give the island a bit of shape, because it looked too much of a coincidence that the island had the same rectangular shape as the viewport :)

PR #1636, opened 2025-11-18
Add Somnus StoryQuest

GustavoCoc:
somnus juego completo beta 1


manuq (changes requested):
@GustavoCoc we decided to keep all StoryQuests PRs open for 3 weeks. Upon that time, if we don't hear back from you, we'll be closing them. If you can, please apply our suggestions so we can merge the PR. Thanks!

Issue #1638, opened 2025-11-19
InkCombatRound4: Implement the Speed Increase Mechanic


russell-cj:
Implement the main mechanic for Level 4: **progressive speed increase**.  
This level contains **3 vases** and **2 mosquitoes**, which must **shoot faster each time a vase is filled**.

### Tasks
- [x] Implement the progressive speed increase.
- [x] Adjust the mosquitoes' shooting rate for Level 4.
- [x] Apply the new mechanic within the existing Level 4 scene.

### Testing
- [x] Test that the mechanic works correctly with the progressive speed increase.
- [x] Adjust values based on testing results.

PR #1639, opened 2025-11-19
InkCombatRound4: Progressive Speed Increase Mechanic  

russell-cj:
This PR implements the progressive speed increase mechanic for Level 4.  
Each time a vase is filled, the mosquitoes increase their shooting speed, creating a more challenging gameplay experience.

### What This Solves  
- Adds progressive speed-increase logic.  
- Adjusts mosquito shooting speed specifically for Level 4.  
- Integrates the mechanic into the existing Level 4 scene.

`res://scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_4.tscn`

[Play Level 4](https://play.threadbare.game/branches/endlessm/issue_1638/#quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_4)

Resolves: #1638


scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/difficulty_ramp.gd
manuq, line 1: Since this script is specific to the round 4, I think it deserves to live inside a new `components` subfolder  scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/

This is also to avoid crowding `scenes/game_logic/`.

scenes/game_logic/difficulty_ramp.gd
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends Node
+
+@export var speed_multiplier: float = 1.5
+
+
+func _ready() -> void:
+	# Wait for the level to fully load
+	await get_tree().process_frame
```
manuq: I'm curious, what happens if you don't add this await line here? I ask because `scenes/game_logic/fill_game_logic.gd` does the same (iterating nodes in group `filling_barrels`) in its `_ready()` method and doesn't need it.

scenes/game_logic/difficulty_ramp.gd
```diff
@@ -0,0 +1,35 @@
...
+
+func _increase_enemies_speed() -> void:
+	var enemies: Array[Node] = get_tree().get_nodes_in_group("throwing_enemy")
+
+	for enemy_node: Node in enemies:
+		if is_instance_valid(enemy_node) and enemy_node is ThrowingEnemy:
+			var enemy: ThrowingEnemy = enemy_node as ThrowingEnemy
+
+			# Decrease the period to make them shoot faster
+			enemy.throwing_period = enemy.throwing_period / speed_multiplier
```
manuq: `speed_multiplier` is an exported float, so a user can set it to zero, leading to a zero division error. Maybe you can actually multiply it? And use `@export_range(0, 1, 0.01)` to make it between zero and one?

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/difficulty_ramp.gd
russell-cj: Sure, that makes sense. I’ll move this script into the new `components` subfolder inside `scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/` to help keep `scenes/game_logic/` organized. Thanks for the suggestion.

scenes/game_logic/difficulty_ramp.gd
russell-cj: You're right. I added it at first as a precaution, but I'll remove it since the nodes should already be ready at that point.

scenes/game_logic/difficulty_ramp.gd
russell-cj: That makes sense. I will switch to using a multiplication factor  to prevent division by zero .

Thanks for the detailed review @manuq! I'm working on these changes right now and will push the updates shortly.

Issue #248, opened 2025-04-17
Collectibles shine particles


pablitar:
This should be subtle and ideally a frame-by-frame animation of a simple shape. A good reference is the dust particles that appear when the player is running.

----

### XP Summary (total 475)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |
| Art: Animation & Effects | Problem Solving | 50 |
| Art: Animation & Effects | Creativity & Design Thinking | 75 |
| Art: Animation & Effects | Technical Literacy | 75 |

manuq:
@Cristhian036 hi, how is this going? Do you need help with the task? Do you understand what needs to be done?

manuq:
@Cristhian036 are you working on this?

Cristhian036:
Hi @manuq, I’ve had health issues that have prevented me from making progress on this task.

manuq:
@Cristhian036 no problem! If you would like to work on it, let us know and we'll assign you again.

Andersonu1290:
I plan to work on this whit @CristhianCastro685

Andersonu1290:
[#248 ESP Hoja de planificación](https://utpedupe-my.sharepoint.com/:w:/g/personal/u23203203_utp_edu_pe/IQCO0PN8VxVBSJG9vud-U7dTAbU7ePwkzTusBUzUukIK-kY?rtime=9OhfUdkn3kg)

manuq:
@Andersonu1290 @CristhianCastro685 I have iterated on your particles effect in https://github.com/endlessm/threadbare/pull/1794 .

PR #1652, opened 2025-11-20
Collectibles shine particles

Andersonu1290:
Modular Shine Effect on ButtonItem Pickup

The player now triggers a modular visual shine effect when collecting a `ButtonItem`. 
This item emits a signal that directly activates the `ShineParticles` node, 
instead of relying on a method inside `player.gd`. This ensures autonomy and consistency 
with other particle effects in the system.

The PNG used for the sparkle is the finalized asset named **Sparkin Sprite PNG**, 
which remains active and referenced.

Fixes [248#](https://github.com/endlessm/threadbare/issues/248)

### DIFF
--- a/scenes/game_elements/fx/shine_button/ShineParticles.tscn
+++ b/scenes/game_elements/fx/shine_button/ShineParticles.tscn
@@ -37,9 +37,11 @@ anim_speed_max = 1.0
 [node name="ShineParticles" type="GPUParticles2D"]
 modulate = Color(1, 1, 0, 1)
 material = SubResource("CanvasItemMaterial_tnibl")
+emitting = false
 amount = 1
 texture = ExtResource("1_vl6ir")
 lifetime = 0.5
+one_shot = true
 randomness = 0.38
 fixed_fps = 10
 process_material = SubResource("ParticleProcessMaterial_5lbww")


scenes/game_elements/characters/player/player.tscn
manuq: Hmm it looks like you forgot to `git add scenes/game_elements/characters/player/components/Brillo2.png` before commiting. Can you add it and push another commit?

Andersonu1290: Okay, can you see the glitter PNG now?

manuq: Yes I can see the player constantly shining particles now if I go to: https://github.com/endlessm/threadbare/pull/1652#issuecomment-3559199032

[Grabación de pantalla desde 2025-11-20 14-51-21.webm](https://github.com/user-attachments/assets/cadbe76f-aa33-44fd-be2a-02eea70e9a6a)

Maybe you forgot to turn them off before saving? So they only shine through the script lines you added.

scenes/game_elements/characters/player/components/player.gd
manuq: You can turn `One Shot` on in the Inspector directly, it doesn't have to be through code.

Andersonu1290: https://github.com/endlessm/threadbare/pull/1652/commits/7b2823086cf233ced52bb7e35e7c3c27209656fb
I'm still trying to figure out why the brightness activates when entering a scene.

Andersonu1290: I've fixed it, please check it. Thank you so much for the feedback, it helped me a lot.

Andersonu1290:
@manuq

scenes/game_elements/characters/player/player.tscn
manuq: This is changing the dust particles effect emitted when the player walks, so please revert this change.

scenes/game_elements/props/button_item/components/button_item.gd
manuq: Note: consider the following if you want the effect to be in the player. Check my note on where to add the effect.

This is a bit hacky. The button script is calling `activate()` in a node inside the player scene. A more elegant approach would be:
- In the player script, create a new method `func item_collected()` that does this.
- Here, use get_tree().call_group("player", "item_collected")

Check this documentation about [get_tree().call_group()](https://docs.godotengine.org/en/stable/tutorials/scripting/groups.html#using-code)

manuq (changes requested):
Thanks for posting a video that showcase the effect! It is much easy to review it that way.

In the video I see that the effect is added to the player. So first of all, can you tell us why do you want the shine effect to be in the player? The other option is that the effect is added "in the air" in the same position that the collectible was. Similar to the ink particles that you can see when an ink blob hits the player.

Unfortunately the task description is a bit vague. My understanding of it was that the 3 threads collectibles would be constantly shining particles. But I like adding a little effect when the buttons are collected! In this case, I think that the effect can work better if it is a little less repetitive. The asset you are using "Sparks_Sprite" has a frame-by-frame animation, but watching the video it looks like you don't have the particle node setup in a way that each frame appears individually. I can see a row of 4 frames appearing at the same time, making the effect a bit repetitive.

On assets:

I see 2 assets added in this pull request, `Brillo2.png` and `Sparks_Sprite.png`. But only the last one is used. So please remove `Brillo2.png` from the PR.

Also please provide the license of `Sparks_Sprite.png‎`. Where did you found the asset? It should be one of our [acceptable asset licenses](https://github.com/endlessm/threadbare/blob/main/docs/LICENSING.md#acceptable-asset-licenses)

Andersonu1290:
The Sparks_Sprite.png file

was created by us; it wasn't taken from anywhere else.

Andersonu1290:
I believe I have already complied with everything you suggested; I just need your confirmation. @manuq

scenes/game_elements/characters/player/components/player.gd
manuq: The player is already in this group! You don't have to add it by code.

Image

manuq: Please use UID for preloading, it is better because if we ever move the particles scene, we don't have to change this line and the script won't break:

```suggestion
		preload("uid://cfm58xiwevc2x")
```

Instead of: `preload("res://scenes/game_elements/characters/player/components/PlayerShineParticles.tscn")`

scenes/game_elements/characters/player/components/PlayerShineParticles.tscn
manuq: Since the shine particles is not in the player anymore, I suggest you rename the scene and the script to `ShineParticles.tscn` and `shine_particles.gd`. And move it to the folder `scenes/game_elements/fx/` where other effects live.

scenes/game_elements/props/button_item/components/button_item.gd
manuq: The button global position is at the bottom of the button (where the contact with the floor is, this is for Y sorting). So the particles appear below:

Image

A better approach would be to use the collision shape position (converting it to global position with `to_global()`):

Image

Another approach would be to try calculating the contact point of the 2 area shapes (the button collision shape and the player hitbox) but that may be too much for a simple effect.

manuq:
@Andersonu1290 sorry, I didn't mean to be a single frame, but a single animated particle! It was also only a suggestion, please consider reverting because it was better before. I loved the spark animation.

Captura desde 2025-11-26 15-53-45

scenes/game_elements/characters/player/components/player.gd
manuq, line 251: This was OK when the particles were part of the player. Since the particles are not part of the player anymore, the button can do this directly.

scenes/game_elements/fx/shine_button/shine_particles.gd
manuq, line 7: There is no need for this script. Istead of calling `activate()`, set `emitting = true`.

Issue #1030, opened 2025-08-11
Implement "click" / select sound effect into the game


saspiers:
Take the sound from #1028  and implement the sound into the game so that when players select an option in a game menu, the sound is triggered.

### XP Summary (total 275)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 50 |
| Game Design: Game Feel & UX | Technical Literacy | 150 |

Un-player0:
I plan to work on this.

wjt:
You might want to look at #1028 for some previous discussion about how to implement this. My best idea is the last one described in https://github.com/endlessm/threadbare/pull/1670#pullrequestreview-3571403859.

Un-player0:
Hoja de planificación: https://docs.google.com/document/d/1aH3jSwcvGClgauLhgV4y_Qjz1y_9uyDskttfQehS358/edit?tab=t.0

Un-player0:
Was wondering whether an autoload or a custom node was the best approach, since there are plenty autoloads already on threadbare. Would like an expert's opinion

manuq:
> Was wondering whether an autoload or a custom node was the best approach, since there are plenty autoloads already on threadbare. Would like an expert's opinion

@Un-player0 yes! An autoload was the last option (and recommended) by Will here: https://github.com/endlessm/threadbare/pull/1670#pullrequestreview-3571403859

Issue #1028, opened 2025-08-11
Create a UI sound effect (click/select)


saspiers:
Create, record, and attach an mp3 file of a sound effect the game makes when players select buttons in the settings menu / game. When you attach this sound affect to this issue, please write a short description of how you made and recorded this sound.

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Art: Music & Sound | Problem Solving | 50 |
| Art: Music & Sound | Creativity & Design Thinking | 75 |
| Art: Music & Sound | Technical Literacy | 50 |

DonSeja06:
I Plan to work on this

DonSeja06:
[Planning Sheet](https://docs.google.com/document/d/111tqpkPq7ieA63ASApkKDrfGor4pNzaKfDbJe1CqlpE/edit?usp=sharing)

PR #1670, opened 2025-11-21
Add AudioManager with click sounds

kphero:
**Description:**  
This PR introduces a global `AudioManager` to handle audio assets consistently across the project. It also adds click sound effects from #1028.

### Changes
- Added the `click` assets from #1028.
- Implemented `AudioManager` as a global utility.
- Designated Click1 as the default click, Click2 as the toggle sound, and Click3 as the back sound.
- Added click sounds to the main menu, options, and credit scenes.

This PR is starting to implement #1030, with only functionality in the main menu currently. But should be easy to implement the click sounds wherever needed.

### DIFF
--- a/scenes/globals/audio/audio_manager.gd
+++ b/scenes/globals/audio/audio_manager.gd
@@ -7,11 +7,14 @@ extends Node
 @onready var ui_click_back_sound: AudioStreamPlayer = %ClickBack
 @onready var ui_click_toggle_sound: AudioStreamPlayer = %ClickToggle
 
+
 func play_ui_click():
 	ui_click_sound.play()
 
+
 func play_ui_click_back():
 	ui_click_back_sound.play()
-	
+
+
 func play_ui_click_toggle():
 	ui_click_toggle_sound.play()


scenes/globals/audio/audio_manager.gd
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
```
manuq, line 3: Does this really need to be a `@tool` ?

scenes/globals/audio/audio_manager.gd
```diff
@@ -0,0 +1,20 @@
...
+extends Node
```
manuq, line 4: I wonder if we should give it a `class_name` to this node and add it to the menu scenes, rather than adding yet another autoload to the game. What do you think? I don't see this being used anywhere else than inside `scenes/menus/`. If so then I would propose `menu_audio.gd` or similar for the file (and `MenuAudio` or similar for the class name) instead of `audio_manager.gd`, which seems more global.

wjt (changes requested):
Following from what Manuel has said: the approach implemented here adds a new autoload but also requires writing code for each button.

If we are prepared to write code for every button in the game then we don't need an autoload. And if we're going to ask the developer to do something for every button in the game, IMO it could be a node that we make a child of each button that plays a sound; or a script that we can attach directly to each control.

A middle ground would be to add a ControlSoundEffects node that we place into every scene that has controls that should make noise. We could either connect every button in the scene's `pressed` signal to a method on this node; or have it walk the scene on ready to find controls.

But a more comprehensive option would be to use an autoload that automatically connects to every control that is ever used in the game. Something like this:

```gdscript
func _ready() -> void:
    get_tree().node_added.connect(_on_node_added)

func _on_node_added(node: Node) -> void:
    if node is CheckButton:
        node.toggled.connect(_on_check_button_toggled)
    elif node is Button:
        node.pressed.connect(_on_button_pressed)
    elif node is Slider:
        node.value_changed.connect(_on_slider_value_changed)

func _on_check_button_toggled(_toggled_on: bool) -> void:
    ui_click_sound.play()

# ... etc ...
```

Then every control in the game will make the desired noises - it's impossible to forget to add the relevant code.

wjt:
Closing this due to inactivity, but feel free to reopen in future having addressed the feedback above!

PR #1671, opened 2025-11-22
Implement Fragile Vase Mechanic (InkCombatRound5)

### LINKED ISSUE #1648: Development (Logic): Implement the script with the health system and color validation.
A new logic system must be implemented for Level 5 vases, expanding the current filling mechanic to include vase fragility and destruction behavior.

### **1. Health System**
- Each vase must have a configurable amount of health (HP).

### **2. Interaction Logic**
- **If the projectile color matches the vase color:**  
  The vase should fill (existing behavior).
- **If the projectile color does NOT match:**  
  The vase must take damage (HP decreases).

### **3. Visual Feedback**
- The vase must visually update to reflect its damage state (cracks intensify as HP decreases).

### **4. Destruction**
- When HP reaches **0**, the vase must:
  - Play a breaking animation.
  - Get removed from the scene.

### **5. Defeat Condition (Game Over)**
- A system must monitor all target vases in the room.
- If **all vases are destroyed**, the level must **restart automatically**.

russell-cj:
This PR implements the complete **Fragile Vase** mechanic for the InkCombat minigame (**InkCombatRound5**). It introduces breakable objects with health, visual damage states, and a game-over condition that triggers a smooth level restart when all protected vases are destroyed.

### Logic
- Added **FragileVase** (inherits from `FillingBarrel`), with HP, crack progression, and destruction.
- Added **VaseDestructionTracker** to detect when all vases are destroyed and trigger a game-over restart.
- Updated `projectile.gd` to support both filling and damaging interactions.
- Implemented a **wipe transition** restart instead of an instant hard reset.

###  Visuals
- Added crack sprites (4 stages) and final shatter animation.
- Debris stays in the level with active collision.

### Level Adjustments
- Replaced old barrels with FragileVase instances.
- Configured color-matching labels.
- Updated ThrowingEnemy to throw multiple ink colors.

## Testing Instructions

1. Hit a vase with the **wrong color** → cracks appear.
2. Hit it **4 times** → vase shatters; debris stays on the ground.
3. Destroy **all 3 vases** →  the level restarts with a transition.
4. Hit with the **correct color** → vase fills normally.

### 🎮 Play Ink Drinker Level 5  
[Play Ink Drinker Level 5](https://play.threadbare.game/branches/endlessm/issue_1648/#quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_5)

resolves  #1648

### DIFF
--- a/scenes/game_elements/props/projectile/components/projectile.gd
+++ b/scenes/game_elements/props/projectile/components/projectile.gd
@@ -158,20 +158,16 @@ func _on_body_entered(body: Node2D) -> void:
 	add_small_fx()
 	duration_timer.start()
 
-	# Retrieve the actual object. If we hit a HitBox, we want its owner.
-	# Usually the root node of the scene.
-	var target_object: Node = body.owner if body.owner else body
-
 	# Logic for Fragile Barrel
 	# We must check for the specific subclass first because it inherits from FillingBarrel
-	if target_object is FragileBarrel:
-		target_object.hit_by_droplet(label)
+	if body.owner is FragileBarrel:
+		body.owner.hit_by_droplet(label)
 		queue_free()
 		return
 
 	# Standard Logic for FillingBarrel
-	if target_object is FillingBarrel:
-		var filling_barrel: FillingBarrel = target_object as FillingBarrel
+	if body.owner is FillingBarrel:
+		var filling_barrel: FillingBarrel = body.owner as FillingBarrel
 		if filling_barrel.label == label:
 			filling_barrel.increment()
 			queue_free()


scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
```diff
@@ -0,0 +1,76 @@
...
+@export var pre_fade_delay: float = 1.5
+
+var total_targets: int = 0
+var destroyed_count: int = 0
+var is_game_over: bool = false
+
+
+func _ready() -> void:
+	# Wait a frame to ensure all nodes are ready
+	await get_tree().process_frame
```
manuq: When you say "all nodes" here, which nodes are you referring? All the nodes in the scene? From past experience, this doesn't actually ensure what it says, and is usually not needed.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,74 @@
...
+
+func break_vase() -> void:
+	if crack_overlay_node:
+		crack_overlay_node.visible = false
+
+	# Play destruction animation
+	# Usamos el acceso directo para evitar variables temporales sin tipo
+	if animated_sprite_2d and animated_sprite_2d.sprite_frames.has_animation("shatter"):
+		animated_sprite_2d.play("shatter")
+		await animated_sprite_2d.animation_finished
```
manuq: I wouldn't add so many checks. You can safely assume that the variables contain what you expect while in game. If you need checks for exported variables, add in-editor warnings by defining a `_get_configuration_warnings()` function. Then in setters call `update_configuration_warnings()`.
```suggestion
	animated_sprite_2d.play("shatter")
	await animated_sprite_2d.animation_finished
```

project.godot
```diff
@@ -89,6 +89,34 @@ guard_enemy=""
...
+"process/fix_alpha_border": true,
+"process/hdr_as_srgb": false,
+"process/hdr_clamp_exposure": false,
+"process/normal_map_invert_y": false,
+"process/premult_alpha": false,
+"process/size_limit": 0,
+"roughness/mode": 0,
+"roughness/src_normal": ""
+}
+
```
manuq: Can you explain this change in the `project.godot` file? My guess is that this would affect the way all textures are imported, so it will affect the whole game.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,74 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name FragileVase
+extends FillingBarrel
+
+# Signal emitted when the vase is destroyed
```
manuq: This could become a [documentation comment](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_documentation_comments.html) (starting with two `##`). And the "signal" word is redundant. It can also be a bit more informative, referencing `current_health`:

```suggestion
## Emitted when [member current_health] reaches 0
```

Then this is how it looks in the editor documentation:

Image

You can see from that documentation page (left click FragileVase node -> Open documentation) that the FragileVase class and other methods and variables would welcome some documentation.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,74 @@
...
+signal vase_destroyed(vase_instance: FragileVase)
+
+@export_group("Fragility")
```
manuq: This "Fragility" is the only group, so I wonder if it's really needed?

I would start grouping exported properties when they become many.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,74 @@
...
+@export_range(3, 5) var max_health: int = 4
+## Reference to the AnimatedSprite2D containing the crack frames.
+@export var crack_overlay_node: AnimatedSprite2D
```
manuq: Why not add access as unique name to the node, and then here just reference it?

```
@onready var crack_overlay: AnimatedSprite2D = %CrackOverlay
```

If you are thinking in modding, maybe is better to export a SpriteFrames like FillingBarrel is doing.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,74 @@
...
+	super._ready()
+	current_health = max_health
+
+	if crack_overlay_node:
+		crack_overlay_node.visible = false
+		crack_overlay_node.stop()
+		crack_overlay_node.frame = 0
+	else:
+		if not Engine.is_editor_hint():
+			push_warning("FragileVase: 'Crack Overlay Node' is not assigned in %s" % name)
```
manuq: I think you can safely assume that the node is there. It is in the scene file!

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,74 @@
...
+func update_cracks() -> void:
+	if not crack_overlay_node:
+		return
+
+	var damage_taken: int = max_health - current_health
+
+	crack_overlay_node.visible = true
+	if damage_taken > 0:
+		# Frame 0 = 1 damage, Frame 1 = 2 damage, etc.
+		crack_overlay_node.frame = damage_taken - 1
```
manuq: So the vase won't display any cracks at the beginning if `max_health` is very high, and then display one frame per damage, depending on the length of the frame-by-frame animation. Is this what you really want? We had something similar in FillingBarrel before, and we changed it to display the frames evenly.

So for instance if you have 3 frames, and max_health is 9, you could show the frames evenly:
- at health 9, frame 2
- at health 8, frame 2
- at health 7, frame 2
- at health 6, frame 1
- at health 5, frame 1
- at health 4, frame 1
- at health 3, frame 0
- at health 2, frame 0
- at health 1, frame 0
- at health 0 --> crash! (trigger animation)

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
```diff
@@ -0,0 +1,76 @@
...
+	if is_game_over:
+		return
+
+	destroyed_count += 1
+	# print("Object destroyed. Progress: %d/%d" % [destroyed_count, total_targets])
+	check_loss_condition()
+
+
+func check_loss_condition() -> void:
+	if destroyed_count >= total_targets:
```
manuq: This loss condition doesn't consider the "barrels to win" set in the FillGameLogic. In the round 5, barrels to win is set to 1. But if I set it to 2, I can reach a dead-end state in which the level doesn't restart and I cannot win:

Image

project.godot
russell-cj: Hi @manuq,
I set the default texture import preset to '2D Pixel'. This is intentional to ensure all new sprites (like the vases and cracks) are imported without filtering, matching the game's pixel art aesthetic

manuq: OK! Thanks for explaining the why. I think we need to discuss it first before adding it to the project. And the change could be delicate, so it should be treated separately. Please:
- Open a issue about it, explaining the problem and why you think the project needs this texture import preset.
- Put this change in a separate PR, targetting the new issue.
- Remove this change from thir PR.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
manuq: @russell-cj this question still stands. I would remove this await.

scenes/game_elements/props/filling_barrel/filling_barrel.tscn
```diff
@@ -163,8 +163,10 @@ script = ExtResource("1_md0e3")
 [node name="AnimatedSprite2D" type="AnimatedSprite2D" parent="."]
 unique_name_in_owner = true
 position = Vector2(0, -21)
-sprite_frames = ExtResource("3_o1oxn")
-animation = &"filling"
+sprite_frames = ExtResource("2_vbewj")
+animation = &"shatter"
+frame = 10
+frame_progress = 1.0
```
manuq: The changes in this scene were made by the editor, but aren't needed. The `filling_barrel.tscn` scene can continue showing the "filling" animation instead of the "shatter" animation. Actually, I think this is important for StoryQuests because all the existing ones aren't using the fragile vasels. Same for the first 3 rounds in the lore quest.

I think it is safe to restore this `filling_barrel.tscn` scene from latest main without affecting the PR.

scenes/game_elements/props/filling_barrel/fragile_vase.tscn
```diff
@@ -0,0 +1,20 @@
+[gd_scene load_steps=4 format=3 uid="uid://bxr2wvo3uqeby"]
+
+[ext_resource type="PackedScene" uid="uid://y8ha8abfyap2" path="res://scenes/game_elements/props/filling_barrel/filling_barrel.tscn" id="1_ab25j"]
+[ext_resource type="SpriteFrames" uid="uid://ch0p31t4upx7w" path="res://scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_cracks/barrel_cracks.tres" id="2_g6wko"]
+[ext_resource type="Script" uid="uid://dnon02l02aw6o" path="res://scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd" id="2_wal8y"]
+
+[node name="FragileVase" instance=ExtResource("1_ab25j")]
```
manuq: Thanks for using the same name in both the root node and the tscn file name!

Nit: I think adding a new noun can be confusing, so I would rename it to "fragile barrel", to match the "filling barrel".

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -157,8 +157,20 @@ func add_small_fx() -> void:
...
+
+	# Retrieve the actual object. If we hit a HitBox, we want its owner.
+	var target_object: Node = body.owner if body.owner else body
+
+	# Logic for Fragile Vase
+	# Delegate hit handling if the object supports it
+	if target_object.has_method("hit_by_droplet"):
+		target_object.hit_by_droplet(label)
+		queue_free()
+		return
```
manuq: It is a bit confusing to use a variable that can be both the `body` and the `body_owner`. So I would just check:
```
if body.has_method("hit_by_droplet"):
    # logic for fragile barrel
elif body.owner is FillingBarrel:
    # logic for filling barrel
```

By the way, since FragileVase extends FillingBarrel, are you sure that it doesn't work using owner? Because checking if the object is an instance of a class is more reliable than using `has_method`. Please check if this works:
```
if body.owner is FragileVase:
    var fragile_vase: FragileVase = body.owner as FragileVase
    fragile_vase.hit_by_droplet(label)
    # ....
elif body.owner is FillingBarrel:
    # ...
```

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
```diff
@@ -0,0 +1,72 @@
...
+@onready var crack_overlay_node: AnimatedSprite2D = %CrackOverlay
+
+
+func _ready() -> void:
+	super._ready()
+	current_health = max_health
+
+	crack_overlay_node.visible = false
+	crack_overlay_node.stop()
+	crack_overlay_node.frame = 0
```
manuq: I don't think this is needed. Unless the AnimatedSprite2D is set to autoplay, you can safely assume that it won't be playing.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
```diff
@@ -0,0 +1,108 @@
...
+
+
+# Helper function to find the logic node in the scene
+func _find_fill_game_logic() -> FillGameLogic:
+	var root: Node = get_tree().current_scene
+	for child in root.get_children():
+		if child is FillGameLogic:
+			return child
+
+	return null
```
manuq: Instead of trying to find the node, please add an `@export` for it, and let the developers do the setup in their levels through the Inspector.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
```diff
@@ -0,0 +1,108 @@
...
+				(
+					"Object %s in group %s is missing 'vase_destroyed' signal."
+					% [target.name, target_group_name]
+				)
+			)
+
+
+func _on_target_destroyed(_target_ref: Node) -> void:
+	if is_game_over:
+		return
```
manuq: If you call `player.defeat()`, then instead of adding a custom game over variable here, you could find the player (should the only node in the "player" group) and check if its mode is DEFEATED.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
```diff
@@ -0,0 +1,108 @@
...
+	# Use the project's transition system (SceneSwitcher)
+	var current_scene_path: String = get_tree().current_scene.scene_file_path
+
+	# Use default wipe transitions similar to Teleporter
+	SceneSwitcher.change_to_file_with_transition.call_deferred(
+		current_scene_path,
+		^"",
+		Transition.Effect.LEFT_TO_RIGHT_WIPE,
+		Transition.Effect.RIGHT_TO_LEFT_WIPE
+	)
```
manuq: Our friend @wjt added a nice `player.defeat()` method in 8b07d831 and I think we should use it here! It avoids taking care of the scene switch, does fade in/out transition that is consistent with the rest of the game, and also set the player to the DEFEATED mode, which you can use instead of the custom game_over variable.

russell-cj:
Hi @manuq!

I just checked the existing StoryQuests as you suggested.

I can confirm that there are no regressions: the logic in `projectile.gd` uses a dynamic check (`if target_object.has_method("hit_by_droplet")`). Since the base `FillingBarrel` class doesn't have this method, the old levels run the standard code path unaffected.

project.godot
russell-cj: Understood, Manu. You're right — it's a global change that should be handled carefully to avoid affecting other assets.

I will revert the changes in `project.godot` in this PR to keep it clean. Then, as you suggested, I’ll open a separate discussion to address the '2D Pixel' import configuration for the future.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
russell-cj: You're right. I'll update that now.
Thanks for reminding me :)

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
russell-cj: Got it,  I’ll make that change.

I’ll add the `@export` and leave the setup for each level through the Inspector.

Thanks!

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/vase_destruction_tracker.gd
russell-cj: Got it, I’ll make that adjustment and use the player's **DEFEATED** mode instead of the custom variable.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_vase.gd
russell-cj: Thanks for the tip! I will remove those lines

scenes/game_elements/props/projectile/components/projectile.gd
manuq: Nice, thanks for switching to check the class with `is FragileBarrel`. Now the confusing `target_object` that can be both the `body` and the `body.owner` still stands. So I confirmed that is not needed and appended a fixup commit, so we can merge it.

Issue #1389, opened 2025-10-20
Implement player lives HUD


wjt:
Implement the design from https://github.com/endlessm/threadbare/issues/1387, using the game state implemented in #1388.

### XP Summary (total 250)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 50 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 50 |
| Game Design: Game Feel & UX | Technical Literacy | 150 |

PR #1679, opened 2025-11-25
Implement player lives and respawn system

### LINKED ISSUE #1388: Implement player health tracking & respawn behaviour
Add current player health to game state (in-memory and saved). Decrement it when the player is defeated. Adjust the defeat code so that, when the player runs out of lives, switch to the start of the current challenge (rather than the current scene), resetting the spawn point (in case the current scene is also the start of the current challenge, as in the stealth challenge), and resetting the number of lives to 3. Refill lives to 3 when returning to Fray's End.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |

Gatorrante:
https://github.com/user-attachments/assets/b93ca10f-21a1-415c-bf14-e35813b8d5b9

## Summary
This PR implements a lives system for **Threadbare**, giving players **3 attempts per challenge**.  
When all lives are lost, players restart from the beginning of the current challenge with lives restored.

Resolves #1388

---

### 1. Player Script (`player.gd`)
- Handles respawn logic based on remaining lives:
  - **Lives remaining:** Reloads current scene/checkpoint  
  - **No lives left:** Triggers game over sequence and restarts from challenge start  
- Includes comprehensive debug logging for tracking lives state  

### 2. Game State (`game_state.gd`)
- Added `current_lives: int` variable (default: 3)  
- Added `MAX_LIVES` constant (3)  
- Added `lives_changed(new_lives: int)` signal for UI updates  
- Implemented new methods:
  - `decrement_lives()` → Reduces lives by 1 when defeated  
  - `reset_lives()` → Restores to 3 lives  
  - `add_life()` → For future *extra life* pickups  
  - `set_challenge_start_scene(scene_path)` → Marks challenge start point  
  - `get_challenge_start_scene()` → Retrieves challenge start for respawn  
- Lives automatically reset to 3 when:
  - Starting a quest  
  - Completing a quest  
  - Abandoning a quest  
  - Returning to Fray's End  
- Lives persist across save/load with proper serialization  
- Added safety checks to prevent errors when erasing non-existent config keys  

### 3. Fray's End (`frays_end.gd`)
- Automatically restores lives to 3 when player enters the safe haven  

### 4. Collectible threads (`collectible_item.gd`)

- Call `GameState.set_challenge_start_scene` before advancing to the next scene.

---

### Lives Flow
- Player starts quest → **3 lives (♥♥♥)**  
- Player defeated → Lives decremented, checkpoint respawn  
- Lives reach 0 → Game over transition → Challenge restart with 3 new lives  
- Return to Fray's End → Lives restored to 3  

### Save/Load Behavior
- Current lives saved to `user://game_state.cfg`  
- Lives restored when loading saved game  
- On game over: checkpoint cleared, lives reset to 3, progress within challenge lost  

### Debug Features
- All life-related operations log to console with `[LIVES DEBUG]` or `[PLAYER LIVES DEBUG]` prefix  
- Tracks: defeats, respawns, resets, challenge restarts, and save/load operations  

---

## Future Extensibility
- `add_life()` method ready for *extra life* collectibles  if is necessary 
- `lives_changed` signal ready for UI heart display  for #1389

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -100,6 +100,7 @@ func _on_interacted(player: Player, _from_right: bool) -> void:
 	queue_free()
 
 	if next_scene:
+		GameState.set_challenge_start_scene(next_scene)
 		SceneSwitcher.change_to_file_with_transition(next_scene)
 
 


scenes/game_elements/characters/player/components/player.gd
```diff
@@ -226,19 +226,56 @@ func _set_walk_sound_stream(new_value: AudioStream) -> void:
...
+	# Get the start of the current challenge
+	var challenge_start_scene: String = GameState.get_challenge_start_scene()
+
+	if challenge_start_scene.is_empty():
+		# Fallback: reload current scene if no challenge start is defined
+		print("[PLAYER LIVES DEBUG] No challenge start found, reloading current scene")
+		SceneSwitcher.reload_with_transition(Transition.Effect.FADE, Transition.Effect.FADE)
+	else:
+		# Clear spawn point to start from the beginning
+		GameState.set_current_spawn_point(^"")
```
wjt: I don't think this should be necessary: the (empty) spawn point (`^""`) is the second parameter to `SceneSwitcher.change_to_file_with_transition` below.

But actually you *do* need it in the block above - in the fallback case of reloading the current scene, we should erase any spawn point within the current scene.

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -226,19 +226,56 @@ func _set_walk_sound_stream(new_value: AudioStream) -> void:
...
 func defeat(falling: bool = false) -> void:
+	# Prevent multiple defeat calls
 	if mode == Player.Mode.DEFEATED:
 		return
 
 	mode = Player.Mode.DEFEATED
 
+	# Decrement lives and save the new count
+	GameState.decrement_lives()
+	print("[PLAYER LIVES DEBUG] Player defeated! Lives remaining: ", GameState.current_lives)
```
wjt: If you put all these `LIVES DEBUG` statements into `game_state.gd` then you could add an (off by default) boolean to that file, and control whether they're printed there.

scenes/globals/game_state/game_state.gd
```diff
@@ -70,6 +81,8 @@ func _ready() -> void:
 	if err != OK and err != ERR_FILE_NOT_FOUND:
 		push_error("Failed to load %s: %s" % [GAME_STATE_PATH, err])
 
+	print("[LIVES DEBUG] GameState initialized with ", current_lives, " lives")
```
wjt:
```suggestion
	prints("[LIVES DEBUG] GameState initialized with", current_lives, "lives")
```

`prints`, unlike `print`, puts a space between its parameters. Just an idea.

scenes/globals/game_state/game_state.gd
```diff
@@ -79,38 +92,73 @@ func set_incorporating_threads(new_incorporating_threads: bool) -> void:
...
 	_save()
 
 
 ## Set the scene path and [member current_spawn_point].
 func set_scene(scene_path: String, spawn_point: NodePath = ^"") -> void:
 	if scene_path in TRANSIENT_SCENES:
 		return
+
+	# Check if returning to Fray's End
+	if "fray" in scene_path.to_lower() and "end" in scene_path.to_lower():
```
wjt: This is ugly and error-prone. It would be clearer to match by the full path - `res://scenes/world_map/frays_end.tscn` - but actually I don't think you need to do this at at all because you're resetting lives to full at the start of a quest. If we want to special-case Fray's End, there could be a node in that scene that calls `reset_lives()`.

scenes/globals/game_state/game_state.gd
```diff
@@ -79,38 +92,73 @@ func set_incorporating_threads(new_incorporating_threads: bool) -> void:
...
+	_save()
+
+
+## Get the challenge start scene, or the first scene of the current quest
+## if no challenge start has been set.
+func get_challenge_start_scene() -> String:
+	var challenge_start: String = _state.get_value(QUEST_SECTION, QUEST_CHALLENGE_START_KEY, "")
+
+	if challenge_start.is_empty() and current_quest:
+		challenge_start = current_quest.first_scene
```
wjt, line 144: Interesting approach!

scenes/globals/game_state/game_state.gd
```diff
@@ -119,7 +167,15 @@ func mark_quest_completed() -> void:
 			_state.set_value(GLOBAL_SECTION, COMPLETED_QUESTS_KEY, completed_quests)
 
 		current_quest = null
-		_state.erase_section_key(QUEST_SECTION, QUEST_PATH_KEY)
+		if _state.has_section_key(QUEST_SECTION, QUEST_PATH_KEY):
+			_state.erase_section_key(QUEST_SECTION, QUEST_PATH_KEY)
+		if _state.has_section_key(QUEST_SECTION, QUEST_CHALLENGE_START_KEY):
+			_state.erase_section_key(QUEST_SECTION, QUEST_CHALLENGE_START_KEY)
```
wjt: This code is duplicated in `abandon_quest` so I suggest you add a `_clear_quest_state` helper function.

You should not need to check `_state.has_section_key(QUEST_SECTION, QUEST_PATH_KEY)` - it should always be set if `current_quest` is defined. (Unless you have found a case where this fails?)

I think it would be simpler to make `QUEST_CHALLENGE_START_KEY` always be set when the player is on a quest - you could do this by initialising it to the first scene of a quest when starting a quest.

It would be neatest of all if we could just `erase_section(QUEST_SECTION)` to remove everything about the current quest. Would this work? Are there any keys in there that would need to be moved to a different section?

scenes/globals/game_state/game_state.gd
```diff
@@ -176,10 +239,43 @@ func _update_inventory_state() -> void:
...
+	_save()
+	lives_changed.emit(current_lives)
+	print("[LIVES DEBUG] Lives reset to: ", current_lives)
+
+
+## Add one life to the player, up to the maximum.
+## This is for future "extra life" pickups.
+func add_life() -> void:
+	if current_lives < MAX_LIVES:
+		current_lives = min(MAX_LIVES, current_lives + 1)
```
wjt: if `current_lives < MAX_LIVES` then `current_lives + 1 <= MAX_LIVES` so you don't need `min`: you can unconditionally add 1 inside the `if`.

scenes/globals/game_state/game_state.gd
wjt: In fact you actually did that in `frays_end.gd`.

wjt (changes requested):
Nice simple implementation, I like it!

scenes/game_elements/characters/player/components/player.gd
Gatorrante: Thanks! I moved the `set_current_spawn_point(^"")` call to only the fallback case where we reload the current scene. When we switch to a different scene, the empty spawn point is already passed as a parameter to `change_to_file_with_transition`

scenes/game_elements/characters/player/components/player.gd
Gatorrante: Done! Added `const DEBUG_LIVES := false` in `game_state.gd` and moved all the lives-related debug statements there. Now we can enable/disable all lives logging by changing just that one constant.

scenes/globals/game_state/game_state.gd
Gatorrante: Changed to `prints()` throughout - didn't realize it auto-spaces parameters.

scenes/globals/game_state/game_state.gd
Gatorrante: You're right that I already handle this in `frays_end.gd`! Removed the redundant check from `set_scene()`.

scenes/globals/game_state/game_state.gd
Gatorrante: Done! Added `_clear_quest_state()` helper that uses `erase_section()` to clean up the entire quest section. Much simpler than checking and erasing individual keys. The duplication is gone and `QUEST_CHALLENGE_START_KEY` is now always set when starting a quest.

scenes/globals/game_state/game_state.gd
Gatorrante: Good catch - the `min()` was redundant since we're already inside the `if current_lives < MAX_LIVES` check. Changed to just `current_lives += 1`.

scenes/game_elements/characters/player/components/player.gd
wjt: You didn't actually guard all the print statements with `if DEBUG_LIVES`. I pushed a patch to fix this.

wjt (approved):
Nice. I pushed one extra commit that sets the start of each challenge - without this, for example, in the Musician quest you can get to the stealth challenge with 2 threads, get caught 3 times, and you find yourself back at the music puzzle but still with the 2 threads.

Issue #1373, opened 2025-10-17
Input hints for aiming grappling thread with joypad are misleading


manuq:
After the refactor https://github.com/endlessm/threadbare/pull/1356 the player can now aim with the joypad right stick. But the input hints in scene `grappling_hook_start.tscn` are misleading: is showing a d-pad, not the right stick graphic.

At first glance, the existing helper scripts `interact_input.gd` and `gamepad_input.gd` don't quite fit for this case, unless we add 4 TextureRect node using the same graphic. Because the right stick is for all four aim actions: `aim_left`, `aim_right`, `aim_top` and `aim_bottom`.

### XP Summary (total 225)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 75 |
| Game Design: Game Feel & UX | Technical Literacy | 75 |

PR #1681, opened 2025-11-25
Input hints: Add right stick aim input hints for all devices

AlealarcViva:
This PR introduces a new system for displaying right-stick aiming hints across multiple input devices. It adds the RightStickHint and RightStickHintManager scripts, along with their supporting resources and textures for keyboard, Xbox, PlayStation, Switch, and Steam Deck.

**Overview of Changes**
Added RightStickHint and RightStickHintManager scripts to manage device-specific right-stick aim prompts.
Added new texture assets and resources for all supported device families.
Implemented unified logic to select the correct prompt based on the active input device.
Updated grappling_hook_start.tscn to use the new aim hint system.
Integrated device-specific input prompt assets to replace hard-coded placeholders.

resolves [#1373](https://github.com/endlessm/threadbare/issues/1373)

### DIFF
--- a/scenes/game_elements/props/hint/input_key/RightStickHintManager.gd
+++ b/scenes/game_elements/props/hint/input_key/RightStickHintManager.gd
@@ -6,21 +6,20 @@ extends Node
 # Device keyword → resource path
 const DEVICE_MAP := {
 	"keyboard": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_keyboard.tres",
-	"xbox": "res://scenes/game_elements/props/hint/resources/stickmovements/stick_xbox.tres",
+	"xbox": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_xbox.tres",
 	"playstation":
-	"res://scenes/game_elements/props/hint/resources/stickmovements/stick_playstation.tres",
-	"ps": "res://scenes/game_elements/props/hint/resources/stickmovements/stick_playstation.tres",
-	"play": "res://scenes/game_elements/props/hint/resources/stickmovements/stick_playstation.tres",
-	"switch": "res://scenes/game_elements/props/hint/resources/stickmovements/stick_switch.tres",
-	"nintendo": "res://scenes/game_elements/props/hint/resources/stickmovements/stick_switch.tres",
-	"steam": "res://scenes/game_elements/props/hint/resources/stickmovements/stick_steamdeck.tres",
-	"steamdeck":
-	"res://scenes/game_elements/props/hint/resources/stickmovements/stick_steamdeck.tres"
+	"res://scenes/game_elements/props/hint/resources/stickmovements/Aim_playsation.tres",
+	"ps": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_playsation.tres",
+	"play": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_playsation.tres",
+	"switch": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_switch.tres",
+	"nintendo": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_switch.tres",
+	"steam": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_steamdeck.tres",
+	"steamdeck": "res://scenes/game_elements/props/hint/resources/stickmovements/Aim_steamdeck.tres"
 }
 
 # Fallback genérico
 const GENERIC_RESOURCE := (
-	"res://scenes/game_elements/props/hint/resources/" + "stickmovements/stick_xbox.tres"
+	"res://scenes/game_elements/props/hint/resources/" + "stickmovements/Aim_xbox.tres"
 )
 
 


scenes/game_elements/props/hint/input_key/RightStickHintManager.gd
```diff
@@ -0,0 +1,53 @@
...
+
+
+# Determine which JoypadButtonTextures resource to use for a given device string.
+func _resource_for_device(device: String) -> JoypadButtonTexturesAim:
+	if device == null or device.is_empty():
+		return ResourceLoader.load(GENERIC_RESOURCE) as JoypadButtonTexturesAim
+
+	var d := device.to_lower()
+	for key in DEVICE_MAP.keys():
+		if key in d:
```
manuq: The "key" here is missing the type (see the warning in the other comment):

```suggestion
	for key: String in DEVICE_MAP.keys():
		if key in d:
```

But even better would be to use [Dictionary.get()](https://docs.godotengine.org/en/stable/classes/class_dictionary.html#class-dictionary-method-get) and avoid the for loop. `get()` returns `null` by default, but you can also provide a default as second parameter. So you can just do something like (untested!):

```
var resource_path: String = DEVICE_MAP.get(device.to_lower(), GENERIC_RESOURCE) 
return ResourceLoader.load(resource_path) as JoypadButtonTexturesAim
```

scenes/game_elements/props/hint/input_key/RightStickHint.gd
```diff
@@ -0,0 +1,72 @@
...
+var current_device: String = ""
+var is_keyboard_mode: bool = true
+
+
+func _ready() -> void:
+	if Engine.has_singleton("InputHelper"):
+		InputHelper.device_changed.connect(_on_input_device_changed)
+		_on_input_device_changed(InputHelper.device, InputHelper.device_index)
+	else:
+		_on_input_device_changed("keyboard", -1)
```
wjt, line 16: I think we can assume that the `InputHelper` singleton is present: we depend on this addon. Is there a particular need to have these checks that I've missed?


```suggestion
	InputHelper.device_changed.connect(_on_input_device_changed)
	_on_input_device_changed(InputHelper.device, InputHelper.device_index)
```

scenes/game_elements/props/hint/input_key/RightStickHint.gd
```diff
@@ -0,0 +1,72 @@
...
+
+
+func _on_input_device_changed(device: String, _device_index: int) -> void:
+	current_device = device
+
+	if Engine.has_singleton("InputHelper"):
+		is_keyboard_mode = (device == InputHelper.DEVICE_KEYBOARD)
+	else:
+		is_keyboard_mode = device.to_lower() == "keyboard"
```
wjt, line 25:
```suggestion
	is_keyboard_mode = (device == InputHelper.DEVICE_KEYBOARD)
```

project.godot
```diff
@@ -33,6 +33,7 @@ CameraShake="*res://scenes/globals/camera_shaker/camera_shake.tscn"
 InputHelper="*res://addons/input_helper/input_helper.gd"
 AspectRatioDebugger="*res://scenes/globals/aspect_ratio_debugger/aspect_ratio_debugger.tscn"
 InputGlobal="*res://scenes/game_elements/props/hint/input_key/InputHintManager.gd"
+RightStickHintManager="*res://scenes/game_elements/props/hint/input_key/RightStickHintManager.gd"
```
wjt, line 36: I don't quite understand why we need a global for this - can't it live in `RightStickHint.gd`?

scenes/game_elements/props/hint/input_key/RightStickHintManager.gd
```diff
@@ -0,0 +1,52 @@
...
+func _resource_for_device(device: String) -> JoypadButtonTexturesAim:
+	if device == null or device.is_empty():
+		return ResourceLoader.load(GENERIC_RESOURCE) as JoypadButtonTexturesAim
+
+	var d := device.to_lower()
+	for key: String in DEVICE_MAP.keys():
+		if key in d:
+			return ResourceLoader.load(DEVICE_MAP[key]) as JoypadButtonTexturesAim
+
+	return ResourceLoader.load(GENERIC_RESOURCE) as JoypadButtonTexturesAim
```
wjt, line 36: I don't understand the need to do these string comparisons. The Input Helper addon defines a small set of device types:

```gdscript
const DEVICE_KEYBOARD = "keyboard"
const DEVICE_XBOX_CONTROLLER = "xbox"
const DEVICE_SWITCH_CONTROLLER = "switch"
const DEVICE_PLAYSTATION_CONTROLLER = "playstation"
const DEVICE_STEAMDECK_CONTROLLER = "steamdeck"
const DEVICE_GENERIC = "generic"
```

These are the possible values for the first parameter of the `InputManager.device_changed` signal, which is where the `device` parameter here comes from. So `DEVICE_MAP` only needs to have these keys (no need for "playstation", "ps" and "play"), and this can be reduced to:


```suggestion
func _resource_for_device(device: String) -> JoypadButtonTexturesAim:
	if device == null or device.is_empty():
		return ResourceLoader.load(DEVICE_MAP.get(device, DEVICE_MAP["generic"])) as JoypadButtonTexturesAim
```

In fact the values could be `preload()`s of the resource files and then we don't need the `load()` call either.

scenes/game_elements/props/hint/resources/stickmovements/JoypadButtonTexturesAIM.gd
```diff
@@ -0,0 +1,29 @@
...
+		"aim_up":
+			return aim_up
+		"aim_down":
+			return aim_down
+		"aim_left":
+			return aim_left
+		"aim_right":
+			return aim_right
+		_:
+			return aim_unpressed
```
wjt, line 29:
```suggestion
func get_texture_for_action(action_name: String) -> Texture2D:

	var texture := (get(action_name) as Texture2D)
	if not texture:
		texture = aim_unpressed
	return texture
```

https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-get

wjt (changes requested):
This is essentially a copy of `movement_input_hints.tscn` and all its supporting machinery - I would have expected it to be another scene that reuses the same scripts, just with a different set of resources applied to it.

I have some specific suggestions below but I would like to try implementing a different approach next week myself – I think that would be the clearest way to describe what I mean!

PR #1686, opened 2025-11-27
Update game_state.gd

Slayer2505:
"Fix inventory persistence: do not save collected items until scene changes"

### DIFF
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -2,8 +2,16 @@
 # SPDX-License-Identifier: MPL-2.0
 extends Node
 
+## Emitted when a new item is collected, even if it wasn't added to the
+## inventory due to it being already there.
 signal item_collected(item: InventoryItem)
+
+## Emitted when an item is consumed, causing it to be removed from the
+## [member inventory].
 signal item_consumed(item: InventoryItem)
+
+## Emitted whenever the items in the inventory change, either by collecting
+## or consuming an item.
 signal collected_items_changed(updated_items: Array[InventoryItem])
 
 const GAME_STATE_PATH := "user://game_state.cfg"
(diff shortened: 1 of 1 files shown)


scenes/globals/game_state/game_state.gd
```diff
@@ -55,6 +55,10 @@ var current_quest: Quest
 var persist_progress: bool
 var _state := ConfigFile.new()
 
+## Internal flag used to delay inventory persistence until the scene changes.
+## This ensures items are not saved the moment they are collected.
```
wjt, line 59: Use `#` for internal fields. Adding a `##` comment causes it to show up in the API documentation for the class; but we don't want this one to be there since it shouldn't be read/modified outside this file.

```suggestion
# Internal flag used to delay inventory persistence until the scene changes.
# This ensures items are not saved the moment they are collected.
```

scenes/globals/game_state/game_state.gd
```diff
@@ -92,7 +98,14 @@ func start_quest(quest: Quest) -> void:
 func set_scene(scene_path: String, spawn_point: NodePath = ^"") -> void:
 	if scene_path in TRANSIENT_SCENES:
 		return
+
 	_do_set_scene(scene_path, spawn_point)
+
+	# If inventory changes were pending, write them now.
+	if _pending_inventory_save:
```
wjt, line 105: I don't think you need this flag. I think you can just unconditionally call `_write_inventory_to_state()` here?

scenes/globals/game_state/game_state.gd
```diff
@@ -81,7 +87,7 @@ func set_incorporating_threads(new_incorporating_threads: bool) -> void:
 ## Set [member current_quest] and clear the [member inventory].
 func start_quest(quest: Quest) -> void:
 	_do_clear_inventory()
-	_update_inventory_state()
+	_mark_inventory_dirty()  # Deferred saving
```
wjt, line 90: This should not be deferred when starting a quest - it should be cleared immediately.

PR #1687, opened 2025-11-28
cat: Don't block player movement while miaowing

wjt:
Previously, when you interacted with the cat, you had to wait for the miaow to finish before you could move again.

Instead, mark the interaction as completed immediately, but disable further interactions with the cat until it has finished miaowing.

Also remove the npc.gd script from the cat's root node: none of that functionality is actually used.


manuq:
Not merging just in case @Alexander3242G wants to test it in the web build.

PR #1695, opened 2025-12-01
Update quest images

Gatorrante:
I have re-drawn the sketches for **The Musicians Quest** and **The Void** storybook.  
Previously, the images exceeded 720p quality, which caused excessive compression and loss of detail.  

- **The Void** now features a clearer, more comprehensible image that better matches the level.  
- The **layout’s top margin** has been slightly adjusted, as it previously appeared too close to the book’s edges.
image
image

### DIFF
--- a/scenes/menus/storybook/components/storybook_page.tscn
+++ b/scenes/menus/storybook/components/storybook_page.tscn
@@ -248,7 +248,7 @@ grow_vertical = 2
 size_flags_horizontal = 3
 theme = SubResource("Theme_rf3dg")
 theme_override_constants/margin_left = 0
-theme_override_constants/margin_top = 0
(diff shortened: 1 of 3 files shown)


scenes/menus/storybook/components/storybook_page.tscn
```diff
@@ -324,7 +325,7 @@ autowrap_mode = 3
 [node name="PlayButton" type="Button" parent="HBoxContainer/VBoxContainer2"]
 unique_name_in_owner = true
 layout_mode = 2
-size_flags_vertical = 8
+size_flags_vertical = 4
```
manuq, line 328: If I read this correctly, it is changing from SIZE_SHRINK_END to SIZE_SHRINK_CENTER, right?

PR #1697, opened 2025-12-03
Add SFX for FragileBarrel

### LINKED ISSUE #1647: Audio:Source/Create and import SFX for impact (cracking) and total destruction.
Source or create sound effects (SFX) for the breakable vase mechanics.

### Requirements
1.  **Crack Impact SFX:** A sound that plays when the vase is hit by a wrong color. It should sound like glass or ceramic cracking.
2.  **Shatter SFX:** A louder, distinct sound for when the vase is completely destroyed.

russell-cj:
This PR implements audio for the FragileBarrel mechanic.

**Changes:**
* Added `cracks` and `breaking` sound effects assets.
* Used `AudioStreamRandomizer` with a 1.1 pitch variation to prevent repetition.
* Updated `FragileBarrel` script and scene to trigger sounds on damage and destruction.

### 🎮 Play Ink Drinker Level 5  
[Play Ink Drinker Level 5](https://play.threadbare.game/branches/endlessm/issue_1647/#quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_5)

resolves #1647

### DIFF
--- a/scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/sounds/breaking.ogg.license
+++ b/scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/sounds/breaking.ogg.license
@@ -1,0 +1,4 @@
+SPDX-License-Identifier: CC0-1.0
+SPDX-FileCopyrightText: 2021 Fission9
+
+Tile Breaking by Fission9 -- https://freesound.org/s/583310/
--- a/scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/sounds/cracks.ogg.license
+++ b/scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/sounds/cracks.ogg.license
@@ -1,0 +1,4 @@
+SPDX-FileCopyrightText: 2020 max2587
+SPDX-License-Identifier: CC-BY-4.0
+
+glass 2.wav by max2587 -- https://freesound.org/s/548232/


scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_barrel.gd
```diff
@@ -41,6 +42,8 @@ func take_damage() -> void:
 	if current_health <= 0:
 		break_barrel()
 	else:
+		if crack_sound:
+			crack_sound.play()
```
manuq: You can remove the `if` conditional here. You can always call `play()` in a AudioStreamPlayer2D. If no stream is set, nothing will be played.


```suggestion
		crack_sound.play()
```

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_barrel.gd
```diff
@@ -61,6 +64,9 @@ func update_cracks() -> void:
 func break_barrel() -> void:
 	crack_overlay_node.visible = false
 
+	if shatter_sound:
+		shatter_sound.play()
```
manuq: Same here. You can remove the `if`.


```suggestion
	shatter_sound.play()
```

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/sounds/sfx_barrel_breaking.tres
```diff
@@ -0,0 +1,8 @@
+[gd_resource type="AudioStreamRandomizer" load_steps=2 format=3 uid="uid://c3iuv5ax8i78v"]
+
+[ext_resource type="AudioStream" uid="uid://74b6dlm4xhr" path="res://scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/sounds/breaking.wav" id="1_kfwyl"]
+
+[resource]
+random_pitch = 1.1
```
manuq, line 6: Good! Randomizing the pitch a bit is a nice touch.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/fragile_barrel.gd
wjt: If the scene does not have a node with unique name `%CrackSound`, or it has one with the wrong type, then `crack_sound` would indeed be `null` and the `if` guard would prevent a crash. But in that case the scene has been misconfigured, the engine will log an error while it is being initialised, and I think it is correct to crash here.

wjt (approved):
I agree that the crack sound is a little low in the mix. I'll spend a few minutes with Audacity…

PR #1701, opened 2025-12-05
credits: Fix reference to keyboard_arrows_none.png

wjt:
The individual input hint assets were removed in commit
745893a32588d42b34294fb9b51e4f7a494c8891, which was developed in parallel with
the scrolling credits scene.

    WARNING: res://scenes/menus/title/components/credits.tscn:14 - ext_resource,
             invalid UID: uid://cexenripopr1m - using text path instead:
             res://assets/third_party/inputs/kenney_input-prompts_1.4/Keyboard & Mouse/Default/keyboard_arrows_none.png
         at: load (scene/resources/resource_format_text.cpp:444)
    ERROR: Failed loading resource:
           res://assets/third_party/inputs/kenney_input-prompts_1.4/Keyboard & Mouse/Default/keyboard_arrows_none.png.
       at: _load (core/io/resource_loader.cpp:343)

Replace the dangling reference to that individual asset with the replacement
AtlasTexture resource.


manuq (approved):
Good catch, Will!

PR #1706, opened 2025-12-09
Play musician's "playing" animation when puzzle is solved

wjt:
@Omar071201 drew a really fun animation where the musician is playing his instrument, but it was so far not used in the game.

Teach `sequence_puzzle_assistant.gd` to start playing an optional `solved` animation when the puzzle is solved. Define that animation in the musician's SpriteFrames.


manuq (approved):
Great idea! Lovely.

PR #1709, opened 2025-12-09
Sueños Nocturnos: Convert large sound files to Ogg Vorbis

### LINKED ISSUE #1708: Sueños Nocturnos: Reduce size of audio assets
This quest is 108MB in the Git repository; of this, 77 MB is music and 21 MB is sound effects. We should be able to reduce this substantially.

wjt:
Before this PR, the `shjourney` folder was 109 MB in the source tree. With this PR, which converts many WAV and large MP3 files to Ogg, it is 23 MB.

Resolves #1708


wjt:
The Actions run for this job has an 88 MB .pck, compared to 110 MB on main: 22 MB smaller. You might ask: why isn't it 86 MB smaller? It's because Godot compresses WAV files during export (though not as well as Ogg Vorbis does).

scenes/quests/story_quests/shjourney/8_shjourney_outro/Outro.tscn
```diff
@@ -1,7 +1,6 @@
-[gd_scene load_steps=43 format=4 uid="uid://b4dotq384vljq"]
+[gd_scene load_steps=42 format=4 uid="uid://b4dotq384vljq"]
 
 [ext_resource type="TileSet" uid="uid://birttlb0sk74i" path="res://scenes/quests/story_quests/shjourney/8_shjourney_outro/tilesetOutro.tres" id="1_0jcvd"]
-[ext_resource type="Script" uid="uid://bdvg77f0s1ymu" path="res://scenes/quests/story_quests/shjourney/8_shjourney_outro/outro.gd" id="1_bov4k"]
```
manuq, line 4: Probably an unused script that went away when saving the scene?

manuq: Ah I see now the last commit "[shjourney: Remove empty script](https://github.com/endlessm/threadbare/pull/1709/commits/0c99361aa1a63efcb4646bf2eaaa12fb2219e848)"

PR #1713, opened 2025-12-10
Remove obsolete properties from PlayStation button texture resource

wjt:
Remove obsolete properties from PlayStation button texture resource

These `move_up` etc. properties were removed from `JoypadButtonTextures`
in commit cef922098728a8ee6895582f405480434dd870a1, in favour of a
sub-resource for `move` and another for `aim` which each have `up` etc.
properties.

The fact that I committed this resource with both sets of properties
reveals my development process: I added the new sub-resources, made the
scripts `@tool` scripts that set the new `move` sub-resource from the
old `move_up` etc. properties, and resaved all these resource files.
Then I removed the old fields, removed `@tool`, and opened and saved
each resource again. Evidently I missed one!


manuq (approved):
Yes! I have this same diff locally.

Issue #1714, opened 2025-12-10
Feature: Add skin tones shader


manuq:
The visual style of Threadbare encourages the cel-shading render technique. Generally there are 3 shades of the same color used for skin, or any other material.

Usually we have color variations of props or characters as separate PNG files. This is OK for trees or for known characters like the StoryVores. But for human skin the amount of possibilities would be too many, if we want to embrace racial and cultural diversity. That is why a shader approach would be better for this particular case.

Provide a shader to let developers remap 3 key colors of a sprite to any color. Provide high-level logic to use known skin tones as the remapping colors.

PR #1715, opened 2025-12-10
Add utility to recolor cel-shading sprites

manuq:
Add a shader, a material with that shader applied, and a custom node with a bit of logic to set the shader parameters in a smart way.

The shader replaces 3 key colors with other 3 instance uniforms. Being instance uniforms is important so that each sprite gets their own colors.

The behavior node controls a CanvasItem with the shader applied. It can derive the high and low shades from the medium shade, and knows a list of skin colors. But is also flexible enough to use any color.

The reference for the skin colors is an image from a blog post, linked in the documentation comment in code.

Also, add a test scene with NPCs of different skin tones.

Usage example:

[Grabación de pantalla desde 2025-12-10 14-26-20.webm](https://github.com/user-attachments/assets/af0149bf-442a-4e14-82de-0b2c7bfbf769)

Resolve https://github.com/endlessm/threadbare/issues/1714


scenes/game_elements/components/cel_shading_recolor.gdshader
```diff
@@ -0,0 +1,69 @@
...
+	vec4 color = texture(TEXTURE, UV);
+	if (distance(color.rgb, shade_medium_key) < tolerance) {
+		COLOR = shade_medium_new;
+	}
+	else if (distance(color.rgb, shade_high_key) < tolerance) {
+		COLOR = shade_high_new;
+	}
+	else if (distance(color.rgb, shade_low_key) < tolerance) {
+		COLOR = shade_low_new;
+	}
```
manuq: From this [more generic palette swap addon](https://github.com/KoBeWi/Godot-Palette-Swap-Shader), it looks like I should be able to do this in a more accurate and performant way:
- Converting the color to int, so the comparison is perfectly accurate.
- Avoiding any texture lookup. Which would make it incompatible with modulate, but worth doing for performance reasons.

Their shader [is here](https://github.com/KoBeWi/Godot-Palette-Swap-Shader/blob/master/addons/PaletteSwapShader/SimplePaletteSwap.gdshader). Is ironic how much boilerplate I'm adding around, when the magic is done by a few lines of OpenGL.

manuq:
I have appended a fixup commit with an improvement, based on the addon mentioned in https://github.com/endlessm/threadbare/pull/1715#discussion_r2607678653.

Issue #1683, opened 2025-11-27
Compress large PNGs that are not pixel art


wjt:
At the time of writing we have 21 MB of PNGs in the game source tree.

Here are some of the largest:

```
$ find -name '*.png' -print0 | xargs -0 du -shc | sort -h | tail -n 20
84K	./scenes/globals/scene_switcher/transitions/Radial.png
84K	./scenes/quests/story_quests/shjourney/4_Laberinto/personajes/mapa3/alfombra1.png
112K	./scenes/game_elements/characters/player/components/blow.png
152K	./scenes/quests/story_quests/shjourney/4_Laberinto/personajes/presentacion/baldozas1.png
188K	./docs/screenshot.png
196K	./scenes/quests/story_quests/shjourney/4_Laberinto/personajes/escuela/pisomayolica2.png
212K	./scenes/game_elements/props/tree/components/wind_noise.png
212K	./scenes/quests/story_quests/renya_beyond_sorrow/1_moscas/tree/components/wind_noise.png
240K	./assets/first_party/intro/sketch_04.png
316K	./assets/first_party/intro/sketch_05.png
320K	./assets/first_party/intro/sketch_01.png
380K	./scenes/ui_elements/story_quest_progress/Textures/book_turn_right.png
392K	./scenes/ui_elements/story_quest_progress/Textures/book_turn_left.png
488K	./assets/first_party/intro/sketch_02.png
604K	./assets/first_party/intro/sketch_03.png
1.3M	./scenes/quests/lore_quests/quest_001/Musician.png
1.4M	./scenes/quests/story_quests/shjourney/7_shjourney_combate/assets/tilesetCombate final.png
1.7M	./scenes/quests/lore_quests/quest_002/Void.png
4.7M	./assets/first_party/intro/background.png
22M	total
```

By default, Godot does not compress PNGs: it stores them losslessly. https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/importing_images.html#compress-mode

Most of the top 10 are in `assets/first_party/intro`. They're not pixel art: they're sketches, and a fabric weave background. We should apply lossy compression to these.

We could do this either by converting the files lossily to WebP, or adjusting their import settings, or both. I am not quite sure based on the documentation linked above what happens if the source file is a lossy WebP - does it get stored as-is? 

### XP Summary (total 225)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Optimization & Performance | Problem Solving | 75 |
[...]

PR #1716, opened 2025-12-10
splash, intro, title, pause: Use sand tile as background texture

wjt:
Previously we had a 1920×1080 close-up photograph of fabric as the
background image for these scenes. In the splash screen it is tinted
with the Endless Access blurple colour; in the pause screen it is a
semi-transparent overlay.

This image was a large PNG; when exported (with lossless WebP
compression) it was around 3 MB of the 98 MB .pck file!

We have a set of tileable fabric textures already in the game: the tiles
for the world!

Use the sand tile (without borders) as the background for these scenes.
Apply the same modulation to the splash & pause instances as before.

I initially tried using a TextureRect with an AtlasTexture as its
texture and the stretch_mode "tile". However as documented
<https://docs.godotengine.org/en/latest/classes/class_atlastexture.html>
this doesn't work:

> AtlasTexture cannot be used in an AnimatedTexture, and will not tile
> properly in nodes such as TextureRect or Sprite2D.

While looking at bug reports about this limitation, I stumbled upon a
workaround: use a NinePatchRect, but set the patch margins (i.e. the 8
tiles around the edges & corners) to 0px, leaving only the 64px central
patch which tiles correctly.

Helps https://github.com/endlessm/threadbare/issues/1683


manuq (approved):
I think this is great and the NinePatchRect with the sand texture does the job. If we want something more similar to the previous background, then we could find a tileable texture to replace the sand. But again the sand does the job.

Issue #1359, opened 2025-10-15
Move template .pngs out of NO_EDIT template


wjt:
Currently the NO_EDIT template has 17 PNGs:

NO_EDIT/0_NO_EDIT_intro/NO_EDIT_intro_components/NO_EDIT_intro_image.png
NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth_components/NO_EDIT_guard_enemy_alerted.png
NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth_components/NO_EDIT_guard_enemy_idle.png
NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth_components/NO_EDIT_guard_enemy_walk.png
NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_projectile.png
NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_target.png
NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_throwing_enemy_attack.png
NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_throwing_enemy_defeated.png
NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_throwing_enemy_idle.png
NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_throwing_enemy_walk.png
NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_object.png
NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sign.png
NO_EDIT/NO_EDIT_player_components/NO_EDIT_player_attack_01.png
NO_EDIT/NO_EDIT_player_components/NO_EDIT_player_attack_02.png
NO_EDIT/NO_EDIT_player_components/NO_EDIT_player_defeated.png
NO_EDIT/NO_EDIT_player_components/NO_EDIT_player_idle.png
NO_EDIT/NO_EDIT_player_components/NO_EDIT_player_walk.png

as well as 4 corresponding Aseprite files:

NO_EDIT/3_NO_EDIT_sequence_puzzle/sign.aseprite
NO_EDIT/NO_EDIT_player_components/NO_EDIT_player.aseprite
[...]

PR #1717, opened 2025-12-10
Renya: Remove duplicated scenes & assets

wjt:
I found these while trying to reduce the size of the exported game. None of these files are big but it does add up. In particular the duplicated wind noise texture does not compress very well.

Helps #568


wjt:
> @wjt nice, how did you found them?

`jdupes` (a faster reimplementation of `fdupes`) told me that the `wind_noise.png` texture was duplicated. I worked backwards from ther.

> Ideally learners should not be adding duplicate files in PRs (should commit them as they customize them) but that's not what we suggest with the existing template (which duplicates other files, not these).

Yes, I was thinking about this too... I think the duplication is more important for resources  which it's easy to edit by accident (e.g. SpriteFrames) so I was previously wondering if we should move the template PNGs out of the NO_EDIT directory https://github.com/endlessm/threadbare/issues/1359

Issue #1710, opened 2025-12-10
Renya Beyond Sorrow includes music under unsuitable license


wjt:
The files in `scenes/quests/story_quests/renya_beyond_sorrow/Complementos/music` have no copyright information or license attached (neither in the repo nor in the PR which added them, https://github.com/endlessm/threadbare/pull/1689) but by ear & name they come from https://alkakrab.itch.io/pianoinstrumental.

The zip file available there includes this license:

[AlkaKrab Music License Info.pdf](https://github.com/user-attachments/files/24074260/AlkaKrab.Music.License.Info.pdf)

In particular the license includes this term:

> **Restrictions:**
> No reselling, sublicensing, or redistribution of the track as-is. If you want to use
these tracks in open-source games, contact me directly for a permission.

There is no record of the copyright owner having been contacted for permission to use these tracks in this open-source game, so at present our use of them violates this license and is hence a copyright infringement.

We prefer for assets in this project to be under terms that allow modification and redistribution: <https://github.com/endlessm/threadbare/blob/main/docs/LICENSING.md>. We can make exceptions but we still need to have the right to distribute the files ourselves, which currently we do not have. We must do the following:

- [x] Contact the author, requesting permission to include these tracks in the game
- [x] If they agree: add licensing information to the repo with the standard license and details of any special permission they grant us.

Aleinadsf:
> Le envié un correo electrónico al autor. Esperaré un par de días; si no recibo respuesta, asumiré que no tenemos un permiso especial y eliminaré las pistas.

Buenas tardes, el link donde sacamos la música de ambiente es esta: https://alkakrab.itch.io/piano-instrumental

wjt:
Yes, but if you look inside the ZIP file you downloaded from that link, it has this license file:

> [AlkaKrab Music License Info.pdf](https://github.com/user-attachments/files/24074260/AlkaKrab.Music.License.Info.pdf)
> 
> In particular the license includes this term:
> 
> > **Restrictions:**
> > No reselling, sublicensing, or redistribution of the track as-is. If you want to use
> > these tracks in open-source games, contact me directly for a permission.

This is the problem discussed in https://github.com/endlessm/threadbare/blob/main/docs/LICENSING.md#asset-licensing. Allowing commercial use is not sufficient:

[...]

PR #1719, opened 2025-12-11
renya: Add license & special permission for music

wjt:
The standard license for this music does not permit use in an
open-source game. Happily the composer has given us special permission.

Add the standard license for the music, and the special permission, to
the repository. Add a REUSE.toml file to mark all these tracks as
covered by that license.

Resolves https://github.com/endlessm/threadbare/issues/1710

### DIFF
--- a/scenes/quests/story_quests/renya_beyond_sorrow/Complementos/music/REUSE.toml
+++ b/scenes/quests/story_quests/renya_beyond_sorrow/Complementos/music/REUSE.toml
@@ -1,0 +1,7 @@
+version = 1
+
+[[annotations]]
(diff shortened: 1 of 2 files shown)


scenes/quests/story_quests/renya_beyond_sorrow/Complementos/music/REUSE.toml
manuq, line 1: Interesting. So far we have a single `REUSE.toml` file at the project root. You are proposing a specific `REUSE.toml` file in this directory to apply the license to all `ogg` files here at once. I like it!

PR #1721, opened 2025-12-11
Compress large non-pixel-art images lossily when exporting

wjt:
By default (in our project) images are stored with lossless WebP compression. This does reduce the size of some PNGs, e.g. Void.png is 612 KB, but in the .pck it previously became a 485 KB texture resource. (But I did see some files which actually get larger; in any case, I am surprised to learn that any time and effort that we spend crushing PNGs has no effect on the size of the exported project, only on the size of the Git LFS object!)

I think lossless compression is appropriate for our pixel-art assets (which tend to compress very well: they are low-resolution with a limited colour palette). However we do have some large images which are not pixel art, and don't compress much under lossless compression. Compressing these lossily saves space in the exported game with minimal visible loss of quality.

Helps #1683

### DIFF
--- a/scenes/quests/story_quests/wizzy_quest/2_combat/combat_components/wizzy_quest_throwing_enemy.png.import
+++ b/scenes/quests/story_quests/wizzy_quest/2_combat/combat_components/wizzy_quest_throwing_enemy.png.import
@@ -15,7 +15,7 @@ dest_files=["res://.godot/imported/wizzy_quest_throwing_enemy.png-864662719e26b3
 
 [params]
 
-compress/mode=0
+compress/mode=1
 compress/high_quality=false
 compress/lossy_quality=0.7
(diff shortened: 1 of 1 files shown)


manuq (approved):
Excellent!

russell-cj:
Hi @manuq! The **"2D Pixel"** preset is located in the **Import Dock**, not in the **Project Settings**.

### To see it:
1. Select any image file in the **FileSystem**.
2. Go to the **Import** tab (next to the **Scene** tab).
3. Click the **Preset** button at the top.
4. Select **2D**.

What I did previously was click **"Set as Default for Texture2D"** in that same menu.  
That action is what writes the `importer_defaults` block into the `project.godot` file, ensuring all future assets import with **Nearest** filter (crisp pixel art) instead of **Linear** (blurry).

image

PR #1722, opened 2025-12-11
Become the random frame script a behavior

manuq:
While working on skin tones I realized that this script is also useful for NPCs, not only for decoration, so it could be handy to make it a behavior. Then I saw the same in the cat today.

@wjt what do you think?

### DIFF
--- a/scenes/game_logic/sprite_behaviors/base_sprite_behavior.gd
+++ b/scenes/game_logic/sprite_behaviors/base_sprite_behavior.gd
@@ -1,0 +1,32 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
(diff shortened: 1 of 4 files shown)


scenes/game_logic/base_character_behavior.gd
```diff
@@ -1,8 +1,7 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 @tool
-# TODO: Prepend "abstract" when switching to Godot 4.5:
-class_name BaseCharacterBehavior
+@abstract class_name BaseCharacterBehavior
```
wjt, line 4: 👍🏻 .

I was holding off on this because Godot Patchwork was 4.4-only, but it's been rebased to 4.6 now.

PR #1726, opened 2025-12-12
Duplicate NO EDIT template andd all files and rename them all by repl…

jostar0101:
---


wjt:
The commit message doesn't describe the commit (which actually creates a tileset). I think you accidentally opened this PR against the wrong repo!

PR #1728, opened 2025-12-12
Basic internationalization localization

manuq:
This PR adds basic internationalization & localization to the project, using Gettext. And translates the title screen (except Credits) to Spanish.

Captura desde 2025-12-12 14-57-49

Captura desde 2025-12-12 14-57-56

Resolve https://github.com/endlessm/threadbare/issues/1727

### DIFF
--- a/locale/es.po
+++ b/locale/es.po
@@ -1,0 +1,58 @@
+# Spanish translation for Threadbare for the following files:
+# res://scenes/menus/title/components/main_menu.tscn
+#
+# Manuel Quiñones <manuq@endlessaccess.org> 2025
+msgid ""
+msgstr ""
+"Project-Id-Version: Threadbare\n"
+"Last-Translator: Manuel Quiñones <manuq@endlessaccess.org>\n"
+"Language-Team: Spanish\n"
+"Language: es\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Poedit 3.8\n"
+
+#: scenes/menus/title/components/main_menu.tscn
+msgid "Continue"
+msgstr "Continuar"
+
+#: scenes/menus/title/components/main_menu.tscn
+msgid "Start"
+msgstr "Iniciar"
+
+#: scenes/menus/title/components/main_menu.tscn
+msgid "Options"
+msgstr "Opciones"
+
+#: scenes/menus/title/components/main_menu.tscn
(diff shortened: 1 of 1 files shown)


wjt:
I don't think we should commit the .mo files; they're generated from the .po files

locale/threadbare.pot
```diff
@@ -0,0 +1,59 @@
...
+#: scenes/menus/title/components/main_menu.tscn
+msgid "Credits"
+msgstr ""
+
+#: scenes/menus/title/components/main_menu.tscn
+msgid "Exit Game"
+msgstr ""
+
+#: scenes/menus/options/options.tscn
+msgid "< back"
```
wjt, line 38: Oops, that shortcut of putting a `<` into the string rather than an arrow asset is coming back to haunt us!

locale/es.po
```diff
@@ -0,0 +1,62 @@
+# LANGUAGE translation for Threadbare for the following files:
```
wjt:
```suggestion
# Spanish translation for Threadbare for the following files:
```

locale/threadbare.pot
manuq: Ha yes and also we have a slightly different one in the Credits scene (capitalized)

locale/es.po
```diff
@@ -0,0 +1,62 @@
...
+# res://scenes/menus/title/components/main_menu.tscn
+#
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
```
wjt:
```suggestion
# Manuel Quiñones <manuq@endlessaccess.org> 2025
```

locale/es.po
```diff
@@ -0,0 +1,62 @@
...
+msgid ""
+msgstr ""
+"Project-Id-Version: Threadbare\n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: \n"
+"Last-Translator: \n"
```
wjt: ?

wjt (approved):

locale/es.po
manuq: Clearly my Poedit is not configured. Thanks for spotting these errors.

manuq: So weird, except for `Last-Translator` here I wasn't able to instruct Poedit from flathub to update these fields. Checking block-coding es.po (coming from Transifex) it doesn't have these date fields at all. So I will just remove them. Still if I save it in Poedit again, the 2 fields appear empty, so annoying.

PR #1730, opened 2025-12-12
Icon arrows

manuq:
Follow up of https://github.com/endlessm/threadbare/pull/1728 in which we found that the buttons had text for arrows. I made 2 asset arrows, and added them as icons to the back buttons. Also I took the opportunity to do the same in the next button of the dialogue balloons.

Oh I also translated the pause overlay to Spanish here. I decided to keep "quest" untranslated on purpose.

### DIFF
--- a/scenes/ui_elements/dialogue/balloon.tscn
+++ b/scenes/ui_elements/dialogue/balloon.tscn
@@ -54,6 +54,7 @@ size_flags_horizontal = 8
 theme_type_variation = &"FlatNextButton"
 text = "next"
 icon = ExtResource("4_qmeow")
+icon_alignment = 2
 
 [node name="ResponsesMenu" type="VBoxContainer" parent="Balloon/PanelContainer/VBoxContainer" node_paths=PackedStringArray("response_template")]
 unique_name_in_owner = true


locale/es.po
```diff
@@ -56,3 +57,23 @@ msgstr "Ajustes de video"
...
+msgid "Resume"
+msgstr "Resumir"
+
+#: scenes/globals/pause/pause_overlay.tscn
+msgid "Abandon Quest"
+msgstr "Abandonar quest"
+
+#: scenes/globals/pause/pause_overlay.tscn
+msgid "Exit to Title"
+msgstr "Salir al título"
```
wjt, line 79: Interesting. In English I chose to use "Title Case" for this menu. This is common in style guides, e.g. https://developer.gnome.org/hig/patterns/controls/menus.html https://developer.apple.com/design/human-interface-guidelines/menus#:~:text=To%20be%20consistent%20with%20platform%20experiences%2C%20use%20title%2Dstyle%20capitalization, and in a few games that I checked. Is the convention different in Spanish? (I don't really mind either way - so long as each language is internally consistent with itself.)

manuq: Yes, the convention is different in Spanish! The title case doesn't really exist here. For titles only the first word is capitalized, and sometimes proper nouns. Examples:
- The Old Man and the Sea -> El viejo y el mar
- One Hundred Years of Solitude -> Cien años de soledad

scenes/ui_elements/dialogue/balloon.tscn
```diff
@@ -54,6 +54,7 @@ size_flags_horizontal = 8
 theme_type_variation = &"FlatNextButton"
 text = "next"
 icon = ExtResource("4_qmeow")
+icon_alignment = 2
```
manuq, line 57: Thanks for adding the alignment. I was wondering about this. Both how to do it, and if it's OK to do it (or use left side consistently in buttons). Since you made it look the same as it was before, is better.

PR #1737, opened 2025-12-15
Remove unused visible_teleporter scene + script

wjt:
This was used in an earlier iteration of Fray's End but has not been
used since commit 4dad9e20f7b6e9b23582d32240c6b8d100e7f972.


scenes/game_elements/props/teleporter/visible_teleporter/components/visible_teleporter.gd
```diff
@@ -1,11 +0,0 @@
...
-# SPDX-License-Identifier: MPL-2.0
-@tool
-extends Teleporter
-
-@onready var blue_fire: AnimatedSprite2D = %BlueFire
-
-
-func _ready() -> void:
-	super._ready()
-	blue_fire.play("burning")
```
wjt, line 11: I think if one wanted to rebuild this scene one could just add an AnimatedSprite2D to a teleporter node and set the relevant animation to autoplay.

manuq: Indeed.

PR #1759, opened 2025-12-16
tiles: Randomize starting frame of foam animation

wjt:
Pixel Frog recently posted a nice tutorial on using the Tiny Swords
tilesets <https://pixelfrog-assets.itch.io/tiny-swords/devlog/1138989/tilemap-guide> on which our tiles are based.

This sentence jumped out at me:

> To make it look more natural, make that all different Water Foam sprites start the animation at a different frame.

I had never noticed it before, but in Godot's Tile Set panel, under Select > Animation, there is a Mode option. The default option is Default but the other available option is Random Start Times.

Enable this option for the five foam tiles.


manuq (approved):
This is a great improvement!

I overlooked the post, haven't paid attention to this. It makes a difference!

Issue #1753, opened 2025-12-15
Void runner: Block forest from walking


manuq:
In the Void runner level (1st level of the Void ques) the player shouldn't be able to traverse the forest. Only should be able to walk on paths.

This was explained as a game language problem: if the player is allowed to enter the forest they will try to find something there. Why is this different from the previous forests (Fray’s End, music puzzle) which you can’t enter? The only way to figure it out is by trying walking into it.

PR #1764, opened 2025-12-17
Void runner: Add forest collider

manuq:
Please note that we can't do the same for the "SouthCopse" area because the player has to traverse it. Do it only for the thick forest.

Resolve https://github.com/endlessm/threadbare/issues/1753

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
@@ -849,15 +849,17 @@ polygon = PackedVector2Array(-221.5, 55.5, -226.5, 102.5, -174.5, 105.5, -158.5,
 
 [node name="CollectibleItem" parent="OnTheGround" instance=ExtResource("22_auh5r")]
 position = Vector2(474, 1058)
+revealed = false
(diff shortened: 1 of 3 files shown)


scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
```diff
@@ -613,7 +601,22 @@ scale = Vector2(1.02937, 1.06714)
 position = Vector2(1031.43, 1481.95)
 scale = Vector2(1.11164, 1.02479)
 
-[node name="Tree48" parent="OnTheGround/Forest" instance=ExtResource("10_epkxq")]
+[node name="ForestExtraTrees" type="Node2D" parent="OnTheGround"]
```
wjt, line 604: You need to add this new node to TileMapLayers/Void consumable_node_holders, otherwise these trees are not initially eaten and can be seen in the animation:

Image

wjt: I amended the commit to fix this; and added 1 more commit to deal with the same bug for the thread (but in a different way: it is revealed only when the void is pushed back).

PR #1768, opened 2025-12-17
tiles: Add single-tile foam TileSet

wjt:
The previous foam tileset had five tiles, arranged in a cross shape:

image

This arrangement is hard to use because when two or more coastlines
border the same tile, you need to have 2 or more foam ripple tiles on
top of each other to get the correct effect. And if you look very
closely at the sprite sheet, there is actually the odd pixel of foam in
the inside corners of the ·.

image

https://pixelfrog-assets.itch.io/tiny-swords/devlog/1138989/tilemap-guide
explains that this is supposed to be used in a different way. The 64×64
tile should use a 128×128 texture, centred on the solid-blue square.

image

With this configuration you place the tile under the coastline, rather
than on top of the water, and the foam animation is shown on the four
neighbouring tiles. The ones that are under the ground are obscured.
Then if you have 2-4 coastlines touching the same tile of water, the
correct effect can be achieved with a single foam layer.

Add this new tileset. Update 3 scenes to use it. The others can be updated gradually, and eventually we can delete `foam.tres`.

Unlike with the previous tileset split, we can't do this automatically because it's not a drop-in replacement.


manuq (approved):
Yes! I agree to have the two foam tilesets during the migration, after which we'll remove the former and rename this new one.

PR #1769, opened 2025-12-17
Void runner: Make forest visibly denser

manuq:
Follow up of https://github.com/endlessm/threadbare/pull/1764

Change the AreaFiller minimum separation to 64 (the default value). Play with the Refill button until satisfied with the result. Slightly reposition a few trees after that.


wjt (approved):
Thanks, much better!

PR #1773, opened 2025-12-17
linux: Add forthcoming 0.1.8 release to releases file

manuq:


wjt:
I wonder if we should do this the other way round: release, then update the file. And then we could automate it with GitHub Actions...!

Issue #1760, opened 2025-12-17
Add cliff shadows


wjt:
https://pixelfrog-assets.itch.io/tiny-swords/devlog/1138989/tilemap-guide has a section on adding shadows below cliffs:

> Shadow: The Shadow sprite is 128x128, but you will place it on a 64x64 tile size as the rest of the regular tiles. This is so you can overlap the edges and create an effect like is shown in the example bellow. Keep in mind that the Shadow sprites are going to be placed below the same area as the walkable part of the Elevated Ground, but to create the illusion of height, it is moved completely one 64x64 pixel square downwards.
>
> <img width="896" height="448" alt="Image" src="https://github.com/user-attachments/assets/cae66a59-98f7-4a22-abf5-c7286ed646c5" />

@manuq and I both read this and thought, yes, shadows would be good!

The technique described above is the same technique as for foam: https://github.com/endlessm/threadbare/issues/1621#issuecomment-3664781236

We do already have shadow tiles in the project from the CC0 version of Tiny Swords, in `assets/third_party/tiny-swords/Terrain/Ground/Shadows.png`:

<img width="192" height="192" alt="Image" src="https://github.com/user-attachments/assets/6026ee4c-7741-4d81-bb02-16aea752bcd1" />

But the technique above doesn't work for this version of the shadow tiles (different to those in the post above). The shadow is of course semitransparent, and when you place two of it side by side you can see them overlapping.

PR #1780, opened 2025-12-18
Add shadows tiles, and update Fray's End

wjt:
This uses the shadow tile from Tiny Swords. Like foam, this is a single oversize tile that should be placed underneath each cliff-face.

Unlike foam, this doesn't quite work correctly because the shadow is semi-transparent, and if you have two next to each other horizontally, they overlap slightly, so there is a darker triangle of shadow where the two cliff tiles meet at the ground. However it's very subtle and you have to be looking very closely to notice it; and the net improvement from adding shadows is worth it.

Resolves https://github.com/endlessm/threadbare/issues/1760
(though we will still need to gradually actually add shadows to each scene). I added shadows to Fray's End as an example.


manuq (approved):
Much better!

Image

manuq:
> Unlike foam, this doesn't quite work correctly because the shadow is semi-transparent, and if you have two next to each other horizontally, they overlap slightly, so there is a darker triangle of shadow where the two cliff tiles meet at the ground. However it's very subtle and you have to be looking very closely to notice it; and the net improvement from adding shadows is worth it.

The easiest option is to tweak the asset so these triangles don't happen. I think that's what PixelFrog did in the updated shadow of non-free TinySwords. But I agree that it's too small and hard to notice. The shadows are worth it.

Issue #1770, opened 2025-12-17
Replace all uses of foam.tres with foam2.tres


wjt:
In #1621 I added a replacement foam tileset.

We need to update all scenes to use it before we can remove the old one (and rename the new one to a less ugly name).

This doesn't have to be done all at once: we can open subtasks for particular (groups of) scenes. The general process is:

1. Find scenes that use `tiles/foam.tres`
   <img width="492" height="243" alt="Image" src="https://github.com/user-attachments/assets/121e86e8-7022-4693-a38c-c8fb8635a122" />
2. Double-click one to open it
3. Identify the `TileMapLayer` nodes that use `foam.tres`. There are normally 2 or more, typically at least one for horizontal and vertical foam
4. Duplicate the topmost foam layer
5. Remove all tiles from it
6. Change its Tile Set to `tiles/foam2.tres`
7. Repaint the foam in this layer to match the existing foam (or, better, to correctly match the coastlines)
8. Delete the old foam layers
9. Rename the remaining layer to `Foam`

- [ ] scenes/game_elements/characters/enemies/mothsache/Mothsache_test_scene.tscn
- [x] scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1.tscn
- [x] scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_2.tscn #1772 
- [x] scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn #1771 
- [x] scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
- [x] scenes/quests/lore_quests/quest_001/4_closing_transition/closing_transition.tscn
- [x] scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
- [x] scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_needles.tscn
- [x] scenes/quests/lore_quests/quest_002/3_void_grappling/void_grappling.tscn
- [x] scenes/quests/lore_quests/quest_002/4_closing_transition/closing_transition.tscn
- [ ] scenes/quests/lore_quests/quest_003/1_stealth_sequence_combo/stealth_sequence_combo.tscn
- [ ] scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_4.tscn
- [ ] scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_5.tscn
[...]

PR #1803, opened 2026-01-06
Use single-tile foam in LoreQuest 1

wjt:
Helps #1770


scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
```diff
@@ -160,6 +160,7 @@ script = ExtResource("1_f6lvl")
 stream = ExtResource("2_p2u32")
 
 [node name="CanvasModulate" type="CanvasModulate" parent="."]
+visible = false
```
wjt: oops!

PR #1805, opened 2026-01-06
Add Champ mini game 2

felixwalberg:
The sequence minigame has Evan continuing his journey along the shore of Lake Champlain. He needs to cross part of the lake via some stepping stones that are submerged underwater. Only certain rocks will allow him to stand on them without fully submerging, forcing Evan to start again! 

This minigame involved the extension of scripts for sequence puzzle logic, hint signs, and sequence puzzle objects, as well as the addition of a new champ_long_rock script.

Assets:
champ_object.png by @felixwalberg 
champ_sign.png by @felixwalberg 
champ_animated_water_rock.png by @felixwalberg 
champ_dry_rock.png by @felixwalberg 
champ_long_rock.png by @felixwalberg 
champ_long_water_rock.png by @felixwalberg 
champ_wet_rock.png by @felixwalberg 
water_rocks_sequence.png by @felixwalberg

Engineering:
Sequence Puzzle Level  by @felixwalberg 

Game Design:
Core Gameplay & Mechanics by @felixwalberg 
Progression & Motivation by @felixwalberg

### DIFF
--- a/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
+++ b/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
@@ -159,7 +159,7 @@ func reset_all() -> void:
 	# sprite frames, I can't just wipe everything
 	for obj in objs:
 		if sequences[0].sequence.has(obj) or sequences[1].sequence.has(obj):
-			obj.rock.play("default")
+			obj.rock.play(obj._idle_animation)
 
 ## Function to move camera position so entire sequence is shown (for second sequence)
 func _on_hint_sign_2_demonstrate_sequence() -> void:
--- a/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_object.tres
+++ b/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_object.tres
@@ -1,10 +1,15 @@
-[gd_resource type="SpriteFrames" load_steps=7 format=3 uid="uid://c2m3tplriqe2s"]
+[gd_resource type="SpriteFrames" load_steps=9 format=3 uid="uid://c2m3tplriqe2s"]
 
 [ext_resource type="Texture2D" uid="uid://c0hyny3m5fap4" path="res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_object.png" id="1_5vycx"]
 [ext_resource type="Texture2D" uid="uid://bmqyby71cbp61" path="res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_wet_rock.png" id="1_62qai"]
+[ext_resource type="Texture2D" uid="uid://chc11kiie54o6" path="res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_dry_rock.png" id="2_upwns"]
 
 [sub_resource type="AtlasTexture" id="AtlasTexture_upwns"]
 atlas = ExtResource("1_62qai")
+region = Rect2(0, 0, 96, 96)
+
+[sub_resource type="AtlasTexture" id="AtlasTexture_tmov7"]
+atlas = ExtResource("2_upwns")
 region = Rect2(0, 0, 96, 96)
 
 [sub_resource type="AtlasTexture" id="AtlasTexture_riqxt"]
@@ -27,6 +32,14 @@ animations = [{
 }],
 "loop": true,
 "name": &"default",
+"speed": 10.0
+}, {
+"frames": [{
+"duration": 1.0,
+"texture": SubResource("AtlasTexture_tmov7")
+}],
+"loop": true,
+"name": &"dry",
 "speed": 10.0
 }, {
 "frames": [{
--- a/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_object.gd
+++ b/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_object.gd
@@ -7,21 +7,20 @@ extends SequencePuzzleObject
 @onready var collision: CollisionShape2D = $CollisionShape2D
 @onready var rock: AnimatedSprite2D = %AnimatedSprite2D
 
+var _idle_animation := &"default"
 
-func _ready() -> void:
-	super._ready()
-	if not Engine.is_editor_hint():
-		sprite_frames = sprite_frames.duplicate()
-
-## Function to change sprite frames from the submerged rock to dry rock
+## Makes the rock passable, and displays a dry idle state.
 func dry_off() -> void:
-	# Current implementation changes file path of "default" animation, since `_stop` in parent sequence puzzle script plays "default" sprite frames 
-	rock.sprite_frames.set_frame("default", 0, preload("res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_dry_rock.png"), 1)
-	if collision:
-		collision.disabled = true
+	_idle_animation = &"dry"
+	collision.disabled = true
 
-## Function to change object to default animation with waves, frees interactive node
+# Overrides superclass to make idle animation dynamic
+func _stop() -> void:
+	if animated_sprite.is_playing() and animated_sprite.animation == "struck":
+		await animated_sprite.animation_looped
+		animated_sprite.play(_idle_animation)
+
+## Submerges the rock and prevents further interactions with it.
 func submerge() -> void:
-	rock.play("default")
-	if interact_area:
-		interact_area.queue_free()
+	rock.play(&"default")
+	interact_area.disabled = true


felixwalberg:
This branch was open for a while (during my winter break), so it fell behind the main branch. In an attempt to rebase, there were many conflict issues with files that I did not touch on this branch (and ensured were nowhere in my git history before committing). 

I decided to reset to before the rebase and am pushing only my changes, as I believe that shouldn't overwrite anything outside the scope of the Champ SQ. If this is incorrect and there is anything I can do to fix it, please let me know!

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_long_rock.tscn
```diff
@@ -0,0 +1,88 @@
...
+[sub_resource type="RectangleShape2D" id="RectangleShape2D_oln2r"]
+size = Vector2(286, 110.000015)
+
+[node name="ChampLongRock" type="Node2D"]
+script = ExtResource("1_f6rm8")
+
+[node name="Sprite2D" type="Sprite2D" parent="."]
+visible = false
+rotation = 1.5707964
+texture = ExtResource("2_rwx2y")
```
wjt, line 73: I think this hidden Sprite2D is unused and can be removed.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_long_rock.gd
```diff
@@ -0,0 +1,36 @@
...
+
+func _ready() -> void:
+	if submerged:
+		rock.play("waves")
+	else:
+		rock.play("default")
+
+func _physics_process(delta: float) -> void:
+	if area_2d.get_overlapping_bodies() != [] and submerged:
+		_on_area_2d_body_entered(Node2D.new())
```
wjt: It took me a while to figure out what this is meant to do. I think what you are trying to deal with is this sequence of events:

1. Rock is not submerged
2. Player walks onto rock
   - _on_area_2d_body_entered() is called, but because the rock is above water `water_entered` should not be emitted
4. Rock becomes submerged

At this point you want to knock the player back, so in essence you fake a `body_entered` signal being emitted.

I think there is a simpler way to do this. I'll push a patch. Actually the result is 1 line longer but I think it's clearer & more elegant.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
```diff
@@ -23,6 +23,1132 @@
...
+"duration": 1.0,
+"texture": SubResource("AtlasTexture_mfero")
+}, {
+"duration": 1.0,
+"texture": SubResource("AtlasTexture_rsmxy")
+}],
+"loop": true,
+"name": &"struck",
+"speed": 10.0
+}]
```
wjt: This SpriteFrames resource is duplicated for each rock. Please save it to a file and assign that to each rock.

In fact you already have such a saved resource. The problem is probably that you have set `resource_local_to_scene` on the resource. This means that the resource will get duplicated each time a scene that uses it is instantiated. There are times when this is useful - for example, if a scene has a CollisionShape that gets modified at runtime, and you have multiple instances of that scene whose collision shapes are independent. I fixed a bug like this recently in another project - https://github.com/endlessm/moddable-platformer/pull/51/changes/6a00fff81c80213bd725e698a94c5c0f3694136c.

But in this case the thing to do is to unset that flag and share the resource... oh, I see! The problem is that you actually do modify the resource in `champ_sequence_puzzle_object.gd`. I don't think that's the right approach but we can do better even without changing that:

1. Duplicate the `sequence_puzzle_object.tscn` as `champ_rock.tscn`.
2. Change the root node on that scene to use `champ_sequence_puzzle_object.gd` (rather than overriding each instance of it in `champ_sequence_puzzle.tscn`
3. Set the `champ_object.tres` as its SpriteFrames
4. Replace the rocks with instances of that scene

Now `resource_local_to_scene` will behave as desired.

I'll cook a quick patch to make this change.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
```diff
@@ -31,115 +1157,328 @@ metadata/_custom_type_script = "uid://bgmwplmj3bfls"
...
 
 [node name="Player" parent="OnTheGround" instance=ExtResource("2_ey61o")]
-position = Vector2(383, 371)
+z_index = 1
+position = Vector2(350, 350)
 sprite_frames = ExtResource("3_6wj4q")
 
+[node name="Camera2D" type="Camera2D" parent="OnTheGround/Player"]
+position = Vector2(-526, -360)
+anchor_mode = 0
```
wjt, line 99: I think you should unset these properties because they make it harder to reason about the camera motion. More on this story in the code that moves the camera around...

```suggestion
```

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
```diff
@@ -0,0 +1,171 @@
...
+			obj.rock.play("default")
+
+## Function to move camera position so entire sequence is shown (for second sequence)
+func _on_hint_sign_2_demonstrate_sequence() -> void:
+	var goal: float = camera.offset.y - CAMERA_MAX_OFFSET_Y
+	while camera.offset.y > goal:
+		await get_tree().create_timer(CAMERA_TIMEOUT_INTERVAL).timeout
+		camera.offset.y = camera.offset.y - CAMERA_OFFSET_Y
+		camera.zoom.y = camera.zoom.y - CAMERA_ZOOM_SCALAR
+		camera.zoom.x = camera.zoom.x - CAMERA_ZOOM_SCALAR
```
wjt, line 171: Adjusting the camera during the demonstration is a really nice touch! I designed around this in stealth_sequence_combo by having each sequence only use the rocks that are visible when the player is standing at that rock, but it was annoying.

However I think you could consider doing this a bit differently:

1. What you are hand-rolling here is effectively a `Tween` https://docs.godotengine.org/en/stable/classes/class_tween.html which smoothly animates 1 or more properties with a given curve, duration, and target value.
2. But Camera2D already knows how to smoothly move itself between positions! I did a little testing and you don't need to adjust the zoom - if you temporarily move the camera to the position of the middle rock of the middle row of the second puzzle, all rocks are in frame. Something like this:


```suggestion
	camera.global_position = $Objects/Middle8.global_position
```

(assuming you have taken my other suggestion of resetting the camera's anchor mode and position). 

Then you can put it back in `_on_hint_sign_hint_sequence_finished` with `camera.position = Vector2.ZERO`.

(The camera still moves with the player so this relies on the fact that the player can't move during the demonstration. You could avoid this theoretical problem by having a second camera that is only used during the demonstration.)

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
wjt: This doesn't work either because the `sprite_frames` property is on the scene root node so is duplicated when the scene is instanced. OK, I'll work around that!

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
wjt, line 1: At this point, I'm not sure that re-using (and subclassing) the scenes & scripts from the vanilla puzzle is really helping all that much. `sequence_puzzle.gd` is only 140 lines and this subclass overrides almost all of it!

TBH I have never been all that happy with how tightly coupled this puzzle is. It was I who wrote it, and who made it "generic", but the second user of it - Stella - ended up working around the design forced by each object being a separate scene, when the Stella instance has one turtle with coloured segments of its shell.

I have the feeling that this level would be simpler & less error-prone if you duplicated the scenes & scripts, deleted the generic stuff that is no longer needed, and made the necessary changes, e.g. having the rocks go into a "solved" state that lets you walk over them.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
```diff
@@ -0,0 +1,171 @@
...
+	platforms = Array(long_rocks.get_children())
+	objs = Array(objects.get_children())
+	if sequences.size() > 0:
+		first_seq_solve_index = sequences[0].sequence.size() - 1
+	super._ready()
+	
+func  _physics_process(delta: float) -> void:
+	if player.position.x > SECOND_SEQ_BOUND:
+		for i in range(ROCK_WIDTH * first_seq_solve_index, ROCK_WIDTH * (first_seq_solve_index + 1)):
+			objs[i].collision.disabled = false
```
wjt, line 42: This is a lot of work to being doing on every frame. You should instead have an Area2D on the middle island that detects the player, and makes whatever  change you want to make. This also avoids you having magic numbers like `SECOND_SEQ_BOUND` which have to be kept in sync with the level  design.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_hint_sign.gd
```diff
@@ -0,0 +1,22 @@
...
+func _ready() -> void:
+	# Ensure the second hint sign camera movement does not cut off a step
+	collision_shape.scale = COLLISION_SCALE
+	collision_shape.position.y = collision_shape.position.y - COLLISION_OFFSET
+	super._ready()
+	
+## Function to emit a signal to champ specific sequence puzzle script so path can be reset
+func demonstration_finished() -> void:
+	hint_sequence_finished.emit()
+	super.demonstration_finished()
```
wjt, line 22: You don't need to override this and handle it in `champ_sequence_puzzle.gd` because it is the SequencePuzzle script that calls `demonstration_finished()` (from `_on_demonstrate_sequence`). Override that function and do the necessary work from `_on_hint_sign_hint_sequence_finished` there.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_hint_sign.gd
```diff
@@ -0,0 +1,22 @@
...
+
+@onready var collision_shape: CollisionShape2D = $InteractArea/CollisionShape2D
+
+const COLLISION_OFFSET : int = 50
+const COLLISION_SCALE : Vector2 = Vector2(0.5, 0.5)
+
+func _ready() -> void:
+	# Ensure the second hint sign camera movement does not cut off a step
+	collision_shape.scale = COLLISION_SCALE
+	collision_shape.position.y = collision_shape.position.y - COLLISION_OFFSET
```
wjt, line 16: I don't understand why you would need to change the collision shape of the sign in order to affect the camera movement.

But if you need to change the collision shape: duplicate the scene and change it there, rather than doing it in code.

With the two changes proposed here you can drop this custom script.

wjt (changes requested):

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
wjt: Pushed some patches to fix this.

wjt (approved):
I pushed some patches that address some of my comments. If you're happy with them feel free to merge and we can iterate on the other points!

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
felixwalberg: I'm planning to completely remove this logic. Instead, once the rock has been successfully "solved", it loses the interact_area but does not retain the Collision2D. This way, the learner *can* walk back across the platforms while not running the risk of interacting with a previous sequence.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_hint_sign.gd
felixwalberg: This was an attempt to force the player to walk closer to the sign so that the camera wouldn't cut off any of the rocks in the hint demo. Now that we center the camera around rock positions, rather than relative to the player, we can completely remove this.

PR #1806, opened 2026-01-06
Add Champ outro animation prompt and script

felixwalberg:
This modifies the outro to change the champ sprite to an AnimatedSprite2D; and to add basic movement with CharacterBody2D, along with hints on other ways the movement could be changed.

### DIFF
--- a/scenes/quests/story_quests/champ/4_outro/champ_animation.gd
+++ b/scenes/quests/story_quests/champ/4_outro/champ_animation.gd
@@ -1,0 +1,24 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends CharacterBody2D
+
+@onready var champ_animation: AnimatedSprite2D = $AnimatedSprite2D
+
+var swim_speed: int = 100
+
+func _ready() -> void:
+	# Give champ a starting swim speed
+	velocity.x = swim_speed
+
+func _physics_process(_delta: float) -> void:
+	# Make sure Champ doesn't swim out of the water! 
+	# If he gets too close, turn him around and start swimming the other direction
+	if position.x > 400 or position.x < 0:
+		champ_animation.scale.x *= -1
+		velocity.x *= -1
+
+	# TODO: Can we make Champ swim in a more interesting pattern?
+	# Can we change the way Champ swims vertically? Can we accelerate?
+
+	# Continues to move the CharacterBody2D Object based on it's velocity
+	move_and_slide()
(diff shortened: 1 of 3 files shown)


scenes/quests/story_quests/champ/4_outro/champ_outro.tscn
```diff
@@ -1,11 +1,31 @@
...
+
+[sub_resource type="SpriteFrames" id="SpriteFrames_6d3de"]
+animations = [{
+"frames": [{
+"duration": 1.0,
+"texture": ExtResource("7_276wt")
+}, {
+"duration": 1.0,
+"texture": ExtResource("7_276wt")
+}, {
```
wjt, line 21: For future reference you could have one frame at 2× length rather than duplicating the frame:

Image

scenes/quests/story_quests/champ/4_outro/champ_animation.gd
```diff
@@ -0,0 +1,24 @@
...
+var swim_speed: int = 100
+
+func _ready() -> void:
+	# Give champ a starting swim speed
+	velocity.x = swim_speed
+
+func _physics_process(_delta: float) -> void:
+	# Make sure Champ doesn't swim out of the water! 
+	# If he gets too close, turn him around and start swimming the other direction
+	if position.x > 400 or position.x < 0:
```
wjt, line 16: Given that this is a CharacterBody2D, you could also do this by having invisible walls that it collides with, and handling `move_and_slide()` returning `true` (i.e. a collision occurred) by flipping the velocity.

(Not saying you have to change this, just an idea.)

scenes/quests/story_quests/champ/4_outro/champ_animation.gd
```diff
@@ -0,0 +1,24 @@
...
+		champ_animation.scale.x *= -1
+		velocity.x *= -1
```
wjt, line 18: Another way you might spell this is:

```suggestion
		velocity.x *= -1
		champ_animation.flip_h = velocity.x < 0
```

I think the difference is that `flip_h` only applies to the texture itself, whereas setting `scale.x = -1` would also flip any child nodes. In this particular case they are equivalent. Just leaving a note about what we do elsewhere in the project, e.g https://github.com/endlessm/threadbare/blob/ecf740c615f9214f4dfe6cb1ef5885d9f48af9fd/scenes/game_elements/characters/player/components/player_sprite.gd#L13-L14

scenes/quests/story_quests/champ/4_outro/champ_animation.gd
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends CharacterBody2D
+
+@onready var champ_animation: AnimatedSprite2D = $AnimatedSprite2D
+
+var swim_speed: int = 100
```
wjt, line 7:
```suggestion
@export var swim_speed: int = 100
```

Perhaps?

wjt (approved):

PR #1820, opened 2026-01-13
storyquest_bootstrap: Add to command palette

wjt:
storyquest_bootstrap: Add to command palette

This is accessed with Ctrl+Shift+P on Linux and some other shortcut on other platforms.


addons/storyquest_bootstrap/plugin.gd
```diff
@@ -17,10 +18,14 @@ var _new_storyquest_dialog: Window
 
 func _enter_tree() -> void:
 	add_tool_menu_item(TOOL_MENU_LABEL, _open_new_storyquest_dialog)
+	EditorInterface.get_command_palette().add_command(
```
manuq, line 21: Nice!

PR #1821, opened 2026-01-13
Move NO_EDIT template assets to assets/

wjt:
Previously we included various template PNGs (and in some cases Aseprite source
files) in the NO_EDIT template quest.

Unlike the SpriteFrames resources and dialogue files (where it's easiest to edit
them in-place rather than duplicate them), we've not seen people inadvertently
editing the template PNG files in-place. It's more common for those who are
drawing their own art to copy a new file from whatever tool they are using (e.g.
Piskel) into their StoryQuest. Many of the StoryQuests I have reviewed have
contained unmodified copies of the various PNGs, either used or unused.

Move these assets to assets so we don't end up with so many unused duplicates.
Update all references.

Resolves https://github.com/endlessm/threadbare/issues/1359


wjt:
@jgbourque Please take a look! This is what we discussed on #1359 but I want to draw your attention to it in case any learning materials will need to be updated.

wjt:
Landing this without further review on the basis of https://github.com/endlessm/threadbare/issues/1359#issuecomment-3739909114

PR #1826, opened 2026-01-14
Stella: Rebuild tortoise as separate scene(s)

wjt:
Unlike all the other sequence puzzles in the game, where the objects to
kick/tap/etc. are visibly separate in the game, in this puzzle the objects are
all segments of the tortoise's shell. This makes it particularly awkward to use the
sequence_puzzle_object.tscn scene because the InteractArea collision shapes have
to be different for each segment; and the segments themselves shouldn't block
player movement (because the main body of the tortoise should do that).

It's made even harder by the fact that sequence_puzzle_object.tscn scales its
sprite by 0.7× (which I consider a bug), so the main body of the tortoise has to
be scaled to match. (And then the tortoise assets are half the desired size so
the whole thing needs to be scaled up 2× on top of that.)

Rebuild the tortoise as a pair of new scenes: one for the segments, which does
not have any collision shapes configured; and one for the tortoise as a whole,
which instantiates six of the segments, enables editable children, and applies
the necessary tweaks. Discard the 0.7× scaling and simply scale up all the
sprites by 2.

Back in the main scene for the challenge, instantiate the tortoise. Now this is
the tricky part: we can still refer to the SequencePuzzleObject nodes within the
Tortoise node on the Step nodes, even though they are not visible when Editable
Children is disabled. This is a bit obscure but it makes the main challenge
scene much more legible IMO so I think it's worth it.

Helps https://github.com/endlessm/threadbare/issues/679

### DIFF
--- a/scenes/quests/story_quests/stella/3_stella_sequence_puzzle/stella_sequence_components/stella_tortoise.tscn
+++ b/scenes/quests/story_quests/stella/3_stella_sequence_puzzle/stella_sequence_components/stella_tortoise.tscn
@@ -1,0 +1,141 @@
+[gd_scene load_steps=15 format=3 uid="uid://cml3bcuq3vmeh"]
+
+[ext_resource type="Texture2D" uid="uid://c6khhgic6yboq" path="res://scenes/quests/story_quests/stella/3_stella_sequence_puzzle/stella_sequence_components/stella_tortoise.png" id="1_pw0mv"]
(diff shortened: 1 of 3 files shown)


hydrolet:
Thanks for the correct link! I didn't scroll up enough and used the one github automatically posts. Oops!

Tried it and it looks great. Glad to hear the change would be clear to learners. It would be great to document that note in the tooltip, as you discussed. Many thanks!!

Issue #1829, opened 2026-01-15
Theme re-written when saving in freshly-opened project with no open files


wjt:
Steps to reproduce:

1. Open project
2. Close all open scenes & script files
3. Reload project
4. Close default scene
5. Monitor the filesystem for changes (e.g. `watchman-wait -m 0 *`)
6. Press Ctrl+S to save

Expected results:

- No files outside `.godot` are rewritten.

Actual result:

```
scenes/ui_elements/components/theme.tres-FRikAt
scenes/ui_elements/components/theme.tres
scenes/ui_elements/components
```

(The file with a 6-char random suffix is a temporary file used for the "save to temporary file then replace original file with `rename()`" pattern to atomically write to a file.)

The theme does not actually have any changes: `git diff` is (for me) clean after doing this.

### XP Summary (total 825)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Optimization & Performance | Problem Solving | 150 |
| Engineering: Optimization & Performance | Creativity & Design Thinking | 150 |
| Engineering: Optimization & Performance | Technical Literacy | 150 |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

wjt:
Surprisingly this is actually related to #1832.

The issue there turned out to be that the pause overlay, which is an autoloaded scene so is initialised very early in the project setup, transitively depends on the credits resources, via `preload()`ing the title screen scene.

With that fixed in #1833, plus the change in #1828, the only file rewritten when saving a freshly-opened project with no open scenes and scripts is `scenes/ui_elements/components/theme.tres`. I remain interested in this problem…

wjt:
I tried moving the default theme to the project settings and removing explicit references from all scenes in the project. (Arguably we should do that anyway!) But the behaviour remained.

I've expended enough time on this. It is harmless so I will stop here and leave this in the backlog.

wjt:
I've tagged this with extremely high XP because I think this will need deep knowledge of the internals of Godot to solve.

My current best idea is to build Godot from source with debug symbols enabled; record a trace of the problem with https://rr-project.org/; replay the trace; set a breakpoint in the codepath that actually writes the file; identify the pointer address of the resource being saved; then set a breakpoint on that object's `set_edited` method and replay the trace.

PR #1831, opened 2026-01-15
Unsaved Files in Git After Loading Godot

felixwalberg:
After attempting to save the files that had been changed so that I could show Will the diff, all the changes vanished on running a commit. 

What worked: completely deleting the repository and cloning again. Now I am able to run my scenes without UID not found errors.

What is still happening: These files (in the diff of this PR) are still appearing as changed and untracked. I added all of them to be tracked by git and committed them. There were no files that I made any modifications to during my work. My local `main` was up to date with `endless/threadbare.git`.

As of now, this doesn't seem to be hindering my ability to make more progress on the VRmont SQ, but of course, once I go to make my next PR, these changes might still be present.


scenes/menus/credits/data/fannypack.csv.import
```diff
@@ -1,14 +1,17 @@
 [remap]
 
-importer="game.threadbare.credits"
-type="Resource"
+importer="csv_translation"
```
wjt, line 3: OK, this is interesting and disappointing. I introduced a new import plugin to parse the CSV in a different way to the default.

felixwalberg:
> > After attempting to save the files that had been changed so that I could show Will the diff, all the changes vanished on running a commit.
> 
> If I understand correctly what happened: you committed the changes (screenshot on #1829) so by definition re-saving the project changed nothing. You could have used `git show` or a similar command to inspect the comimt.

That makes sense. Once I committed, my git status was up to date with the remote branch, and I was unaware that there were other ways to find commits if they didn't show up as "ahead" of the remote branch there.

wjt:
> That makes sense. Once I committed, my git status was up to date with the remote branch, and I was unaware that there were other ways to find commits if they didn't show up as "ahead" of the remote branch there.

You can compare your current branch to another one with a command like `git diff main...my-lovely-branch`. You can use `git diff main...HEAD` to compare main to the checked-out commit. Also you may find a Git GUI useful to explore the commit graph.

I'll close this PR and track the problem with the CSV import plugin in #1832. Thanks for reporting this!

PR #1841, opened 2026-01-21
storyquest_bootstrap: Free scenes after duplicating them

wjt:
In order to duplicate a .tscn file, we load it as a PackedScene,
instantiate it, modify that instance, pack it, and save the new
PackedScene. Previously the instance was leaked. Free it.

This is not needed for any other types handled by this tool because
Resources are RefCounted; PackedScene is a Resource, but the instanced
scene is a Node which is not a RefCounted.

### DIFF
--- a/addons/storyquest_bootstrap/copier.gd
+++ b/addons/storyquest_bootstrap/copier.gd
@@ -108,6 +108,8 @@ func copy_packed_scene(packed_scene: PackedScene, copy_path: String) -> Resource
 	result = ResourceSaver.save(copied)
 	assert(result == OK, error_string(result))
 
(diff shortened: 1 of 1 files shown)


addons/storyquest_bootstrap/copier.gd
```diff
@@ -108,6 +108,8 @@ func copy_packed_scene(packed_scene: PackedScene, copy_path: String) -> Resource
 	result = ResourceSaver.save(copied)
 	assert(result == OK, error_string(result))
 
+	scene.free()
```
manuq, line 111: 👍

PR #1844, opened 2026-01-22
Move generic sequence puzzle object scene to NO_EDIT template

wjt:
Previously we had one scene with the sequence_puzzle_object.gd script attached
at its root, with a generic white circle as the sprite. This scene was then
instantiated in every scene that uses this puzzle, and on every instance in each
scene, the SpriteFrames were overridden appropriately.

The problem with doing this is that it only really works if the sprite matches
the collision & interaction shapes of the generic asset. (In fact these shapes
did not match the generic asset: they were created based on the musical rocks!)
We either ignored this issue, or used the Editable Children feature to tweak the
collision shapes as needed.

In recent commits this has changed:

- musical_rock.tscn represents the specific rock that is used in the musician
  quest; it is configured with the rock SpriteFrames and matching shapes.

- champ_sequence_puzzle_object.tscn uses a subclass of the
  sequence_puzzle_object.gd script, and has its own collision shapes matching the
  sprite in that quest.

- In Stella, the tortoise is a scene containing six instances of
  stella_tortoise_shell_gem.tscn. Each gem has sequence_puzzle_object.gd attached
  but unlike the other instances it is a Node2D rather than a SolidBody2D. The
  tortoise itself is a single SolidBody2D.

The pattern is that it is more flexible to duplicate the scene so that it can be
customized for the specific puzzle at hand, while reusing the script.

To guide learners in this direction, move the generic object scene into the
NO_EDIT template. Our StoryQuest duplicator will duplicate the scene (but not
the script!) when duplicating the NO_EDIT template. This will make it easier for
the team creating a StoryQuest to adjust the object scene as needed to match
their assets without affecting the broader game.

(In practice not every team has replaced the generic object sprite, so in this
case we will have a duplicate scene for no upside. But scenes are small!)

See https://github.com/endlessm/threadbare/discussions/1150 for discussion of
this general approach.

---
  - https://app.asana.com/0/0/1211209074317274


addons/storyquest_bootstrap/copier.gd
wjt, line 149: With Godot 4.5.1 from Flathub the engine segfaults at this line in the case where `scene` referenced another scene. I can't reproduce it with a local build of the development branch with debug symbols enabled, but I don't know whether this is because some change between 4.5.1 and 4.6-rc2 has fixed whatever the underlying engine bug is, or because I am building with a different toolchain in a way that masks the bug.

I have tried to boil this down to a smaller, reproducible test case, and found some problems along the way, but ultimately I have spent too long on this and have to set it aside. I'm pretty sure my logic above is correct and the engine is buggy (at some level that is obvious because GDScript should not be able to cause a segfault in the engine).

Issue #1830, opened 2026-01-15
Add exported properties for sound effects where contestants may wish to add them in Champ


wjt:
@felixwalberg you mentioned a number of places that you imagined a competitor might add sound effects. If these places do not already have a good way to drop in a sound effect, can you list some such places here, and we can collaborate on making it easy to add SFX?

PR #1851, opened 2026-01-22
Champ learner assets

felixwalberg:
This PR addresses some of the additions for learner contributions outlined in the Champy GDD. 
- AudioStreamPlayer nodes in each scene for learners to add background music
- Renaming of nodes to match instructions for changing/inserting art pngs
- Graying out obstacles in combat level to imply the need for learner contributions
- Replacing audio file path for sequence puzzle rock object (there might be a better way to set this up)

### DIFF
--- a/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_object.tscn
+++ b/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_object.tscn
@@ -2,7 +2,6 @@
 
 [ext_resource type="Script" uid="uid://br3of3tj8lb8p" path="res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_object.gd" id="1_flhem"]
 [ext_resource type="SpriteFrames" uid="uid://c2m3tplriqe2s" path="res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_object.tres" id="2_as5uj"]
-[ext_resource type="AudioStream" uid="uid://cg57q82pb243w" path="res://assets/third_party/nepalese_hand_bells/handBells-c4.ogg" id="3_2vrsm"]
 [ext_resource type="Script" uid="uid://du8wfijr35r35" path="res://scenes/game_elements/props/interact_area/components/interact_area.gd" id="4_g5jna"]
 [ext_resource type="AudioStream" uid="uid://b1qhmlf8hcbnv" path="res://scenes/quests/story_quests/champ/3_sequence_puzzle/champ_rock_sound.ogg" id="5_as5uj"]
 
@@ -44,7 +43,7 @@ debug_color = Color(0.600391, 0.54335, 0, 0.42)
 
 [node name="AudioStreamPlayer2D" type="AudioStreamPlayer2D" parent="." unique_id=661787463]
 unique_name_in_owner = true
-stream = ExtResource("3_2vrsm")
+stream = ExtResource("5_as5uj")
 bus = &"SFX"
 
 [connection signal="interaction_started" from="InteractArea" to="." method="_on_interaction_started"]


scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
```diff
@@ -114,93 +112,123 @@ position = Vector2(356, 453)
 
 [node name="Top" parent="OnTheGround/SequencePuzzle/Objects" instance=ExtResource("5_tf43i")]
 position = Vector2(190, -194)
+audio_stream = ExtResource("7_mtlrl")
```
wjt: This shouldn't be necessary because you've also changed it in the scene that this is instancing!

scenes/quests/story_quests/champ/0_intro/champ_intro.tscn
```diff
@@ -102,6 +103,10 @@ _data = {
 
 [node name="Intro" type="Node2D"]
 
+[node name="Background Music" type="AudioStreamPlayer2D" parent="."]
+stream = ExtResource("1_m7ukv")
+autoplay = true
```
wjt: We have a specific node for background music. Please make these instances of `scenes/game_elements/props/background_music/background_music.tscn`. This node handles cross-fading between levels. It also makes sure that the music is in the correct channel so that it responds to the music volume control in settings.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle_object.tscn
wjt: If you change it here as well, on the root node of `champ_sequence_puzzle_object.tscn`, then you won't need to re-set it in every scene that uses it.

scenes/quests/story_quests/champ/2_combat/champ_combat.tscn
```diff
@@ -481,3 +481,7 @@ frame_progress = 0.7079184
 
 [connection signal="cinematic_finished" from="Cinematic" to="FillGameLogic" method="start"]
 [connection signal="goal_reached" from="FillGameLogic" to="OnTheGround/CollectibleItem" method="reveal"]
+
+[editable path="OnTheGround/Objects/Object1"]
+[editable path="OnTheGround/Objects/Object2"]
+[editable path="OnTheGround/Objects/Object3"]
```
wjt: Why do you need editable children here? What properties are expected to be modified? Could they be exported from `buoy.tscn` instead? Or can they be edited in `buoy.tscn` for all instances?

(These are all just questions.)

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_long_rock.gd
```diff
@@ -4,6 +4,12 @@ extends Node2D
 @onready var rock: AnimatedSprite2D = $AnimatedSprite2D
 @onready var area_2d: Area2D = $AnimatedSprite2D/Area2D
 
+## Audio Stream Player for rock sound effects
+@export var audio_player: AudioStreamPlayer2D
```
wjt: I think you should add this to the `champ_long_rock` scene and assign it with `$` as with the other nodes above.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_long_rock.gd
```diff
@@ -23,15 +29,27 @@ func set_submerged(is_submerged: bool) -> void:
 
 
 func _ready() -> void:
+	if not audio_player:
+		audio_player = AudioStreamPlayer2D.new()
+	if not splash_sound:
+		print("Defauklt")
+		audio_player.stream = preload("res://assets/third_party/sounds/collectibles/MemoryCollectable.ogg")
```
wjt: You should make this the default value of the `splash_sound` property, and then unconditionally set it on the audio_player node.

wjt: Or alternatively default to no sound. `AudioStreamPlayer2D.play()` does the right thing if no stream is set (I think).

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
```diff
@@ -50,48 +50,58 @@ tile_set = ExtResource("2_6wj4q")
 [node name="LongRocks" type="Node2D" parent="."]
 
 [node name="ChampLongRock" parent="LongRocks" instance=ExtResource("3_bvcb0")]
+unique_name_in_owner = true
```
wjt: Why do these need unique names?

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
```diff
@@ -291,6 +301,8 @@ position = Vector2(1442, 368)
 dialogue = ExtResource("19_27hsp")
 sprite_frames = ExtResource("20_ey61o")
 
+[node name="Sound Effects" type="AudioStreamPlayer2D" parent="."]
```
wjt: What is this one for?

When you add new `AudioStreamPlayer(2D)` nodes please add them to the appropriate bus. In this case, `SFX`.

wjt: Image

scenes/quests/story_quests/champ/2_combat/champ_combat.tscn
felixwalberg: I will remove these. It would make more sense to direct the learner to the `buoy.tscn` scene where they can replace the sprite frames for all instances. If they want to add more variation, they could copy the entire scene and give it different sprite frames.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
felixwalberg: This node no longer seems necessary, so I removed it. The long rock audio player is now linked to the `SFX` bus.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.tscn
felixwalberg: These should no longer be unique.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_long_rock.gd
felixwalberg: I can completely remove the `splash_sound` property. As you mentioned, `AudioStreamPlayer2D.play()` will not break when there is no sound, and then the learners can just set this property in the inspector.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_long_rock.gd
felixwalberg: Done!

felixwalberg:
I think this could also satisfy #1830 unless you think more things should be exported rather than nodes in the scenes. The other additions to sound effects in the stealth mini game will come after more changes have been made to that scene (to avoid conflicts!).

wjt (approved):

PR #1860, opened 2026-01-23
Allow quests to contain 0-6 threads, not always 3

### LINKED ISSUE #1234: Allow quests to contain 0, 1, or 2 collectibles rather than 3
Currently, all quests have the player collect exactly three threads – memory, imagination, and spirit – which they take back to the Eternal Loom in Fray's End.

We would like to change this to allow quests to have fewer threads, for 2 reasons:

* Storytelling: A tale about a grandparent in WWI might invoke only Memory an Spirit, while a tale about the [Cailleach](https://en.wikipedia.org/wiki/Cailleach) might invoke Imagination only.
* Contributing: We want people to be able to contribute shorter quests, to reduce the workload & further increase the quality.

It should be possible to collect 2 or more of the same kind of thread (e.g. two or three Memory). In principle it should be possible to have a 4- or more-thread quest, though I don't think this is so high a priority

Right now we hardcode the 3-thread requirement in a few places:

- [x] The HUD assumes that exactly three items will be collected (of any type)
- [x] The Eternal Loom only allows you to drop off threads [if you have collected at least 3  threads](https://github.com/endlessm/threadbare/blob/dc0ae5f15bff7ac57e0341f3a579414d429434dd/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd#L54-L55) (of any type)
- [x] The game's saved state does not save which threads (M, I, S) have been collected, only the number of threads collected. When restoring the state, if you have (say) 2 threads, it restores Memory & Imagination
- [ ] #1861
- [ ] The Sokoban puzzle sets are a progression with 1, then 2, then 3 threads

### XP Summary (total 500)

| Subdomain | Skill | XP |
| - | - | - |

wjt:
Previously we assumed that all quests required the player to collect exactly 3 threads. This has proven to be limiting in practice: some stories are best told with fewer threads, other more.

Add a field to `quest.gd` to customise this. Make the HUD react to this field. Guess the quest when deep-linking to a scene (via URL hash, or F6 in editor) so that the HUD can rely on `GameState.current_quest` being set. 

Resolves #1234


scenes/quests/lore_quests/quest_003/quest.tres
```diff
@@ -7,4 +7,5 @@ script = ExtResource("1_onpv7")
 title = "Placeholder"
 description = "A quest to hold scenes that don't yet have a permanent home."
 first_scene = "uid://d2ejk3qrh0fo3"
+threads_to_collect = 2
```
manuq, line 10: The dialogue removed in this 1st commit was granting 2 threads (imagination, spirit). But yes I see that it wasn't updated when the ink_combat_round_6.tscn added a 4th thread. So the threads to collect are indeed 2.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -70,4 +70,8 @@ func on_offering_succeeded() -> void:
 
 
 func is_item_offering_possible() -> bool:
-	return GameState.items_collected().size() >= InventoryItem.ItemType.size()
+	return (
+		GameState.current_quest
+		and GameState.current_quest.threads_to_collect > 0
+		and GameState.items_collected().size() >= GameState.current_quest.threads_to_collect
```
manuq, line 76: 👍

PR #1865, opened 2026-01-26
InkCombatRound6: Implement Sequential Filling Mechanic

### LINKED ISSUE #1847: InkCombatRound6: Implement Sequential Filling Logic
Description: This task covers the implementation of the core mechanic scripts for the Sequential Filling in Level 6.

Requirements:

- Create logic to randomize the order of the vases at the start of the level.

- Ensure only one vase is "Active" at a time. All others must be "Locked".

- Upon filling the active vase, immediately unlock/illuminate the next vase in the generated sequence.

russell-cj:
This implements the **Sequential Combo** mechanic for **Level 6**. Barrels must now be filled in a specific order.

- **Visuals:** Added a glow animation to the barrel to indicate the currently active target.
- **Barrel Logic:** Barrels can now be locked. When locked (gray state), they do not receive ink.
- **New Mechanic:** Introduced the `BarrelUnlockSequence` component:
  - Optionally, shuffles the barrel order when the level starts.
  - Keeps all barrels locked except one.
  - Automatically unlocks the next barrel when the current one is completed.

Resolves  #1847

### DIFF
--- a/scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
+++ b/scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
@@ -18,6 +18,7 @@ signal completed
 
 const DEFAULT_SPRITE_FRAMES: SpriteFrames = preload("uid://dlsq0ke41s1yh")
 const FILLING_NAME_ANIMATION: StringName = &"filling"
+const LOCKED_COLOR := Color(0.5, 0.5, 0.5, 1)
 
 ## Determines how the FillingBarrel looks.[br]
 ## It is required that the [member sprite_frames] have an animation called filling.[br]
@@ -37,11 +38,13 @@ const FILLING_NAME_ANIMATION: StringName = &"filling"
 @export var color: Color:
 	set = _set_color
 
-var is_locked: bool = false
+## Controls whether the barrel can receive ink. When locked, it turns gray and ignores hits.
+var is_locked: bool = false:
+	set = set_is_locked
 
 var _amount: int = 0
 
-@onready var barrel_glow: AnimatedSprite2D = $BarrelGlow
+@onready var barrel_glow: AnimatedSprite2D = get_node_or_null("BarrelGlow")
 @onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
 @onready var animation_player: AnimationPlayer = %AnimationPlayer
 @onready var collision_shape_2d: CollisionShape2D = %CollisionShape2D
@@ -62,8 +65,7 @@ func _set_sprite_frames(new_sprite_frames: SpriteFrames) -> void:
 	sprite_frames = new_sprite_frames if new_sprite_frames else DEFAULT_SPRITE_FRAMES
 	if not is_node_ready():
 		return
-	if animated_sprite_2d:
-		animated_sprite_2d.sprite_frames = sprite_frames
+	animated_sprite_2d.sprite_frames = sprite_frames
 	update_configuration_warnings()
 
 
@@ -77,15 +79,15 @@ func _ready() -> void:
 		barrel_glow.visible = false
 
 
-func set_locked_state(locked: bool) -> void:
+func set_is_locked(locked: bool) -> void:
 	is_locked = locked
 
 	if is_locked:
-		modulate = Color(0.5, 0.5, 0.5, 1)
+		modulate = LOCKED_COLOR
 		if barrel_glow:
 			barrel_glow.visible = false
 	else:
-		modulate = Color(1, 1, 1, 1)
+		modulate = Color.WHITE
 		if barrel_glow:
 			barrel_glow.visible = true
 			barrel_glow.play("glowing")
--- a/scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
+++ b/scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
@@ -5,30 +5,30 @@ extends Node
 
 @export var barrels: Array[FillingBarrel]
 @export var auto_start: bool = true
+@export var randomize_barrel_order: bool = false
 
 var current_target_index: int = 0
 
 
 func _ready() -> void:
 	if auto_start:
-		call_deferred("start_sequence")
+		start_sequence.call_deferred()
 
 
 func start_sequence() -> void:
-	randomize()
-
 	if barrels.is_empty():
 		push_warning("BarrelUnlockSequence: No barrels assigned.")
 		return
 
-	barrels.shuffle()
+	if randomize_barrel_order:
+		barrels.shuffle()
 
 	for barrel in barrels:
 		if not barrel.completed.is_connected(_on_barrel_completed):
 			barrel.completed.connect(_on_barrel_completed)
 
 		# Initial state: all locked
-		barrel.set_locked_state(true)
+		barrel.is_locked = true
 
 	unlock_next_barrel()
 
@@ -36,9 +36,7 @@ func start_sequence() -> void:
 func unlock_next_barrel() -> void:
 	if current_target_index < barrels.size():
 		var target: FillingBarrel = barrels[current_target_index]
-
-		if is_instance_valid(target):
-			target.set_locked_state(false)
+		target.is_locked = false
 
 
 func _on_barrel_completed() -> void:


russell-cj:
You’re absolutely right, @wjt. Originally, Level 6 was focused only on the sequential barrel mechanic, but later Mothsache was added to test it, which significantly increased the difficulty.

In fact, I had an idea that I didn’t get a chance to share before the internship ended. I’ll open a discussion and share the idea I had in mind with you.

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -37,8 +37,11 @@ const FILLING_NAME_ANIMATION: StringName = &"filling"
 @export var color: Color:
 	set = _set_color
 
+var is_locked: bool = false
```
wjt: Please document this field.

wjt: Also make the `set_locked_state` function a setter for this field:


```suggestion
var is_locked: bool = false:
	set = set_is_locked
```

(I suggest renaming the method below.)

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -70,10 +73,29 @@ func _ready() -> void:
 	animated_sprite_2d.animation = FILLING_NAME_ANIMATION
 	animated_sprite_2d.frame = 0
 
+	if barrel_glow:
+		barrel_glow.visible = false
+
+
+func set_locked_state(locked: bool) -> void:
```
wjt:
```suggestion
func set_is_locked(locked: bool) -> void:
```

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -70,10 +73,29 @@ func _ready() -> void:
...
+	is_locked = locked
+
+	if is_locked:
+		modulate = Color(0.5, 0.5, 0.5, 1)
+		if barrel_glow:
+			barrel_glow.visible = false
+	else:
+		modulate = Color(1, 1, 1, 1)
```
wjt:
```suggestion
		modulate = Color.WHITE
```

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
```diff
@@ -0,0 +1,46 @@
...
+
+@export var barrels: Array[FillingBarrel]
+@export var auto_start: bool = true
+
+var current_target_index: int = 0
+
+
+func _ready() -> void:
+	if auto_start:
+		call_deferred("start_sequence")
```
wjt:
```suggestion
		start_sequence.call_deferred()
```

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
```diff
@@ -0,0 +1,46 @@
...
+
+
+func start_sequence() -> void:
+	randomize()
+
```
wjt:
```suggestion
```

https://docs.godotengine.org/en/stable/tutorials/math/random_number_generation.html#the-randomize-method says:

> Since Godot 4.0, the random seed is automatically set to a random value when the project starts. This means you don't need to call randomize() in _ready() anymore to ensure that results are random across project runs. However, you can still use randomize() if you want to use a specific seed number, or generate it using a different method.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
wjt: But why do you need to defer this?

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
```diff
@@ -0,0 +1,46 @@
...
+	unlock_next_barrel()
+
+
+func unlock_next_barrel() -> void:
+	if current_target_index < barrels.size():
+		var target: FillingBarrel = barrels[current_target_index]
+
+		if is_instance_valid(target):
+			target.set_locked_state(false)
```
wjt: If you add the `set = ...` syntax I suggest above, then this can be:

```suggestion
			target.is_locked = false
```

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
```diff
@@ -0,0 +1,46 @@
...
+
+
+func start_sequence() -> void:
+	randomize()
+
+	if barrels.is_empty():
+		push_warning("BarrelUnlockSequence: No barrels assigned.")
+		return
+
+	barrels.shuffle()
```
wjt: I don't think the order should necessarily be shuffled. The order of the barrels is part of the level design - for example, you could imagine designing a level where it would be impossible to get across a bridge until the first barrel is filled.

Another way of looking at this is: if the order is random each time the level is played, then why do I (the level designer) have to manually assign the barrels to this node, rather than this component discovering the barrels in the scene by itself?

https://github.com/endlessm/threadbare/blob/29424894f3b4c09282415fbdba354f52b42a1b01/scenes/game_logic/fill_game_logic.gd#L42

Some ideas:

1. Remove the `shuffle()` call
2. Keep it, but add an `@export var randomize_barrel_order: bool` variable (adjusting the default value to taste :) )
3. If you think the mechanic is best enjoyed with a random order: keep the shuffle, but remove the `@export` from `var barrels` and instead populate that array in `_ready` in the same way that `fill_game_logic.gd` does

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
```diff
@@ -0,0 +1,46 @@
...
+		barrel.set_locked_state(true)
+
+	unlock_next_barrel()
+
+
+func unlock_next_barrel() -> void:
+	if current_target_index < barrels.size():
+		var target: FillingBarrel = barrels[current_target_index]
+
+		if is_instance_valid(target):
```
wjt: I'm always a bit suspicious of `is_instance_valid` calls because I think they tend to mask bugs. In what situation would you expect a barrel to be freed from the scene before it has been enabled? Or is there another case I have not considered which would cause an element of the array to be freed?

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -70,10 +73,29 @@ func _ready() -> void:
...
 
+	if barrel_glow:
+		barrel_glow.visible = false
+
+
+func set_locked_state(locked: bool) -> void:
+	is_locked = locked
+
+	if is_locked:
+		modulate = Color(0.5, 0.5, 0.5, 1)
```
wjt: Perhaps define a `const LOCKED_COLOR := Color(0.5, 0.5, 0.5, 1)` constant for this?

scenes/game_elements/props/filling_barrel/components/filling_barrel.gd
```diff
@@ -70,10 +73,29 @@ func _ready() -> void:
 	animated_sprite_2d.animation = FILLING_NAME_ANIMATION
 	animated_sprite_2d.frame = 0
 
+	if barrel_glow:
```
wjt, line 78: You have this line above:

```gd
@onready var barrel_glow: AnimatedSprite2D = $BarrelGlow
```

`$x` is a shorthand for `get_node("x")`. [docs](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#:~:text=There%20are%20also%20two%20constructs%20that%20look%20like%20literals%2C%20but%20actually%20are%20not%3A).
[get_node()](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-get-node) raises an error if the node does not exist.

You should do one of the following two things:

1. In the `@onready ...` line, replace `$BarrelGlow` with `get_node_or_null("BarrelGlow")`, which does not raise an error if the node does not exist https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-get-node-or-null
2. Remove all the `if barrel_glow:` checks - by using `$BarrelGlow` we are requiring that the scene that this script is attached to has a node of that name.

(You may ask, why does `_set_sprite_frames` have a similar `if animated_sprite_2d:` check? Good question! One answer is: setters can run before a node is ready, so `@onready` variables may not have been populated yet. But then you might say: but doesn't that setter check `if not is_node_ready():` first? Yes it does. The `if animated_sprite_2d:` check is redundant.)

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
russell-cj: I think I understand your point of view.

scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/components/barrel_unlock_sequence.gd
russell-cj: I used `call_deferred` because, at the time, I thought it was the best alternative to fix an issue where the glow was disappearing. Deferring the execution solved that bug for me.

wjt (approved):

PR #1869, opened 2026-01-30
throwing_enemy: Fix property grouping

wjt:
The intention is for all projectile_ properties to be grouped together
in a "Projectile" group via:

```gdscript
@export_group("Projectile", "projectile")
```

The second argument is a prefix to strip from properties in the group,
so `projectile_follows_player` becomes "Follows Player". However, per its documentation:

> The grouping will break on the first property that doesn't have a
> prefix.

In commit 117e713d7220f7fc629c567623bfe2e9e19c309a, `var distance` was
added right below this `@export_group`. `distance` does not have the
prefix, so the group ends.

Move the `distance` property out of the group section of the file. Move
the `projectile_scene` property into that group.


manuq:
Thanks @jgbourque for finding the issue.

PR #1879, opened 2026-02-04
void runner: Fix forest barrier collision layers

wjt:
The dense forests are now StaticBody2D nodes with a collision shape that
prevents the player from trying to walk between the trees at all, even
if the geometry of the trees would let them.

However the collision layer & mask were left at the default values:
layer 1, i.e. the "player" physics layer.

Also weirdly, SanctuaryWoods was a bare Node2D with the trees and a
StaticBody2D as a child. I looked back through the git history and it
was always a Node2D. I think what happened is: I planted the trees with
an Area2D + AreaFiller, then for some reason I removed those. Then we
later added the barrier.

Make the SanctuaryWoods be the StaticBody2D, rather than the trees being
siblings of the StaticBody2D.

Update this and another forest to use the correct collision layer:
non_walkable_floor. (There is an argument that they are in fact walls
but non_walkable_floor is what the other barriers in this specific level
e.)

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
@@ -411,6 +411,8 @@ sprite_frames = ExtResource("13_b5w51")
 
 [node name="Forest" type="StaticBody2D" parent="OnTheGround" unique_id=1750666820]
 y_sort_enabled = true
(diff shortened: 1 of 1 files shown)


manuq (approved):
Great catch. Whoa how easy it is to get these things wrong, reviewing textual diffs of scene files is prone to errors.

PR #1880, opened 2026-02-04
Guards: Use walk behaviors

### LINKED ISSUE #1265: Guard: Integrate with character behavior

manuq:
Refactor the guards to use walk behaviors.

To maintain the existing implementation as much as possible, add a new PointsWalkBehavior. In a PointsWalkBehavior, the character moves in straight lines between the points of a Path2D.

Use a PointsWalkBehavior for the patrolling state.

Use another PointsWalkBehavior for the returning state. This one is created dynamically from the breadcrumbs dropped while investigating.

Use a FollowWalkBehavior for the investigating state. And add a helper Node2D that is repositioned to the player's last seen position. The Node2D is used as target for the FollowWalkBehavior.

Use a Timer node for the waiting state.

Note: The detecting state is a bit confusing because it uses the patrolling walk behavior until the player is out of sight for the first time, and the investigating walk behavior for the rest. In the future this "detecting" state could be merged into both patrolling and investigating.

Fix https://github.com/endlessm/threadbare/issues/1265


manuq:
Closing, one more attempt on refactoring this failed due to more important things to do. I'll keep the branch.

Issue #1868, opened 2026-01-29
Champ sequence puzzle: repeatedly interacting with second sign breaks camera


Fauche-r:
### Description

On the second half of the puzzle in champ sequence puzzle , if you keep on interacting with the sign, you stay zoomed out, and keep on shifting upward and to the right.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

felixwalberg:
Addressed in #1888

PR #1888, opened 2026-02-05
Champ puzzle sequence : Hint sign

Fauche-r:
- changed behavior of the cinematic, in order to have it well centered
- fixed a bug in which you could interact repeatedly with the sign and have the camera break out of the boundary of the map

Addresses  Issue #1868 and part of feedback on #1805 pull request for camera behavior

### DIFF
--- a/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
+++ b/scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
@@ -3,10 +3,6 @@
 class_name ChampSequencePuzzle
 extends SequencePuzzle
 
-const CAMERA_MAX_OFFSET_Y: int = 475
-const CAMERA_OFFSET_Y: int = 25
-const CAMERA_TIMEOUT_INTERVAL: float = 0.03
-const CAMERA_ZOOM_SCALAR: float = 0.005
 const CHECKPOINT_POS: Vector2 = Vector2(1400,180)
 const RESPAWN_DELAY: float = 1.5
 const ROCK_WIDTH: int = 3


scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
```diff
@@ -163,9 +160,10 @@ func reset_all() -> void:
...
-		await get_tree().create_timer(CAMERA_TIMEOUT_INTERVAL).timeout
-		camera.offset.y = camera.offset.y - CAMERA_OFFSET_Y
-		camera.zoom.y = camera.zoom.y - CAMERA_ZOOM_SCALAR
-		camera.zoom.x = camera.zoom.x - CAMERA_ZOOM_SCALAR
+	player._set_mode(player.Mode.DEFEATED)
+	camera.global_position = $Objects/Middle8.global_position
+
+
+func _on_hint_sign_1_demonstrate_sequence() -> void:
+	player._set_mode(player.Mode.DEFEATED)
```
felixwalberg: We had talked about how it looks a little funny to have the defeated animation, particularly for this first sequence hint sign, since the player is fully visible. If we wanted to, we could replace this line with: 
```
player._toggle_player_behavior(player.player_interaction, false)
```
then in the sequence completed signal (in place of line 142), we could follow up with:
```	
player._toggle_player_behavior(player.player_interaction, true)
```

Not necessarily something we need to change but I wanted to leave it here as an idea. It might also be a better way to approach restricting the player's inputs when they select the wrong rock and are standing on a submerged one (before respawning). Either that, or we could leave it as DEFEATED and have the learners create an animation of the character splashing in the water or something along those lines.

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
```diff
@@ -136,13 +137,9 @@ func _on_champ_long_rock_water_entered() -> void:
...
 func _on_hint_sign_hint_sequence_finished() -> void:
 	await get_tree().create_timer(RESPAWN_DELAY).timeout
-	if _current_step == 1:
-		var goal: float = camera.offset.y + CAMERA_MAX_OFFSET_Y
-		while camera.offset.y < goal:
-			await get_tree().create_timer(CAMERA_TIMEOUT_INTERVAL).timeout
-			camera.offset.y = camera.offset.y + CAMERA_OFFSET_Y
-			camera.zoom.y = camera.zoom.y + CAMERA_ZOOM_SCALAR
-			camera.zoom.x = camera.zoom.x + CAMERA_ZOOM_SCALAR
+	#if _current_step == 1:
```
felixwalberg:
```suggestion

```

scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
Fauche-r: I think for now we leave it as DEFEATED for when the player select the wrong rock, as it would be a nice way to have learner find a way to create an animation that happen when something else happens (in contrast to the one animation of champ that is at the end of the level, that happen no matter what). I have however implemented the change of behavior of sign when interacted with, with the solution you have suggested.

felixwalberg (approved):

Issue #1874, opened 2026-02-03
Fill game: Allow varying projectile (appearance) by label


wjt:
Currently the fill game (or rather the throwing enemy) has a few different ways to vary the projectiles that are thrown:

- Each enemy can throw a different projectile scene (and apply various properties to it)
- Projectiles can be recolored (via modulate) to match the colour of a barrel. Barrels have an optional label and optional colour. It is assumed that all barrels in the scene that have the same label also have the same colour.

Projectiles can only fill barrels with a matching label. When spawning a projectile, the enemy assigns it a label (and corresponding colour, if any) out of the labels for the remaining barrels in the scene.

In two quests, the creators have wanted to do something that is not supported by this:

1. In Stella, the idea was that four kinds of sea creature fly out of the water: 3 different regular fish, or a starfish. Only the starfish should be accepted by the boat (barrel). This was not possible without custom logic. One thing we tried was having offscreen barrels with 3 different labels but what's missing is the ability to use different spriteframes (or scenes) for different labels.
2. In Champ, there is a sign on the shore that tells you which colour of fish needs to go into each of 3 targets. There is custom logic to swap the `SpriteFrames` of the projectile based on the target label.

wjt:
It does make for a lot of duplication to have 3 scenes in champ which are identical except for the colour of the fish. Actually I think the modulate logic would be fine for that scene - make the fish greyscale, modulate for the different targets - but if a learner wanted to make the fish more different, then it would be the stella case again. (@felixwalberg I guess that's why you added this custom logic rather than modulating a single white fish.)

Issue #1873, opened 2026-02-03
Change throwing enemy projectile customisation


wjt:
Currently the pattern we use is that on each `throwing_enemy.tscn` instance, we override properties like `projectile_sprite_frames`, `projectile_hit_sound_stream`, etc. These are all in turn assigned to the instanced projectile.

Since commit 402f2abe0597f8a38c70b6ee2a963c81340e985b it has been possible to override the projectile scene that is instanced. A logical further improvement is to remove those properties from the `throwing_enemy.gd` script. Instead, to customise these properties, a separate scene should be created. For example, the inkdrinkers would throw an `ink_blob_projectile.tscn` scene.

(This is the approach taken in https://github.com/endlessm/threadbare/pull/1308 but we didn't yet wire up this behaviour outside the test scene. Removing the other properties from `throwing_enemy.gd`, and having custom projectile scenes, is a further step in that direction.)

PR #1890, opened 2026-02-06
Rework how projectiles are customised

wjt:
Previously, all instances of the "fill game" used the same Projectile
scene. To customise its appearance, the ThrowingEnemy scene had a whole
series of properties that would be assigned each instantiated
projectile, in addition to assigning a label & optional modulate color
based on the target barrel.

This is not a very "Godot-ish" way to do it. Conceptually,
different projectiles should be different scenes that have similar
structures and reuse common scripts. Instead, we have had to manually
re-export properties from deep inside the projectile scene, and
propagate their values from the ThrowingEnemy scene.

Since commit 402f2abe0597f8a38c70b6ee2a963c81340e985b
("ThrowingEnemy: Export the projectile scene") it has been possible to
customise which projectile scene is instantiated by any specific enemy,
but this isn't enough either for Stella or for Champ. In Stella, we
wanted the invisible enemies to throw, each time, a random choice
between a starfish (which matches the ship "barrel") or one of three
fish (which don't). But this wasn't possible. Similarly in Champ we want
the three different colours of fish to potentially be different sprites;
they currently happen to be recoloured versions of the same asset that
could be done with the projectile modulate logic but we don't want to
limit contestants to this.

So:

1. For each StoryQuest that previously reused the generic projectile
   scene, create one (or more!) scenes specific to the quest. This does
   introduce some duplication but I decided that the scenes are small
   enough that using inherited scenes is more trouble than it's worth.

2. Similarly, make a dedicated ink blob projectile scene for the core
   game.

3. Move the generic projectile scene to the NO_EDIT template, so that it
   will be duplicated when creating a new StoryQuest.

4. Remove the sprite_frames and hit_sound_stream properties from
   projectile.gd: these can be set by modifying the scene.

5. Remove the projectile_sprite_frames, projectile_hit_sound_stream,
   projectile_small_fx_scene, projectile_big_fx_scene, and
   projectile_trail_fx_scene properties from throwing_enemy.gd: these
   can be set by modifying the projectile scene that it spawns.

6. Add the ability to spawn different projectile scenes based on the
   target label. This is adapted from similar code in the Champ quest to
   use different sprite frames per label. Use this in Champ and Stella.
   In both cases define a custom throwing enemy scene, this time as an
   inherited scene: there are huge animations in the enemy scene so in
   this case I thought this was worth it.

7. Break the projectile-spawning code up a little. Having done this,
   simplify some of the StoryQuest-specific projectile customisations.

Resolves https://github.com/endlessm/threadbare/issues/1873
Resolves https://github.com/endlessm/threadbare/issues/1874


scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
```diff
@@ -65,8 +70,25 @@ const WALK_TARGET_SKIP_RANGE: float = 0.25
...
+	set(new_value):
+		projectile_scene = new_value
+		update_configuration_warnings()
+
+## Alternative projectile scenes to spawn based on the target label.
+## [br][br]
+## When launching a projectile, if the chosen label is found in this dictionary,
+## the associated scene will be used. If not, the default [member projectile_scene]
+## will be used.
+@export var projectile_scene_for_label: Dictionary[String, PackedScene]:
```
wjt, line 88: This is a "colo(u)r of bikeshed" situation but maybe this should be `projectile_scene_per_label`. I want `projectile_scene` to be at the start of the property name because all properties in the `Projectile` group must begin with `projectile_` for their names to be shortened in the inspector (#1869).

scenes/quests/story_quests/champ/2_combat/champ_combat.tscn
```diff
@@ -24,79 +20,6 @@
...
-extends FillGameLogic
-#
-#@export var lab: String = \"\"
-#
-#func _ready() -> void:
-	#var color_per_label = {\"a\": Color(.3,.5,6), \"b\": Color(.9,.1,6),\"c\": Color(.1,.8,.2)}
-	#for enemy: ThrowingEnemy in get_tree().get_nodes_in_group(\"throwing_enemy\"):
-		#enemy.allowed_labels = allowed_labels
-		#enemy.color_per_label = color_per_label
-"
```
wjt, line 40: This embedded script doesn't do anything so I removed it in passing.

scenes/quests/story_quests/champ/2_combat/champ_combat.tscn
```diff
@@ -316,36 +239,13 @@ position_smoothing_enabled = true
 editor_draw_limits = true
 
 [node name="ThrowingEnemies" type="Node2D" parent="OnTheGround" unique_id=1014610845]
-visible = false
```
wjt, line 324: I moved this to the `AnimatedSprite2D` inside the enemy scene.

scenes/quests/story_quests/champ/2_combat/combat_components/champ_throwing_enemy.gd
```diff
@@ -1,40 +0,0 @@
...
-	var projectile: Projectile = projectile_scene.instantiate()
-	projectile.direction = projectile_marker.global_position.direction_to(player.global_position)
-	scale.x = 1 if projectile.direction.x < 0 else -1
-	projectile.label = allowed_labels.pick_random()
-	if projectile.label in project_sprite_frames:
-		projectile.sprite_frames = project_sprite_frames[projectile.label]
-	else:
-		# Default to original projectile sprite frames
-		projectile.sprite_frames = preload("uid://b00dcfe4dtvkh")
-	if projectile.label in color_per_label:
```
wjt, line 27: @felixwalberg I like this design. I moved the spirit of this into `throwing_enemy.gd` - but selecting a per-label *scene* not SpriteFrames.

scenes/quests/template_quests/NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_projectile.tscn
wjt, line 1: To do:

1. Resave every scene that references this scene to update the path
2. Test that this this actually get duplicated & relinked correctly by the tool - I assume so but actually this will be the first PackedScene property.

wjt: 2. It does work, though I had to fix #1894.

wjt:
I gave a brief screenshare demo of this change to @jgbourque. Even if we think this is an improvement for learners building StoryQuests, the issue is that this will mean that some guides & videos are now out of date. And we cannot afford to re-record every video for every change to the game.

We agreed that it's not necessary. @jgbourque's plan is to add some general note to this effect, and also explicitly call out to learners that we are gradually changing the customisation model to be "edit a scene" not "change properties on a shared scene". So I think from a learning team perspective we can move ahead with this change.

scenes/quests/story_quests/wizzy_quest/2_combat/combat_components/wizzy_throwing_enemy.gd
```diff
@@ -1,105 +1,36 @@
 # SPDX-FileCopyrightText: 2025 Game-Lab-5-0-UTP-Group-1-Team-1 Contributors
 # SPDX-License-Identifier: MPL-2.0
-
+@tool
```
wjt, line 3: Without this Godot warns that a non-`@tool` script extends a `@tool` script.

scenes/quests/story_quests/wizzy_quest/2_combat/combat_components/wizzy_projectile.gd
```diff
@@ -116,11 +113,7 @@ func _stop_projectile_motion() -> void:
 
 ## Changes sprite to disabled appearance
 func _apply_disabled_visual() -> void:
-	if disabled_sprite_frames:
-		animated_sprite_2d.sprite_frames = disabled_sprite_frames
```
wjt, line 120: I found that in the one place this is used, the two `SpriteFrames` resources are the same.

PR #1900, opened 2026-02-09
Make volume sliders control volume_linear

wjt:
Previously the volume sliders adjusted volume_db linearly in the range
-30 dB to 0 dB, with a special case to mute the audio at -30 dB. The dB
scale is logarithmic, not linear.

Godot provides linear_to_db() and db_to_linear(), which map linear
values in the range 0.0 to 1.0 to/from the logarithmic range -inf dB
(silence) to 0 dB, exactly to implement volume sliders. AudioServer
provides convenience methods to set and get linear volume, implemented
in terms of linear_to_db() and db_to_linear().

Connect the volume sliders to adjust the linear volume. Adjust the range
to 0.0 to 1.0 to match. Remove the mute special case: setting the linear
volume to 0.0 is the same as setting the decibel volume to -inf.

Since players will have out of range values saved if they have ever
adjusted the volumes, add versioning to the settings file, and discard
any previous settings. This will reset the volume levels to the
defaults. There are no other settings stored in the file at present.

While we're here: add ticks to the volume sliders.

### DIFF
--- a/scenes/globals/settings/settings.gd
+++ b/scenes/globals/settings/settings.gd
@@ -4,10 +4,15 @@ extends Node
 
 const SETTINGS_PATH := "user://settings.cfg"
 
+const META_SECTION := "Meta"
+const VERSION_KEY := "Version"
+const VERSION := 1
+
 const VOLUME_SECTION := "Volume"
-const MIN_VOLUME := -30.0
+const MIN_VOLUME := 0.0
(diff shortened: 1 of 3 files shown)


manuq (approved):
Nice improvement. I like the versioning to be used for clearing the old settings, rather than trying to keep multiple implementations working.

PR #1901, opened 2026-02-09
Added blink movement ability in the champ quest stealth mini-game

### LINKED ISSUE #1892: Create a more advanced "ability" to traverse the level
This could be some kind of power-up, speed, sprinting ability, or the "blink" method defined by Justin. We want this to be loosely defined in the `champ_incomplete_character.gd` script with some of the logic either commented out or written in natural language.

Blink method:
func _apply_blink():
    # Step 1: Check if the blink key is pressed (for example, "ui_blink").
    #         Only run the blink when this key is just pressed.

    # Step 2: Decide how far you want to blink.
    #         Hint: you might want a variable like blink_distance.

    # Step 3: Figure out which direction to blink in.
    #         Options:
    #           - Use the current input direction (where the player is trying to move), OR
    #           - Use the last non-zero movement direction you stored.

    # Step 4: Calculate a new position in that direction:
    #         new_position = current_position + (direction * blink_distance)

    # Step 5: Move the player to the new position instantly.

    # Step 6: (Optional) Add a cooldown so you can’t blink every frame.
    #         - Start a timer or use a variable that counts down.

    # Step 7: (Optional) Add a small visual change when blinking
    #         such as a brief color change or flash.
    pass

shadeydo:
This ability teleports the character instantly in the direction they are running whenever the player presses C (Closes #1892).
The ability is local to the stealth mini-game character, and the hope is that the script will later be partially redacted so that the competitors can learn to code player movement.

Changes:
 - Added a new keybind “champ-blink” to the input map in project settings
 - Added an island in the stealth mini-game that can only be visited with the blink ability
 
The hope is that the player will see this new island and be incentivized to code a working blink ability.
This still needs some work, mainly because now that the player can go out of bounds, the system of using a hidden border tile layer becomes a bit awkward.

### DIFF
--- a/scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
+++ b/scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
@@ -25,7 +25,7 @@ var mode: Mode = Mode.COZY
 @export_range(10, 100000, 10) var moving_step: float = 4000.0
 
 ## How many 64px tiles the player should teleport when the blink key (c) is pressed.
-@export_range(0,24,0.5) var blink_distance_tiles = 3.0
+@export_range(0,24,0.5,"suffix:tiles") var blink_distance = 3.0
 
 
 ## Function that is called every "tick" that is constantly listening
@@ -45,25 +45,23 @@ func _physics_process(delta: float) -> void:
 	move_and_slide()
 
 
-# TODO: Make this blink fn more readable and redact some parts of it so the player can write their own.
+# TODO:Redact parts of the blink ability so the learners can write their own.
 # Blink ability
 
-
 func _apply_blink():
-
 # Convert tiles to pixels for blink distance.
-	var blink_distance_pixels = blink_distance_tiles*64;
+	var blink_distance_pixels = blink_distance*64
 # Figure out which direction to blink
 	var direction := input_vector.normalized()
 # Calculate new coordinates
 	var target_coordinates: Vector2 = position + (direction * blink_distance_pixels)	
 # Move dummy to coordinates and see if its safe to go
-	$BlinkCheck.position = target_coordinates;
+	$BlinkCheck.position = target_coordinates
 	# Force physics
 	$BlinkCheck.force_update_transform()
-	# If there are no collisions, move the player there.
+	# If there are no collisions, move the player to the new position.
 	if not $BlinkCheck.has_overlapping_bodies():
-		position = target_coordinates;
+		position = target_coordinates
 
 
 # (Optional) Add a cooldown so you can’t blink every frame.


project.godot
```diff
@@ -263,6 +263,11 @@ throw={
 , Object(InputEventMouseButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"button_mask":1,"position":Vector2(92, 25),"global_position":Vector2(92, 54),"factor":1.0,"button_index":1,"canceled":false,"pressed":true,"double_click":false,"script":null)
 ]
 }
+champ-blink={
```
wjt: The implied convention for built-in Godot input actions, that we have followed for all other actions in this project, is to separate words with `_` not `-`.

scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn
```diff
@@ -369,6 +375,7 @@ dialogue = ExtResource("15_sfcyb")
 metadata/_custom_type_script = "uid://x1mxt6bmei2o"
 
 [node name="Incomplete Character" type="CharacterBody2D" parent="." unique_id=94949845]
+z_index = 1
```
wjt: You shouldn't need to explicitly set the z-index of the character if you have the y-sorting in the level configured correctly.

scenes/quests/story_quests/champ/1_stealth/stealth_components/lake decorations/water ripples/water.gd
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends AnimatedSprite2D
+
+
+# Called when the node enters the scene tree for the first time.
+func _enter_tree() -> void:
+	play("1")
```
wjt: You should be able to make animation `"1"` autoplay by setting this property on the node. You shouldn't need a script.

https://docs.godotengine.org/en/stable/classes/class_animatedsprite2d.html#class-animatedsprite2d-property-autoplay

scenes/game_elements/props/decoration/water_rock/water_rock.tscn
wjt, line 1: Did you mean to change the water_rock scene that is shared with the whole project, rather than the one in `champ`?

scenes/quests/story_quests/champ/1_stealth/stealth_components/lake decorations/water ripples/water.gd
shadeydo: Oh, the auto-play button in the gui is easy to miss! I addressed your comments in the latest commit, thanks for all the help getting things up to standards.

scenes/game_elements/props/decoration/water_rock/water_rock.tscn
```diff
@@ -4,16 +4,24 @@
...
 
+[sub_resource type="CircleShape2D" id="CircleShape2D_7gr2h"]
+radius = 20.09975
+
 [node name="WaterRock" type="Node2D" unique_id=200189883]
 script = ExtResource("1_7gr2h")
 sprite_frames = ExtResource("1_2lr2g")
 
-[node name="AnimatedSprite2D" type="AnimatedSprite2D" parent="." unique_id=1238515801]
+[node name="StaticBody2D" type="StaticBody2D" parent="." unique_id=35559795]
```
wjt: This water rock is deliberately not a StaticBody2D. In general, in the rest of the game, it is placed in bodies of water where the character can't walk anyway. But there are some cases where it is placed partially under a bridge, e.g. in `res://scenes/quests/lore_quests/quest_003/1_stealth_sequence_combo/stealth_sequence_combo.tscn`:

Image

As a result, if you can make it across to this bridge, the rock blocks the player's movement. (It happens to be positioned high enough that you can walk under it, but it's still wrong for it to collide with the player.)

You have also not configured its collision layers correctly. It is in the default layer, 1, `players`:

Image

As a result, enemies in this level detect & attack water rocks with this change:

Image

I think your `champ_stealth` scene would have the same problem if the enemies moved (which they don't for unknown reason).

I fixed a similar bug in commit 8e7c25b494d456b9903be83898a448b8acceb282.

Please do the following:

1. Duplicate this scene, with your changes, into somewhere in the `champ` folder with a name of your choice - it could be simply `champ_water_rock.tscn`
2. Replace all uses of `water_rock.tscn` in champ with your copy - I suggest that the easiest way to do this is to close the scene in Godot, then open `scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn` in a text editor and change this line:

```ini
[ext_resource type="PackedScene" uid="uid://dv4f232y8w8dv" path="res://scenes/game_elements/props/decoration/water_rock/water_rock.tscn" id="29_rik0t"]
```

to this (removing the `uid` and changing the `path` to match wherever you save the rock):

```ini
[ext_resource type="PackedScene" path="res://scenes/quests/story_quests/champ/1_stealth/stealth_components/champ_water_rock.tscn" id="29_rik0t"]
```

Now if you open and save the scene in the editor, it should add the UID for your copy of the scene.

3. In your copy, fix the collision layers. You can look at `res://scenes/game_elements/props/decoration/rock/rock.tscn` for an example of how they should look.
4. Finally, undo your changes to the shared file:

```console
git restore --source=3e2a4ecb scenes/game_elements/props/decoration/water_rock/water_rock.tscn
git add scenes/game_elements/props/decoration/water_rock/water_rock.tscn
git commit -m 'Revert changes to shared water rock'
```

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +41,48 @@ func _physics_process(delta: float) -> void:
 	velocity = velocity.move_toward(input_vector, step * delta)
 	move_and_slide()
 
+
+# TODO: Make this blink fn more readable and redact some parts of it so the player can write their own.
+# Blink ability
+@export_range(64, 32*64) var blink_distance_pixels = 192
```
wjt: It is conventional to put all `@export`ed variables together at the top of the file, above function definitions.

https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +41,48 @@ func _physics_process(delta: float) -> void:
...
+func _apply_blink():
+
+# Step 1: Decide how far you want to blink.
+#         Hint: you might want a variable like blink_distance.
+	
+# Step 2: Figure out which direction to blink in.
+	var direction: Vector2 = Vector2(input_vector.x,input_vector.y).normalized()
```
wjt:
```suggestion
	var direction := input_vector.normalized()
```

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +41,48 @@ func _physics_process(delta: float) -> void:
...
+
+# Step 3: Move the player in that direction instantly
+	position = Vector2(position.x+(direction.x*blink_distance_pixels), position.y+(direction.y*blink_distance_pixels));
```
wjt:
```suggestion
	position = position + (direction * blink_distance_pixels)
```

scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn
```diff
@@ -288,20 +84,27 @@ y_sort_enabled = true
...
+visible = false
 y_sort_enabled = true
 color = Color(0.481789, 0.48179, 0.481789, 1)
 metadata/_edit_lock_ = true
 
 [node name="TileMapLayers" type="Node2D" parent="." unique_id=1370239493]
 z_index = -1
 
+[node name="Decorations" type="TileMapLayer" parent="TileMapLayers" unique_id=422508235]
+z_index = 1
```
felixwalberg: I noticed that when the player is on the blink island, the top border of the water ripple decorations renders above them. I think if you reset the z_index to 0 and then reorder the node tree so that the `Decorations` TileMapLayer node is below the `Ground` TileMapLayer, then they should respect your y_sorting while also appearing above the base tilemap.

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +44,54 @@ func _physics_process(delta: float) -> void:
...
+func _apply_blink():
+
+# Convert tiles to pixels for blink distance.
+	var blink_distance_pixels = blink_distance_tiles*64;
+# Figure out which direction to blink
+	var direction := input_vector.normalized()
+# Calculate new coordinates
+	var target_coordinates: Vector2 = position + (direction * blink_distance_pixels)	
+# Move dummy to coordinates and see if its safe to go
+	$BlinkCheck.position = target_coordinates;
```
felixwalberg: Image
<br>I noticed that it is possible to blink right behind/on top of the NPC here in the corner. It isn't an issue now since you can blink away, but if the learner were to add a cooldown or limit to blinks, then this could result in someone getting stuck, as you can't walk. Not sure if it is really something that needs to be fixed now, as the tool works well overall!

scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn
shadeydo: This whole time I thought the node ordering just wasn’t working... Turns out the ordering is just the opposite way than what I assumed!

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +44,54 @@ func _physics_process(delta: float) -> void:
...
+	# Force physics
+	$BlinkCheck.force_update_transform()
+	# If there are no collisions, move the player there.
+	if ($BlinkCheck.get_overlapping_bodies().size() == 0):
```
wjt:
```suggestion
	if not $BlinkCheck.has_overlapping_bodies():
```

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +44,54 @@ func _physics_process(delta: float) -> void:
...
+		position = target_coordinates;
```
wjt:
```suggestion
		position = target_coordinates
```

I actually didn't know that semicolons are legal in  GDScript!

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -24,6 +24,9 @@ var mode: Mode = Mode.COZY
 ## How fast does the player transition from stopped to walking/running.
 @export_range(10, 100000, 10) var moving_step: float = 4000.0
 
+## How many 64px tiles the player should teleport when the blink key (c) is pressed.
+@export_range(0,24,0.5) var blink_distance_tiles = 3.0
```
wjt:
```suggestion
@export_range(0, 24, 0.5, "suffix:tiles") var blink_distance_tiles = 3.0
```

(I'm indifferent as to whether the `_tiles` suffix on the variable name is necessary. Certainly it makes the code using it clearer!)

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
wjt:
```suggestion
	$BlinkCheck.position = target_coordinates
```

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -41,19 +44,54 @@ func _physics_process(delta: float) -> void:
...
 
+
+# TODO: Make this blink fn more readable and redact some parts of it so the player can write their own.
+# Blink ability
+
+
+func _apply_blink():
+
+# Convert tiles to pixels for blink distance.
+	var blink_distance_pixels = blink_distance_tiles*64;
```
wjt:
```suggestion
	var blink_distance_pixels = blink_distance_tiles*64
```

PR #1904, opened 2026-02-10
Prototype: Grapple to climb

manuq:
Here is a quick branch to check the status of the grappling hook to climb.

This is currently not possible because the grapple collides with walls. Otherwise, the player would be able to reach buttons or pins in a diagonal, making the use of needles less important:

Captura desde 2026-02-10 09-50-38

Captura desde 2026-02-10 09-50-50

Also the player is colliding with walls while pulling. This branch disables the collision with walls while pulling, as it was happening with the non-walkable floor (water, void). And because the grapple ray may start in the middle of the wall (when the player is touching the wall) it kind of works for going up one level. But as you can see, it doesn't work for going down. I added some pins to the needles level to test it:

[recording.webm](https://github.com/user-attachments/assets/6c1c474a-affc-41db-bac2-d9541e762ae6)

[recording.webm](https://github.com/user-attachments/assets/aa63b9d0-8f6d-4fe6-b5fa-baf4d662e37d)

My initial feeling about this is: trying to do this kind of things into a game that doesn't have actual elevation is hard!


manuq:
I couldn't hack around it with custom data for tiles. Because the collider is a single one for all the tiles. This doesn't seem possible to me without actual elevation data! At least this surfaced the "actual elevations VS fake elevations" topic. Closing for now.

Issue #1887, opened 2026-02-05
Walk on water ability


felixwalberg:
Partway along the Causeway, we want there to be an option for the player to "walk on water". 

One possible idea:
- make the traditional path tedious, something that takes a roundabout way to get to the end, where it would clearly be more efficient to walk over a conveniently placed "gap" in the Causeway filled with water. 

We will need to make an incomplete function inside the champ_incomplete_character.gd script that listens for some keybind associated with bypassing the collisions. This keybind could then disable/change the collision layer of the player (the one associated with colliding into the border of the shoreline) so that they can walk on the water directly. 

There might be an issue with the player letting go of the keybind while on the water. 

Another problem with this would be resolving the collision layers on all mini games. Mainly, the collision of projectiles on the combat minigame would need to be altered so that the tilemap can include collisions rather than the workaround of using a separate (hidden) border layer.

PR #1909, opened 2026-02-10
Champ walk on water

Fauche-r:
Created the ability to walk on water (should be key mapped to 'f' (for Fly))for the champ stealth scene. I brute forced it a little, however, it seems to be working as intended, and does not seems to be creating issues with other scene.

Resolves issue #1887


project.godot
```diff
@@ -263,6 +263,11 @@ throw={
 , Object(InputEventMouseButton,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"button_mask":1,"position":Vector2(92, 25),"global_position":Vector2(92, 54),"factor":1.0,"button_index":1,"canceled":false,"pressed":true,"double_click":false,"script":null)
 ]
 }
+walk_on_water={
```
wjt: Please prefix this champ-specific action with `champ_`. (We don't normally have StoryQuest-specific actions here but I don't see a way around it.)

Issue #1911, opened 2026-02-11
Prefix NO_EDIT root nodes with quest name


wjt:
Currently the root nodes of the NO_EDIT template have a mixture of names: some of them contain the word "Template", some do not:

```console
$ find scenes/quests/template_quests/NO_EDIT/ -name '*.tscn' | xargs rg --pcre2 '^\[node name="([^"]+)"(?!.*parent=)' -r '$1' -o
scenes/quests/template_quests/NO_EDIT/4_NO_EDIT_outro/NO_EDIT_outro.tscn
9:Outro

scenes/quests/template_quests/NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sequence_puzzle.tscn
30:SequencePuzzleTemplate

scenes/quests/template_quests/NO_EDIT/0_NO_EDIT_intro/NO_EDIT_intro.tscn
100:Intro

scenes/quests/template_quests/NO_EDIT/1_NO_EDIT_stealth/NO_EDIT_stealth.tscn
36:StealthTemplateLevel

scenes/quests/template_quests/NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat.tscn
21:Combat
```

A pattern we have tried to establish is that the root node name should be the filename converted to camelcase. So for example `scenes/quests/template_quests/NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sequence_puzzle.tscn` should have a root node `NoEditSequencePuzzle`, and when duplicated as `scenes/quests/story_quests/unicorn_paradise/3_sequence_puzzle/unicorn_paradise_sequence_puzzle.tscn` the root node should become `UnicornParadiseSequencePuzzle`.

Currently we allow the quest name to be (almost) anything, but the filename to be only a-z, 0-9, or underscore - we want the filenames to be consistently lowercase, and easy to type on a US English keyboard & at the shell with no possibility of ambiguous filename encoding (there are multiple ways to represent ñ in Unicode) or homoglyphs (Cyrillic "а" and Latin "a" are not the same character). (Effectively the transformation is s/\W+/_/g` plus lowercasing and stripping leading & trailing underscore.) The rules for node names can be less restrictive: it might be nice to allow, e.g. `SueñosDeHeladoSequencePuzzle`.

There are two parts to this change that IMO should be done together:

1. Rename the root node of all `.tscn` files in the `NO_EDIT` template to have the prefix `NoEdit` and not contain the words `Template` or `Scene`
[...]

PR #1912, opened 2026-02-11
Prefix template scene root nodes with quest name

wjt:
For all scenes in the NO_EDIT template, change the root node name to be NoEdit followed by the quest name, matching the filename.

When instantiating the template, rewrite the root node name to replace NoEdit with the PascalCase name of the quest, so that, again, the root node matches the filename.

Resolves https://github.com/endlessm/threadbare/issues/1911

### DIFF
--- a/addons/storyquest_bootstrap/copier.gd
+++ b/addons/storyquest_bootstrap/copier.gd
@@ -5,6 +5,7 @@ extends RefCounted
 
 const STORYQUESTS_PATH := "res://scenes/quests/story_quests/"
 const TEMPLATE_PREFIX := "NO_EDIT"
+const TEMPLATE_ROOT_NODE_PREFIX := "NoEdit"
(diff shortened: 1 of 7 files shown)


addons/storyquest_bootstrap/copier.gd
```diff
@@ -100,6 +101,11 @@ func copy_packed_scene(packed_scene: PackedScene, copy_path: String) -> Resource
 
 		await maybe_copy_properties(node, path)
 
+	if scene.name.begins_with(TEMPLATE_ROOT_NODE_PREFIX):
+		var suffix := scene.name.substr(TEMPLATE_ROOT_NODE_PREFIX.length())
+		var pascal_case_name := quest_name.capitalize().replace(" ", "")
+		scene.name = pascal_case_name + suffix
```
manuq, line 107: 👍

PR #1914, opened 2026-02-11
Champ scenes are now in the proper order

izzykriz:
Scene navigation triggers have been adjusted for the Champ storyquest to go in the following order:

1: Intro
2: Combat
3: Sequence Puzzle
4: Stealth
5: Outro


scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn
wjt, line 1: I personally think it is very confusing that `1_stealth` comes after `3_sequence_puzzle`. I suggest renumbering all the folders in the `champ/` folder so that they are in gameplay order.

wjt: If you rename resources in Godot (as opposed to your file manager), then (I believe) any references will be updated when you hit save. But in fact what is stored in the `.tscn` files when linking scenes in this way is the UID of the target scene - which does not change.

felixwalberg: Made this change, but I am going to have to go back and deal with merge conflicts. I am also noticing the zoom of the camera feels different. More of the tilemap (and the gray background) is visible, particularly in the intro/outro.

wjt: Sorry, bad recommendation from me. Revert the rename, so the branch can be merged; then rename in a new branch.

wjt:
```
git revert e4fb403f370a72177fa1debc0b97e1e527b64d46
git push
```

wjt: The zoom will not be anything to do with this  branch.

felixwalberg: Commit reverted! After the team has got some other changes merged in I will look at adjusting tilemaps for the camera.

PR #1916, opened 2026-02-13
Champ: sequence puzzle refinement

felixwalberg:
This addresses the remaining feedback from #1805. I was able to completely remove the custom script for the champ_sequence_puzzle_hint_sign.tscn node. 

There is no longer a check on each physics process of the player reaching the middle island. Instead, collisions on "solved" rocks are now disabled. So the player can now walk back across the solved sequence of rocks.

The player mode is set to DEFEATED on an incorrect selection. This is an animation that the learner could replace with a customized one. 

The remaining comment from the original PR is an idea of scaling difficulty. We can implement this by adding a third sequence with sparser hints. The second sequence could lose the demo, and the third could be based on environmental clues or something else that is less explicit.


scenes/quests/story_quests/champ/3_sequence_puzzle/champ_sequence_puzzle.gd
```diff
@@ -128,14 +127,21 @@ func _on_champ_long_rock_water_entered() -> void:
...
 			objs[i].collision.disabled = false
+			if sequences[0].sequence.has(objs[i]) or sequences[1].sequence.has(objs[i]):
+				objs[i].interact_area.disabled = false
 		solve_progress = 0 # redundant?
+	player.mode = Player.Mode.COZY
+
+## Overrides the parent function so we can continue to do logic after sequence ends
+func _on_demonstrate_sequence(step: SequencePuzzleStep) -> void:
+	await super._on_demonstrate_sequence(step)
+	_on_hint_sign_hint_sequence_finished()
```
wjt, line 138: Why not instead connect to `sequence_finished` on each sign in `_ready()`?

PR #1917, opened 2026-02-13
Champ: walking on water ability

### LINKED ISSUE #1887: Walk on water ability

Fauche-r:
Created the ability to walk on water (should be key mapped to 'f' (for Fly))for the champ stealth scene.

Resolves #1887.

### DIFF
--- a/scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn
+++ b/scenes/quests/story_quests/champ/1_stealth/champ_stealth.tscn
@@ -53,6 +53,7 @@
 [ext_resource type="Resource" uid="uid://b7ln3pvi43t77" path="res://scenes/quests/story_quests/champ/1_stealth/stealth_components/champ_incomplete_stealth_decorations_hint.dialogue" id="27_oswsn"]
 [ext_resource type="Resource" uid="uid://dl6n1ifa5ktnp" path="res://scenes/quests/story_quests/champ/1_stealth/stealth_components/champ_incomplete_stealth_movement_hint.dialogue" id="28_fqjr6"]
 [ext_resource type="PackedScene" uid="uid://drew54gcfpekn" path="res://scenes/quests/story_quests/champ/1_stealth/stealth_components/champ_water_rock/champ_water_rock.tscn" id="29_rik0t"]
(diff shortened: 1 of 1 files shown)


Fauche-r:
There is an npc next to the bridge that hint at the possibility of flying/blinking, we could be less subtle and just say "blah blah blah you can try walking on water with f " or something like that.

wjt:
See  scenes/ui_elements/input_hints/repel_hint.tscn as used in scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1.tscn for an example of an on-screen hint? }

scenes/quests/story_quests/champ/1_stealth/champ_incomplete_character.gd
```diff
@@ -63,35 +63,40 @@ func _apply_blink():
...
-		#axis.x = 1
-	
+	var axis: Vector2 = Vector2(0,0)
+	if(Input.is_action_pressed(&"move_left")):
+		axis.x = -1
+	if(Input.is_action_pressed(&"move_right")):
+		axis.x = 1
+	input_vector = axis * SPEED
+
+ 	# TODO: Question: how can we make diagonal speed the same as walking in a straight line?
```
wjt, line 90: I think you'll want a different TODO: how can we make it possible to walk up and down?

felixwalberg: This is true, I think we need a more uniform approach to how we "hint" at ideas for the learners. Up/Down movement is mentioned in the dialogue from a hint NPC at the start, but it would be good to include it here.

Issue #637, opened 2025-05-15
Improve background music crossfading between scenes


wjt:
Currently, the background music always fades out during the scene-exit transition, then fades in during the scene-enter transition. This has two negative consequences:

1. It sounds kind of janky when the successive scenes have the same music (e.g. the three waves of ink combat);
2. It means the opening chord of each new piece of music is guaranteed to be inaudible since the 1-second fade-in is still ongoing.

I think it would be better to do this:

- When leaving a scene, continue playing the music at normal volume
- When entering a new scene:
  - If it has the same music, no further action needed
  - Otherwise, fade the old music out for the duration of the 1-second fade-in; then start the new music, if any.

This avoids needing to be able to look "into the future" to find out what the background music will be before switching to the new scene. If we could do that (I'm not sure how to do that easily…) then we could align the fade-out of the music with the fade-out transition, and start the new track at the blank point.

### XP Summary (total 425)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |
| Art: Music & Sound | Problem Solving | 75 |
| Art: Music & Sound | Creativity & Design Thinking | 75 |
| Art: Music & Sound | Technical Literacy | 50 |

manuq:
> The problem is that, in the current implementation, we don't know what the background sound will be for the next scene during the fade-out.

Sorry, when I said "before" I meant to say "at the same time". Just without audio fading (only visual fading).

wjt:
I was thinking about what other features we would like a background music player to have (without going too far towards recreating fmod). Some thoughts:

- When we go inside the Eternal Loom to play Sokoban, it might be nice to apply a low-pass filter to the music, reducing the higher frequencies, to give the impression of being inside a building with the music playing from outside.
- I think the final round of ink combat would sound better if the combat music faded out (perhaps replaced by some ambient background sounds rather than other music) when you defeat the enemies and the collectable thread is revealed.
- Similarly, we don't have appropriate music, but if we had a piece of music that incorporated the melodies from the music puzzle (either by composing new music, or changing the puzzle, or both), it might be nice to fade *in* some background music when the puzzle is solved.

wjt:
[...]

PR #1920, opened 2026-02-17
Only fade background music when changing streams

wjt:
Previously, the background music always faded out during the scene-exit
transition, then faded in during the scene-enter transition. This had:
two negative consequences:

1. It sounds bad when the successive scenes have the same music (e.g.
   the three waves of ink combat);
2. It means the opening chord of each new piece of music is guaranteed
   to be inaudible since the 1-second fade-in is still ongoing.

Instead:

- Do not fade the music when the scene fades out;
- When the new scene arrives, with a BackgroundMusic node:
  - If the stream is the same, do nothing (fixing point 1 above);
  - Else, fade out the old stream for 1 second, then start playing the
    new stream at full volume (fixing point 2 above).

To do this, move most of the logic from BackgroundMusic to MusicPlayer.
Remove the ability to pause/resume BackgroundMusic in the editor for
simplicity, leaving only Play/Stop.

Improve documentation of the two cooperating scripts.

Resolves https://github.com/endlessm/threadbare/issues/637

### DIFF
--- a/scenes/game_elements/props/background_music/components/background_music.gd
+++ b/scenes/game_elements/props/background_music/components/background_music.gd
@@ -1,17 +1,22 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 @tool
+class_name BackgroundMusic
 extends Node
+## Specify the background music to play in a scene.
+##
+## This controls the singleton MusicPlayer node, which outlives every scene, so
+## that music can keep playing between scenes.
+##
+## There should be at most one BackgroundMusic component in a scene.
 
-@export_tool_button("Play") var play_button: Callable = _play
-@export_tool_button("Stop") var stop_button: Callable = _stop
-@export_tool_button("Pause/Resume") var pause_resume_button: Callable = _pause_resume
+@export_tool_button("Play") var play_button: Callable = play
+@export_tool_button("Stop") var stop_button: Callable = stop
+
 @export var stream: AudioStream:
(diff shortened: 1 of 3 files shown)


scenes/game_elements/props/background_music/components/background_music.gd
```diff
@@ -22,63 +27,26 @@ func _get_configuration_warnings() -> PackedStringArray:
...
-		return
-
-	if stream:
-		_fade_in()
+	if not Engine.is_editor_hint():
 		_play()
-	else:
-		_stop()
-
-	Transitions.started.connect(self._fade_out)
```
wjt, line 34: By the way, this should also fix the issue in `res://scenes/quests/lore_quests/quest_002/4_closing_transition/closing_transition.tscn` where the music stops when the screen fades during dialogue.

scenes/quests/lore_quests/quest_001/4_closing_transition/closing_transition.tscn
```diff
@@ -134,6 +134,20 @@ tracks/2/keys = {
...
+tracks/3/enabled = true
+tracks/3/path = NodePath("../BackgroundMusic")
+tracks/3/interp = 1
+tracks/3/loop_wrap = true
+tracks/3/keys = {
+"times": PackedFloat32Array(1),
+"transitions": PackedFloat32Array(1),
+"values": [{
+"args": [],
+"method": &"stop"
```
manuq, line 148: 👍

Issue #380, opened 2025-04-29
Fray's End loads slowly due to generating cloud texture


wjt:
Fray's End takes (on my 2019 Dell XPS laptop) around 4 seconds to load. The vast majority of this time is spent generating the noise texture for the clouds.

Having read the [NoiseTexture2D documentation](https://docs.godotengine.org/en/stable/classes/class_noisetexture2d.html) I feel like it should be possible to create the noise asynchronously and fade the clouds in when they are ready.

wjt:
The problem was mitigated in #377 by:

1. Loading the next scene asynchronously during the fade-out transition; then
2. Blocking until complete at the blank point of the transition, so the pause is not so jarring.

wjt:
I added some very primitive instrumentation to SceneSwitcher.change_to_file_with_transition() and did some rough timings running the game in the editor on my Dell XPS 13 from 2019.

Here are three tests of the transition from the intro to Fray's End:

```
Transition to uid://cufkthb25mpxy took 7.24s
Transition to uid://cufkthb25mpxy took 7.14s
Transition to uid://cufkthb25mpxy took 7.26s
```

These times include 2 seconds for the fade in and fade out.

No timing for the switch to the music puzzle because it uses `change_to_packed_with_transition()` but interestingly that is also slow, and the music puzzle is the other level that has clouds.

Switching to each combat level, and to the stealth challenge, is essentially instant:

```
Transition to uid://c764br4tplkb2 took 2.03s
Transition to uid://cgemeabsewy1f took 2.02s
Transition to uid://bo6qfusag3fae took 2.02s
Transition to uid://db6vhi7s2f37c took 2.00s
```


Then when transitioning from the closing cutscene back to Fray's End:

```
Transition to uid://cufkthb25mpxy took 6.88s
```

@pablitar you had wondered whether it's fast the second time you go to Fray's End because the scene is cached. The short answer is no.

wjt:
After I removed the clouds from Fray's End:

```
Transition to uid://cufkthb25mpxy took 4.37s
```

And after I removed the clouds from the music puzzle as well, Fray's End loads instantaneously:

```
Transition to uid://cufkthb25mpxy took 2.04s
```

(Note that the music puzzle is referenced from Fray's End as a PackedScene!)

wjt:
[...]

PR #1921, opened 2026-02-17
Reinstate clouds shadow

manuq:
Replace the clouds overlay effect with a new implementation, and reinstate it in Fray's End.

See node editor descriptions or script comments for implementation details.

Fix https://github.com/endlessm/threadbare/issues/380

### DIFF
--- a/scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gd
+++ b/scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gd
@@ -5,17 +5,27 @@ extends Parallax2D
 
 @export_tool_button("Random Clouds") var randomize_button: Callable = randomize
 
+@export var _seed: int:
+	set = _set_seed
+
 var texture: NoiseTexture2D
 var noise: FastNoiseLite
 
 @onready var color_rect: ColorRect = %ColorRect
 
 
+func _set_seed(new_seed: int) -> void:
+	_seed = new_seed
+	if noise:
+		noise.seed = _seed
+		await texture.changed
+
+
 func _ready() -> void:
 	texture = (color_rect.material as ShaderMaterial).get_shader_parameter("shadow_texture")
 	noise = texture.noise
+	_set_seed(_seed)
 
 
 func randomize() -> void:
-	noise.seed = randi()
-	await texture.changed
+	await _set_seed(randi())


scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gdshader
```diff
@@ -0,0 +1,61 @@
+/**
+ * SPDX-FileCopyrightText: The Threadbare Authors
+ * SPDX-License-Identifier: MPL-2.0
+ */
+shader_type canvas_item;
+
+/**
+ * The shader is set to substract because it's a shadow.
```
wjt:
```suggestion
 * The shader is set to subtract because it's a shadow.
```

scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gd
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends Parallax2D
+
+@export_tool_button("Random Clouds") var randomize_button: Callable = randomize
```
wjt, line 6: I don't quite follow how this button is supposed to be used.

Suppose that I have `res://scenes/world_map/frays_end.tscn` open in my editor. If I click this button, the clouds in the editor change, but when I run the scene they are back to the seed=0 clouds.

However, if I have both `res://scenes/world_map/frays_end.tscn` and `res://scenes/game_elements/fx/clouds_shadow/clouds_shadow.tscn` open. Then:

1. Switch to the frays end tab
2. Click Random Clouds
3. Save Fray's End - no change on disk
4. Switch back to the `clouds_shadow.tscn` scene; observe that the clouds have changed to match Fray's End
5. Save *that scene*; observe that the seed has been persisted here.

If I set all of the ShaderMaterial, NoiseTexture2D and FastNoiseLite as `resource_local_to_scene = true` (I don't understand why all three have to be) then this doesn't happen: changing the seed in Fray's End does not persist. But then it does nothing in the editor.

I think it's fine either way - the worst that happens is that the random clouds are inadvertently changed occasionally - but maybe we want to do the `NOTIFICATION_EDITOR_PRE_SAVE` trick in clouds_shadow.gd to reset the seed? Or we can just leave it.

scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gdshader
```diff
@@ -0,0 +1,61 @@
...
+	vec2 uv = UV + speed * TIME;
+
+	float shadow = texture(shadow_texture, uv).r;
+
+	// Add amount value.
+	shadow = amount + 0.5 - shadow;
+
+	// Apply contrast to define the shadow silhouette.
+	// The result may be outside of the 0-1 range.
+	shadow = (shadow - 0.5) * contrast + 0.5;
```
wjt: I've read this a few times and I find it a bit hard to understand the `0.5`s being added and subtracted. I wonder if this would be clearer:


```suggestion
	// Add amount value.
	shadow = amount - shadow;

	// Apply contrast to define the shadow silhouette.
	// The result may be outside of the 0-1 range.
	shadow = shadow * contrast + 0.5;
```

I think it's equivalent because on line 49 you add 0.5, and then subtract it immediately in line 53.

scenes/game_elements/fx/clouds_shadow/clouds_shadow.tscn
```diff
@@ -0,0 +1,45 @@
...
+repeat_size = Vector2(1024, 1024)
+repeat_times = 3
+script = ExtResource("1_vhj68")
+
+[node name="ColorRect" type="ColorRect" parent="." unique_id=1045851952]
+unique_name_in_owner = true
+editor_description = "Start with a small noise texture that is then upscaled to a square. The size of this ColorRect.
+
+The texture width is about 2/3 of its height to look flattened, matching the game perspective. This way the shadow seems to be projected on the ground.
+"
```
wjt, line 37: I like this use of editor description!

scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gdshader
```diff
@@ -0,0 +1,61 @@
...
+	vec2 uv = UV + speed * TIME;
+
+	float shadow = texture(shadow_texture, uv).r;
+
+	// Add amount value.
+	shadow = amount - shadow;
+
+	// Apply contrast to define the shadow silhouette.
+	// The result may be outside of the 0-1 range.
+	shadow = shadow * contrast + 0.5;
```
manuq, line 53: @wjt very good point, I was keeping the contrast operation separate so I can comment it out, but is better to simplify and not add 0.5 and then immediately substract it. My reference for the contrast operation was https://github.com/patriciogonzalezvivo/lygia/blob/main/color/contrast.glsl

wjt: Ah OK! Sorry, If you think it's better as it was, that's fine. That reference is useful.

scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gd
manuq: Good catch, since in my prototype branch I was calling to the randomize() function directly, and testing the button only in the clouds_shadow.tscn scene, I didn't notice. I guess there is no other way than exporting a proxy "seed" property and then the noise resource is set to this value.

manuq: @wjt I appended a fixup commit. Let me know what do you think. If we want the seed hidden we could change the export to be `@export_storage`, but I think is good to expose it. I had something similar in my townies exploration https://github.com/endlessm/threadbare/pull/1801

scenes/game_elements/fx/clouds_shadow/components/clouds_shadow.gdshader
manuq: Not at all, shader code should be optimized. The same happens when doing brightness + contrast operation, is preferred to do them both at once https://github.com/patriciogonzalezvivo/lygia/blob/main/color/brightnessContrast.glsl

wjt: TBH I'm sure the shader compiler is smart enough to optimise out the `+ 0.5 - 0.5`; I was more concerned about the shader interpreter in my mind :)

PR #1926, opened 2026-02-17
Champ: renamed directories to match order of gameplay

felixwalberg:
Renamed the directories for the combat, sequence puzzle, and stealth minigames so they follow the order of how they are played within the quest.


shadeydo (approved):
Looks good

PR #1927, opened 2026-02-17
Champ stealth collision changes

shadeydo:
### Replaced outer hidden border wall with a polygon
This simplifies the bounds collision and enables blink bounds checking
### Blink ability Improvements
- Fixed bug where dummy could become misaligned with the player, messing up blink calculations
- Added simple blink position indicator
- Player can no longer blink out of bounds
- Player can no longer blink to water
- Began work on removing parts of the function, so learners can code it themselves.

All of these changes are to make blinking a more robust ability, and to increase the distinction between walking on water vs blinking.

### DIFF
--- a/scenes/quests/story_quests/champ/4_outro/champ_animation.gd
+++ b/scenes/quests/story_quests/champ/4_outro/champ_animation.gd
@@ -4,7 +4,7 @@ extends CharacterBody2D
 
 @onready var champ_animation: AnimatedSprite2D = $"Champ Animation"
 
-var swim_speed: int = 100
+@export var swim_speed: int = 100
 
 func _ready() -> void:
 	# Give champ a starting swim speed


shadeydo:
Accidentally changed the audio, reopening with a new commit.

project.godot
```diff
@@ -20,6 +20,10 @@ boot_splash/use_filter=false
 boot_splash/image="uid://dvfpk2yguwjfv"
 config/icon="uid://dhwqkdin6refx"
 
+[audio]
+
+buses/default_bus_layout="uid://c7thbop54thnf"
```
wjt, line 25: I don't think this change is intentional?

wjt: It's harmless - it configures this explicitly to the default - but I don't think it should be part of this PR.

wjt: I opened https://github.com/endlessm/threadbare/pull/1928 to move `res://default_bus_layout.tres` to a better home and, in passing, make this same change.

PR #1935, opened 2026-02-18
docs: Remove pages which have been imported to the wiki

wjt:
We previously had an awkward split between documentation stored in the
repository, and pages on the wiki (which were previously in the
Discussions section).

I have imported all four files in the docs/ folder into the wiki. Remove
them from this repository.

I mostly am not worried about dead links - we can fix them as we find
them - but in the specific case of the CONTRIBUTING.md file, this has
special status on GitHub: if a CONTRIBUTING.md file is found in the repo
root, docs/ folder, or .github/ folder, then it is linked to prominently
in several places by GitHub, including
https://github.com/endlessm/threadbare/contribute (which is linked from
https://github.com/endlessm/threadbare/issues if you have not previously
contributed). Add a stub file to .github/ that links to the
corresponding wiki page.

https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors

Remove the CODEOWNERS rules for these files.

This leaves the only thing in the docs/ folder being a screenshot used
by the Flatpak build's AppStream metainfo. We can address this another
day.


manuq (approved):
Yes!

Issue #1613, opened 2025-11-16
Make painting cliffs terrains easier


manuq:
From Jackie Codes [presentation in GodotFest](https://youtu.be/VrwDPpLefUU?t=968), there is an easier setup for cliff tiles. It is well explained in [this other video](https://www.youtube.com/watch?v=Q8EC7womtnk) of Jackie.

Our levels had issues in the past because the cliff terrains leave inconsistent tiles that had to be adjusted one by one.

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 150 |
| Game Design: Core Gameplay & Mechanics | Creativity & Design Thinking | 150 |
| Game Design: Core Gameplay & Mechanics | Technical Literacy | 150 |

wjt:
This is in my notes from that talk! You beat me to filing the bug.

wjt:
Problems I see for migrating the main elevation tileset:

1. I bet that there are places where we have the clifftop tile without the clifffront; and where we have the clifffront tile without the clifftop. Maybe we can ignore these but we'll need to migrate them somehow.

2. There are cases where we have the "wrong" cliff-front tile for cosmetic reasons; for example, the door at the end of lorequest 1 stealth:

<img width="512" height="509" alt="Image" src="https://github.com/user-attachments/assets/650c594d-875d-407b-9f0d-4f846dd2b8de" />

3. The real spanner in the works is the cliff_mines tiles:

<img width="256" height="384" alt="Image" src="https://github.com/user-attachments/assets/7ea37aa7-49dc-4060-80ff-ebf6d9a28f9a" />

For these you probably don't want to always have both the clifftop and the clifffront. Not sure what to do about these

PR #1938, opened 2026-02-18
Add simpler elevation tileset

manuq:
Plus a separate tiles/decoration.tres tileset for decoration tiles
that can be added in a layer on top of other tiles.

----

Duplicate tiles/elevation.tres into tiles/elevation_2.tres and
simplify it:

Set size in atlas of tiles with walls below to be 1x2 (twice taller)
and adjust their rendering texture origin to be -32 pixels.

Adjust the collision polygons of the now taller tiles to be the same
as before. Both in physics layer 0 (matching "walls") and in physics
layer 1 (matching "sight occluders"). Also remove physics layer 2
that was matching "non-walkable floor" but didn't have any polygons
in tiles.

Leave a single "Stone" terrain that can paint all at once and with
consistency for the internal wall corners.

Add a new alternative tile for the center tall tile, so it can
be used eg. in Fray's End to let the player walk the bridge.

Remove cliff mines tiles by removing the PNG from tile sources.

----

For decoration tiles:

Create a new PNG using GIMP to remove the duplicated pixels between
cliffs with and without mines. This is done by putting the cliffs
at a layer on top with the Color Erase blending mode.

Although this PNG can be improved in GIMP or Aseprite, it is not
worth the effort. It would be better to gain access to the source
artwork.

----

Port Fray's End to new elevation tileset.

Change the tileset resource in the 3 elevation TileMapLayer
nodes. Remove the now missing wall tiles. Add the alternative
collision-less tile behind the two bridges.

Also add a new 2 TileMapLayer for decorations:
- On level 0: a mine entry.
- On level 1: a staircase and a mine entry blocked by rocks.

Resolve https://github.com/endlessm/threadbare/issues/1613


manuq:
The cliffs at the east of Fray's End has been fixed by this. Before/after:

Captura desde 2026-02-18 18-22-34
Captura desde 2026-02-18 18-22-58

PR #1939, opened 2026-02-18
Champ stealth blink fixes

shadeydo:
The player can no longer:
- Blink to the water in outside corners
- Blink when in a defeated state

### DIFF
--- a/scenes/quests/story_quests/champ/3_stealth/champ_incomplete_character.gd
+++ b/scenes/quests/story_quests/champ/3_stealth/champ_incomplete_character.gd
@@ -66,7 +66,8 @@ func _apply_blink() -> void:
 	if $"../Bounds/Blink Bounds".overlaps_area(dummy):
 	# If there are no collisions, move the player to the new position.
 		if not dummy.has_overlapping_bodies():
-			global_position = target_coordinates
+			#TODO: Teleport the player instantly to the desired location
+			return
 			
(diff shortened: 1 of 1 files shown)


scenes/quests/story_quests/champ/3_stealth/champ_incomplete_character.gd
```diff
@@ -69,14 +72,17 @@ func _apply_blink() -> void:
...
 
-	# (Optional) Add a cooldown so you can’t blink every frame.
+	# TODO: (Optional) Add a cooldown so you can’t blink every frame.
 	# Start a timer or use a variable that counts down.
 
-	# (Optional) Add a small visual change when blinking,
+	# TODO: (Optional) Add a small visual change when blinking,
 	# such as a brief color change or flash.
 
 # TODO: Remove this dev ability before release.
```
felixwalberg: I'm wondering if we are getting to the point where we should try to keep the dev tools out of main. I'm not sure how late Ben will wait to pull it for the learners, but it might be better to leave main ready to go whenever that may happen.

shadeydo: Sounds good, I can remove the _dev_apply_blink function. I’d also like to start removing more stuff from the partially complete blink version for the learners. What do you think is the best way to do that while maintaining the balance of learning?

Issue #1906, opened 2026-02-10
Add cliff shadows: quest_001 - 3_stealth_level


russell-cj:
This sub-issue focuses on a visual update to the stealth level in Quest 001 to include the new base cliff shadows.

### Scene to modify
- `res://quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn`

PR #1943, opened 2026-02-19
Add cliff shadows to quest_001 - 3_stealth_level

russell-cj:
Update the stealth level scene in Quest 001 to use the new base cliff shadows.

resolve: #1906


russell-cj:
>If you have two cliffs stacked on top of each other, on top of the water, I thought it would look weird if the upper cliffs cast a shadow on the thin air. But actually I think it's better:

Analyzing the examples, I think that in the first case (same size), the shadow doesn't provide a clear perspective, so I would consider removing it and leaving the top part without a shadow. However, in the second case, where the upper cliff is narrower, the shadow does provide a fairly acceptable sense of depth. I believe that distinction is key in deciding where to apply it.

manuq:
> It makes the ground shadow rather subtle but still better than no shadow.

Thanks for this example, @wjt . The "substract" idea doesn't work at all, I was confused by other shadow layer in the same stealth scene *facepalm*.

Thanks for your opinion too @russell-cj . I would keep it like this, consistently adding shadows to the elevation tiles. If we can't find a case in which the shadow is not needed, then we should open a task to add the shadow to the cliffs PNG. I agree with @wjt on https://github.com/endlessm/threadbare/pull/1943#issuecomment-3928088621 that a shadow is better too in the "cliff over cliff" case.

russell-cj:
Hi @manuq  and @wjt ,

I’d like to know what the final decision will be regarding the cliff shadows over water. Should I continue doing it the way I was before, or use the `CanvasItemMaterial` that Manu added and apply the suggested blend mode?

PR #1950, opened 2026-02-19
Altering champ combat game to make changing sprites more important

izzykriz:
Since changing the bait sprites is so avoidable in the combat minigame, altered it so that the sprites are very difficult to tell apart at first, and encourages the players to edit the sprite pngs so they can easily distinguish them.


felixwalberg (approved):
I think this is great! The closeness of the dull colors makes it difficult for me to determine which target the bait is associated with. I'm sure this will provide extra motivation for the learners. The only change I made was adding the HitSound of the projectile scenes back to the SFX bus (my understanding is that this is managed globally in Threadbare and will ensure it is mixed properly with the other audio). 

One other thing I will mention is that the dialogue prompt "color in the `champ_bait_projectile` pngs" can also be addressed by learners with the `Modulate` attribute of the AnimatedSprite2D node of the projectile via the inspector. This allows them to change the color of the entire png without having to leave Godot. If they want to get more creative, as your dialogue encourages, they can replace the entire sprite!

Issue #1925, opened 2026-02-17
Champ: Standardize Learner Hints


felixwalberg:
Develop a convention/standard format for how we hint that learners need to make changes. Should these all be TODO comments in the code? Should it be mentioned in the narrative? Both? Whatever we decide, we will make sure it is consistent to avoid confusion.

PR #1952, opened 2026-02-19
Champ standardize hint

Fauche-r:
As discussed in Issue #1925, we needed some more hint, as well as standardizing hint as a whole. I believe keeping some hint in game is useful,(in conjunction with in code TODO and hint) as it allow player to get a good idea of what they need to change on their first playthrough, instead of having to play the game, and then find out on their own what problem exist.

### DIFF
--- a/scenes/quests/story_quests/champ/1_combat/combat_components/champ_combat.dialogue
+++ b/scenes/quests/story_quests/champ/1_combat/combat_components/champ_combat.dialogue
@@ -12,6 +12,6 @@ Of course, he could always experiment with something else that might be more ent
 ~ well_done
 Evan has fed whatever swims below the dark surface of the lake, but he only got a glimpse of something he thought looked green and scaly. 
 It swims so fast! He must push forward to keep up.
-He believe if those buoy had been more flamboyant and bold in their color, the thing that lurked beneath might have been more attracted.
-If you want to help the next poor soul who come through here, open the buoy.tscn scene, and replace the Sprite Frames resource with an image worthy of being a true buoy.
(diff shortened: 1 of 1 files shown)


scenes/quests/story_quests/champ/1_combat/combat_components/champ_combat.dialogue
```diff
@@ -12,4 +12,6 @@ Of course, he could always experiment with something else that might be more ent
 ~ well_done
 Evan has fed whatever swims below the dark surface of the lake, but he only got a glimpse of something he thought looked green and scaly. 
 It swims so fast! He must push forward to keep up.
+He believe if those buoy had been more flamboyant and bold in their color, the thing that lurked beneath might have been more attracted.
```
felixwalberg: There might be a tiny conflict with #1950, but it will be an easy fix as it's just dialogue text.

scenes/quests/story_quests/champ/3_stealth/champ_incomplete_character.gd
```diff
@@ -82,9 +82,11 @@ func _apply_blink() -> void:
 
 ## Function to remove collisions, allowing the player to walk on water tiles
 func _walk_on_water() -> void:
+	#remove the collision of the Water_border, allowing player to "walk" on "water"
 	$"../TileMapLayers/Water_border".enabled = false
```
wjt, line 86: In general, `$../` is an anti-pattern. If there are nodes elsewhere in the scene that this script wants to reference, it should have an `@export` and you should assign the nodes in the inspector.

Fauche-r: We addressed this issue in PR #1977.

Issue #1903, opened 2026-02-10
Lore Stealth level: Stones duplicated in two tilemap layers


manuq:
### Description

The same stones are duplicated in 2 tilemap layers:

https://github.com/user-attachments/assets/b268de0b-b7a8-4aa4-94e7-2586520be08e

At first glance, it looks like it would be enough to remove the "Stones Level 3" layer entirely. But this has to be confirmed first, by carefully comparing the tiles in the two layers.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

Issue #1417, opened 2025-10-23
Weird light occlusion in stealth level


wjt:
### Description

Something weird is going on with the light here in `res://scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn`:

<img width="448" height="272" alt="Image" src="https://github.com/user-attachments/assets/400f7b25-e319-4243-821b-fc1e53a0c5cf" />

On the left-hand side of the cliff it's being blocked by something. If I had to guess, I would guess that there is another tile on a lower level of the tilemap that is blocking the light. Or is it the bush? I have to say: I don't actually know how the light occlusion works!

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

Issue #1940, opened 2026-02-18
Migrate scenes to new elevation


manuq:
Follow up of https://github.com/endlessm/threadbare/issues/1613

Once the task above is done we'll need to manually migrate all scenes using `tiles/elevation.tres/` to the new `tiles/elevation_2.tres` and maybe also add decorations (mines) on top.

The list of scenes referencing the tileset to replace is (`git grep -l dfp36ffpanjq2`):

- [ ] scenes/game_elements/characters/components/gym_area_test.tscn
- [ ] scenes/game_elements/characters/enemies/mothsache/Mothsache_test_scene.tscn
- [x] scenes/game_elements/props/area_filler/components/area_filler_test.tscn
- [ ] scenes/game_elements/props/void/components/void_chromakey_test.tscn
- [x] scenes/quests/lore_quests/quest_000/5_void_runner/tutorial_void_runner.tscn
- [x] scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn
- [x] scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_1.tscn
- [ ] scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_2.tscn
- [ ] scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
- [x] scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
- [x] scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
- [ ] scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_needles.tscn
- [ ] scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_powerup.tscn
- [x] scenes/quests/lore_quests/quest_003/1_stealth_sequence_combo/stealth_sequence_combo.tscn
- [ ] scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_4.tscn
- [ ] scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_5.tscn
- [ ] scenes/quests/lore_quests/quest_003/2_ink_drinker_levels/ink_combat_round_6.tscn
- [ ] scenes/quests/story_quests/el_juguete_perdido/1_stealth/stealth.tscn
- [ ] scenes/quests/story_quests/el_juguete_perdido/2_combat/combat.tscn
[...]

PR #1953, opened 2026-02-19
Lore quest 1 Stealth: Use new elevations

manuq:
For this to work occlussion rendering needs to be disabled in the "Stones" layer below the player and guards. Just as physics is disabled for the same tilemap layer.

Also remove layer Stones Level 2 which had repeated tiles from Stones Level 1.

Also remove Stones Cover layer. This fixes occluding the light from the first guard.

Helps https://github.com/endlessm/threadbare/issues/1940
Fixes https://github.com/endlessm/threadbare/issues/1903
Fixes https://github.com/endlessm/threadbare/issues/1417


manuq:
I migrated the Stealth level because of it's complexity, wanting to test if the new elevations could handle it. It wasn't that hard.

Issue #1958, opened 2026-02-20
Replace Ink Drinkers “spit” terminology & sound effect


manuq:
Ink drinkers spitting has been found to be inappropriate in some cultures.
- Replace the sound effect.
- Change dialogue to say "splat" instead.

manuq:
> [InkDrinker_Pluurgh_02.wav](https://github.com/user-attachments/files/25444378/InkDrinker_Pluurgh_02.wav) [InkDrinker_Pluurgh_01.wav](https://github.com/user-attachments/files/25444379/InkDrinker_Pluurgh_01.wav) [InkDrinker_Apuhuh_01.wav](https://github.com/user-attachments/files/25444381/InkDrinker_Apuhuh_01.wav) [InkDrinker_AirPop_01.wav](https://github.com/user-attachments/files/25444380/InkDrinker_AirPop_01.wav)

I tried all of them and went with "AirPop" in the pull request. I think this one maintains the same feeling as before, and match the cartoony style. But I'm open to suggestions! My other candidate was "Apuhuh". The two "Pluurgh" ones seemed too much of revealing a human voice.

https://github.com/user-attachments/assets/6284049a-36b4-4fdc-a31f-842aefeca9b6

PlayMatters:
Love this, Manuel. Huge thanks.

PR #1959, opened 2026-02-20
Ink combat dialogue: Use "splat" term for the enemy action

manuq:
Fix https://github.com/endlessm/threadbare/pull/1958

### DIFF
--- a/scenes/quests/lore_quests/quest_001/2_ink_combat/components/dialogues/ink_well.dialogue
+++ b/scenes/quests/lore_quests/quest_001/2_ink_combat/components/dialogues/ink_well.dialogue
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 ~ start
 Musician: The songwriters wrote such colorful songs, before. But the InkDrinkers consumed all of the ink. Restore the ink to the matching jars.
-Musician: But be careful, they spit!
+Musician: But be careful, they splat!
 => END
 ~ well_done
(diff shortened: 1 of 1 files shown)


scenes/quests/lore_quests/quest_001/2_ink_combat/components/dialogues/ink_well.dialogue
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 ~ start
 Musician: The songwriters wrote such colorful songs, before. But the InkDrinkers consumed all of the ink. Restore the ink to the matching jars.
-Musician: But be careful, they spit!
+Musician: But be careful, they splat!
```
wjt, line 5: I don't think “they splat” makes sense. Splatting is what the blob of ink does when it hits you.

Squirt, spritz, spray, by contrast, seem like verbs that could mean "expel a blob of ink".

manuq: Ohh I didn't know! "Splat" is what I got from an internal conversation. @saspiers @PlayMatters please resolve.

manuq: What about: "Musician: But be careful, they do splats!"

wjt: Let's discuss further on https://github.com/endlessm/threadbare/issues/1958

Issue #1957, opened 2026-02-20
Don't kick player back to an earlier scene when they run out of lives


wjt:
Currently, when the player runs out of lives, they are kicked back to the start of the current challenge. “Current challenge” is defined to be:

- If 1 or more threads have been collected on this quest: the scene the last thread they collected took them to
- Otherwise: the first scene of the quest

For example:

1. If they are defeated by the void 3 times in res://scenes/quests/lore_quests/quest_002/3_void_grappling/void_grappling_round_2.tscn, they are returned to the start of res://scenes/quests/lore_quests/quest_002/3_void_grappling/void_grappling.tscn
2. If they are caught by a guard three times in the stealth challenge in NO_EDIT, they are taken back to... the opening cutscene
3. Similarly, if they are caught by the terrifying monster three times in the stealth challenge at the start of Sueños Nocturnos, they are taken back to the opening cutscene. (There are two cutscenes.)

2 and 3 are the result of a simplistic definition of "start of challenge".

By contrast, 1 is intentional behaviour. But playtesting feedback such as in #1944, and my own personal experience from testing the game, is that it's incredibly frustrating to be kicked back several levels. I think we need to change this.

PR #1963, opened 2026-02-20
Grant the player effectively infinite lives

wjt:
We currently have a partial implementation of lives in the game. The player starts with 3 lives; they lose a life each time they are defeated; and when they run out of lives, they are kicked back to the start of the current challenge.

However, we have no visible indication of the number of lives the player has remaining. As a result, the game feels broken when the player unexpectedly loses their progress after being defeated.

In addition, some recent playtesters felt that the game-over behaviour is unfair, even if it were visible.

For now, set the number of lives to be effectively infinite (the largest positive integer that Godot can represent), to effectively disable the feature. This hexadecimal number is 9,223,372,036,854,775,807 in decimal: 9 quintillion in the short scale.

Helps https://github.com/endlessm/threadbare/issues/1957

### DIFF
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -29,7 +29,7 @@ const GLOBAL_SECTION := "global"
 const GLOBAL_INCORPORATING_THREADS_KEY := "incorporating_threads"
 const COMPLETED_QUESTS_KEY := "completed_quests"
 const LIVES_KEY := "current_lives"
-const MAX_LIVES := 3
+const MAX_LIVES := 0x7fffffffffffffff
 const DEBUG_LIVES := false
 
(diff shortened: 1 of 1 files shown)


scenes/globals/game_state/game_state.gd
```diff
@@ -29,7 +29,7 @@ const GLOBAL_SECTION := "global"
 const GLOBAL_INCORPORATING_THREADS_KEY := "incorporating_threads"
 const COMPLETED_QUESTS_KEY := "completed_quests"
 const LIVES_KEY := "current_lives"
-const MAX_LIVES := 3
+const MAX_LIVES := 0x7fffffffffffffff
```
manuq, line 32: Yes, smart. Simpler than reverting.

manuq (approved):

PR #1965, opened 2026-02-23
MouseManager: Set process_mode to ALWAYS

wjt:
This allows the mouse to be reliably used when the game is paused.


scenes/globals/mouse_manager/mouse_manager.tscn
```diff
@@ -3,6 +3,7 @@
 [ext_resource type="Script" uid="uid://rw3yh3185lpn" path="res://scenes/globals/mouse_manager/mouse_manager.gd" id="1_u8i10"]
 
 [node name="MouseManager" type="Node" unique_id=1563614058]
+process_mode = 3
```
manuq, line 6: What about adding an editor description to this node with the same explanation as in the commit message? As we moved to the squash & merge strategy on GitHub, I'm trusting less the Git history as a place for finding decisions made.

PR #1968, opened 2026-02-23
NO_EDIT sequence puzzle: Add foam

wjt:
I also reset the zoom to 100%. I do not see the need for the level to be zoomed in.


scenes/quests/template_quests/NO_EDIT/3_NO_EDIT_sequence_puzzle/NO_EDIT_sequence_puzzle.tscn
```diff
@@ -51,7 +56,6 @@ y_sort_enabled = true
 position = Vector2(383, 371)
 
 [node name="Camera2D" type="Camera2D" parent="OnTheGround/Player" unique_id=1732237457]
-zoom = Vector2(1.333, 1.333)
```
manuq, line 54: 👍

Issue #1347, opened 2025-10-14
Musician's "success" dialogue doesn't play when you solve the puzzle


wjt:
### Description

In the music puzzle, the musician has this dialogue:

> ~ item_hint
> Musician: You did it!... That's the magical melody... how wonderful. It's been so long since I heard it.
> Musician: You have opened the portal to the Ink Well. Let's go there to collect ink so songs and stories can be written again.
> Musician: But we must be careful, InkDrinkers dwell there!
> Musician: The melody awakened the memory of song. Collect the Memory threads before you go!
> => END

However this dialogue only plays if you go and talk to the musician after solving the puzzle. If you don't, and go straight to the collectible, then you only get this dialogue:

> ~ well_done
> StoryWeaver: Ink Well, here we come!
> => END

I think the first block of dialogue should play as soon as you solve the puzzle. If you talk to the musician after that, the last line of dialogue should play (or perhaps just the last sentence).

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

wjt:
I tried having the musician spontaneously talk to you but I think it's simpler to just reveal the thread from the dialogue, and have the player go and talk to them manually. The new (since this ticket was filed) "puzzle solved" animation on the musician is a clear signal that something has happened.

PR #1972, opened 2026-02-23
Ink combat dialogue: Use "spray" term for the enemy action

manuq:
Revert 715a11a8 and use "spray" instead.

Helps https://github.com/endlessm/threadbare/issues/1958

### DIFF
--- a/scenes/quests/lore_quests/quest_001/2_ink_combat/components/dialogues/ink_well.dialogue
+++ b/scenes/quests/lore_quests/quest_001/2_ink_combat/components/dialogues/ink_well.dialogue
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 ~ start
 Musician: The songwriters wrote such colorful songs, before. But the InkDrinkers consumed all of the ink. Restore the ink to the matching jars.
-Musician: But be careful, they splat!
+Musician: But be careful, they spray ink!
 => END
 ~ well_done
 Musician: Well done StoryWeaver, you have collected enough ink for new songs and stories. Everyone will rejoice.


scenes/quests/lore_quests/quest_001/2_ink_combat/components/dialogues/ink_well.dialogue
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 ~ start
 Musician: The songwriters wrote such colorful songs, before. But the InkDrinkers consumed all of the ink. Restore the ink to the matching jars.
-Musician: But be careful, they splat!
+Musician: But be careful, they spray ink!
```
manuq, line 5: I find the "ink" word appearing way too many times in 2 lines of dialogue. But it was the case before too.

> Musician: The songwriters wrote such colorful songs, before. But the **Ink**Drinkers consumed all of the **ink**. Restore the **ink** to the matching jars.
> Musician: But be careful, they spray **ink**!
>

wjt: The funny thing is: most people won't see this dialogue. https://github.com/endlessm/threadbare/issues/1347

manuq: > The funny thing is: most people won't see this dialogue. #1347

This is the intro dialogue of `ink_combat_round_1` and is mandatory.

wjt: Oh, sorry, I mixed up the dialogue files.

PR #1973, opened 2026-02-23
Replace throwing enemy action sound effect

manuq:
Rename it from "spit" to "spray". And move it to the first-party directory. Removing the previous license information.

Also adjust the AnimationPlayer node in the enemy scene to delay a bit playing this sound effect, so it matches the frame-by-frame animation.

The sound was recoded by @PlayMatters as InkDrinker_AirPop_01.wav. So attribution is Threadbare's Authors.

There are unrelated changes in ink combat rounds 4, 5, 6 because of the recent projectile changes.


assets/first_party/sounds/throwing_enemy/Spit.ogg
wjt, line 1: Let's rename this file? It's no longer a "spit".

assets/first_party/sounds/throwing_enemy/Spit.ogg.import
```diff
@@ -0,0 +1,19 @@
+[remap]
+
+importer="oggvorbisstr"
+type="AudioStreamOggVorbis"
```
wjt: There is a school of thought that sound effects should be `.wav`, which Godot compresses, lossily, at export, to a format which is computationally cheaper than Ogg Vorbis to decompress. How big was the .wav of this sound effect?

(The existing sound effect was `.ogg` so this is not a merge blocker of course.)

assets/first_party/sounds/throwing_enemy/Spit.ogg
manuq: Oh I thought I had renamed it as the commit message says, but taking a look today, clearly I didn't :) . Fixup coming.

assets/first_party/sounds/throwing_enemy/Spit.ogg.import
manuq: Ah yes, I remember now https://github.com/endlessm/threadbare/wiki/Audio-Style-Guide . I'll use the original WAV then.

manuq:
@wjt thanks for the feedback. Ready for review again!

PR #1976, opened 2026-02-24
Video settings: Use text property of CheckButton

wjt:
Video settings: Use text property of CheckButton

Previously the "Fullscreen" label for the corresponding toggle was a
separate label. But CheckButton has a built-in text property. Using this
has the advantage that clicking the label toggles the button; and that
the focus rectangle covers the label as well as the button.

Also define the hover_pressed style for CheckButton: without it the
label alignment changes in the case where the button is both enabled,
and being hovered over.

### DIFF
--- a/scenes/menus/options/components/video_settings.tscn
+++ b/scenes/menus/options/components/video_settings.tscn
@@ -23,13 +23,10 @@ text = "Video Settings"
 layout_mode = 2
 columns = 2
 
-[node name="MusicLabel" type="Label" parent="GridContainer" unique_id=353222812]
-layout_mode = 2
-text = "Fullscreen"
-horizontal_alignment = 2
-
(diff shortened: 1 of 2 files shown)


scenes/menus/options/components/video_settings.tscn
```diff
@@ -23,13 +23,10 @@ text = "Video Settings"
...
 
-[node name="MusicLabel" type="Label" parent="GridContainer" unique_id=353222812]
-layout_mode = 2
-text = "Fullscreen"
-horizontal_alignment = 2
-
 [node name="CheckButton" type="CheckButton" parent="GridContainer" unique_id=409098937]
 layout_mode = 2
+size_flags_horizontal = 3
+text = "Fullscreen"
```
manuq, line 29: Text property of CheckButton, excellent!

Issue #1934, opened 2026-02-18
Prototype unlocking a new area based on completed quests


wjt:
We want to have more areas in the world unlock as you progress through the game.

To start with, I will prototype this by building some areas attached to Fray's End that unlock when certain conditions are met:

1. A particular quest has been completed
2. Some number of quests (?) have been completed

PR #1984, opened 2026-02-25
Add a new area of Fray's End unlocked by completing the Void quest

wjt:
Resolves https://github.com/endlessm/threadbare/issues/1934


scenes/globals/game_state/game_state.gd
```diff
@@ -230,6 +229,26 @@ func abandon_quest() -> void:
 	clear_inventory()
 
 
+## Updates [member completed_quests] to include [param quest] if [param
+## is_completed] is true, or remove [param quest] if [param is_completed] is
+## false.
+func set_quest_completed_state(quest: Quest, is_completed: bool) -> void:
```
wjt, line 243: `gdlint` is sad that this class has so many public methods. It does have a point!

scenes/world_map/components/ToggleableTileMapLayer.gd
```diff
@@ -0,0 +1,53 @@
...
+##
+## A [Toggleable] that disables a [TileMapLayer] when enabled, and enables it
+## when disabled. Use this, for example, together with [QuestProgressUnlocker]
+## to block off sections of Fray's End with water or void until the player has
+## completed certain quests. (Depending on the level design, a door may be more
+## appropriate.)
+## [br][br]
+## This script may be attached to a [TileMapLayer] directly; instantiated as a
+## child of a [TileMapLayer]; or instantiated elsewhere in the tree, in which
+## case [member target] must be set.
```
wjt, line 16: I think this is probably overengineered but I saw an opportunity to use my new favourite hammer. Toggleable should be a trait; GDScript doesn't support traits; but Toggleable and TileMapLayer have a common ancestor type, so this script can be attached to a TML. But it appears in the "add node" panel so I wanted it to support the "behavior" pattern we use elsewhere too.

scenes/world_map/components/gardener_a.dialogue
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+if Unlocker.is_satisfied():
+	Marie: You must have been to visit Moss!
+	Marie: We saw the Void retreating from the West End path!
+else:
+	Marie: Oh, are you trying to get to the West End?
+	Marie: That path has been blocked by the Void for years.
+	Marie: Our cat is really upset about it. She used to love going over there.
```
manuq, line 10: I love the cat reference.

scenes/world_map/components/ToggleableTileMapLayer.gd
manuq: This is interesting, as you said it can act as a behavior node or added directly to the node.

PR #1988, opened 2026-02-25
stella: Improve starfish collision shape & let it spin

wjt:
stella: Improve starfish collision shape & let it spin

Previously the starfish projectile sprite was offset slightly from the
centre, and the collision shape was inherited from the generic
projectile.

Zero the positions. Give it a star-shaped collision polygon. Unset
lock_rotation so that it can spin when ricocheting off obstacles.

I made this change while demonstrating to Justin why the new approach of
having separate projectile scenes is worth the trade-off of having more
copies of very similar scenes, then decided it is actually worth pushing.

### DIFF
--- a/scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_projectile_starfish.tscn
+++ b/scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_projectile_starfish.tscn
@@ -7,27 +7,19 @@
 friction = 0.0
 bounce = 1.0
 
-[sub_resource type="RectangleShape2D" id="RectangleShape2D_o6xl0"]
-size = Vector2(44, 44)
-
 [node name="StellaProjectileStarfish" type="RigidBody2D" unique_id=1611529190 groups=["projectiles"]]
 collision_layer = 256
 collision_mask = 80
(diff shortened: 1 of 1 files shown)


wjt:
https://play.threadbare.game/branches/endlessm/wjt/stella-improve-starfish-collision-shape-let-it-spin/#quests/story_quests/stella/2_stella_combat/stella_combat

scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_projectile_starfish.tscn
```diff
@@ -7,27 +7,19 @@
...
-[sub_resource type="RectangleShape2D" id="RectangleShape2D_o6xl0"]
-size = Vector2(44, 44)
-
 [node name="StellaProjectileStarfish" type="RigidBody2D" unique_id=1611529190 groups=["projectiles"]]
 collision_layer = 256
 collision_mask = 80
 mass = 0.3
 physics_material_override = SubResource("PhysicsMaterial_tbgi4")
 gravity_scale = 0.0
-lock_rotation = true
```
manuq, line 19: 💯

PR #1989, opened 2026-02-26
Add rain overlay effect

manuq:
A rain effect that can be used in levels. Like the clouds shadow or fog visual effects, it has a randomize method.

### DIFF
--- a/scenes/game_elements/fx/rain/components/rain.gd
+++ b/scenes/game_elements/fx/rain/components/rain.gd
@@ -8,6 +8,10 @@ extends CanvasLayer
 @export var _seed: int:
 	set = _set_seed
 
+## The probability distribution for the amount of rain. The y-axis is the amount
+## of rain for a given (randomly-chosen) point on the x-axis.
+@export var rain_distribution: Curve
+
 var _random_number_generator := RandomNumberGenerator.new()
 
 @onready var gpu_particles_2d: GPUParticles2D = %GPUParticles2D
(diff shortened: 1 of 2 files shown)


scenes/game_elements/fx/rain/rain.tscn
```diff
@@ -0,0 +1,69 @@
...
+	if rain_type < 0.25:
+		# 25% chances of having a drizzle.
+		amount_ratio = randf_range(0, 0.1)
+	elif rain_type >= 0.75:
+		# 25% chances of having a steady rain.
+		amount_ratio = randf_range(0.8, 1)
+	else:
+		# 50% chances of having something in between.
+		amount_ratio = randf_range(0.1, 0.8)
+"
```
wjt: I think you meant to adjust whatever uses this to use the standalone `rain.gd` script?

scenes/game_elements/fx/rain/components/rain.gd
```diff
@@ -0,0 +1,18 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends CanvasLayer
+
+@onready var gpu_particles_2d: GPUParticles2D = %GPUParticles2D
+
+
+func randomize() -> void:
+	var rain_type := randf()
```
wjt: Image

wjt: I don't think you need to change this, but you could instead represent this as a Curve, domain and codomain both in the range [0, 1]; then you pick a random number between 0 and 1, index the curve with it, and the amount_ratio is the y-axis value; and the curve could be edited visually.

manuq: Oh that is a great idea!

scenes/game_elements/fx/rain/rain.tscn
manuq: Aha good catch! Thanks.

scenes/game_elements/fx/rain/components/rain.gd
wjt: I implemented this at https://github.com/endlessm/threadbare/commit/f12befe4e5a41e5d9abc05ae05ad7112846f94c5. I'm actually not entirely convinced it is better!

scenes/game_elements/fx/rain/components/rain.gd
```diff
@@ -0,0 +1,18 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends CanvasLayer
+
+@onready var gpu_particles_2d: GPUParticles2D = %GPUParticles2D
+
+
+func randomize() -> void:
```
wjt: If one wanted to have a static amount of rain in a level, rather than a random amount, what would one do?

You could imagine refactoring this to have a `rain_amount` exported property that gets propagated through to the particles, and then the `randomize` function would set that property rather than directly setting `gpu_particles_2d.amount_ratio`.

But I'm not sure it's necessary for now.

scenes/game_elements/fx/rain/components/rain.gd
manuq: > I implemented this at [f12befe](https://github.com/endlessm/threadbare/commit/f12befe4e5a41e5d9abc05ae05ad7112846f94c5). I'm actually not entirely convinced it is better!

Oh cool let me check. I was also resuming work on this branch:

- Added fade_in() fade_out() methods which may be called (duck-typing) from TimeAndWeather. I verified it works if I make it rain instead of be foggy in the night.
- Added same Randomize tool button with seed value persisted as the other overlays.

And was about to jump into the curve idea. Good timing! I'll check yours first.

PR #1991, opened 2026-02-26
Add time of day to Fray's End

manuq:
## GameState: Add lights on/off state

A (non persisted) state used to communicate when lights
should be turned on or off between components.

----

## Add TimeAndWeather node
A drop-in node that adds a day-night cycle animation. Using
post-processing and overlay visual effects.

It's possible to continue with the same time-of-day in the next scene.

Nothing is persisted. Instead, the in-game time is derived from the
system time.

If two scenes have TimeAndWeather nodes with "use system time" enabled
and same setting for time scale, they'll be in sync. See the weather
museum scene for an example.

Duck typing is used to call methods in the visual effects:
randomize, fade_in, fade_out.

For now every day is the same cloudy and every night is the same foggy.
Later more weather conditions and randomization can be added.

This is crying for sound effects too!

----

## Add ArtificialLightBehavior

A behavior node that can manage lights and turn them
on/off according to the game state.

----

## Player: Use artificial light

No delay because we want the area around the player to immediately
be illuminated.

----

## Houses and loom: Add artificial light

For houses, add some random delay so there is variation in
stacks of houses.

----

## Add ModulateAsSkyBehavior

To be used in CanvasItem nodes with reflective surfaces.

----

## Add time and weather museum

Two scenes to show how the next scene continues at the same time.

----

## Fray's End: Use TimeAndWeather

Add a TimeAndWeather node with the default settings. And
remove the clouds shadow overlay effect, because the new
node provides it now.

Add a ModulateAsSkyBehavior to the water tilemap layer.

----

Initial prototype: https://github.com/endlessm/threadbare/pull/1913

### DIFF
--- a/scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
+++ b/scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
@@ -33,7 +33,7 @@ const SECONDS_PER_DAY: float = 24 * 60 * 60
 ## - 16: 4 PM.[br]
 @export_range(0.0, 24.0, 0.1, "suffix:h") var start_time: float = 10.0
 
-## How fast should time pass in the game.[br]
+## The number of game-days in a real-world day.[br]
 ## - 1: One day in game matches one day in reality. Don't do this![br]
 ## - 24: One hour in game matches one day in reality.[br]
 ## - 144: 10 minutes in game matches one day in reality (default).[br]
@@ -85,6 +85,7 @@ func _set_time_scale(new_time_scale: float) -> void:
 	# The current animation length (in seconds) corresponds to 1 day.
 	var animation_scale := SECONDS_PER_DAY / animation_player.current_animation_length
 	animation_player.speed_scale = time_scale / animation_scale
+	set_time(get_current_time())
 
 
 func _set_show_debug_label(new_show_debug_label: bool) -> void:


scenes/globals/game_state/game_state.gd
```diff
@@ -46,6 +51,9 @@ const TRANSIENT_SCENES := [
 ## Current number of lives the player has.
 var current_lives: int = MAX_LIVES
 
+## Current state of artificial lights.
+var lights_on: bool
```
wjt, line 55: Interesting that this (correctly) adds another non-persisted field to the state. `intro_dialogue_shown` is already not persisted.

To me this suggests that we should (later) split the state up into:

- This singleton
- A resource holding the persisted state
  - and some functions to load and save it
- Another resource holding transient state

Don't change it in this branch!

scenes/game_elements/fx/time_and_weather/time_and_weather.tscn
```diff
@@ -0,0 +1,253 @@
...
+	if time > 5 and time <= 12:
+		period = \"morning\"
+	elif time > 12 and time <= 18:
+		period = \"afternoon\"
+	elif time > 18 and time <= 22:
+		period = \"evening\"
+	else:
+		period = \"night\"
+	text = \"%.1f %s\" % [time, period]
+"
```
wjt: Did you mean to leave this here? It's just a debug node so probably OK to have it be embedded in the scene.

scenes/game_logic/light2d_behaviors/artificial_light_behavior.gd
```diff
@@ -0,0 +1,95 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name ArtificialLightBehavior
+extends Node
+## @experimental
+##
+## Make the light automatically turn OFF during daytime and ON during nighttime,
+## according to the game state, for energy-saving and cosmetic reasons.
```
wjt, line 9: > energy-saving

I love it

scenes/game_logic/canvasitem_behaviors/modulate_as_sky_behavior.gd
```diff
@@ -0,0 +1,47 @@
...
+	if Engine.is_editor_hint():
+		return
+	time_and_weather = get_tree().current_scene.get_node_or_null("TimeAndWeather")
+	if not time_and_weather:
+		process_mode = Node.PROCESS_MODE_DISABLED
+
+
+func _process(_delta: float) -> void:
+	if Engine.is_editor_hint():
+		return
```
wjt: Something I learnt recently is that it is better to use `set_process(false)` in `_ready` rather than have this guard here, because calling a method 60 times per second in the editor has a cost even if it just `return`s immediately.

(This could be fixed later.)

scenes/game_elements/fx/time_and_weather/components/2_time_and_weather_museum.tscn
```diff
@@ -0,0 +1,150 @@
...
+[node name="House_3" parent="OnTheGround" unique_id=1887721550 instance=ExtResource("8_cvgjt")]
+position = Vector2(539, 150)
+texture = ExtResource("10_h3871")
+
+[node name="Sign" parent="OnTheGround" unique_id=1579530966 instance=ExtResource("11_2owc0")]
+position = Vector2(307, 387)
+direction = 1
+text = "Use rocks to set the time of day.
+Warning! This will unsync it with
+the scene at west."
```
wjt, line 94: Rocks as buttons! 😀

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
```diff
@@ -0,0 +1,208 @@
...
+## - 16: 4 PM.[br]
+@export_range(0.0, 24.0, 0.1) var start_time: float = 10.0
+
+## How fast should time pass in the game.[br]
+## - 1: One day in game matches one day in reality. Don't do this![br]
+## - 24: One hour in game matches one day in reality.[br]
+## - 144: 10 minutes in game matches one day in reality (default).[br]
+## - 1440: One minute in game matches one day in reality.[br]
+## - 3600: 24 seconds in game matches one day in reality.[br]
+@export_range(1.0, 3600.0, 1.0) var time_scale: float = 144.0:
```
wjt: I find these numbers a bit hard to think about.

One option is to describe this number in a different way: this number is the number of game-days in a real-world day! I only realised this while writing this comment.

The alternative would be to store the reciprocal: the length of an in-game day, in seconds. So the new default would become `600.0` seconds.

Or store the fractional number of real-world minutes in an in-game day, so the new default would be `10.0` minutes.

We don't have to change this - I'm mainly thinking aloud.

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
```diff
@@ -0,0 +1,208 @@
...
+
+func _set_time_scale(new_time_scale: float) -> void:
+	time_scale = new_time_scale
+	if Engine.is_editor_hint():
+		return
+	if not animation_player:
+		return
+	# The current animation length (in seconds) corresponds to 1 day.
+	var animation_scale := SECONDS_PER_DAY / animation_player.current_animation_length
+	animation_player.speed_scale = time_scale / animation_scale
```
wjt, line 87: Adjusting this while the game is running does not update the various timers for the new time scale. I think that's OK but it took me a little while to understand what was going on.

I think this would fix it but haven't tested it: re-set the time when changing the scale.

```gdscript
	_set_time(get_current_time())
```

scenes/game_elements/fx/time_and_weather/time_and_weather.tscn
manuq: Hmm maybe I move this script to `components/`.

scenes/game_logic/canvasitem_behaviors/modulate_as_sky_behavior.gd
manuq: Good point. I'll fix it before merging!

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
manuq: This is the feedback I most welcome, because it's the part I had more doubts. Let me think about it.

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
manuq: Good point, let me try it.

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
wjt: I wonder if we can have both... Have one real property that is adjustable and stored; and another which is not stored, readonly, shown in the editor, that converts to the other number. So you adjust e.g. this time_scale factor, and below it you would see "In-Game Day Length: 10.0 minutes".

wjt: That's almost certainly overengineering though!

scenes/globals/game_state/game_state.gd
manuq: Sounds like a plan. It has become a bag of cats. The separation between persisted and transient is a great idea.

scenes/game_elements/fx/time_and_weather/time_and_weather.tscn
manuq: Done!

scenes/game_logic/canvasitem_behaviors/modulate_as_sky_behavior.gd
manuq: Done! I remember what happened here. Yesterday I had:
```
func _ready() -> void:
	if Engine.is_editor_hint():
		process_mode = Node.PROCESS_MODE_DISABLED
		return
```
But this effectively disabled the node in the editor, and the scene was saved with the node disabled. So I removed it. Using `set_process(false)` is the correct thing to do.

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
manuq: > One option is to describe this number in a different way: this number is the number of game-days in a real-world day! I only realised this while writing this comment.

I've been scratching my head and haven't thought about this. Well wording!

scenes/game_elements/fx/time_and_weather/components/time_and_weather.gd
manuq: Your single line does the trick. I confirmed with the running game and changing the time scale in the Remote tree dock.

Issue #1974, opened 2026-02-24
Integrate Void music


wjt:
John Wright has composed three new pieces for the Void quest which need to be integrated.

PR #2003, opened 2026-03-02
Integrate Void quest music

wjt:
Add three new tracks by John Wright: Still, Creep, and Bloom.

Roughly speaking, I used Still for the sections in void_runner and
void_grappling when no enemy is chasing you, and Bloom during the chase
sequences. Creep is played throughout the grappling levels, which to me
suits the spirit of discovery in a strange place.

Creep is split into three sections: Intro (first four bars), Outro (last
four bars), and Loop (everything in between). This is because it sounds
weird to me to loop the whole piece.

I created an AudioStreamInteractive with these five clips, and a number
of transitions.

For void_runner and void_grappling: transitions between Still and Bloom
return to the last position in the target clip. This is used when
switching between the two (when you trigger the enemy, and either are
caught or reach the goal) in void_runner. To me, it sounds bad if you
return to the start of each track each time. (The other option would be
to keep playing Bloom once the chase is triggered, even across respawns,
until you reach safety.)

There is a second clip of Bloom, "Bloom 2", which is used in
void_grappling. This is because there is no way to reset the hidden
"previous position" state; but when we get to that third challenge we
would really like to start Bloom from the start again.

For Creep:

- Intro auto-advances to Loop: i.e. if Intro finishes playing
  "naturally", switch to Loop.
- If a transition from Intro to Loop is explicitly requested, wait till
  the end of Intro anyway then start Loop with no fade.
- When a transition from Loop to Outro is requested, wait for Loop to
  finish before starting Outro (with no fade).
- When Outro finishes, switch to Still.
- If we are still in Loop when it comes time to switch to Still, just do
  a quick cross-fade.

The first round of grappling triggers Creep Intro at level start. Since
this is only four bars long, it's expected that the music will switch to
the Creep Loop before you clear this level. (Another way to achieve this
would be to have the intro & body of the track in one file, and set the
loop_offset field to the timestamp corresponding to bar 5. This would be
simpler, but unfortunately loop_offset is not supported on the web
platform, hence this workaround.)

All subsequent levels trigger Creep Loop if it is not already playing,
until grappling_hook_end, which triggers a transition to Creep Outro.
The loop is about 2 minutes long so it may not finish in time. If it
does, we get a nice 4-bar outro, and then auto-advance to Still, ready
for the start of void_grappling. This last transition doesn't sound
quite perfect but I think it's fine. It would be even better with a
filler clip, perhaps just the downbeat of Creep Outro.

All these files are compressed with Ogg Vorbis -q4, approximately
128kbps.

Add new API to BackgroundMusic to make it simpler to trigger clip
switches within a scene, including a ClipSwitcher child node that can be
connected to an Area2D. Make these triggers one-shot by default. This
avoids a problem where retracing your steps in void_runner back to the
starting island would re-trigger Bloom, even though no chase is
occurring.

Resolves https://github.com/endlessm/threadbare/issues/1974

---
  - https://app.asana.com/0/0/1212154478746738


assets/first_party/music/The_Void_Still.ogg.import
```diff
@@ -0,0 +1,19 @@
...
+[deps]
+
+source_file="res://assets/first_party/music/The_Void_Still.ogg"
+dest_files=["res://.godot/imported/The_Void_Still.ogg-7e65310155f40ce2f414bd20a7e726cd.oggvorbisstr"]
+
+[params]
+
+loop=true
+loop_offset=0.0
+bpm=0.0
```
wjt, line 17: TODO

scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
```diff
@@ -61,7 +65,20 @@ metadata/_custom_type_script = "uid://bgmwplmj3bfls"
...
 
 [node name="BackgroundMusic" parent="." unique_id=649811497 instance=ExtResource("1_g2rto")]
-stream = ExtResource("2_g2rto")
+stream = ExtResource("2_kgem8")
+clip = &"Still"
+
+[node name="ChaseClipSwitcher" type="Node" parent="BackgroundMusic" unique_id=1356095710 node_paths=PackedStringArray("trigger")]
+script = ExtResource("3_epkxq")
+clip = &"Bloom"
+trigger = NodePath("../../EnemyTrigger")
```
wjt, line 74: Ah, I found a problem: if you retrace your steps and walk over the trigger, you re-trigger the transition to the chase music!

Needs re-thinking.

wjt: Resolved this by making `ClipSwitcher`'s signal connection to the Area2D ONE_SHOT by default.

PR #2007, opened 2026-03-03
grappling_hook_powerup: Fix shadowing warning

wjt:
grappling_hook_powerup: Fix shadowing warning

Playing this scene would issue this warning from line 18:

> The local variable "player" is shadowing an already-declared variable
> at line 9 in the current class.

There is a member variable `player` holding a reference to `%Player`, so
there is no need to look up the player again in the "collected"
callback.


manuq (approved):
Awesome!

PR #2008, opened 2026-03-03
Add wildlife characters

manuq:
And use them in the Void quest `closing_transition.tscn` outro.

----

## Add wildlife assets

Took the ZIP files delivered by PixelFrog, simplified the nesting,
removing MACOS-specifics, and converting the filenames to snake_case.

----

## Create SpriteFrame resources for bunny, capycorn, and sheep idle animations.

Using the PNGs containing the frame-by-frame spritesheet animations from
`assets/first_party/wildlife/`, create SpriteFrame `.tres` resources for each
animal color variant.

Steps:
- Read the PNG dimensions to determine frame size (128x128) and frame count
  (14 for bunny, 11 for capycorn, 29 for sheep)
- Extracted the texture UIDs from the `.import` files
- Generated `.tres` files following the existing storyweaver SpriteFrame format,
  with AtlasTexture sub-resources slicing each spritesheet into individual frames
- Each resource contains a single "idle" animation, looping at 10 FPS

----

## Void closing transition: Add animals

To both the occluded part and the town. Compose them around the
player and townies, which should be the where the player's eye
should be guided. Add variation by selecting SpriteFrames of
different colors and by randomizing the start frame.

Also extend the reveal animation duration by 4 seconds to give
more time to the player to catch it.

----

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/scenes/quests/lore_quests/quest_002/4_closing_transition/closing_transition.tscn
+++ b/scenes/quests/lore_quests/quest_002/4_closing_transition/closing_transition.tscn
@@ -25,12 +25,12 @@
 [ext_resource type="SpriteFrames" uid="uid://2ek86nvw6y28" path="res://scenes/game_elements/props/tree/components/tree_spriteframes_yellow.tres" id="20_7wffi"]
 [ext_resource type="PackedScene" uid="uid://crqjcicx0vdu" path="res://scenes/game_elements/props/decoration/bush/bush.tscn" id="21_y3es2"]
 [ext_resource type="SpriteFrames" uid="uid://bapks76u4hipj" path="res://scenes/game_elements/props/decoration/bush/components/bush_spriteframes_green_small.tres" id="22_2yns8"]
-[ext_resource type="SpriteFrames" uid="uid://cshxugl83a6kf" path="res://scenes/game_elements/characters/components/sprite_frames/capycorn_purple.tres" id="23_2yns8"]
+[ext_resource type="SpriteFrames" uid="uid://we3g65bpiba4" path="res://scenes/game_elements/characters/components/sprite_frames/capycorn_purple.tres" id="23_2yns8"]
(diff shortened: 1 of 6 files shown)


manuq:
@wjt I didn't add SpriteFrame resource files for the butterfly because I think that animation can be done in Godot. And the butterfly and flower layers should be exported separated from Aseprite.

scenes/quests/lore_quests/quest_002/4_closing_transition/closing_transition.tscn
```diff
@@ -514,6 +520,55 @@ sprite_frames = ExtResource("22_2yns8")
...
+y_sort_enabled = true
+
+[node name="AnimatedSprite2D6" type="AnimatedSprite2D" parent="OnTheGround/Animals" unique_id=668378154]
+position = Vector2(936, 216)
+sprite_frames = ExtResource("25_c4nxe")
+animation = &"idle"
+autoplay = "idle"
+flip_h = true
+
+[node name="RandomFrameSpriteBehavior" type="Node2D" parent="OnTheGround/Animals/AnimatedSprite2D6" unique_id=290214502 node_paths=PackedStringArray("sprite")]
```
wjt, line 533: What's weird is - this doesn't seem to be working, the sheep are mostly synchronised. But I can't see why.

manuq: Oh good catch! It is because the sprite property is not the parent in each RandomFrameBehavior. My error was in duplicating them in the editor.

manuq: @wjt I just appended a fixup. It's weird but the UIDs changed too.

PR #2012, opened 2026-03-04
Update Still BPM and transitions

wjt:
Set The Void Still's BPM to 52.

Previously the transition from Still to Bloom/Bloom 2 was a crossfade
over 2 "beats" of whatever Godot defaults to when a clip's BPM is not
set (seconds, I think); change this to 1 beat for a slightly snappier
sound.

Previously the transition from Still to Creep Intro was immediate at
whatever point in Still was playing. Change this to transition at the
next bar of Still. Start Creep Intro immediately and fade out Still for
1 beat.


manuq:
I did another playthrough and sounds as great as before. I couldn't notice a difference, it might be subtle.

PR #2014, opened 2026-03-04
FollowWalkBehavior: Stop when target is reached

manuq:
Add a boolean property to stop the character when the target is reached, and set it to enabled by default.

Previously the behavior was that the controlled character stays walking around the target. This is still possible by setting the new property to false.


manuq:
I have setup a scene with 3 characters for testing. One is controlled with input, the second follows the first, and the third follows the second. I have removed the 2nd and 3rd from any collision layers, so they are basically ghosts.
  
This is the existing behavior: 

https://github.com/user-attachments/assets/9f2e6662-20b5-4ebc-8aa5-0f095e15e3b1

This would be the new default:

https://github.com/user-attachments/assets/d14cb571-8f16-4332-a8c7-9d44eecc5d10

It allows a crew of characters like seen in some RPGs.

It was very nice to make this change with the game running ("hot reloading"). The two videos are taken in the same run, during which I changed the script.

PR #2015, opened 2026-03-04
Add fence tileset

manuq:
The wildlife assets came with this. Move it to the corresponding directory and create a tileset for the fence. With a wall collision layer and a terrain to draw it.


wjt:
> Should this be part of elevation_2.tres instead?

Personally I don't think so. One thing we could do is to set the tilesetsource ID to a different number to the one in elevation_2 in case we want to merge them later.

> I decided to not add an occlussion on purpose. Guards can see you behind a fence!

I think that's correct! But will they be able to navigate around the fence to chase you? 😁

Issue #1375, opened 2025-10-17
Revisit player modes


manuq:
Currently the player is in one of these modes:
- Cozy
- Fighting
- Hooking
- Defeated

"Cozy" let's the player interact with other things or creatures in the world using the `interact` input action. Talk with an Elder, read a sign post, tap a musical rock.

"Fighting" allows the player to repel projectiles using the `repel` input action.

"Hooking" allows the player to throw the grappling hook using the `throw` input action, and to aim with the `aim_*` input actions or with the mouse.

Some levels currently switch the player modes just to let them pick a collectible. Sometimes after an event (like [after filling all barrels](https://github.com/endlessm/threadbare/blob/34992ef67d802b4da3ebec9c09e8bb2c7d5afe8c/scenes/game_logic/fill_game_logic.gd#L78)). Sometimes in [even more hacky ways](https://github.com/endlessm/threadbare/blob/34992ef67d802b4da3ebec9c09e8bb2c7d5afe8c/scenes/quests/lore_quests/quest_002/2_grappling_hook/components/switch_mode_area.gd). This may lead to confusion. The same joypad buttons or keys are mapped to all "interact / repel / throw (except for "throw" in joypad, but you get the idea). Proposal:

- Make the player always interact, using a dedicated key or joypad button.
- Make the repelling and grappling hook the actual two player ~~modes or "weapons"~~ abilities. Considering that more abilities can be added in the future. ~~Only one weapon~~ Multiple abilities can be active at the same time.

wjt:
I would like to try allowing all three of interact/repel/grapple simultaneously. I take your point, though, that this wouldn't scale to more "weapons".

wjt:
Straw-man idea:

- Interact: keyboard Space, xbox A (bottom action button)
- Repel: keyboard B (mnemonic "Block"), mouse right-click, xbox B (right action button)
- Grapple: keyboard G (mnemonic "Grapple"), mouse click, xbox R2/RT (as it is today)

Or another control scheme for interaction I saw in Death by Scrolling: to talk to someone you just bump into them.

wjt:
I think if we added an offensive attack action, that could be switched with grappling. It would seem weird to not be able to use the shield/button if StoryWeaver is carrying it all the time. So then the mode would be: what does my needle do?

Issue #806, opened 2025-06-17
Add touch input controller


manuq:
Add a touch input system to make the player walk, run, and trigger an action.

It is common in mobile to add a virtual controller for moving in one screen side (left side, for example), and one or more actions in the other side (right side, for example).

For the movement controller:
- A circle appears whenever the user press and hold the screen, with the pressed position as the center of the circle.
- While holding, moving the finger far from the circle center makes the player walk in that direction.
- When the finger reaches the border of the circle, the player runs.
- Releasing the finger the player stops.

Currently the sole action for interacting (in cosy mode) and attacking (in combat mode) is "ui_accept". But that may change by this task. In that case, a tap anywhere in the right side could trigger the action.

Could an addon like [Godot Virtual Joystick](https://github.com/MarcoFazioRandom/Virtual-Joystick-Godot) suit our needs, without being a big dependency to maintain? This one also gives a vector `Input.get_vector(&"ui_left", &"ui_right", &"ui_up", &"ui_down")` from the virtual controller.

PS Sokoban is another story, maybe we implement that one in a separate issue? One idea is to swipe up, down, left and right for moving in the player inside the sokoban.

manuq:
De-assigning after no response.

PR #2032, opened 2026-03-06
Switch from player modes to simultaneous player abilities

manuq:
### Context

Prior to this change, the player actions (besides walking, running) depended of which mode the Player instance was:
- "Cozy" let the player interact with NPCs and other elements. For example: start a dialogue, collect a magical thread.
- "Fighting" let the player repel projectiles.
- "Hooking" let the player use a grappling hook.
- "Defeated" forbid any action.

https://github.com/user-attachments/assets/0fc6e6af-1968-4d43-bd03-9741958c1112

On one hand this was simple enough, and allowed quests to have levels that focused on a single mechanic. But on the other hand it was too limiting:
- The player needed to switch from one mode to the other in the same level. For example, to collect the magical thread after an ink fight.
- The actions can't be conbined.

It is also confusing to use the same input mapping for different actions during a play run.

Most importantly, modes don't allow player progression. We want the player to gain these abilities during the game. And maybe enhace (even temporarily debuff) these abilities during the game.

The abilities could be gained in different order depending on the player decisions. For example, getting involved in a quest before another. This allows some replayability and non-linearity to the otherwise linear quests.

On the learning / user-contributed content side, a StoryQuest (considering it an isolated narrative unit or minigame) can also have a similar player progression as the main game.

### Changes for the main game

- Add flags for abilities and ability modifiers to the game state.
- Remove player modes. Instead, check the game state to turn on/off mechanics ("repel", "grapple") and mechanic changes (longer thread powerup for "grapple").
- The player starts the game without abilities.
- Grant "repel" in Musician quest, at the beginning of `ink_combat_round_1` (2nd) level.
- Grant "grapple" in the Void quest, at the beginning of `grappling_hook_start` (2nd) level.
- Grant "longer thread" modifier for "grapple", in the Void quest, when collecting the powerup item in `grappling_hook_powerup` (5th) level.
- Grant "longer thread" modifier for "grapple" too in the same Void quest `void_grappling_round_2` last level. Although it should be already given at this point, just to avoid redesigning the levels.

Note: we want these abilities to be powerups that the player collects, not granted like in here. But that will be addressed separately.

https://github.com/user-attachments/assets/a6c65319-e385-4598-b780-8a7c52c88d80

### Changes for StoryQuests

Change for learners: The Player doesn’t have a “mode” to set per-level anymore. The StoryQuests can have player progress through the levels by gaining abilities (“repel”, “grapple”). This could be granted by the level or by collecting an item ([in the future](https://github.com/endlessm/threadbare/issues/2057), but will be the desired option). Learners could potentially create their own player abilities.

To help with transitioning StoryQuests, there is a new PlayerMode node. This holds the same options as before "COZY/FIGHTING/HOOKING" and will set a single ability ("repel" for fighting, "grapple" for hooking) and discard all abilities gained in previous levels. New StoryQuest creators are ancouraged to add player progress to their stories instead of using this node.

### Saved state / Difference between World Map, Lore quests and StoryQuests

There are two saved states. Player abilities are saved globally for the whole game run, and also for the current quest. How this happens depends of the quest being a lore quest or a StoryQuest.

- When a quest is resumed (lore or SQ), the abilities already obtained (or potentially lost) during this quest should be reflected.

For lore quests:

- Lore quests start with a copy of the global abilities.
- When a lore quest is abandoned, the abilities in World Map are the same as before starting. Any obtained (or lost) ability during the abandoned quest are dropped. 
- When a lore quest is completed, the global abilities are updated / copied from the current quest. So any ability obatained (or lost) are reflected in the full game.

For StoryQuests:
- They start with no abilities at all (see PlayerMode node above for how the existing ones are being transitioned).
- When a StoryQuest is abandoned or completed,  the abilities in World Map are the same as before starting.

### New input controls mapping

For keyboard + mouse (best experience):
- WASD for moving
- Shift for sprint
- Space bar to interact
- Mouse move for aiming
- Left click to grapple
- Right click to repel (new)

For just keyboard:
- Arrows for moving + aiming
- Shift for sprint
- Z to grapple (new)
- X to repel (new)
- Space bar to interact

For joypad (let's say XBox):
- Left stick or D-pad for moving
- Right stick for aiming
- LB for sprint
- RB to repel (new)
- RT to grapple
- A to interact

### Example gameplay
- Start the game normally. There should be no repel or grapple abilities.
- Get involved in The Void quest. Reach level 2. The player is now able to grapple.
- Abandon the quest. No grapple anymore.
- Get involved in The Void quest again. This time complete it. The player is now able to grapple in the world map too!
- Get involved in Stella StoryQuest. The player starts with no abilities in 1st (stealth) level.
- Reach Stella level 2 (combat), the player should be able to repel.
- Abandon quest. The player should have the grapple ability.
- Get involved in The Musician quest. The player should be able to grapple during the 1st (sequence puzzle) level.
- Reach level 2 (combat). The player should be able to repel + grapple.
- Complete the quest. The player should now be able to repel + grapple in the world map.

----

Resolve https://github.com/endlessm/threadbare/issues/1375


manuq:
I have written Example gameplay section in the description, and already found a bug. Checking!

scenes/game_elements/characters/player/components/player_hook.gd
```diff
@@ -247,6 +250,9 @@ func pull_string() -> void:
 func stop_pulling() -> void:
 	character.set_collision_mask_value(Enums.CollisionLayers.NON_WALKABLE_FLOOR, true)
 	pulling = false
+	# After pulling, return control to the user.
+	if character.has_method("return_control"):
+		character.return_control(self)
```
manuq, line 255: The alternative to this I was handling is using the mode directly:

```
if character is Player:
	# Return control to the user:
	(character as Player).mode = Player.Mode.PLAYING
```

I guess duck-typing is better so a learner can do their own Player scenes from scratch, not necessarily extending the Player class.

scenes/globals/game_state/game_state.gd
```diff
@@ -257,6 +317,66 @@ func _do_set_quest_completed_state(quest: Quest, is_completed: bool) -> void:
...
+## Enable or disable a player ability.
+## [br][br]
+## This behaves differently outside quests than when involved in a quest.
+func set_ability(ability: Enums.PlayerAbilities, is_enabled: bool) -> void:
+	if is_enabled:
+		if not has_ability(ability):
+			if _use_global_abilities():
+				game_player_abilities |= ability
+			else:
+				quest_player_abilities |= ability
```
manuq, line 355: This seems over-engineered. Could be simplified if we design the game in a way that abilities can be obtained only during lore quests.

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -6,15 +6,13 @@ extends CharacterBody2D
...
-	COZY,
-	## Player is engaged in combat. Player can use combat actions.
-	FIGHTING,
-	## Player is using the grappling hook.
-	HOOKING,
+	## Player is reacting to user input.
+	PLAYING,
+	## Player is being controlled by other means: interacting,
+	## pulling the grappling hook, being put on rails, etc.
+	CONTROLLED,
```
manuq: I hope this CONTROLLED enum is not misunderstood as "controlled by the player/user". It is also confusing that "player" can be the user playing or the player character (common issue in gamedev). If so here is an alternative option:
- PLAYER_CONTROLLED / USER_CONTROLLED # Player is reacting to user input.
- SYSTEM_CONTROLLED # Player is being controlled by other means: interacting, pulling the grappling hook, being put on rails, etc.
- DEFEATED

scenes/menus/debug/debug_settings.tscn
manuq, line 1: I'm not happy with having 2 thin scrolling text areas:

Image

But that was the easiest to do. In the future we could have tabs in the Debug Settings, maybe?

scenes/game_elements/characters/player/components/animation_player.gd
```diff
@@ -6,69 +6,62 @@ const REPEL_ANTICIPATION_TIME: float = 0.3
...
 
+func _on_player_repel_repelling_changed(repelling: bool) -> void:
+	if not repelling:
+		return
+
+	# Repel animation is being played for the first time. So skip the anticipation and go
+	# directly to the action.
+	speed_scale = original_speed_scale
+	play(&"repel")
+	seek(REPEL_ANTICIPATION_TIME, false, false)
```
wjt, line 63: It's not new in this PR but it's a little weird how this animation "starts" in the middle! I can't quite figure out why it is that way around. (No need to change it.)

scenes/globals/game_state/game_state.gd
```diff
@@ -77,6 +98,18 @@ var persist_progress: bool
...
+func _validate_property(property: Dictionary) -> void:
+	match property["name"]:
+		# Treat the player abilities as bit flags.
+		# The @export_flags would be ideal but it expects constant
+		# strings, and we want to use the PlayerAbilities enum keys
+		# as hint strings.
+		# This also requires this script to be a @tool.
+		"lore_player_abilities", "storyquest_player_abilities":
+			property.hint = PROPERTY_HINT_FLAGS
+			property.hint_string = ",".join(Enums.PlayerAbilities.keys())
```
wjt, line 123: I spent some time trying to find an alternative to this (as I would guess you did too) but could not.

I did find this:

```
PROPERTY_USAGE_CLASS_IS_BITFIELD = 512
```

I wonder if that can be used somehow. (Not a blocker.)

scenes/menus/debug/debug_settings.tscn
wjt: Another option would be to have just one ScrollContainer, enclosing both lists of options. But I think tabs are likely better. Let's leave this for now and come back to it.

scenes/quests/lore_quests/quest_002/2_grappling_hook/components/grappling_hook_start.gd
```diff
@@ -2,6 +2,24 @@
...
 func _ready() -> void:
-	GameState.set_ability(Enums.PlayerAbilities.ABILITY_B, true)
+	if _play_cinematic:
+		cinematic.start()
+
+
+func _on_spawn_point_from_powerup_player_teleported() -> void:
+	# The player is coming back from the east, so play the
+	# animation directly. This is actually letting the Void
+	# add tiles.
```
wjt, line 20: Another approach here would be to have a more general way to save state in scenes, then the enemy could remember its position and the Void layer could remember what has been eaten... I think this is fine for now.

scenes/game_logic/player_mode.gd
wjt, line 1: Normally we would name this file `player_mode.gd`.

scenes/game_elements/characters/player/components/player.gd
wjt: I agree that the name is a bit confusing. I slightly prefer:

- USER_CONTROLLED (because we named this script Player not PlayerCharacter -defensible, it's so much shorter!)
- SYSTEM_CONTROLLED
- DEFEATED

More verbose, but unambiguous.

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -177,12 +175,8 @@ func _physics_process(delta: float) -> void:
...
 		return
 
-	# While pulling the grappling hook, the movement is handled in PlayerHook._physics_process.
-	if player_hook.pulling:
-		return
-
-	if player_interaction.is_interacting or mode == Mode.DEFEATED:
-		velocity = Vector2.ZERO
+	# TODO: Use InputWalkBehavior instead, and enable/disable it when the mode changes.
+	if mode in [Mode.CONTROLLED, Mode.DEFEATED]:
```
wjt: This is a nice change. Previously if you trigger dialogue by means other than an interaction while the player is moving, then they stop responding to input but keep their velocity, so would keep walking across the screen! IIRC I ran into this while trying to fix https://github.com/endlessm/threadbare/issues/1347

scenes/game_logic/player_mode.gd
manuq: Oh good catch! I'll fix it.

scenes/game_elements/characters/player/components/player.gd
manuq: Great I'll do the rename in a fixup.

scenes/menus/debug/debug_settings.tscn
manuq: OK! Yes I don't think we have Threadbare theme that supports tabs, I'll check.

scenes/game_elements/characters/player/components/animation_player.gd
manuq: This really needs some explanation. The animation we got from PixelFrog has 1 frame of anticipation, which is great. From [Wikipedia](https://en.wikipedia.org/wiki/Anticipation_(animation)):

> For example, a baseball pitcher must "wind up" before throwing the ball. This is called the anticipation for the main action.

But in videogames the controls feel slow if the action doesn't happen immediately. So the idea is that the action is immediate (starts without anticipation) when the user press the input, but subsequent ones do display the anticipation, to also space the repeaded actions.

I've seen this in videogames, let me come up with a reference.

scenes/globals/game_state/game_state.gd
manuq: Yes I dedicated a while to try different things. I haven't seen a way to avoid hardcoding the strings (which on the other side are generic names, not "repel"). I can also open an issue upstream and first check if there is one.

scenes/game_elements/characters/player/components/player.gd
manuq: Yes at one point I said "wait, the dialogue captures the input, no need to system-control it". But then I saw StoryWeaver continue walking haha.

wjt: That's a downside of handling the input in `_unhandled_input`!

wjt: And as I discovered in https://github.com/endlessm/threadbare/issues/806#issuecomment-3914668785 if we want to use 4.7's VirtualJoystick then we'll have to move the input handling back to `_physics_process`.

scenes/game_elements/characters/player/components/animation_player.gd
manuq: Reference in Super Mario Wonder (is hard to see at normal speed because it's just one frame) https://youtu.be/3kIULxgnXOg?t=274

Also this GDC talk coming from my Youtube history https://youtu.be/PKZJmHrG4Yw?t=832

PR #2033, opened 2026-03-07
mejoras

JeessonYman:


scenes/tileset.tres
wjt, line 16: Please do not modify this shared tileset. I see that you already have a copy in your storyquest folder - ensure that this is used in your quest then revert the changes to this file.

scenes/quest.tres
wjt, line 1: This file is in the wrong folder. It needs to be in your StoryQuest's folder, i.e. in the `quests/story_quests/estation _echo_9` folder.

PR #2040, opened 2026-03-10
Remove teleporter scene

wjt:
teleporter.gd is a subclass of Area2D. teleporter.tscn is just an Area2D
node with teleporter.gd attached and the collision layers configured
correctly.

Previously the game used a mix of instances of teleporter.tscn, and bare
Area2Ds with teleporter.gd attached. In fact grappling_hook_needles.tscn
used both in a single scene.

teleporter.gd is a `@tool` and configures the correct layers on itself,
so I don't really see the value of having the scene. The script is right
there in the "add node" dialogue, and works correctly if used like that.

Replace all instances of the scene with Area2Ds with the script
attached. Delete the scene. Simplify the directory hierarchy.


scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
```diff
@@ -37,7 +37,6 @@
 [ext_resource type="Resource" uid="uid://bykip1b51w8x" path="res://scenes/quests/lore_quests/quest_002/1_void_runner/components/collected.dialogue" id="27_g2rto"]
 [ext_resource type="PackedScene" uid="uid://dv4f232y8w8dv" path="res://scenes/game_elements/props/decoration/water_rock/water_rock.tscn" id="28_oo13r"]
 [ext_resource type="PackedScene" uid="uid://cfcgrfvtn04yp" path="res://scenes/ui_elements/hud/hud.tscn" id="29_gbo6l"]
-[ext_resource type="PackedScene" uid="uid://cf3cbl12wt6vr" path="res://scenes/game_elements/props/buildings/temple/temple.tscn" id="37_kgem8"]
```
wjt, line 40: Hmm this seems wrong.

wjt:
Good point - they should probably be consistent. Personally I prefer the approach here, because it's not possible to get it wrong: rather than a warning, `_ready()` just sets up the layers correctly:

https://github.com/endlessm/threadbare/blob/d8bdbd58b31159a18115ce69e86397305236544f/scenes/game_elements/props/teleporter/teleporter.gd#L44-L48

(It should be changed to use the enum members.)

manuq (approved):

PR #2046, opened 2026-03-11
Overworld prototype

### LINKED ISSUE #2021: Prototype an overworld
We are exploring linking the different areas of the Threadbare together into an overworld/world map. Prototype this. Questions to answer:

- How does it feel?
- What new components would we need?
- Can we effectively gate the player moving around the overworld based on the quests they have completed?
- What impact would this have on existing scenes & quests?

wjt:
Introduce a QuestTeleporter that, when entered, starts a quest. Add
staging areas representing the Song Sanctuary and Linenville with
QuestTeleporters to the musician quest and the void quest respectively.

Take the player back to a defined spawn point when they abandon a quest,
rather than to Fray's End. Save progress within each quest so that if a
quest is abandoned then resumed, the player picks up where they left
off.

Add an overworld, a zoomed-out space for fast travel between regions of
the world. Link Fray's End to the Song Sanctuary and Linenville.
Gate access to Linenville (represented by the temple) with a blocking
area of void, removed when you complete the musician quest. Gate access
to a third island by grappling, enabled only when the void quest is
complete. (Imagine that the placeholder quest can be accessed here.)

Remove the LoreQuest Elder's storybook. Have them give you directions to
the next quest.

Resolves https://github.com/endlessm/threadbare/issues/2021


wjt:
> 2. Mechanically, I think that the collected threads should now be per-quest and tracked that way to just keep it easier for managing quest conditionals. (Possibly an amorphous object?.... In case we ever have a quest not just be retrieving threads?)

When this has been discussed before (in the context of StoryQuests where having these brightly-coloured threads does not mesh with the surrounding art style) the use of threads has been considered sacrosanct. Of course all things can change but just for background!

manuq:
Should we merge the overworld scene plus the hub areas that are the new starting point to existing lore quests? And leave the resuming of quests, which seem to need more discussion, out for now?

wjt:
I'll open some tasks to extract bits of this to mergeable code.

wjt:
Superseded by #2287

PR #2049, opened 2026-03-12
WIP: In-game push blocks

wjt:


push_block.gd
```diff
@@ -0,0 +1,66 @@
...
+	# For this to work, the water layer's physics quadrant size has to be set to
+	# 1, i.e. no chunking of water physics - probably bad for performance...
+	# But if this became a practical problem we could split up the water layer
+	# based on whether it's placed somewhere where blocks can be pushed into it, or not.
+	var coords := tile_map.get_coords_for_body_rid(body_rid)
+
+	# This doesn't work because at the point when the 1×1px area in the centre of a block
+	# travelling rightwards touches a block of water at (x, y),
+	# its centre is still in tile (x - 1, y).
+	# var coords := tile_map.local_to_map(tile_map.to_local(global_position))
```
wjt, line 55: This is why the previous version didn't work. Actually what I originally implemented was the section just above that uses `get_coords_for_body_rid` but then I thought, well, how hard can it be to use `local_to_map`? and didn't even test it before pushing.

wjt:
I have tried several different ways to move the blocks in response to the repel action:

- Keep the blocks as a CharacterBody2D, set velocity in the closest cardinal direction of the repel; decelerate at a constant rate, with the two constants chosen (with some dodgy maths) to move the block almost exactly 64px in that direction.
- Make the blocks be a RigidBody2D. What I couldn't make work here is that we want the player pushing against the body to never move it (as if it's a wall); but I couldn't make that work.

PR #2054, opened 2026-03-16
storybook: Refactor building list of authors

wjt:
Factor out the code that turns ["a", "b", "c"] into "a, b and c". Add an Easter egg.

### DIFF
--- a/scenes/menus/storybook/components/storybook_page.gd
+++ b/scenes/menus/storybook/components/storybook_page.gd
@@ -18,6 +18,40 @@ var quest: Quest = preload("uid://dwl8letaanhhi"):
 @onready var play_button: Button = %PlayButton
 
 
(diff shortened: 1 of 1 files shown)


scenes/menus/storybook/components/storybook_page.gd
```diff
@@ -18,33 +18,50 @@ var quest: Quest = preload("uid://dwl8letaanhhi"):
 @onready var play_button: Button = %PlayButton
 
 
-func _set_quest(new_quest: Quest) -> void:
-	quest = new_quest
+# This is an in-joke/Easter egg for the Endless Access team
+func _chadify(name: String) -> String:
```
wjt, line 22:
```suggestion
static func _chadify(name: String) -> String:
```

PR #2060, opened 2026-03-17
web: Require a click on the page to play the game

wjt:
On the web, you can only begin playing audio in direct response to a
user gesture. This means that the sewing-machine sound effect for the
splash screen does not reliably play, and nor does the music on the
title screen until you happen to click on the game (or press a keyboard
key with the canvas focused).

Change the web export HTML template to replace the engine.startGame()
method with an inlined version of the chain of promises it runs. Add a
new state to the page, with a "Play Threadbare" pseudo-link. Trigger
this state once the game has been loaded; wait for a click before
continuing the promise chain to run engine.start().

Use the same Jersey font as is used in-game for the "Play Threadbare"
link. Load it from Google Fonts, even though we have a copy in this
repo, because it's not immediately obvious how to add additional files
to the web export of the project, and this is just a couple of lines.

Update .editorconfig to use tabs to indent HTML files, matching the
existing indentation of this file.

Resolves https://github.com/endlessm/threadbare/issues/376

### DIFF
--- a/web/full-size.html
+++ b/web/full-size.html
@@ -26,7 +26,7 @@ body {
 	outline: none;
 }
 
-#status, #status-splash, #status-progress {
+#status, #status-splash, #status-progress, #status-ready {
 	position: absolute;
 	left: 0;
 	right: 0;
@@ -44,6 +44,10 @@ body {
 	justify-content: center;
 	align-items: center;
 	visibility: hidden;
+}
+
+#status.is-ready {
+	cursor: pointer;
 }
 
 #status-splash {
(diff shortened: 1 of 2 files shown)


web/full-size.html
```diff
@@ -105,6 +124,7 @@
 			<img id="status-splash" class="$GODOT_SPLASH_CLASSES" src="$GODOT_SPLASH" alt="">
 			<progress id="status-progress"></progress>
 			<div id="status-notice"></div>
+			<div id="status-ready">Play Threadbare</div>
```
wjt, line 127: This is not translated. Should it be?

manuq (approved):
This works! The extra step is a downside for me but I understand the reason.

We can iterate on the click button later, make it translatable, etc.

PR #2064, opened 2026-03-18
linux: Add v0.1.11 and v0.1.10 to releases file

wjt:


.editorconfig
```diff
@@ -12,6 +12,7 @@ indent_size = 4
 [*.{yaml,yml,xml}]
 indent_style = space
 indent_size = 2
+max_line_length = 100
```
manuq: 👍

wjt: Sadly this won't fly because every other XML file in the repo has incredibly long lines. I just reverted this for now.

PR #2070, opened 2026-03-19
github: Add version & scene to bug report template

wjt:
As well as being useful information to add, by making these into form
fields we can have a button in the game that launches the "new bug" form
with the fields pre-filled.


export_presets.cfg
wjt, line 1: This shouldn't be here...

wjt:
I'm going to go ahead and merge this without @endlessm/community review. Some internal discussion suggests we may need to revise `CODEOWNERS`

PR #2071, opened 2026-03-19
Player: Use InputWalkBehavior

manuq:
Keep slowing down the player while the grappling hook is being used. For this, add a new aiming_changed signal to the PlayerHook.

Also use the running_changed signal of InputWalkBehavior instead of checking constantly for:
- Emitting the dust particles.
- Doubling the frame-by-frame animation speed or using a special run animation.

----

Adapt the scenes and scripts in StoryQuests that had the player speeds set.

Resolve https://github.com/endlessm/threadbare/issues/1151

### DIFF
--- a/scenes/game_elements/characters/player/components/player.gd
+++ b/scenes/game_elements/characters/player/components/player.gd
@@ -61,8 +61,7 @@ const DEFAULT_SPRITE_FRAME: SpriteFrames = preload("uid://vwf8e1v8brdp")
 @export var walk_sound_stream: AudioStream = preload("uid://cx6jv2cflrmqu"):
 	set = _set_walk_sound_stream
 
-var _initial_walk_speed: float
-var _initial_run_speed: float
+var _initial_speeds: CharacterSpeeds
 
 @onready var input_walk_behavior: InputWalkBehavior = %InputWalkBehavior
 @onready var player_interaction: PlayerInteraction = %PlayerInteraction
@@ -150,9 +149,7 @@ func _ready() -> void:
 
 func _set_speeds(new_speeds: CharacterSpeeds) -> void:
 	speeds = new_speeds
-	# speeds.changed.connect()
-	_initial_walk_speed = speeds.walk_speed
-	_initial_run_speed = speeds.run_speed
+	_initial_speeds = new_speeds.duplicate()
 	if not is_node_ready():
 		return
 	input_walk_behavior.speeds = speeds
@@ -262,5 +259,7 @@ func _handle_game_over() -> void:
 
 
 func _on_player_hook_aiming_changed(is_aiming: bool) -> void:
-	input_walk_behavior.speeds.walk_speed = aiming_speed if is_aiming else _initial_walk_speed
-	input_walk_behavior.speeds.run_speed = aiming_speed if is_aiming else _initial_run_speed
+	input_walk_behavior.speeds.walk_speed = (
+		aiming_speed if is_aiming else _initial_speeds.walk_speed
+	)
+	input_walk_behavior.speeds.run_speed = aiming_speed if is_aiming else _initial_speeds.run_speed


manuq:
This works but is draft for now because StoryQuests have to be maintained, as a quick grep shows:

```
❯ git grep walk_speed scenes/quests/story_quests/
scenes/quests/story_quests/after_the_tremor/3_combat/after_the_tremor_combat.tscn:walk_speed = 10.0
scenes/quests/story_quests/after_the_tremor/3_combat/after_the_tremor_combat.tscn:walk_speed = 10.0
scenes/quests/story_quests/after_the_tremor/3_combat/scripts/combat.gd: p_node.walk_speed = 0.0
scenes/quests/story_quests/after_the_tremor/3_combat/scripts/combat.gd:         bryan.walk_speed = 300.0
scenes/quests/story_quests/shjourney/4_Laberinto/Laberinto.tscn:walk_speed = 200.0
scenes/quests/story_quests/shjourney/5_shjourney_sequence_puzzle/shjourney_sequence_puzzle.tscn:walk_speed = 190.0
scenes/quests/story_quests/shjourney/8_shjourney_outro/Outro.tscn:walk_speed = 200.0

❯ git grep run_speed scenes/quests/story_quests/
scenes/quests/story_quests/after_the_tremor/3_combat/after_the_tremor_combat.tscn:run_speed = 10.0
scenes/quests/story_quests/after_the_tremor/3_combat/after_the_tremor_combat.tscn:run_speed = 10.0
scenes/quests/story_quests/after_the_tremor/3_combat/scripts/combat.gd: p_node.run_speed = 0.0
scenes/quests/story_quests/after_the_tremor/3_combat/scripts/combat.gd:         bryan.run_speed = 500.0
scenes/quests/story_quests/shjourney/4_Laberinto/Laberinto.tscn:run_speed = 350.0
scenes/quests/story_quests/shjourney/5_shjourney_sequence_puzzle/shjourney_sequence_puzzle.tscn:run_speed = 200.0
scenes/quests/story_quests/shjourney/8_shjourney_outro/Outro.tscn:run_speed = 300.0
```

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -143,48 +142,20 @@ func _get_configuration_warnings() -> PackedStringArray:
...
-
-func _physics_process(delta: float) -> void:
-	if Engine.is_editor_hint():
-		return
-
-	# TODO: Use InputWalkBehavior instead, and enable/disable it when the mode changes.
-	if mode in [Mode.SYSTEM_CONTROLLED, Mode.DEFEATED]:
+func _set_speeds(new_speeds: CharacterSpeeds) -> void:
+	speeds = new_speeds
+	# speeds.changed.connect()
```
wjt: Did you mean to come back and connect something to `speeds.changed`?

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -288,3 +259,8 @@ func _handle_game_over() -> void:
 		SceneSwitcher.change_to_file_with_transition(
 			challenge_start_scene, ^"", Transition.Effect.FADE, Transition.Effect.FADE
 		)
+
+
+func _on_player_hook_aiming_changed(is_aiming: bool) -> void:
+	input_walk_behavior.speeds.walk_speed = aiming_speed if is_aiming else _initial_walk_speed
+	input_walk_behavior.speeds.run_speed = aiming_speed if is_aiming else _initial_run_speed
```
wjt: Another way to do this would be to swap the whole `input_walk_behavior.speeds` value out with a copy that is initialised in `_ready` with these changes. This is fine I think.

scenes/game_elements/characters/player/components/player_dust_particles.gd
```diff
@@ -5,11 +5,7 @@ extends GPUParticles2D
...
-	if not player:
-		return
-	if player.is_running() and not player.velocity.is_zero_approx():
-		emitting = true
+func _on_input_walk_behavior_running_changed(is_running: bool) -> void:
+	emitting = is_running
+	if emitting:
 		process_material.direction = Vector3(player.velocity.x, player.velocity.y, 0.0)
-	else:
-		emitting = false
```
wjt, line 15: Nice improvement here!

wjt: I'm always kind of surprised that you can't do:

```gdscript
process_material.direction = Vector3(player.velocity, 0.0)
```

i.e. a Vector3 ctor that takes a Vector2 and then 1 additional component.

scenes/game_elements/characters/player/player.tscn
```diff
@@ -658,5 +676,8 @@ unique_name_in_owner = true
 root_node = NodePath("../../PlayerRepel")
 libraries/ = SubResource("AnimationLibrary_1n58k")
 
+[connection signal="running_changed" from="InputWalkBehavior" to="PlayerDustParticles" method="_on_input_walk_behavior_running_changed"]
+[connection signal="running_changed" from="InputWalkBehavior" to="AnimationPlayer" method="_on_input_walk_behavior_running_changed"]
 [connection signal="body_entered" from="PlayerRepel/AirStream" to="PlayerRepel" method="_on_air_stream_body_entered"]
+[connection signal="aiming_changed" from="PlayerHook" to="." method="_on_player_hook_aiming_changed"]
 [connection signal="body_entered" from="PlayerHarm/HitBox" to="PlayerHarm" method="_on_hit_box_body_entered"]
```
wjt, line 683: I am still not totally sure (c.f. https://github.com/endlessm/threadbare/wiki/Contributing#connect-signals-to-callbacks-through-the-editor-interface) that this is preferred. These components are tightly coupled - `player.gd` doesn't work properly if these signals aren't connected, and it has references to all (?) these child nodes and calls methods on them, so IMO it would be clearer to connect the signals in code. But this isn't a firm conviction. At present it's no big deal because player.tscn and player.gd are also tightly coupled, so this is fine.

scenes/quests/story_quests/shjourney/5_shjourney_sequence_puzzle/shjourney_sequence_puzzle.tscn
```diff
@@ -34,7 +36,12 @@
 [ext_resource type="Resource" uid="uid://m3wugmcg8r6m" path="res://scenes/quests/story_quests/shjourney/5_shjourney_sequence_puzzle/shjourney_sequence_components/shjourney_sequence_puzzle.dialogue" id="32_1fi4m"]
 [ext_resource type="AudioStream" uid="uid://wh36l3rfb7xa" path="res://scenes/quests/story_quests/shjourney/Otros_componentes/musicas/Mind's Eye Full.ogg" id="33_at0ry"]
 [ext_resource type="PackedScene" uid="uid://dpbsi065gmh8a" path="res://scenes/game_elements/fx/fog/fog.tscn" id="35_at0ry"]
-[ext_resource type="Script" uid="uid://pk3ucq7e2eah" path="res://scenes/game_logic/player_mode.gd" id="1_pmz81"]
+
+[sub_resource type="Resource" id="Resource_8ckgm"]
+script = ExtResource("9_it68j")
+walk_speed = 190.0
+run_speed = 200.0
+metadata/_custom_type_script = "uid://csev4hv57utxv"
```
wjt, line 44: I was about to say that the speeds resource should be saved so it can be shared between scenes, but the speeds are actually different in each scene!

scenes/game_elements/characters/player/components/player.gd
wjt: I think the reason to listen to this signal is to update the `_initial_walk_speed` and `_initial_run_speed` if the character's speed is changed in code, e.g. this is done in after the tremor. But... you would then need to ignore the signal when changing the speeds in `_on_player_hook_aiming_changed`.

manuq: Oh thanks for finding the leftover. Yes indeed, I started questioning myself when I went to see after the tremor and thinking: what if we want to change `speeds.walk_speed` directly in a level? Should it update the initial walk speed that is being saved just to change it back after aiming with the grappling hook? Maybe we can revisit this when the problem arise, but we should think in decoupling the grapple.

scenes/game_elements/characters/player/components/player.gd
manuq: Actually I had that at one point, using `Resource.duplicate()`. Let me check again if that's better before merging.

scenes/game_elements/characters/player/components/player_dust_particles.gd
manuq: Indeed, it would be a nice addition to the Vector3 constructor.

scenes/game_elements/characters/player/player.tscn
manuq: I'm glad you raise this! What I wrote there is open to discussion. So maybe I should open a post in Discussions instead. Yes if we want to decouple the mechanics from player, this is not helping.

scenes/game_elements/characters/player/components/player.gd
manuq: By the way for actually using `speeds.changed.connect()` I think we'll need to add setters to every exported property of Speeds and call [`emit_changed()`](https://docs.godotengine.org/en/stable/classes/class_resource.html#class-resource-method-emit-changed) in each.

Issue #2057, opened 2026-03-16
Add powerup collectible to gain abilities


manuq:
It is much better to get the ability by interacting and collecting an item, than to grant it at the beginning of the scene.

Create a powerup item, with a property for the ability that it gives, plus visuals that can be customized per ability. And instantiate it in the 2 lore levels that give the abilities:
- `ink_combat_round_1.tscn` gives Throw
- `grappling_hook_start.tscn` gives Repel

Plus in the template quest which is exemplary:
- `NO_EDIT_combat.tscn` gives Throw

Similar to the existing "long thread" powerup. Use placeholder assets for now.

PR #2073, opened 2026-03-20
Add powerup for abilities

manuq:
### InteractArea: Emit signal when characters start/stop seeing it

Maintain an array of CharacterSight areas and emit the signal when
the array changes.

### Add powerup for abilities

The powerup scene gives one ability when picked upon interaction.

It displays a dialogue telling the player about the new ability.
The default dialogue just says "new ability: ABILITY NAME".

It has a collected signal that can be used to display input hints
for the corresponding ability. Or doing other side effects, like
zooming out the camera when the longer thread powerup is picked.

It has a ripple effect that can be colored, so we can have
color codes for the abilities. The asset is a SpriteFrames that
is proxied so it can be customized.

### Use powerup in lore and template quests

- Add Repel powerup to ink_combat_round_1 and NO_EDIT_combat. And display the input hints when collected.
- Add Grapple powerup to grappling_hook_start. And display the input hints when collected, instead of from cinematic dialogue.
- Replace longer thread powerup in grappling_hook_powerup with this new component.

### Projectile: Export hit speed

The speed after the projectile gets hit.

### unify_abilities_test: Add powerups and improve test level

Add the 2 powerups needed. Make the level a bit easier by using a projectile scene with the hit speed set to a low value. Also:
- Use lore assets: ink drinker character, ink blob, inkwell.
- Remove unneeded HUD and sign post.
- Use the new elevation_2 tileset and repaint it as terrain.
- Paint Void tiles as terrain, so they have the embroidery.
- Add more Void tiles around to constrain the player.
- Make the island at the north a 3x2 tiles (was a 2x2 square) so the player has more space for aiming.

----

Resolve https://github.com/endlessm/threadbare/issues/2057

### DIFF
--- a/scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_powerup.tscn
+++ b/scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_powerup.tscn
@@ -360,6 +360,4 @@ shape = SubResource("RectangleShape2D_s7yva")
 
 [node name="HUD" parent="." unique_id=1267842834 instance=ExtResource("11_7v23v")]
 
-[connection signal="collected" from="OnTheGround/LongerThreadPowerup" to="." method="_on_longer_thread_powerup_collected"]
-
 [editable path="OnTheGround/VoidPatrolling"]
--- a/scenes/quests/lore_quests/quest_002/2_grappling_hook/components/grappling_hook_powerup.gd
+++ b/scenes/quests/lore_quests/quest_002/2_grappling_hook/components/grappling_hook_powerup.gd
@@ -2,10 +2,21 @@
 # SPDX-License-Identifier: MPL-2.0
 extends Node2D
 
+var zoom_tween: Tween
 
-func _on_longer_thread_powerup_collected() -> void:
-	# Zoom out the camera when collecting the powerup, because now the player
-	# can throw a longer thread:
+
+func _ready() -> void:
+	GameState.abilities_changed.connect(_on_abilities_changed)
+	_on_abilities_changed()
+
+
+func _on_abilities_changed() -> void:
+	var has_longer_thread := GameState.has_ability(Enums.PlayerAbilities.ABILITY_B_MODIFIER_1)
+	var camera_zoom := Vector2(0.5, 0.5) if has_longer_thread else Vector2(1.0, 1.0)
+	# Zoom out when the player can throw a longer thread:
 	var camera: Camera2D = get_viewport().get_camera_2d()
-	var zoom_tween := create_tween()
-	zoom_tween.tween_property(camera, "zoom", Vector2(0.8, 0.8), 1.0)
+	if camera.zoom != camera_zoom:
+		if zoom_tween:
+			zoom_tween.kill()
+		zoom_tween = create_tween()
+		zoom_tween.tween_property(camera, "zoom", camera_zoom, 1.0)
--- a/scenes/quests/lore_quests/quest_002/3_void_grappling/components/void_grapping_round_2.gd
+++ b/scenes/quests/lore_quests/quest_002/3_void_grappling/components/void_grapping_round_2.gd
@@ -2,14 +2,31 @@
 # SPDX-License-Identifier: MPL-2.0
 extends Node2D
 
+var zoom_tween: Tween
+
 @onready var player: Player = %Player
 @onready var void_chasing: CharacterBody2D = %VoidChasing
 
 
-func _on_longer_thread_powerup_collected() -> void:
-	# Zoom out the camera when collecting the powerup, because now the player
-	# can throw a longer thread:
+func _ready() -> void:
+	GameState.abilities_changed.connect(_on_abilities_changed)
+
+	# TODO: This level removes the longer thread modifier when started.
+	# Otherwise the Void chase can't be triggered.
+	GameState.set_ability(Enums.PlayerAbilities.ABILITY_B_MODIFIER_1, false)
+
+
+func _on_abilities_changed() -> void:
+	var has_longer_thread := GameState.has_ability(Enums.PlayerAbilities.ABILITY_B_MODIFIER_1)
+	var camera_zoom := Vector2(0.5, 0.5) if has_longer_thread else Vector2(1.0, 1.0)
+	# Zoom out when the player can throw a longer thread:
 	var camera: Camera2D = get_viewport().get_camera_2d()
-	var zoom_tween := create_tween()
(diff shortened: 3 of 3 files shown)


scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -37,6 +37,9 @@ extends RigidBody2D
 ## The speed of the initial impulse and the bouncing impulse.
 @export_range(10., 100., 5., "or_greater", "or_less", "suffix:m/s") var speed: float = 30.0
 
+## The speed after the projectile gets hit.
+@export_range(10., 100., 5., "or_greater", "or_less", "suffix:m/s") var hit_speed: float = 100.0
```
wjt, line 41: 👍🏻

scenes/quests/lore_quests/quest_002/2_grappling_hook/intro_start.dialogue
```diff
@@ -7,5 +7,4 @@ do animation_player.animation_finished
 Moss the Monk: Oh, look! The Void is blooming fast here.
 Moss the Monk: You’ll need a [b]grappling hook[/b] to get to LinenVille. Take mine.
 do GameState.set_ability(Enums.PlayerAbilities.ABILITY_B, true)
```
wjt: I think you meant to remove this line so that the player-character has to pick it up?

scenes/game_elements/props/powerup/components/powerup.gd
```diff
@@ -0,0 +1,115 @@
...
+	GameState.set_ability(ability, true)
+	collected.emit()
+
+
+func _on_interact_area_observers_changed() -> void:
+	if _tween:
+		_tween.kill()
+	_tween = create_tween()
+	var width: float = 0.8 if interact_area.is_being_observed else 0.4
+	_tween.tween_property(highlight_effect, "material:shader_parameter/width", width, 1.0)
```
wjt, line 117: Because the material is not local to scene, observing any powerup causes every powerup to glow:

https://github.com/user-attachments/assets/74bf8929-5542-471d-8b20-43b84673d72a

I'll push a fixup.

scenes/game_elements/characters/player/components/unify_abilities_test.tscn
```diff
@@ -73,31 +70,37 @@ position = Vector2(21, 262)
...
-label = ""
-color = Color(1, 1, 1, 1)
+position = Vector2(-1197, 1032)
+label = "Cyan"
+color = Color(0, 1, 1, 1)
 
-[node name="ScreenOverlay" type="CanvasLayer" parent="." unique_id=1976326993]
+[node name="RepelPowerup" parent="OnTheGround" unique_id=1795462717 instance=ExtResource("17_w66m4")]
+position = Vector2(-307, 1622)
+highlight_color = Color(0, 0, 1, 1)
```
wjt: This colour appears darker than the yellow one below (see in the video above). Not a blocker, just an observation.

scenes/game_elements/props/powerup/components/powerup.gd
```diff
@@ -0,0 +1,115 @@
...
+	match what:
+		NOTIFICATION_EDITOR_PRE_SAVE:
+			# Since this is a tool script that plays the animations in the
+			# editor, reset the frame progress before saving the scene.
+			sprite.frame_progress = 0
+
+
+func _on_abilities_changed() -> void:
+	var has_ability := GameState.has_ability(ability)
+	ground_collision.disabled = has_ability
```
wjt, line 83: I was hoping to see this. I have been sceptical of objects in the world that you can walk through (e.g. arguing that the collectible threads should have a collider) but in this case I think it makes sense for the powerup to become a ghost once you have collected it, because it lets us barricade a pathway with a mandatory powerup.

scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_powerup.tscn
```diff
@@ -257,15 +221,12 @@ position = Vector2(837, -280)
...
-[node name="ButtonSprite" parent="OnTheGround/LongerStringPowerUp/ButtonItem" parent_id_path=PackedInt32Array(987233957, 1967107361) index="0" unique_id=1194360136]
-sprite_frames = SubResource("SpriteFrames_g2ubw")
-
-[node name="HookableArea" parent="OnTheGround/LongerStringPowerUp" index="2" unique_id=1267513436]
-weight = 0.3
+[node name="LongerThreadPowerup" parent="OnTheGround" unique_id=1795462717 instance=ExtResource("17_a36ca")]
+position = Vector2(-549, 805)
+ability = 32
+ability_name = "Longer Thread"
+interact_action = "Collect Longer Thread"
```
wjt, line 228: I quite liked the way that you previously had to use the grappling hook to pick up the powerup! This is fine though.

How about `res://scenes/quests/lore_quests/quest_002/3_void_grappling/void_grappling_round_2.tscn`? It has another "make the thread longer" powerup.

scenes/quests/lore_quests/quest_002/2_grappling_hook/intro_start.dialogue
wjt: Or if not, the powerup should not be in the scene?

manuq: Oh I thought I did it, but it looks like I forgot to add the change to Git. Thanks!

scenes/game_elements/props/powerup/components/powerup.gd
manuq: Good catch!

manuq: Thanks for the fixup.

scenes/game_elements/characters/player/components/unify_abilities_test.tscn
manuq: Yes I will make it a lighter blue. It is hard to pick colors for these because we also have the color code for the memory/imagination/spirit (yellow/green/red). Maybe when we have actual assets for the powerups it will be better.

scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_powerup.tscn
manuq: It will be trivial to add a HookableArea to it, let me do it!

Also good catch, I totally forgot about the same powerup in the following level. In the current form it will appear as ghost, because the ability has been already granted. I will just do that rather than removing it. As discussed in private, in the future we may split this quest in 2, for collecting the longer thread modifier in a separate quest.

PR #2078, opened 2026-03-24
Rework and document CharacterSight and PlayerInteraction

manuq:
Use body_entered and body_exited signals for obtaining the best area, instead of doing it in the process loop.

Rename the node InteractZone in player to CharacterSight and give it a class name, moving it to game_elements.

champ & after_the_tremor StoryQuests: Adapt to the recent changes in player.

### DIFF
--- a/scenes/game_elements/props/interact_area/interact_area.gd
+++ b/scenes/game_elements/props/interact_area/interact_area.gd
@@ -7,7 +7,7 @@ extends Area2D
 ##
 ## To make an interactable object, add an [InteractArea] to it, and handle
 ## [signal interaction_started]. When the interaction is complete, call [method
-## end_interaction]. This area is detected by the player scene's [InteractZone];
+## end_interaction]. This area is detected by character's [CharacterSight];
 ## the player scene is typically responsible for calling [method
 ## start_interaction] in response to player input.
 ## [br][br]


manuq:
This is extracted from https://github.com/endlessm/threadbare/pull/2073, in which I'm trying to highlight the object that can be interacted. That is the object currently being observed by the player (the one showing the interact label).

scenes/game_elements/props/character_sight/character_sight.gd
wjt, line 1: I think it's a bit weird for this to live in `props` rather than `player`.

manuq: Hmm, I did it to match interact_area. But also this can potentially be attached to a CharacterBody2D NPC.
Maybe I move it to scenes/game_logic/ ?

wjt: Could work. Honestly I think it's also quite strange that this script (not a scene) lives in `scenes/`, like many many others (e.g. area_filler.gd) but I can't bring myself to propose rearranging all the folders in the project again :)

wjt: Don't feel the need to address this now, it's just a path, we can change it later. I was just observing that it seems a bit strange

manuq: Yes the whole structure is strange to me as well!

Issue #2080, opened 2026-03-25
Add in-game problem reporting


wjt:
We would like to make it possible to report problems from within the game.

Fancier solutions are possible (c.f. https://github.com/endlessm/threadbare/discussions/2076) but for starters we can add a menu item to the pause screen that opens a URL like the following in your web browser:

https://github.com/endlessm/threadbare/issues/new?template=bug.yml&version=v0.1.11-14-g36082b47&scene=res%3A%2F%2Fscenes%2Fworld_map%2Ffrays_end.tscn

with the `version` and `scene` fields filled in dynamically.

PR #2081, opened 2026-03-25
Add “Report a Problem” button to pause screen

wjt:
This opens your browser at a bug report form, with the game version and current scene pre-filled into the appropriate fields. This does mean you need a GitHub account to report a problem. In future you could imagine directing users somewhere else, but this is an simple start based on infrastructure that already exists.

To do this, refactor the code that determines the version from `git describe` so that it can also be used when running the game from the editor. For good measure, put the version number on the pause screen too.

Resolves https://github.com/endlessm/threadbare/issues/2080

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -26,9 +26,12 @@ body:
       label: Affected scene
       placeholder: e.g. Fray's End, quests/lore_quests/quest_002/3_void_grappling/void_grappling_round_2.tscn
       description: >
-        In which scene in the game is the problem? If it affects multiple
-        scenes, give an example. If you're not sure, describe the scene
-        visually, or leave it blank.
+        Give an example of a scene in the game where you see this problem. If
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -26,9 +26,12 @@ body:
...
       description: >
-        In which scene in the game is the problem? If it affects multiple
-        scenes, give an example. If you're not sure, describe the scene
-        visually, or leave it blank.
+        Give an example of a scene in the game where you see this problem. If
+        you're not sure, describe the scene visually, or leave this blank.
+
+        If you report a problem from within the game, this will have been filled
+        automatically with the current scene. You can change this if you are
+        reporting a problem you saw in a different scene.
```
wjt, line 34: @manuq Do you think this change captures your point?

manuq: @wjt yes, well explained!

PR #2095, opened 2026-04-02
SceneSwitcher: Add input to copy / paste current scene path

manuq:
Rationale: Although the web build can load a specific scene by copying the scene path to the URL hash, people is not used to edit the URL.

With this change we could say: "Copy the following line that starts with res:// and then press Ctrl + V in the game".

In the future, two Copy / Paste buttons could be exposed in the Debug Settings.


manuq:
@wjt just an idea, what do you think?

PR #2109, opened 2026-04-13
shaker: Remove unused scene

wjt:
shaker: Remove unused scene

Every use of the shaker component uses shaker.gd directly, not
Shaker.tscn. Remove the scene. Simplify the folder hierarchy.


manuq (approved):
Nice, I thought this was because of the global. But the global autoload is for CameraShake, which is using Shaker as a node already.

PR #2112, opened 2026-04-14
grappling_hook_start: Switch to Creep Outro when returning from powerup

### LINKED ISSUE #2110: Wrong section of “Void Creep” played when returning to grappling_hook_start
### Threadbare version

v0.1.11-38-g9866dd86

### Affected scene

res://scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_start.tscn

### Description

Void Creep is played through `quest_002/2_grappling_hook`. The music is split into three clips and the progression is meant to be as follows:

1. Player enters `grappling_hook_start`. The opening dialogue plays and the background music switches to Creep Intro
2. Creep Intro auto-advances to Creep Loop
3. Every other scene of this section sets the clip to "Creep Loop", so it keeps playing
4. When the player returns to the starting area after collecting the power-up, queue up a transition to "Creep Outro". This transitions waits for the end of Creep Loop (up to 106 seconds) then switches cleanly to Creep Outro, which is a 10-second outro. The hope is that this will happen around the time the player reaches the collectible.
5. Creep Outro auto-advances to Still, setting things up for `3_void_grappling`.

Prior to #2032 this was very simple: step 4 was actually not the same scene as step 1, but a copy, `grappling_hook_end.tscn`, so the `BackgroundMusic` node was just set to play `Creep Outro`.

Since #2032 the copy was eliminated and the player actually does return to the same scene. There is some custom logic to skip the opening dialogue. However the Creep Outro logic was lost. As a result the music abruptly cuts back to Creep Intro at step 4.

### Checklist

wjt:
The desired behaviour is that when the player returns to this scene
after collecting the powerup, a music transition to Creep Outro is
triggered. Creep Loop will be playing, so this allows it to reach its
natural end, then play an outro, and transition to Still in preparation
for the next scene, void_grappling.

Previously this was achieved by having a second copy of this scene,
grappling_hook_end, which set BackgroundMusic.clip to "Creep Outro".
However since commit b86ea32df3ff7638049ccf3ac6af7d55392f7f19 this scene
was removed. As a result the music was switched to Creep Intro when
returning to this scene.

Change BackgroundMusic.switch_to_clip() into a setter for
BackgroundMusic.clip, so that it can be used before
BackgroundMusic.play() has been called (from _ready()). (This has the
nice side-benefit that you can now test AudioStreamInteractive
transitions in the editor by hitting Play in the inspector then
adjusting the Clip property!)

In grappling_hook_start, add a ClipSwitcher, triggered by the player
being teleported, in the same way that we disable the opening dialogue
in that case.

Update ink_combat_round_6 to use ClipSwitcher, matching
ink_combat_round_3.

Resolves https://github.com/endlessm/threadbare/issues/2110

### DIFF
--- a/scenes/game_elements/props/background_music/components/background_music.gd
+++ b/scenes/game_elements/props/background_music/components/background_music.gd
@@ -10,7 +10,7 @@ extends Node
 ## [br][br]
 ## If [member stream] is an [AudioStreamInteractive], you can control the clip
 ## to play at different points in the scene using [member clip], [method
-## switch_to_clip], and [ClipSwitcher] child nodes.
+## set_clip], and [ClipSwitcher] child nodes.
 ## [br][br]
 ## There should be at most one BackgroundMusic component in a scene.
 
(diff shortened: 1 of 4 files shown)


manuq (approved):
I didn't realize that I broke this, sorry.

Is very nice to test the music transitions in the editor. I spent some time clicking Play and then changing the Clip.

The clips I see now in a playthrough are:
- Creep Intro in grappling_hook_start, the first time.
- Creep Loop in the levels in between.
- Creep Outro in grappling_hook_start, when returning.

PR #2115, opened 2026-04-14
Add new Inn building

manuq:
With an asset provided by PixelFrog.

Use 2 rectangle shapes for collision.

Add 2 artificial lights to this building.

Also add the aseprite source file to the revision.


manuq:
This is how it looks:

https://github.com/user-attachments/assets/9da103ce-09bb-48dd-bc16-ef509300bb31

wjt (approved):
Cute!

PR #2116, opened 2026-04-14
Integrate new grappling hook assets

### LINKED ISSUE #2117: Void: Replace pin and needle assets
With new assets by PixelFrog.

### LINKED ISSUE #2118: Void: Replace arrow indicator
With new assets provided by PixelFrog.

manuq:
## Integrate new pin and needle assets

As delivered by PixelFrog.

Update the position of elements in the corresponding scenes too. Move the sprite and adjust the marker and control for hooking. They only differ in a few pixels so it shouldn't be any gameplay regression.

## Integrate new arrow indicator

PixelFrog has delivered 2 new assets: an indicator pointing down and another
pointing up. The "up" should be used in the player. And this commit does so. The
"down" will be used as indicator in things that can be hooked, and potentially
for indicating other things on screen.

Use the "up" indicator in the HookControl, rotated by 90 degrees.

Put these assets in assets/first_party/ because they can be multipurpose.

Remove the previous arrow asset.

Resolve https://github.com/endlessm/threadbare/issues/2117
Resolve https://github.com/endlessm/threadbare/issues/2118


manuq:
This is the usage explanation by PixelFrog:

image

scenes/game_elements/props/hookable_pin/hookable_pin.tscn
```diff
@@ -13,7 +13,7 @@ collision_layer = 16
 collision_mask = 0
 
 [node name="Sprite2D" type="Sprite2D" parent="." unique_id=1242315016]
-position = Vector2(0, -32)
+position = Vector2(0, -24)
```
wjt, line 16: As an aside - I read somewhere that it's better to use `Sprite2D.offset` rather than adjusting its `position`, but I can't remember why. Not specific to this PR of course.

wjt (approved):

PR #2120, opened 2026-04-14
Integrate powerup asset

### LINKED ISSUE #2121: Void: Replace powerup assets
With assets delivered by PixelFrog.

manuq:
This is a ball of thread animated, for the grappling hook. Use it in both the grapple ability and the longer thread modifier.

Merge the Aseprite file provided by PixelFrog with the existing one, because we still need a default powerup asset. Also keep the shadow, which wasn't in the delivered files. Also desaturate the highlight sprite provided, because modulate the highlight is an easy way to distinguish the powerups. Finally, export multiple layers:
- A shadow layer
- The highlight effect
- The powerup asset itself

Change the powerup scene accordingly. Remove the shader used before for highlighting.

Resolve https://github.com/endlessm/threadbare/issues/2121


manuq:
For reference this was the spritesheet provided:

Grappling Hook

I have changed the Aseprite file and exported things individually. This is how it looks in the unify abilities level:

https://github.com/user-attachments/assets/8318bdad-ff3b-4270-9eb7-a889c11bf824

scenes/game_elements/props/powerup/components/powerup.gd
```diff
@@ -21,7 +21,7 @@ signal collected
 	set = _set_interact_action
 
 ## Asset of this powerup.
-@export var sprite_frames: SpriteFrames:
+@export var sprite_frames: SpriteFrames = preload("uid://cualgrcaaggcm"):
 	set = _set_sprite_frames
 
 ## The powerup shines with this color through a shader.
```
wjt: I think this comment is no longer true.

wjt (approved):
LGTM except for a comment that needs updating.

scenes/game_elements/props/powerup/components/powerup.gd
manuq: Oh, yes! Thanks.

PR #2127, opened 2026-04-15
Add project setting to skip the sokobans

manuq:
The Sokobans may be repurposed in the future. For now, add a setting so developers can skip them with a debug setting, in project.godot or in an override.cfg local file.

In the settings addon, add a class name so scripts can reference the settings by the constant. For that, move the "threadbare/" prefix to the constants and add an assertion for the prefix. Also move the comment to hint strings.

### DIFF
--- a/addons/threadbare_project_settings/threadbare_project_settings.gd
+++ b/addons/threadbare_project_settings/threadbare_project_settings.gd
@@ -1,16 +1,24 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 @tool
+class_name ThreadbareProjectSettings
 extends EditorPlugin
 
-## Debug aspect ratio while playing the game.
-const DEBUG_ASPECT_RATIO = "debugging/debug_aspect_ratio"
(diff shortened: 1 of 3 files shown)


manuq:
Cherry-picked from the townies system exploration.

wjt (approved):
OK! This isn't quite the same as just skipping the Sokobans because you also skip the elder dialogue thanking the player that would play if the player goes to the Sokobans and back again. However we can let this slide because more changes to the quest loop are coming.

addons/threadbare_project_settings/threadbare_project_settings.gd
```diff
@@ -1,16 +1,24 @@
...
-const DEBUG_ASPECT_RATIO = "debugging/debug_aspect_ratio"
+const DEBUG_ASPECT_RATIO = "threadbare/debugging/debug_aspect_ratio"
+const SKIP_SOKOBANS = "threadbare/debugging/skip_sokobans"
 
 static var setttings_configuration = {
 	DEBUG_ASPECT_RATIO:
 	{
 		value = false,
 		type = TYPE_BOOL,
+		hint_string = "Display a letterbox overlay in the game, to debug aspect ratio issues.",
```
wjt, line 15: 👍🏻

PR #2129, opened 2026-04-17
Add townies system with occasional helpers

### LINKED ISSUE #2119: Prototype Townie System
@JoniCeceri has come with an idea for bridging the gameplay and the narrative together. We are going to explore it.

## Townie System

An “upgrade” system where at the end of quests, players will recount the quest to the townies via a dialogue tree, and will have multiple upgrade options that shape how the world expands.

- Players collect sacred element threads in the quests, and are able to “spend” them on new features during these retellings.
  - Can use other options besides the threads to embody the elements! 
- Memory: Upgrades focused on lore and gameplay hints. (Townies start remembering more about the world and its history.)
- Imagination: Upgrades to the environment, townies expand the town and fix up long disused ruins/temples/shops/etc. Opens up shortcuts, new shops, many options.
- Spirit: Townies show up in combat/puzzle encounters, right alongside the player during future quests.

----

- Focus of this system is to bridge the gameplay and narrative together in inventive ways. Give the player incentive to care about the NPCs!
- Many directions this can go in, can have established NPCs that build up over time.
- Can also have bespoke upgrades for specific narrative beats! Doesn’t have to lock us into a certain set of upgrades.
- Also set up the player to think more impactfully about their actions during quests, direct commentary on the game play. 
  - Locked out of certain options depending on playstyle?
  - Various gags we can do. (if a player resets a quest and plays it again, townies respond “We already heard this one!”)

----

## Companions

- Leaving design space for us to include them, but current thinking is keep them simple. 
  - Mechanically, should just give player stat buff/debuffs and/or ability modifications. 
  - Open to them evolving as we solidify more of the game! Especially as we define the townie system, thinking of ways to make the player be more invested in the NPCs both narratively and mechanically should be our goals.

manuq:
This adds a help system to the game. Quests creators can now define a dialogue tree for retelling the adventures recently lived before offering the magical threads collected to the Eternal Loom. People from the town gathers around to listen the retelling. If the player has collected a thread that matches one of the options, that option will be highlighted. And if that option is picked, one of the villagers will stand up, and "equipped" (indicated in the HUD) for helping later in the game. The thread collected indicates the type or category of the help, which can be interpreted by level creators in any way.

Level creators can add helper characters to the level. They will appear only if the player got help, and if their helper type matches the one obtained during the retelling. When the player interacts with this helper character, a dialogue is displayed to offer the help. If the player accepts, it is expected that the level creator does something, and then consume the help (disappearing from the HUD). For example: An imagination helper at the stealth level can fix the bridge, which is a shortcut to skip the most difficult part of the level, full of guards.

Details below.

### Void quest: Add helpers

Add the following helpers:
- 1_void_quest: Memory helper that displays a placeholder dialogue.
- 1_void_quest: Spirit helper that slows down the enemy.
- 2_grappling_hook: Imagination helper that makes needles appear in the optional loop.

### Musician quest: Add helpers

Add the following helpers:
- 1_music_puzzle: Memory helper that triggers a dialogue saying "Lore lore lore" as placeholder.
- 2_ink_combat: Spirit helper that fills the hard to reach inkwell.
- 3_stealth_level: Imagination helper that fixes the bridge.

### Musician and Void quests: Add retelling dialogue

The dialogue lines should be considered placeholders.

### Frays End, Eternal Loom: Add retelling and remove sokobans

Change the interaction of the eternal loom when the player is ready to offer the
threads. Now if the quest has retelling dialogue, display it. A retelling may
result in the player obtaining help, that can be used later in the game.

After the retelling, or if no retelling happened, trigger the incorporating
threads cutscene and let the corresponding elder congratulate the player.

Remove the sokoban puzzles, which are going to be repurposed. More cleanup on
this later.

### Add HelperCharacter

Node that will be used to make characters become helpers in levels, for
consuming the help obtained during the retelling at the eternal loom. There is
no actual implementation, this help could be spent in imaginative ways and is
left to the level creators. A default dialogue is provided as example.

### Add RetellingManager and RetellingTownie

The RetellingManager coordinates the retelling with the EternalLoom, to avoid
cluttering that node. It does so by listening to signals and calling a method.

### StoryQuestProgress: Add helper character face to HUD

Listen to the global state for doing so. If there is helper state, instantiate
the face of the character according to its seed, and set the background color to
the type of help.

### Quest: Add optional retelling dialogue to the resource

This dialogue will be displayed when interacting with the eternal loom, before
the threads offering.

### CharacterRandomizer: Vary the animation when randomizing

If not, when randomizing multiple characters at the same time, they all end up
moving in sync. This is the case of the retelling townies in Fray's End.

### CharacterRandomizer: Allow accessing the head sprite.

By setting an unique name to this node. This will be used to show the head in
the HUD.

### InventoryItem: Add color constants

The colors will be used in the HUD as background to indicate the type of help.

### Add global state for helper characters

Persist the type of help, matching one of the magical threads, and a character
seed for the visual features of the character.

Add methods to obtain and consume the help. And a signal to monitor changes.

Resolve https://github.com/endlessm/threadbare/issues/2119


manuq:
This is implementation for an idea that @wjt and I had this morning: a townie helps you with the most difficult part of the Stealth level, by fixing a bridge that opens a shortcut.

Here is one `game_state.cfg` to try it:

[game_state-fix_bridge.zip](https://github.com/user-attachments/files/26902294/game_state-fix_bridge.zip)

Video:

[recording.webm](https://github.com/user-attachments/assets/10db7a80-1383-40e6-9e96-03125a2f80a0)

manuq:
Update: I added 6 townies to quests now:

In musician quest:
- Memory helper at `music_puzzle.tscn`: Tells a story. Just narrative / lore.
- Imagination helper at `stealth_level.tscn`: Fixes a bridge to skip the difficult part.
- Spirit helper at `ink_combat_round_3.tscn`: Helps the player by automatically filling one barrel (tries the hard cyan first)

In void quest:
- Memory helper at `void_runner.tscn`: Tells a story. Just narrative / lore.
- Imagination helper at `grappling_hook_needles.tscn`: Appears extra needles. The only way to collect certain buttons.
- Spirit helper at `void_runner.tscn`: Helps the player by slowing down the Void (this should be reused or repurposed for the difficult last chasing scene!)

Useful game states to try:

[game_states.zip](https://github.com/user-attachments/files/26908576/game_states.zip)

manuq:
I rebased this and adapted it to the flawless new game state! Here are the new game states "snapshots" I took for testing this:
[saved_games.zip](https://github.com/user-attachments/files/28398748/saved_games.zip)

I´'m now polishing this to put it for review.

manuq:
This is now finally ready for review! Things left out of this change:
- Animations and camera movement when using the help in levels.
- All the dialogue should be considered placeholder.
- Centering the camera around the loom when the retelling happens.

Also there is a small bug in the walking animations, when the character that gets closer to the player leaves the loom.

scenes/globals/game_state/inventory/inventory_item.gd
```diff
@@ -8,6 +8,13 @@ enum ItemType {
 	MEMORY,
 	IMAGINATION,
 	SPIRIT,
+	NONE,
```
wjt, line 11: I think you put `NONE` at the end because you didn't want to have to update every instance in the game. OK! Perhaps we can later reorder this and update all resources.

wjt: The other option is to have `GameState.global.helper` be `null` when there is no helper. Then you would not need to introduce the `NONE` member here, which is a potential source of bugs if someone sets `NONE` on a `CollectibleItem`.

scenes/ui_elements/story_quest_progress/components/story_quest_progress.gd
```diff
@@ -31,6 +38,22 @@ func _ready() -> void:
 	GameState.global.item_consumed.connect(self._on_item_consumed)
 
 
+func _on_helper_state_changed() -> void:
+	var has_helper := bool(GameState.global.helper.character_seed)
```
wjt: So here you're not using the `NONE` state but the magic `0` value for the seed to indicate "no helper". I think having `helper` be `null` would be clearer.

wjt: Oh… but it's so that you can connect to the `changed` signal. OK; but that could also be done by monitoring `GameState.global.changed` and emitting that when the `helper` property is assigned to.

Maybe it's clear enough as you have it, don't feel the need to rewrite it all.

scenes/ui_elements/story_quest_progress/components/story_quest_progress.gd
```diff
@@ -31,6 +38,22 @@ func _ready() -> void:
...
+	helper_container.visible = has_helper
+	if has_helper:
+		var helper_character: CharacterRandomizer = TOWNIE.instantiate()
+		add_child(helper_character)
```
wjt, line 46: It seems weird at first glance to instantiate a scene, fish out its head, and delete it again. I guess this is because the seed is used for the townie as a whole.

scenes/quests/lore_quests/quest_001/quest_001_retelling.dialogue
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+Townies: Tell us a story about your last adventure!
+StoryWeaver: I went on a quest with the musician and I...
+- Learned a forgotten song {{ memory_text() }} => retell_memory
+- Slipped past the guards {{ imagination_text() }} => retell_imagination
+- Fought off some InkDrinkers {{ spirit_text() }} => retell_spirit
```
wjt, line 8: We can iterate on this but I wonder if we should make the options conditional on the thread being available. Something like this:

```
- Learned a forgotten song [if has_memory()] => retell_memory
- Slipped past the guards [if has_imagination()] => retell_imagination
- Fought off some InkDrinkers [if has_spirit()] => retell_spirit
- Can't really remember => END
```

scenes/quests/lore_quests/quest_001/3_stealth_level/components/fix_bridge.gd
```diff
@@ -0,0 +1,16 @@
...
+
+var _tiles_to_fix: Array[Vector2i] = [Vector2i(86, 21), Vector2i(108, 19)]
+
+@onready var bridges: TileMapLayer = $"../../TileMapLayers/Bridges"
+
+
+func fix_bridge() -> void:
+	for t in _tiles_to_fix:
+		bridges.set_cell(t, BRIDGE_TILESET_SOURCE_ID, FIXED_BRIDGE_ATLAS_COORDS)
+	await Engine.get_main_loop().create_timer(1).timeout
```
wjt:
```suggestion
	await get_tree().create_timer(1).timeout
```

scenes/quests/lore_quests/quest_002/1_void_runner/components/void_slower.gd
```diff
@@ -0,0 +1,19 @@
...
+
+func is_void_alive() -> bool:
+	return is_instance_valid(void_spreading_enemy)
+
+
+func slowdown_void() -> void:
+	if not is_void_alive():
+		return
+	var result := void_spreading_enemy.get_node_and_resource(
+		"NavigationFollowWalkBehavior:speeds:walk_speed"
```
wjt, line 16: Fancy!

wjt (approved):
Nice!

I think we will want to revisit the logic that hides the thread HUD when you're not on a quest. What happens as of this branch is that you tell the townie a story, they offer their help (and appear in the HUD), but then the threads go to the Loom and then the HUD is hidden.

scenes/globals/game_state/inventory/inventory_item.gd
manuq: I agree, I couldn't find another way in the prior game state. But now with resources, it looks like I should be able to get rid of the NONE enum and use null on the whole helper state. Thanks! Let me try it.

scenes/ui_elements/story_quest_progress/components/story_quest_progress.gd
manuq: Let me try without the NONE before merging.

scenes/ui_elements/story_quest_progress/components/story_quest_progress.gd
manuq: Yes, is only because of that. Do you think this could be done with a separate scene for the head?

scenes/quests/lore_quests/quest_001/quest_001_retelling.dialogue
manuq: This comes from the spec by @JoniCeceri, but now I see that's not part of the text I copy/pasted in the issue. The retelling of each option should be available anyways. If the player has the corresponding thread, they receive a reward.

Currently, the 2 quests to which I added retelling will have all 3 threads, so is impossible to test at this point.

scenes/ui_elements/story_quest_progress/components/story_quest_progress.gd
wjt: Having thought about it a bit, I think this is fine – it ensures that the HUD head is the same as the corresponding townie.

You could imagine splitting the townie scene up into a resource for the components + corresponding colours which is input for the scene... but I don't think it's worth it.

scenes/quests/lore_quests/quest_001/quest_001_retelling.dialogue
wjt: I guess I disagree with the spec then 😀 but indeed – impossible to test without making threads optional.

scenes/ui_elements/story_quest_progress/components/story_quest_progress.gd
manuq: @wjt I made it work identifying null as the lack of helper, and is much better. I added a new signal `GameState.global.helper_changed` for monitoring the change, similar to the existing `completed_quests_changed`. Otherwise listening to `GameState.global.changed` would trigger for other things.

Sorry that it took me some time. I had to run the whole game again, because my state snapshots for quick testing didn't work anymore.

wjt: Something I took away from a Bippinbits talk (perhaps that you shared?) was the value of a custom "run" button that lets you choose the state of the game to run. That might be useful!

BTW you can also change the GameState of the running game in the inspector quite easily now.

PR #2131, opened 2026-04-18
CharacterRandomizer: Use own random number generator

### LINKED ISSUE #2130: CharacterRandomizer makes the whole game predictable
### Threadbare version

main

### Affected scene

_No response_

### Description

Instantiating a townie (`scenes/game_elements/characters/npcs/townie.tscn`) into a scene and then calling `randomize_character()` on them changes the seed of the base random number generator. This means that any random method will be predictable from now on. For instance, in townie system exploration https://github.com/endlessm/threadbare/pull/2129 I'm adding a group of townies to Fray's End to listen the retelling, and then trying to pick one of them randomly with `_townies.pick_random()`. Because the base seed is set, the same townie is always picked.

The CharacterRandomizer should use its own instance of RandomNumberGenerator, to make the townie look&feel predictable, not the entire game!

### Checklist

manuq:
Avoid changing the seed of the base random number generator, which makes the rest of the game predictable. Instead, using its own RNG in this class.

Pass the RNG as argument to the CelShadingRecolor and RandomTextureSpriteBehavior, which now accept one, while still falling back to use the base RNG by default. This way, the character is consistent with its character_seed.

Luckly, characters randomized this way stay the same, because textures and skin colors are immediately called with the same seed as before.

Also: Add a class name and make it inherit from CharacterBody2D. The script is attached to CharacterBody2D nodes in all cases.

Fix https://github.com/endlessm/threadbare/issues/2130

### DIFF
--- a/scenes/game_elements/characters/components/character_randomizer.gd
+++ b/scenes/game_elements/characters/components/character_randomizer.gd
@@ -1,7 +1,8 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 @tool
-extends Node2D
+class_name CharacterRandomizer
+extends CharacterBody2D
 ## @experimental
 ##
 ## Provide a single button to randomize various aspects of a character.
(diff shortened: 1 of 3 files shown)


wjt (approved):
Oops, well debugged!

PR #2132, opened 2026-04-20
project: Refer to InputHud global scene by path

wjt:
project: Refer to InputHud global scene by path

This is consistent with the other scenes and scripts listed here.
Without this the following error is emitted when the project is first
imported:

    ERROR: Unrecognized UID: "uid://dfu3rocpande8".
       at: get_id_path (core/io/resource_uid.cpp:208)


manuq (approved):
Aha, I've been seeing this output too, and didn't have time to fix it.

PR #2133, opened 2026-04-20
Settings: Add option to disable the input HUD

wjt:
Settings: Add option to disable the input HUD

I wanted this for a slightly cleaner screenshot.

Rather than duplicating the code from the fullscreen toggle, make the
script more generic. It now toggles a named property on the Settings
singleton. Use the existing fullscreen methods as the getter/setter of a
new property, and add a new property for whether the input HUD is
visible.

Remove a stale signal connection from the video_settings scene - the
target method does not exist.


scenes/ui_elements/input_hints/components/input_hud.gd
```diff
@@ -63,7 +63,12 @@ func _on_scene_changed() -> void:
 
 
 func _update_visibility() -> void:
-	visible = not get_tree().paused and not Transitions.is_running() and (player or sokoban_ruleset)
+	visible = (
+		Settings.show_input_hud
```
manuq, line 67: 👍

manuq (approved):
Yes, good idea.

PR #2134, opened 2026-04-21
Storybook: Hide navigation arrows at page boundaries

### LINKED ISSUE #1815: Storybook: "previous" from first page, "next" from last page
Currently in the storybook, you can turn to the "previous" page (left arrow) on the first page:

Image

This takes you to the last page, where you can turn to the "next" page (right arrow) to go back to the front page:

Image

This is not how books work. There should be no left-arrow when on the first page, and no right-arrow when on the last page.

### XP Summary (total 275)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Technical Literacy | 150 |

dkg-0717:
Hide the left arrow when on the first page of the storybook, and the right arrow when on the last page.

Previously, the arrows were always visible and navigating past the boundaries would wrap around to the other end of the book, which is not how books work.

Resolves https://github.com/endlessm/threadbare/issues/1815

### DIFF
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -84,6 +84,9 @@ func _ready() -> void:
 
 ## Show/hide index or detail pages
 func _update_page_visibility() -> void:
+	left_button.visible = _current_spread_index > 0
+	right_button.visible = _current_spread_index < _quests.size()
+
 	if _current_spread_index == 0:
 		quest_container.visible = true
 		storybook_page.visible = false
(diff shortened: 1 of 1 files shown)


wjt (approved):
Looks good to me!

wjt:
Thanks for this fix!

You are already in the credits but if you want to adjust your name or add your GitHub username so you get linked in the credits: https://github.com/endlessm/threadbare/blob/main/scenes/menus/credits/data/endless_access.csv

PR #2139, opened 2026-04-21
Fix accessing threadbare_project_settings in non-editor builds

### LINKED ISSUE #2138: ThreadbareProjectSettings parser error in exported game
```
E 0:00:00:732   GDScript::reload: Parse Error: Could not resolve class "ThreadbareProjectSettings", because of a parser error.
  <GDScript Source>aspect_ratio_debugger.gd:7 @ GDScript::reload()
```

I think this regression is due to commit ef05fceba00844256e4129545af9cc75d40cb4e5, investigating.

wjt:
Previously ThreadbareProjectSettings extended EditorPlugin, which means
it is not available in release builds. As a result any code which
accessed ThreadbareProjectSettings did not work. This includes the
splash screen, main menu, and (transitively) the story book.

Extend Node instead, and move the actual plugin piece to a separate
script. Also fix an excessive use of the letter 't' in the word
'setttings'.

Resolves https://github.com/endlessm/threadbare/issues/2138


wjt:
(Found while testing #2134.)

addons/threadbare_project_settings/threadbare_project_settings.gd
```diff
@@ -2,13 +2,13 @@
...
 class_name ThreadbareProjectSettings
-extends EditorPlugin
+extends Node
 
 const DEBUG_ASPECT_RATIO = "threadbare/debugging/debug_aspect_ratio"
 const SKIP_SOKOBANS = "threadbare/debugging/skip_sokobans"
 const SKIP_SPLASH = "threadbare/debugging/skip_splash"
 
-static var setttings_configuration = {
+static var settings_configuration = {
```
manuq, line 11: Typo fixed even!

manuq (approved):
Thanks for the fix.

manuq:
"an excessive use of the letter 't' " :100:

PR #2142, opened 2026-04-21
shjourney: Remove unintentional TileMapCover

wjt:
The Laberinto scene includes a TileMapLayer named TileMapCover with the
tile_map_cover.gd script attached from void runner. The script is
misconfigured - the configured terrain to cover with doesn't exist, and
no containing nodes are configured - so it does nothing more than a
plain TileMapLayer.

Then looking at the tile data, one is a patch of grass that is hidden
below another TileMapLayer, and the rest are way outside the visible
area.

And the embedded TileSet references many cells in atlases that don't
exist, causing a page of errors when the scene runs.

Remove it!


manuq (approved):
How this ended up here?

manuq:
Looks like it was already present in the first submission f5f43933.

PR #2143, opened 2026-04-22
CharacterRandomizer: Change hair and clothes colors too

### LINKED ISSUE #2204: Townie: Generalize skin recolor to full palette swap
Because we want to change the color not only of the skin, but also of hair and clothes. Generalize the shader to be a full palette swap, mapping colors from a key/source palette to colors of another new/output palette.

The randomize script should still understand about skin colors, and in addition about hair/clothes colors. And export buttons so they can easily be tweaked from the Inspector.

manuq:
Before, the shader swapped 3 colors: light, medium and dark. The script had a
way to derive the light and dark colors from the medium one. And the key colors
were 3 shades of green.

Go back to the palette-swap shader reference [1] and only change variable names
and comments.

Townie asset:

- Change the colors so skin is yellow and clothes are green.

- Manually paint the contour lines of shapes, to a darker value of the fill
  color. This adds a 4th color to the cel shading, which will be mapped to a
  dark line or a light line, for more readability.

- Export all PNG spritesheets.

CelShadingRecolor:

- Pass the two palettes to the shader.

- Add lists of hair tones and clothes tones. The latter is rather arbitrary and
  could be improved in the future.

- Export buttons to randomize the first 4 colors of the palette for the skin,
  the next 4 for the hair, and the last 4 for the clothes.

[1] https://github.com/KoBeWi/Godot-Palette-Swap-Shader/blob/master/addons/PaletteSwapShader/SimplePaletteSwap.gdshader

Resolve https://github.com/endlessm/threadbare/issues/2204


manuq:
There is a problem with this approach: it doesn't work with the effects in Fray's End:

Check this 2 townies in editor:
Captura desde 2026-04-24 09-43-54

And in game, the colors change:
Captura desde 2026-04-24 09-43-44

manuq:
This is now on top of https://github.com/endlessm/threadbare/pull/2163

The palette swap works now, but:
- To avoid the existing characters to have different look, I had to apply the new randomizations (hair, clothes) at the bottom in the script. I left a TODO about it.
- Because the exported ShaderMaterial in Townie is set to "local to scene" (to make each instance unique), now the shader parameters data is duplicated in every scene. Check Arun, who appears in several scenes of the tutorial.

manuq:
Hmm from main it looks like Marie has changed compared with [3 weeks ago](https://github.com/endlessm/threadbare/pull/2143#issuecomment-4313251834) so maybe is fine to remove the TODO note and just call `cel_shading_recolor.randomize(_random_number_generator)`. And play with the randomize buttons again to find characters that suit better.

image

scenes/game_elements/components/cel_shading_recolor.gdshader
```diff
@@ -1,64 +1,26 @@
...
  *
- * This shader can replace 3 shades from a sprite that uses the cel-shading technique.
- * It is based on this more generic shader: [link]https://github.com/KoBeWi/Godot-Palette-Swap-Shader/blob/master/addons/PaletteSwapShader/SimplePaletteSwap.gdshader[/link]
+ * This shader replaces each color that appears in the key_colors palette,
+ * with the one of same index in the new_colors palette.
  *
+ * It is copied from: [link]https://github.com/KoBeWi/Godot-Palette-Swap-Shader/blob/master/addons/PaletteSwapShader/SimplePaletteSwap.gdshader[/link]
  *
  * SPDX-FileCopyrightText: The Threadbare Authors
  * SPDX-License-Identifier: MPL-2.0
```
wjt: Comparing this to the source shader - they're identical except for comments and variable names. So:

```suggestion
 * SPDX-FileCopyrightText: 2025 KoBeWi
 * SPDX-License-Identifier: MIT
```

manuq: You are right!

wjt (approved):

PR #2145, opened 2026-04-27
Add walk animation to townies

manuq:
### RandomTextureSpriteBehavior: Allow idle + walk animation

Assets:

- Rename existing part assets to "idle".
- Add walk animation for the legs part.

RandomTextureSpriteBehavior:

- Export an Array of SpriteFrames resources. Instead of using the custom addon
  to change the textures.

Townie scene:

- Add a CharacterSpriteBehavior to each part. And configure the legs part with
  the new animations.

- Add one .tres file per body part variant. And configure each Array of
  SpriteFrames resources with them.

Townie demo:

- Add a FollowWalkBehavior to one townie, and remove the TalkBehavior.

### CharacterSpriteBehavior: Reset frames when back to idle

The townies "idle" has a subtle up and down animation.

For a character made of parts, since the last commit one part (like the head)
may keep playing "idle" while another (like the legs) may play "walk". But when
the character is back to idle, all parts should be set in sync again. Otherwise,
the head may go up when the legs go down, making a funny but not desired (for
now) dance.

### CharacterSpriteBehavior: Fallback to idle when walk is not available

When the SpriteFrames animation doesn't have an animation named "walk", use its
"idle" animation for walking. This makes the modding easier, since a single
animation is now needed for start testing a SpriteFrames in a character with any
of the walk behaviors. We also use "idle" at double speed for running.

And also, when a character is made of multiple parts, some parts can stay in the
"idle" animation. Meaning there are less parts to draw.


manuq:
[recording.webm](https://github.com/user-attachments/assets/c772b9b4-0d88-4b89-8217-1b9ff04824f6)

And this is the funny bug I've been fighting today live on Discord:

https://github.com/user-attachments/assets/c4dd66a9-2ac9-4f06-af38-39f7e512eaca

The lower part of the body is out of sync after a walk. This was fixed by resetting frame and progress of all animations when back to idle (either if they are reusing the idle animation when walking or not):
```
sprite.set_frame_and_progress(0, 0.0)
```

manuq:
If I find time today/tomorrow I will also animate the body and hair (skipping the faces) for the walk cycle.

scenes/game_logic/sprite_behaviors/random_texture_sprite_behavior.gd
```diff
@@ -47,9 +52,15 @@ func _offset_child_sprites(offset: Vector2) -> void:
...
-	var offset := _get_offset_from_texture_filename(new_texture)
+	var offset := _get_offset_from_spriteframes_filename(new_sprite_frames)
 	_offset_child_sprites(offset)
-	SpriteFramesHelper.replace_texture(null, new_texture, sprite.sprite_frames)
+
+	var previous_animation := sprite.animation
+	sprite.sprite_frames = new_sprite_frames
+	if not Engine.is_editor_hint():
+		if previous_animation in sprite.sprite_frames.get_animation_names():
+			sprite.play(&"idle")
```
wjt: Is this supposed to be re-playing `previous_animation`? If not I don't quite understand what the `previous_animation` check is for.

wjt (approved):
This is great. Having walking animations for townies will make it much easier to make the world feel alive.

scenes/game_logic/sprite_behaviors/random_texture_sprite_behavior.gd
manuq: You are right, this should be the previous_animation

PR #2146, opened 2026-04-28
Desaturate musical rock sprite

### LINKED ISSUE #1739: Add multicolour, or monochrome/modulate-able, rock assets
We currently have this very nice rock:

Image

Image

(and a version that's in the water that we don't currently use)

In `res://scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn`, where this rock is used, we recolour 6 instances of it with the `modulate` property:

Image

The problem is, because the original rock is quite dark, the rocks all come out in quite muted colours. For example, the second rock has this modulate colour:

Image

The rock uses four shades of turquoise, plus a slightly-blue near-black grey for the border.

I think these colourful rocks would look better if either:

1. We actually had a collection of rocks in different colours, hand-picked, and stopped using modulate to make them visually distinct
2. We recoloured the asset to be monochrome, with the brightest colour being set to white and the other three being appropriate shades of grey, so that they can be recoloured arbitrarily with `modulate`.

I am no artist so I am the wrong person to say which approach would be better!

### XP Summary (total 225)
| Subdomain | Skill | XP |
| - | - | - |
| Art: Aesthetics & Visual Design | Problem Solving | 75 |
| Art: Aesthetics & Visual Design | Creativity & Design Thinking | 75 |
| Art: Aesthetics & Visual Design | Technical Literacy | 75 |

wjt:
In the musician puzzle (and sequence/stealth combo) we make the 6 (4)
rocks different colours using `modulate`. But the underlying asset is a
shade of blue. The result is that the modulate colour is not what
appears in-game; and that all the rocks are rather dark.

Add a second copy of the rock spritesheets that have been converted to
greyscale in GIMP. Adjust the SpriteFrames resource to refer to these too.
(fabric_rock.tscn remains unmodified for a static rock in the nice
grey-blue colour.)

Resolves https://github.com/endlessm/threadbare/issues/1739


wjt:
| Before | After |
| ------ | ----- |
| image | image |

manuq (approved):
Yes, much better!

PR #2147, opened 2026-04-28
SequencePuzzleObject: Add sprite_modulate property

wjt:
Most instances of the sequence puzzle use the same sprite for all the
objects, with their colour adjusted using `modulate`. But we modulate
the entire tree of nodes (rooted at a StaticBody2D) rather than just the
AnimatedSprite2D.

Back when the puzzle was first prototyped, modulate was applied just to
the AnimatedSprite2D, based on the pitch of the note. I removed this in
commit ffd20d011aeaa275247e8d9c9cb0dda6c4a37016
("MusicalRock: Remove automatic modulation by note") which has this note:

> This looks a bit weird in the editor because the rock's collision shapes
> and "kick" interaction label also get modulated. But at runtime the
> collision shapes are invisible, and the label is not actually used – it
> just sets the position where the label belonging to the player will be
> positioned. And I consider modulating the sprite to be a temporary
> measure; it would be better to recolour the rocks.

The great thing about temporary measures is they tend not to be temporary!

I am working on replacing the interaction label (not used at runtime)
with an indicator floating above each interactable object, which *is*
visible at runtime, and so modulating the whole scene will no longer
suffice.

Add a `sprite_modulate` property which proxies through to the
`AnimatedSprite2D` within. Update every scene to use it. (Not all
sequence puzzles use this approach: special mention to After the Tremor
which has per-object sprites.)

In effect, this is a manual version of the approach I rejected in commit
ffd20d0:

> (I tried using PropertyUtils to export the modulate property of the
> AnimatedSprite2D. It solved this problem but in my opinion the cure was
> worse than the disease: it added more complexity to what is otherwise a
> very simple scene, and the resulting "Sprite Modulate" property in the
> inspector had no documentation.)

If only Godot had a built-in way for a scene to expose properties of its
children without going all the way to editable children!

### DIFF
--- a/scenes/game_elements/props/sequence_puzzle_object/components/sequence_puzzle_object.gd
+++ b/scenes/game_elements/props/sequence_puzzle_object/components/sequence_puzzle_object.gd
@@ -19,6 +19,12 @@ const REQUIRED_ANIMATIONS: Array[StringName] = [&"default", &"struck"]
 @export var sprite_frames: SpriteFrames = DEFAULT_SPRITE_FRAMES:
 	set = _set_sprite_frames
 
+## Color to apply to the object's sprites. Use this rather than
+## [Node2D.modulate] to avoid tinting any other visible children of this node,
+## such as particles.
+@export var sprite_modulate: Color = Color.WHITE:
+	set = _set_sprite_modulate
+
 ## Sound to play when the player strikes this object.
 @export var audio_stream: AudioStream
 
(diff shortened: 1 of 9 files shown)


manuq (approved):
Makes sense to use a proxy in this case. Ugh you had to do it in all StoryQuests!

wjt:
It was quite an easy search-and-replace across the StoryQuests!

wjt:
@jgbourque I presume that in at least 1 guide about the sequence puzzle, you show people how to adjust `modulate` to change the object's appearance. It is better now to use this new `sprite_modulate` property. `modulate` will still work for now, but assuming https://github.com/endlessm/threadbare/pull/2150 is completed it will also apply to the "interact with this" arrow in an ugly way. `sprite_modulate` does not have this property.

PR #2150, opened 2026-04-28
Move interact verb/action to HUD

### LINKED ISSUE #2114: Move interact prompt label to HUD
Currently, when the player-character is moved near to something that can be interacted with, we show two indications:

1. Near the object, a verb or phrase, such as "Tap", "Talk", or "Talk to Oscár". In theory this label is meant to float above the target so that neither the player-character nor the target is obscured.
2. At the bottom of the screen, "Interact" appears along with the button to press.

Image

Image

Image

I think it would be clearer if the verb were shown in the bottom bar, with a glow or arrow behind/above the target to highlight it. Something like this quick mockup:

https://github.com/user-attachments/assets/9fe7ff88-46a3-4a70-b662-2f1ab735f2ff

though I wonder if an arrow, exclamation mark, etc. floating above the object would be more practical to implement & look better.

This would also allow us to WONTFIX https://github.com/endlessm/threadbare/issues/725.

### XP Summary (total 600)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 150 |

wjt:
Instead of showing text over the object in the world, show it on the HUD
at the bottom of the screen, in place of the "Interact" label.

Show a bouncing red arrow above the currently-targetted interactable
entity. Show nothing above interactable entities that are not currently
targetted, but define an "idle" animation for this case so that we could
in theory change this later just by adjusting the animation.

Define the arrow position with a Marker2D attached to the InteractArea.
Migrate all existing interact_label_position values to a new Marker2D
node on each - these will not all be correct, but some of them might be.
Manually adjust some, including the tricky case of the Stella tortoise.
(Doing this by moving a Marker2D around in the editor is so much easier
than when I did it by editing Vector2 values by hand!) For some cases
where the default position - 64px above the InteractArea - is good
enough, I removed the Marker2D for a simpler scene & smaller diff.

Make all InteractArea nodes visible. (In a number of cases we previously
hid them to reduce the visual clutter, but this would hide the arrow.) I
may have missed some but we can fix them as we find them.

At runtime, if no Marker2D is provided, place one 64px
above the InteractArea node - better than nothing.

Adjust the logic in character_sight.gd to only emit
interact_area_changed if it actually does change, rather than just
because the player moved so that a second object is in range without
changing the "best" object. (This has no visible effect but was
necessary to add a sound effect on change, which I tried then removed.)

Resolves https://github.com/endlessm/threadbare/issues/2114

### DIFF
--- a/scenes/game_elements/characters/player/components/player_interaction.gd
+++ b/scenes/game_elements/characters/player/components/player_interaction.gd
@@ -3,8 +3,9 @@
 class_name PlayerInteraction
 extends Node2D
 
-## Emitted when [method can_interact] will return a different value.
-signal can_interact_changed
+## Emitted when the entity the player is able to interact with changes (possibly
+## to nothing).
+signal interact_action_changed
 
 ## The character that gains interaction.
 ## [br][br]
@@ -14,8 +15,6 @@ signal can_interact_changed
 	set = _set_character
 
 @onready var character_sight: CharacterSight = %CharacterSight
-@onready var interact_marker: Marker2D = %InteractMarker
-@onready var interact_label: FixedSizeLabel = %InteractLabel
 
 @onready var player: Player = self.owner as Player
 
@@ -25,10 +24,15 @@ func _set_character(new_character: CharacterBody2D) -> void:
 	update_configuration_warnings()
 
 
-## Returns [code]true[/code] if the character is currently able to interact with
-## something in the environment.
-func can_interact() -> bool:
-	return character_sight.interact_area != null
+## Returns the human-readable text of [member character]'s current interact
+## action (e.g. [code]"Talk"[/code]), or [code]""[/code] (empty string) if
+## [member character] cannot currently interact with anything. Use [signal
+## interact_action_changed] to monitor for changes.
+func get_interact_action() -> String:
+	if character_sight.interact_area != null:
+		var action := character_sight.interact_area.action
+		return action if action else tr("Interact")
+	return ""
 
 
 func _is_interacting() -> bool:
@@ -47,14 +51,6 @@ func _ready() -> void:
 	_on_character_sight_interact_area_changed()
 
 
-func _process(_delta: float) -> void:
-	if not character_sight.interact_area:
-		return
-	interact_marker.global_position = (
-		character_sight.interact_area.get_global_interact_label_position()
-	)
-
-
 func _unhandled_input(_event: InputEvent) -> void:
 	if _is_interacting():
 		return
@@ -68,7 +64,6 @@ func _unhandled_input(_event: InputEvent) -> void:
 
 		get_viewport().set_input_as_handled()
 		character_sight.monitoring = false
-		interact_label.visible = false
 		interact_area.interaction_ended.connect(_on_interaction_ended, CONNECT_ONE_SHOT)
 		interact_area.start_interaction(player, character_sight.is_looking_from_right)
 
@@ -81,13 +76,4 @@ func _on_interaction_ended() -> void:
 
 
 func _on_character_sight_interact_area_changed() -> void:
-	if _is_interacting():
-		return
-
-	if not character_sight.interact_area:
-		interact_label.visible = false
-	else:
-		interact_label.visible = true
-		interact_label.label_text = character_sight.interact_area.action
-
-	can_interact_changed.emit()
+	interact_action_changed.emit()
--- a/scenes/game_elements/props/interact_area/interact_area.gd
+++ b/scenes/game_elements/props/interact_area/interact_area.gd
@@ -20,20 +20,26 @@ signal interaction_ended
 ## Emitted when characters start or stop seeing this area for interaction.
 signal observers_changed
 
-const EXAMPLE_INTERACTION_FONT = preload("uid://c3bb7lmvdqc5e")
-const EXAMPLE_INTERACTION_FONT_SIZE = 34
+const INDICATOR_SCENE := preload("uid://d252j2mhya0kq")
 
-## Vector2 that approximates the position in which the interact label would
-## appear when a player is close.
-@export_custom(PROPERTY_HINT_RANGE, "-200,200,1,suffix:px,or_greater,or_less")
-var interact_label_position: Vector2:
+## Position at which to show an arrow when this area is being observed. This
+## should generally be centred, slightly above the visible extents of the parent
+## object. If not set, a default position will be used, which may be good enough.
+@export var marker: Marker2D:
 	set(new_value):
-		interact_label_position = new_value
-		queue_redraw()
+		if marker and _indicator:
+			marker.remove_child(_indicator)
+		marker = new_value
+		if marker and _indicator:
+			marker.add_child(_indicator)
+
 @export var disabled: bool = false:
 	set(new_value):
 		disabled = new_value
 		set_collision_layer_value(Enums.CollisionLayers.INTERACTABLE, not disabled)
+		if _indicator:
+			_indicator.visible = not disabled
+
 @export var action: String = "Talk"
 
 ## Whether this area is being observed by one or more characters.
@@ -43,6 +49,8 @@ var is_being_observed: bool:
 
 var _observers: Array[CharacterSight] = []
 
+var _indicator: Node2D
+
 
 func start_interaction(player: Player, from_right: bool) -> void:
 	interaction_started.emit(player, from_right)
@@ -50,10 +58,6 @@ func start_interaction(player: Player, from_right: bool) -> void:
 
 func end_interaction() -> void:
 	interaction_ended.emit()
-
-
-func get_global_interact_label_position() -> Vector2:
-	return to_global(interact_label_position)
 
 
 ## A [CharacterSight] calls this when it starts seeing this area.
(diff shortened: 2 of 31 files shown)


wjt:
This is a draft for a number of reasons, not least that `interact_label_position` is not a good property name for the arrow position, and because there is a bug where the arrows appear much higher for the rocks in situ than the arrow does when editing the rock scene itself and I don't know why.

But another thing I realised while testing this… I had really hoped that we could avoid needing the ScreenOverlay CanvasLayer, on the basis that we seem to get away OK with the grapple direction arrow on the player being smaller when the view zooms out, and because we could then just WONTFIX #725.

However while testing this I happened to look at `res://scenes/quests/story_quests/stella/3_stella_sequence_puzzle/stella_sequence_puzzle.tscn` and discovered that the Guardian NPC in that scene is scaled to 1.5×1.5, and so due to the way I have implemented this feature, so is the arrow above her head. It looks really bad.

When I first prototyped this I put a static, dimmed/semitransparent arrow above all interactable objects, with the currently-targetted one (if any) bouncing and fully opaque. This looked really busy in areas with many interactable objects, such as sequence puzzles. If we are happy to only show a single arrow at a time (not sure!) then we could instead have the `PlayerInteraction` node get the position to show the arrow from the `InteractArea` it is targeting, and have a single arrow that gets moved around the screen as a child of the player node. A bit weird but it would mean the arrow follows the player scale (i.e. hopefully 1×, but if someone does scale the player for some reason it would also scale the grapple arrow...), leaving only the camera zoom problem.

Otherwise... maybe we are stuck with the ScreenOverlay (but could we use the InputHud autoload which is guaranteed always to be present?) and then we'll have to solve the dizzying effect still.

wjt:
Some scenes to look at:

- https://play.threadbare.game/branches/endlessm/wjt/move-interact-verb-action-to-hud/#res://scenes/world_map/frays_end.tscn
- https://play.threadbare.game/branches/endlessm/wjt/move-interact-verb-action-to-hud/#scenes/quests/lore_quests/quest_001/1_music_puzzle/music_puzzle.tscn
-  https://play.threadbare.game/branches/endlessm/wjt/move-interact-verb-action-to-hud/#scenes/quests/story_quests/stella/3_stella_sequence_puzzle/stella_sequence_puzzle.tscn

manuq:
> This is a draft for a number of reasons, not least that `interact_label_position` is not a good property name for the arrow position, and because there is a bug where the arrows appear much higher for the rocks in situ than the arrow does when editing the rock scene itself and I don't know why.

I see that in code you create a Marker2D and then instantiate the `interact_indicator.tscn` . Can we just remove the `interact_label_position` offset add a Marker2D directly as child of InteractArea in the scenes using it? Similar to `@export var anchor_point: Marker2D` in HookableArea. Or just instantiate the arrow directly in scenes, and make it an export of the InteractArea? Something like:

```
## Node2D used to display to indicate that this area can be interacted.
@export var indicator: Node2D
```

> 
> But another thing I realised while testing this… I had really hoped that we could avoid needing the ScreenOverlay CanvasLayer, on the basis that we seem to get away OK with the grapple direction arrow on the player being smaller when the view zooms out, and because we could then just WONTFIX #725.
> 
> However while testing this I happened to look at `res://scenes/quests/story_quests/stella/3_stella_sequence_puzzle/stella_sequence_puzzle.tscn` and discovered that the Guardian NPC in that scene is scaled to 1.5×1.5, and so due to the way I have implemented this feature, so is the arrow above her head. It looks really bad.

This is what I see: 
image

In my opinion it doesn't look that bad. This Guardian is an instance of Talker. Maybe it can just be a StaticBody2D with the AnimatedSprite2D (scaled at 1.5), plus the TalkBehavior, plus the InteractArea as children. Like a one-off node for this particular scene.

I also would like to say goodbye to ScreenOverlay :)

> When I first prototyped this I put a static, dimmed/semitransparent arrow above all interactable objects, with the currently-targetted one (if any) bouncing and fully opaque. This looked really busy in areas with many interactable objects, such as sequence puzzles. If we are happy to only show a single arrow at a time (not sure!) then we could instead have the `PlayerInteraction` node get the position to show the arrow from the `InteractArea` it is targeting, and have a single arrow that gets moved around the screen as a child of the player node. A bit weird but it would mean the arrow follows the player scale (i.e. hopefully 1×, but if someone does scale the player for some reason it would also scale the grapple arrow...), leaving only the camera zoom problem.

I remember watching a screenshare with the multiple semi-transparent arrows and yes, my first impression was that it was a bit busy. I like it one at the time, like in this draft.

When you say you would like to reuse a single arrow, is it for consistency? For optimization? I wouldn't mind having each InteractArea with their own arrow that gets disabled/enabled, at least as a first step.

> 
> Otherwise... maybe we are stuck with the ScreenOverlay (but could we use the InputHud autoload which is guaranteed always to be present?) and then we'll have to solve the dizzying effect still.

manuq:
After playing some more... I really like where this is going. Maybe the "[Space] Talk to..." in the HUD can have a visual/audio clue when it changes (for later). Also I wonder if the arrow is too tiny for the Eternal Loom or other big things to interact, but I guess it's fine!

wjt:
> I see that in code you create a Marker2D and then instantiate the `interact_indicator.tscn` . Can we just remove the `interact_label_position` offset add a Marker2D directly as child of InteractArea in the scenes using it? Similar to `@export var anchor_point: Marker2D` in HookableArea. 

That's what I plan to do. (And in fact you're on the right track that this was the cause of the misalignment bug - in one code path I set the position on the marker, and in the other I set the position on the indicator.)

I was hoping to have a clever tool script migrate all the scenes that set `interact_label_position` to instead add a marker at that position. Even if we have to (gradually!) go through and fix all their positions later.

> Or just instantiate the arrow directly in scenes, and make it an export of the InteractArea? Something like:
> 
> ```
> ## Node2D used to display to indicate that this area can be interacted.
> @export var indicator: Node2D
> ```

The reason I didn't do this is that we want (I think?) the indicator to be consistent between every interactable thing, so it would be manual work each time to put the indicator in place, and then a nightmare if we want to change how it works later!

> In my opinion it doesn't look that bad.

OK!

> This Guardian is an instance of Talker. Maybe it can just be a StaticBody2D with the AnimatedSprite2D (scaled at 1.5), plus the TalkBehavior, plus the InteractArea as children. Like a one-off node for this particular scene.

True!

> I remember watching a screenshare with the multiple semi-transparent arrows and yes, my first impression was that it was a bit busy. I like it one at the time, like in this draft.

👍🏻 

> When you say you would like to reuse a single arrow, is it for consistency? For optimization? I wouldn't mind having each InteractArea with their own arrow that gets disabled/enabled, at least as a first step.

It was for showing a single arrow in ScreenOverlay/InputHud, moved around to the current target. I don't think there is likely to be a performance issue with one arrow per target but all bar one being invisible.

OK, I will continue down this path.

wjt:
I finished migrating all existing `interact_label_position` values to `Marker2D`s. To do this I took advantage of the power of `@tool` scripts: in `InteractArea._ready`, if running in the editor, I would:

1. Add a `Marker2D` child, assigning it the correct owner to be saved to the scene
2. If `interact_label_position` was non-zero, I assigned its position to the `Marker2D` then cleared its value
3. Now I just opened every scene that uses `interact_area.gd` and saved all scenes.

I tweaked the Stella tortoise by hand.

I also tried adding a little sound effect whenever the interact target changes. I think it's good? I wondered about making the "(A) Talk to Sigurd" label emit particles in this case too but decided that's too much, and visual design is not my forté.

wjt:
One more problem I noticed: the arrow is in the world layer so is affected by weather.

scenes/ui_elements/input_hints/components/input_hud.gd
```diff
@@ -77,8 +79,22 @@ func _notification(what: int) -> void:
 			_update_visibility()
 
 
+func _on_interact_action_changed() -> void:
+	if (
+		not player_interaction.get_interact_action().is_empty()
+		and not interact_changed_sound.playing
+	):
+		interact_changed_sound.play()
```
wjt: This doesn't play on the web export and I have no idea why not!

wjt: Weirdly, it works in a local web build!

wjt: Works now. It's a mystery.

scenes/game_elements/characters/player/components/player_interaction.gd
```diff
@@ -3,8 +3,9 @@
 class_name PlayerInteraction
 extends Node2D
 
-## Emitted when [method can_interact] will return a different value.
-signal can_interact_changed
+## Emitted when the entity the player is able to interact with changes (possibly
```
manuq, line 6: At first read I though that you were trying to say either "the entity" or "the player", not both. But after a second read, I got it!

scenes/game_elements/characters/player/components/player_interaction.gd
```diff
@@ -25,10 +24,15 @@ func _set_character(new_character: CharacterBody2D) -> void:
...
 
-## Returns [code]true[/code] if the character is currently able to interact with
-## something in the environment.
-func can_interact() -> bool:
-	return character_sight.interact_area != null
+## Returns the human-readable text of [member character]'s current interact
+## action (e.g. [code]"Talk"[/code]), or [code]""[/code] (empty string) if
+## [member character] cannot currently interact with anything. Use [signal
+## interact_action_changed] to monitor for changes.
+func get_interact_action() -> String:
```
manuq, line 31: 👍

scenes/game_elements/props/interact_area/interact_area.gd
```diff
@@ -20,20 +20,26 @@ signal interaction_ended
...
+## object. If not set, a default position will be used, which may be good enough.
+@export var marker: Marker2D:
 	set(new_value):
-		interact_label_position = new_value
-		queue_redraw()
+		if marker and _indicator:
+			marker.remove_child(_indicator)
+		marker = new_value
+		if marker and _indicator:
+			marker.add_child(_indicator)
```
manuq, line 34: 👍

scenes/game_elements/props/interact_area/interact_area.gd
```diff
@@ -78,32 +82,19 @@ func _ready() -> void:
 	# Initialise interactable bit in collision_layer
 	disabled = disabled
 
+	if not marker and not Engine.is_editor_hint():
+		marker = Marker2D.new()
+		marker.name = "Marker"
+		# Vaguely sensible default position
+		marker.position = Vector2(0, -64)
```
manuq, line 89: 👍 as you say in the PR description, is better than nothing.

scenes/game_elements/props/interact_area/interact_indicator.gd
```diff
@@ -0,0 +1,18 @@
...
+var bouncing := false:
+	set = set_bouncing
+
+@onready var animation_player: AnimationPlayer = %AnimationPlayer
+
+
+func set_bouncing(new_value: bool) -> void:
+	bouncing = new_value
+	if animation_player:
+		animation_player.play(&"bounce" if new_value else &"idle")
```
manuq, line 14: I like that the bounce is done through AnimationPlayer!

scenes/ui_elements/input_hints/components/input_hud.gd
```diff
@@ -78,7 +78,12 @@ func _notification(what: int) -> void:
 
 
 func _update_player_state() -> void:
-	interact_input_hint.visible = player_interaction and player_interaction.can_interact()
+	if player_interaction:
+		var action := player_interaction.get_interact_action()
+		interact_input_hint.visible = not action.is_empty()
+		interact_input_hint.text = action
+	else:
+		interact_input_hint.visible = false
```
manuq, line 86: 👍

manuq (approved):

scenes/game_elements/characters/player/components/player_interaction.gd
wjt: Good point, it's a tricky sentence...

PR #2157, opened 2026-05-06
Add tutorial quest

### LINKED ISSUE #2123: Implement Tutorial Quests in Overworld
Reference Sarah's design spec from #2122

wjt:
Define a new LoreQuest, `quest_000`, which is triggered from the title
screen rather than the previous intro sequence. This quest has a new NPC
teach you the various mechanics of the game, with cut-down versions of
each challenge. Eventually you reach Fray's End, just as with any other
quest.

This is a first iteration: I would expect several further rounds of
scene-dressing and level design.

When I was looking for a name for the NPC, I learnt that “Arun” means
“dawn” in Sanskrit, which seemed apt for a first chapter. But I am not
attached to the name.

Resolves https://github.com/endlessm/threadbare/issues/2123

Co-authored-by: Sarah Spiers <sarah@endlessaccess.org>
Co-authored-by: Manuel Quiñones <manuq@endlessaccess.org>


wjt:
By the way, the erratic naming of the scene files and folders is because Backstitch does not handle renaming folders very well https://github.com/inkandswitch/backstitch/issues/210.

wjt:
> Play this branch at https://play.threadbare.game/branches/endlessm/wjt/tutorial-quest/.
> 

For once this PR is actually best tested by launching the game from the start and choosing "Restart" or "Start" (rather than "Continue").

manuq (approved):
Excellent! I love the tutorializations of both abilities, repel and grapple. It is impossible to fill the ink well by standing behind it, you must repel. And the places where you must grapple are evident because of the buttons. And the Arun character is guiding the player, explaining all.

About flow: I like that this replaces the introduction cinematic. Abandoning the quest sends the player to Fray's End, so maybe the text in the menu should be "Skip tutorial"?

Only detail: I see the second ability is named "grapple" and the HUD says "repel". We should change the text of the HUD hint too.

Image
Image

wjt:
I also noticed that the link to Fray's End is not quite right - IMO StoryWeaver should arrive on the lower island and walk across the river to Fray's End, but I forgot to choose that spawn point.

wjt:
> Only detail: I see the second ability is named "grapple" and the HUD says "repel". We should change the text of the HUD hint too.

I think the "repel" ability text is consistent between the powerup and the HUD.

The discrepancy is between "grapple" on the powerup and "throw" in the HUD.

PR #2158, opened 2026-05-06
Music puzzle: Remove 1.2× scaling on altars

wjt:
I think it looks better at 1× scale; and, I generally think it's better
practice to have the assets be the size we want rather than drawing them
at one size and scaling them to another.


wjt:
| Before | After |
| ------ | ----- |
| Screenshot From 2026-05-06 15-01-55 | Screenshot From 2026-05-06 15-02-19 |

manuq (approved):
Yes! From the screen capture, definitely better.

PR #2162, opened 2026-05-06
Townies: Apply shader to the group of sprites

manuq:
Add a CanvasGroup node for applying the same shader to all the canvas nodes at once.

Adapt the shader to the builtin shader of CanvasGroup.

Export properties for passing the root sprite, the one that is traversed to find other nodes to randomize. And another one for passing the single CelShaderRecolor node.

Update the test scene.

### DIFF
--- a/scenes/game_elements/characters/components/character_randomizer.gd
+++ b/scenes/game_elements/characters/components/character_randomizer.gd
@@ -28,14 +28,17 @@ extends CharacterBody2D
 @export var look_at_side: Enums.LookAtSide = Enums.LookAtSide.LEFT:
 	set = _set_look_at_side
 
+## The root sprite of the hierarchy. Usually the legs.
+@export var root_sprite: AnimatedSprite2D
+
+## The node that recolors the character.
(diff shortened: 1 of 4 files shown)


manuq:
I did this just to confirm that the problem in https://github.com/endlessm/threadbare/pull/2143#issuecomment-4313251834 is actually because of the CanvasGroup being used, and the way the color is being picked from the screen with:

```
	vec4 c = textureLod(screen_texture, SCREEN_UV, 0.0);

	if (c.a > 0.0001) {
		c.rgb /= c.a;
	}
```

https://github.com/user-attachments/assets/b33f0f35-cc6c-4ba3-bde3-29cc97562390

manuq:
In https://github.com/endlessm/threadbare/pull/2163 I will:
- Switch to regular shader parameters
- The same material will be shared by all the parts of the townie
- Each townie will have a unique material (a duplicate of the cel_shading_recolor_material.tres) to have different parameters.

I'll port the current shader (skin only) in #2163. Once that's done I'll do the full palette swap (hair, clothes).

PR #2163, opened 2026-05-06
Cel shading shader: Switch to regular shader parameters

manuq:
Do not rely on instance shader parameters for recoloring. Instead, make the material unique when a different recolor is needed. For instance, by duplicating the cel_shading_recolor_material.tres.


manuq:
I had to refresh in my mind what I was attempting here. It seems to work now, I will clean and put it for review tomorrow!

manuq:
Ready for review. This one was tricky, but is a first step towards a more generic palette swap (hair, clothes).

scenes/game_elements/characters/npcs/townie.tscn
wjt, line 3: This is still set as the `material` property on the root node. I'm a little confused with how this interacts with the `shader_material` custom property.

wjt:
Merged as part of #2143.

scenes/game_elements/characters/npcs/townie.tscn
wjt: In #2143 we removed both the `material` from the root node, and the `shader_material` from the character randomizer script.

PR #2166, opened 2026-05-07
Input HUD: Use consistent input button icons

wjt:
Previously we had a mixture of outline and filled icons; and of colour
and monochrome icons for Xbox and PlayStation.

Refactor the way that these icons are assigned so that they are stored
in the same JoypadButtonTextures resource as directional inputs. This
means that the textures for a device type can be configured all in one
place.

Consistently use outline icons, which are more legible and obscure less of the world.

Consistently use monochrome icons for Xbox. It is true
that official Xbox controllers do have colours on the A/B/X/Y buttons
but looking at some photos in an image search, the colours are either
used for the text (not the surrounding button), or are printed as a
little directional key in between the four buttons. I own two Xbox-style
controllers made by 8BitDo; they use the XBox RB/RT labels for shoulder
buttons rather than R1/R2 so the use of xbox textures for generic
controllers seems appropriate, but they do not use colours.

Similarly, use monochrome icons for PlayStation. Some PlayStation
controllers have coloured buttons, but many more do not, and I don't
think the colours are the strongest association.

Some of the Switch button textures were inconsistent with the input map;
fix these in the process. Unfortunately I was unable to pair my Switch 2
JoyCons with my laptop so I can't actually test these. (If we wanted to
support Switch controllers "properly", IMO we should swap the mappings
so that the Switch "A" button is interact, but that's one for another
lifetime.)

For champ_stealth, rather than adding these champ_specific actions to
the global input map, make interact_input.gd skip the per-device mapping
if it is not provided, which improves on the previous behaviour of
showing blank icons when using a controller.

For Sueños Nocturnos, which has a custom "Saltar cinemática." hint, use
the "interact" button (matching the previous hardcoded button textures),
and adjust the action it looks for to match so that joypad buttons
actually work.

### DIFF
--- a/scenes/game_elements/props/hint/input_key/interact_input.gd
+++ b/scenes/game_elements/props/hint/input_key/interact_input.gd
@@ -3,14 +3,11 @@
 extends TextureRect
 
 @export var action_name: StringName
-@export var keyboard_texture: Texture2D
-@export var xbox_controller_texture: Texture2D
-@export var playstation_controller_texture: Texture2D
-@export var nintendo_controller_texture: Texture2D
-@export var steam_controller_texture: Texture2D
+@export var alternative: bool = false
+@export var devices: InputHintManager
 
 
-func _physics_process(_delta: float) -> void:
+func _process(_delta: float) -> void:
 	if Input.is_action_pressed(action_name):
 		modulate = Color.GRAY
 	else:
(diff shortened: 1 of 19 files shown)


wjt:
| Before | After (filled) | After (outline) |
| ------ | ----- | ---- |
| image | Screenshot From 2026-05-07 10-38-49 | image  |
| image | image | image |
| image | image | image |
| image | image | image |

manuq:
Yes great, the button icons should be consistent!

> image	Screenshot From 2026-05-07 10-38-49
>

By looking at these screenshots and playing the branch, I find the outline versions better to the eye for readability and less intrusive, as they let seeing more of the world behind them.

In particular the filled icon for Shift is a bit unreadable:

image

Than:

image

But yes in general I agree with making all of the buttons consistent. Initially I thought that color icons were used because there weren't monochrome counterparts to use. But they are there!

manuq (approved):
@jofilizola you may have feedback for this as well.

wjt:
OK, I defer to more visually-skilled people on whether the filled or outline versions appear better - your pair of screenshots has convinced me. I'll revise this to use outline versions. We can try #2167 too - it looks good!

wjt:
I'll land this with consistent monochrome, outline icons; we can always iterate.

PR #2167, opened 2026-05-07
HUD hint buttons: Project shadow for better contrast

manuq:
Use the same offset and color than the text label shadow.

## Mouse and keyboard atlas: Remove rectangle of solid pixels

Which actually don't contain any icons. Because when cropping slices of this
atlas right below this area, the drop shadow shader does drop a shadow for the
pixels above.

### DIFF
--- a/assets/third_party/inputs/atlas_kenney_input_prompts_1.4/keyboard/keyboard-&-mouse_sheet_default.png
+++ b/assets/third_party/inputs/atlas_kenney_input_prompts_1.4/keyboard/keyboard-&-mouse_sheet_default.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:1846ff7e8ee22c3a4fb88efc8e569e98edef5b127bd4a6fbae11e16888578c92
-size 29862
+oid sha256:e7d1d25ca8515c413d2cbb675e11a2c792f2f97f13267433df247e1398dc4d6d
+size 24004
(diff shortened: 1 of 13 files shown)


manuq:
@wjt I remember we got feedback on having a better contrast in the HUD some time ago. I think it was in Discord. We could do the change in the assets themselves. Here I did it with a shader, at least for a quick try.

image

manuq:
This is how hints on the right side look:

Captura desde 2026-05-07 12-58-07

There is a problem with the mouse hint, it is projecting shadow at the top border:

Captura desde 2026-05-07 12-59-13

The region is correct, I think. But the texture has pixels painted above. One way to fix it is by making those unused pixels transparent:

Captura desde 2026-05-07 12-53-22

manuq:
Fixed by just removing the solid rectangle from the atlas asset:

image

wjt:
I think I saw another way to avoid this oversampling of the texture - let me check.

wjt:
Found it. AtlasTexture has a `filter_clip` property:

image


I think that would work. But maybe it has a cost; and just deleting the extraneous pixels also works.

manuq:
> Found it. AtlasTexture has a `filter_clip` property:

Oh good to know!

> image
> I think that would work. But maybe it has a cost; and just deleting the extraneous pixels also works.

As you wish. I can revert the texture change. But yes we really don't need those pixels.

wjt:
I think this breaks the effect where the hint gets subtly tinted while the button is pressed. From main:


https://github.com/user-attachments/assets/9bbaa78b-77f1-48db-a630-ba8f57a18530

manuq:
> I think this breaks the effect where the hint gets subtly tinted while the button is pressed. From main:
> 
>  Screencast.From.2026-05-08.15-29-00.mp4

Oh good catch! Back to draft.

manuq:
I ended up extracting the modulate color from the vertex shader as in this comment: https://github.com/godotengine/godot/issues/49781#issuecomment-2876900754

It worked!

manuq:
Ouch my local push failed for networking reasons and I ended up merging without the fixup. Done in https://github.com/endlessm/threadbare/pull/2194 now.

wjt:
> I ended up extracting the modulate color from the vertex shader as in this comment: [godotengine/godot#49781 (comment)](https://github.com/godotengine/godot/issues/49781#issuecomment-2876900754)
> 
> It worked!

We do that in the tree and void shaders too!

PR #2170, opened 2026-05-07
Integrate Void particles

### LINKED ISSUE #1857: Void: Replace placeholder void bloom

manuq:
As delivered by PixelFrog, and composed the 3 of them in a row with GIMP. And change the amount of rows to 3, the new amount of particle variations.

Resolve https://github.com/endlessm/threadbare/issues/1857

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/components/dust-cloud.png
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/components/dust-cloud.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5b8a01835dedcdc441a3ade4663745e3547744af0526aa0c98a7fd855d317c8b
-size 603
+oid sha256:2c1e1bb5298eaca60a0aaae074c09e23ea31e873e80e7299263d93833c3c6e96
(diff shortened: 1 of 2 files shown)


wjt (commented):
As discussed in playtest: I think we need to change the particle emission parameters a bit with this. @manuq you also thought it would look good to apply an outline shaedr.

manuq:
> As discussed in playtest: I think we need to change the particle emission parameters a bit with this. @manuq you also thought it would look good to apply an outline shaedr.

Yes, I experimented with both in this draft pull request, what do you think? https://github.com/endlessm/threadbare/pull/2171#issuecomment-4399478891

The problematic places, in my opinion, are the ones in which the enemy is patrolling on top of Void tiles:

[recording.webm](https://github.com/user-attachments/assets/d14bad69-98c9-414e-8104-8b72e58bf657)

wjt:
I agree that the patrolling void is more problematic. and I think the outline-shader version looks amazing.

manuq:
> I agree that the patrolling void is more problematic. and I think the outline-shader version looks amazing.

Ok! Closing this one in favour of https://github.com/endlessm/threadbare/pull/2171

PR #2171, opened 2026-05-07
Void particles outline

### LINKED ISSUE #1857: Void: Replace placeholder void bloom

manuq:
As delivered by PixelFrog, and composed the 3 of them in a row with GIMP. And change the amount of rows to 3, the new amount of particle variations.

## Void enemy: Add outline to all particles

To do it, because the GPUParticle2D needs a CanvasItemMaterial for the 3
variations, add a CanvasGroup and apply the outline shader to it.

Also, change the particles to:
- Animate the scale animation now starting at zero and ending at zero too.
- Do not animate the alpha of particles anymore.
- Emit from the sphere surface (not from sphere).

Resolve https://github.com/endlessm/threadbare/issues/1857

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/components/dust-cloud.png
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/components/dust-cloud.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5b8a01835dedcdc441a3ade4663745e3547744af0526aa0c98a7fd855d317c8b
-size 603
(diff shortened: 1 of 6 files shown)


manuq:
The difficult part of making this outline is that, for it to be applied to all the particles, a CanvasGroup has to be used, and the shader has to be adapted to use screen coordinates:

[void_with_outline.webm](https://github.com/user-attachments/assets/4e5f70ae-ca43-4247-941c-d38f866d7b1e)

[void_with_outline2.webm](https://github.com/user-attachments/assets/d776ef1f-ebc8-401d-ae6b-dd3f9d7c2e45)

wjt (approved):

PR #2172, opened 2026-05-08
Add butterfly scene

wjt:
Add butterfly scene

This is just a decoration. I only defined SpriteFrames for the
butterflies that sit on the ground. The ones that fly will need more
thought.

(cherry picked from commit df903223a6b0a0ab4998d4221a7e77e30d9c229f)


manuq (approved):
Excellent! A butterfly ready to be instantiated!

PR #2173, opened 2026-05-08
Void decoration tiles

### LINKED ISSUE #2164: Void: Integrate affected-by-void decoration tiles
Here are the tiles delivered by PixelFrog:

[Void Tiles.zip](https://github.com/user-attachments/files/27461619/Void.Tiles.zip)

manuq:
Integrate "affected by void" tiles

As delivered by PixelFrog.

Add new decoration tiles to the Void tileset. This has the advantage that the
shader work on them too, because they have the red mask.

For the animated ones, set them to start playing at random frames, like the
foam. And use the 10 frames per second that is the common framerate of the
project.

Paint polygons with a small corner collision for the "non walkable floor" layer,
in most of the corner tiles. Otherwise, the StoryWeaver can look like standing
on the Void.

## Use them in scenes

This is just a start to see them in game.

Use the static ones in the grappling_hook_powerup scene, at the sides of the
enemy patrolling.

Use the animated ones (and some static ones) in the grappling_hook_start scene,
for the big gap separation. Hopefully players will identify the place when going
back to finally grab the magical thread.

Resolve https://github.com/endlessm/threadbare/issues/2164


manuq:
Some captures:

image


https://github.com/user-attachments/assets/9785c9fa-fff0-455f-9e0d-79b80727a1c3

assets/first_party/tiles/void_affected_tiles_animated/void_animated-top.png
wjt, line 1: I think this one is missing the chromakey

wjt (approved):
This looks great but I wonder if we should go further and replace the "sewn" right-angle edges of the void entirely with the (static) new void?

manuq:
> This looks great but I wonder if we should go further and replace the "sewn" right-angle edges of the void entirely with the (static) new void?

This is a call to @PlayMatters ! I considered them decoration tiles to paint on top. If we really want to make this a replacement for the void carpet terrain, then it must be redone by the artist. Because this is the opposite than the existing one: you are painting the land behind (transparent) tiles, while the Void is at the borders of the tiles. In the existing texture, you paint the Void (red) tiles. I tried editing the delivered texture to flip the tiles but it breaks the seamless of the texture. Also if we want to replace the existing texture, there are some missing tiles. I sketched them in yellow here:

Captura desde 2026-05-13 09-08-14

But if we want to replace the existing texture then the land (transparent) should be at the borders (quickly sketched below). Which in my opinion won't look as good as the ones delivered. Because the land areas have to be thinner:

Captura desde 2026-05-13 09-07-582

wjt:
OK, I suggest we integrate these tiles as-is as decorations. Even if we later invert them. As you say it's an artistic decision!

PR #2174, opened 2026-05-10
[1/1] Adjust player collisions while walking on bridges

wjt:
**Stack:**

* https://github.com/endlessm/threadbare/pull/2174

---

Adjust player collisions while walking on bridges

Currently adding a bridge is tedious and error-prone, because the player
collides with cliff-edges and water - exactly what the bridge is
supposed to be crossing! So we have alternative tiles for water & bits
of cliff which do not have collisions - and worse, in the case of water,
we have a second alternative which only has a bottom collision so that
we can have a pontoon that ends in the middle of the water.

Add two new named physics layers, bridge and bridge_edge. Adjust the
bridges tileset to have bridge shapes along the walkable area of bridges
(including their ends) and bridge_edge on the bits of bridges that the
player should not be able to walk off.

Add an Area2D to the player which detects when the player's feet are
touching "bridge"; while the player is on a bridge, it adjusts the
player's collision_mask so that the player does not collide with wall
(cliff edges) and non_walkable_floor (water) and instead collides with
bridge_edge.

The void was previously non_walkable_floor, like water, but because it
sits conceptually over the world, the player shouldn't be able to walk
across it while on a bridge. (This is contradicted by the closing
cutscene of quest_001 but is otherwise always the case.) Define yet
another layer, "void", and move void collisions to that. Update the
player's collision_mask accordingly.

In Fray's End, replace the alternative water and cliff tiles with
regular ones.

To do the pontoon, we need an alternative lower end of bridge tile that
has bridge_edge at its bottom and sides so that the player can't walk
off into the water. This needs to be done manually - annoying but at
least it's on the same layer as the rest of the bridge, unlike the
alternative cliff and water tiles.

Teach the guard to walk across bridges.

Add a new bridge gym scene for testing.

Thanks to Lilith Duncan for helping me think through better approaches
to elevation and bridges!


wjt:
Hmm, this will break grappling across a bridge, because both the grappling logic and this bridge logic adjust the "collide with water" bit. I guess it will need a little more thought about where the player's physics layers should be controlled.

manuq:
@wjt nice! I think a better approach to the "hidden alternative tiles" will be very much welcome.

Besides, grapping, I was able to walk back into the void:

image

wjt:
Working on making it interact better with grappling I realised I can't make it depend on `Player` because NPCs also need to be able to walk across bridges!

wjt:
[npc_bridge.webm](https://github.com/user-attachments/assets/41434e47-4fa9-4113-b076-96b3d09b5a8b)

OK that wasn't so har.

wjt:
Another problematic case: bridge perpendicular to water. When your feet are on the start of the bridge, you gain the ability to walk on water.

https://github.com/user-attachments/assets/c1ab9074-cb1a-40de-9308-e142d95f9f78

In this case there is a level design problem: no cliff beneath the path.

image

But because we also disable cliff collisions, adding the cliff doesn't fix it!

wjt:
Maybe this can be fixed by making the "is a bridge" box a bit smaller, rather than having it be right up against the edge of the tile…

wjt:
I started building a gym with all the (literal) edge cases and found still more with bridges at corners. Need to finish adding all the combinations of these, tomorrow...

What's hard is that cliff edges protrude into the visually-walkable area of the clifftop.

wjt:
image

wjt:
I have given up on this for now.

Another less-invasive approach would be a script attached to bridge TileMapLayers that adjusts the other layers to put the appropriate water/cliff alternates in the right places.

PR #2177, opened 2026-05-13
Add and use sheep decoration scene

wjt:
This is currently only used in frays end west, but will be used in the tutorial quest.


wjt:
y-sort origin is wrong, fixing.

manuq (approved):

PR #2178, opened 2026-05-13
Improve tutorial quest scenes

wjt:
This primarily dresses the scenes with props and decorations,
but also includes some minor tweaks to the routes the player should follow.

Co-authored-by: Stephen Reid <stephen@endlessaccess.org>


manuq:
Looking great! I love the butterflies.

After defeating the InkDrinkers in the tutorial, I can't walk where they were. Is that intended?

https://github.com/user-attachments/assets/6ceafd17-3cd7-46cf-9e9c-16be22077003

wjt:
No, it's not. I'll fix that.

wjt:
There was water hiding under the grass. I fixed that; fixed the bottom row of cliffs to be a kind that blocks you from walking rather than a kind that needs water to block your movement; and tweaked the camera position 16 pixels to the left, which IMO is an improvement.

wjt:
And removed a duplicate door!

manuq (approved):

PR #2181, opened 2026-05-13
Replace "Abandon Quest" with "Skip Tutorial" for tutorial quest

### LINKED ISSUE #2160: Change "Abandon Quest" to "Skip Tutorial" in pause screen
It is valuable that you can abandon the tutorial quest, e.g. because we want to get someone straight to Fray's End to talk to the StoryQuest Elder.

But it would be clearer to have the pause menu item be labelled "Skip Tutorial" rather than "Abandon Quest".

wjt:
Unlike abandoning a quest, when we skip the tutorial we must grant the
player the abilities that they would have gained by completing the
tutorial.

This is a bit overengineered but it felt better to list the abilities in
the tutorial's quest.tres rather than to hardcode them in the pause
menu.

Resolves https://github.com/endlessm/threadbare/issues/2160

### DIFF
--- a/scenes/menus/storybook/components/quest.gd
+++ b/scenes/menus/storybook/components/quest.gd
@@ -43,7 +43,19 @@ const FILENAME := "quest.tres"
 @export_range(0, 6, 1, "suffix:threads") var threads_to_collect: int = 3
 
 ## Whether this is a lore quest (part of the main storyline).
-@export var is_lore_quest: bool = false
+@export var is_lore_quest: bool = false:
+	set(new_value):
+		is_lore_quest = new_value
+		notify_property_list_changed()
+
+## Whether this quest is skippable (i.e. is the tutorial). Only supported for lore quests.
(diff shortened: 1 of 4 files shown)


scenes/menus/storybook/components/quest.gd
```diff
@@ -43,7 +43,19 @@ const FILENAME := "quest.tres"
...
+		notify_property_list_changed()
+
+## Whether this quest is skippable (i.e. is the tutorial). Only supported for lore quests.
+@export var skippable: bool = false:
+	set(new_value):
+		skippable = new_value
+		notify_property_list_changed()
+
+## Which abilities to award if the quest is skipped
+@export var skip_abilities: Array[Enums.PlayerAbilities] = []
```
manuq, line 58: Oh yes I understand now why this is not as simple as I was expecting. Abilities must be awarded.

manuq (approved):

Issue #1186, opened 2025-09-14
It is hard to control the Sokoban player with an analogue stick


wjt:
### Description

It is hard to control the player in Sokoban puzzles using an analogue stick. As you move the stick around slightly (common even if you're trying to hold one direction) the player moves very rapidly. I noticed this during a quick playtesting session with my daughter so was not able to dig into why this was at the time.

If I had to guess, I would guess that slight movements of the joystick get understood by `directional_input.gd` to be releasing and then pressing the same action, repeatedly. I'm not entirely sure how to fix this (without debugging further) but one idea might be that rather than using `is_action_pressed` and `is_action_released`, the logic could be changed to (roughly):

- Get direction vector with `Input.get_vector()`
- Snap it to length 0 or 1 (I think the length will either be 0 or between 0.5 and 1 because the deadzone on all the directional input actions is 0.5)
- Quantise it to one of the 8 directions (or possibly 4 - perhaps better to ignore the diagonals)
- If it has changed from the previous stored direction, treat this as a new directional input and process it immediately
- Otherwise, consider it as holding down the same direction, and use the existing repeat-after-delay logic

Some experimentation will be necessary to find logic that feels good.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #2183, opened 2026-05-13
Use left-stick not dpad in input hints

wjt:
Also adjust the Playstation L1/R1 hints to the "alternative" version,
which looks a little truer to life.

The one downside of showing the left-stick rather than dpad is that the
stick works poorly in Sokoban
(https://github.com/endlessm/threadbare/issues/1186) but that is just a
bug to fix.


JoniCeceri (commented):
Tested in-engine, looks good to me!

wjt:
Thanks!

PR #2184, opened 2026-05-13
Add tree affected by Void

### LINKED ISSUE #2165: Void: Integrate affected-by-void trees
The ZIP file attached in https://github.com/endlessm/threadbare/issues/2164 also contains two animations for the tree trunk being affected by the Void. Let's integrate them.

manuq:
Add an Aseprite file for the trees, using the existing PNG files. Crop the
treetop from the trunk in separate layers, leaving a single trunk. Then add
layers for void-affected trunks, as delivered by PixelFrog. Also split the
shadow in it's own layer, since the 2 trunks delivered didn't have one.

Create a new scene, duplicating the existing tree scene. This one has 2
AnimatedSprite2D nodes: one for the tree trunk and one for the treetop. They
also share the same material with the wind shader.

Note that the treetop spritesheets have empty frames, because the trunk
animation is larger (14 frames) than the treetop one (6 frames). Pick only the
first 6 frames in the SpriteFrames resources for the treetop.

## grappling_hook_needles.tscn: Use trees with void

Replace all yellow trees with affected-by-void trees, also yellow for the
treetop, and vary the trunk with the two existing variations.

Resolve https://github.com/endlessm/threadbare/issues/2165


manuq:
Ouch I made this branch on top of the tiles one, but it doesn't have to! I'll fix it tomorrow.

scenes/game_elements/props/tree/components/treetop_spriteframes_yellow.tres
wjt, line 1: Wrong folder, the others are in `components/`

scenes/game_elements/props/tree/components/tree_void_affected.gd
```diff
@@ -0,0 +1,49 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends Node2D
+
+@export var treetop_sprite_frames: SpriteFrames = preload("uid://d2i5boaxe3n3g"):
+	set = _set_treetop_sprite_frames
+
+@export var trunk_sprite_frames: SpriteFrames = preload("uid://euk5lvoroc5"):
+	set = _set_trunk_sprite_frames
```
wjt: No need to change this, just an idea: maybe in this case we could instead have 

```
@export_enum("Blue", "Green", "Purple", "Red", "Yellow") var treetop_color: int
@export_range(0, 1) var trunk_variant: int
```

and then code to map these to SpriteFrames. It might be more level-designer-friendly than finding the SpriteFrames by hand. OTOH what you have here matches the rest of our components!

wjt: We could even combine this and the regular tree, with a `var void_level: int` where `0` means "plain trunk" and `1` and `2` mean "some amount of void". Then we could change it on the fly in code as the void gets closer to or further from the tree. (Again, just an idea for the future.)

wjt (commented):
Looks great!

wjt:
image

https://play.threadbare.game/branches/endlessm/integrate-void-trees/#quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_needles

scenes/game_elements/props/tree/components/treetop_spriteframes_yellow.tres
manuq: Ouch sorry! I just pushed a fixup commit.

scenes/game_elements/props/tree/components/tree_void_affected.gd
manuq: Oh yes more specific and friendly exports is a great idea!

As per combining it with the regular tree, that was my initial idea. It makes sense for the main game. But then I though, this may be used in StoryQuests and a simple texture change could be easier than changing 2 textures.

Now, looking at existing StoryQuests, none of them are customizing the tree sprite. they all are using the existing spriteframe resources:

| Scene | Override |
|--------|--------|
| `stella/3_stella_sequence_puzzle/stella_sequence_puzzle.tscn` | `tree_spriteframes_yellow.tres` (one tree only) |
| `renya_beyond_sorrow/1_combat/renya_combat_round_1.tscn` | `tree_spriteframes_purple.tres` (4 trees) |
| `el_abrigo/3_el_abrigo_runner/el_abrigo_runner.tscn` | `tree_spriteframes_red.tres` (29 trees) |
| `el_abrigo/4_el_abrigo_outro/el_abrigo_outro.tscn` | `tree_spriteframes_red.tres` (9 trees) |
| `el_abrigo/0_el_abrigo_intro/el_abrigo_intro.tscn` | `tree_spriteframes_red.tres` (12 trees) |

manuq:
@jgbourque what do you think a learner would like to do, for adding their own tree with a single texture?
1. Intantiate the tree scene, and customize the sprite frames (which is a proxy to the sprite frames of the AnimatedSprite2D node).
2. Duplicate the tree scene, and customize the sprite frames.
3. Create their own tree scene, and reuse script and wind shader from the main tree scene as needed.

Back to draft to do the friendlier exports. I'll probably leave it as a separate scene for now until I get some info from Learning.

PR #2188, opened 2026-05-14
NO_EDIT outro: Add a forest

wjt:
Talking with Justin, we thought this would be a suitable place to drop in an example of using StaticBody2D + AreaFiller to make a forest.


manuq (approved):
Great idea!

Issue #2029, opened 2026-03-06
butterfly_air_loop_*.png textures too large for Godot


wjt:
`res://assets/first_party/wildlife/butterfly/butterfly_air_loop_blue.png` and the other colours are 16896×192. If you try to use them you will see the following error:

```
ERROR: Condition "p_width <= 0 || p_width > 16384" is true.
   at: texture_set_size_override (drivers/gles3/storage/texture_storage.cpp:1591)
```

The 16384×16384 texture size limit could be charitably described as "poorly documented". https://docs.godotengine.org/en/latest/tutorials/3d/3d_rendering_limitations.html#texture-size-limits advises against using textures larger than 8192×8192 on desktop or 4096×4096 on mobile in 3D. I'm not sure if the same applies to 2D.

In the specific case of these butterfly textures we will need to wrap them to be more than one row.

PR #2197, opened 2026-05-15
shjourney: Wrap intro animation sprite sheet

wjt:
Previously 12804 x 80, now 1584 x 720 which is less than 4096 x 4096.

Helps https://github.com/endlessm/threadbare/issues/2029


manuq (approved):
I didn't realize this cinematic was done frame by frame, ugh.

Better than documenting the texture size limit in the wiki, we could have a pre-commit check.

wjt:
yes we should! I didn't find a prewritten one in a quick search earlier. We would also have to change the pre-commit check to run in a git lfs-enabled checkout.

PR #2200, opened 2026-05-15
Remove unused intro scene

wjt:
This was replaced with the tutorial quest.


manuq (approved):
Bye intro! The sketches are still displayed in the credits section.

PR #2201, opened 2026-05-15
Miscellaneous tutorial quest tweaks

wjt:
Mostly visual changes - adding shadows, tweaking transitions – but also a slight layout change to the sequence puzzle to address feedback that the NPC is too close to the objects.


manuq (approved):
Nice! I did a playthough, the NPC position in the sequence puzzle is better.

PR #2202, opened 2026-05-15
void_spreading_enemy: Reduce walking speed slightly

wjt:
The enemy runs when it is more than a certain distance from the player,
and walks when it is close. This makes it possible to outrun the enemy
without sprinting, and gives a sense of urgency while also giving you a
bit of grace if you need to pause and grapple.

Tweak both values slightly to make it a bit easier:

- Walk when the player is within 256 px (4 horizontal tiles), not 192 px
  (3 horizontal tiles);
- Reduce the walk speed, which was already a little lower than the
  player's.

This should make the initial void runner sequences much less
challenging, and the last void_grappling sequence a bit fairer.

### DIFF
--- a/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.tscn
+++ b/scenes/quests/lore_quests/quest_002/1_void_runner/components/void_spreading_enemy.tscn
@@ -21,7 +21,7 @@ metadata/_custom_type_script = "uid://csev4hv57utxv"
 
 [sub_resource type="Resource" id="Resource_pumxu"]
 script = ExtResource("3_45we0")
-walk_speed = 270.0
+walk_speed = 250.0
 metadata/_custom_type_script = "uid://csev4hv57utxv"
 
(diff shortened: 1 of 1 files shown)


wjt:
This is mainly in response to https://github.com/endlessm/threadbare/discussions/2179. But the difficulty of the void runner sequences is a recurring problem.

I have to say that I think this makes res://scenes/quests/lore_quests/quest_000/5_void_runner/tutorial_void_runner.tscn and res://scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn way too easy, but res://scenes/quests/lore_quests/quest_002/3_void_grappling/void_grappling.tscn and particularly res://scenes/quests/lore_quests/quest_002/3_void_grappling/void_grappling_round_2.tscn feel less unfair.

manuq:
I think is fine for now. In the future, the townie helping system could be the answer to make a more accessible game, while keeping it difficult for player that enjoy the challenge.

manuq (approved):

PR #2205, opened 2026-05-16
Fix running Storybook scene directly

wjt:
Previously the default value for `Storybook.quest_directory` was
`res://scenes/quests/story_quests`.

In commit 226ff248b I moved the call to `_enumerate_quests()` (which
scans that folder to feast on the quests within) from `_ready()` (which
is only called once the node enters the tree – but the Elder wants to
check if there are any quests before showing the book) to the setter of
the `quest_directory` property.

This broke running the scene directly because of a GDScript quirk that
it's taken me nearly two years to notice: setters are not called for
default values!

The StoryQuest Elder only works because it explicitly re-sets the
`quest_directory` property rather than relying on the default.

Change the default value in the script to empty string. Set it to
`res://scenes/quests/story_quests` in storybook.tscn instead.


manuq (approved):
This is good to know for me as well.

PR #2209, opened 2026-05-18
Rename filenames containing ' and %

wjt:
Rename filenames containing ' and %

I am currently looking at a number of project directories where these
files have been renamed by something. I am not sure what they are being
renamed by.

While these characters are (as far as I know) legal on all operating
systems, ' is annoying to quote at the shell, and % is annoying in
URLs.


manuq (approved):
Yes, thank you.

PR #2213, opened 2026-05-18
tutorial stealth: Adjust hiding places

wjt:
tutorial stealth: Adjust hiding places

Sarah found that in the previous design it was possible to be caught
while hiding in the hiding space.

Redesign the first passageway. Add some sand paths to mark important
areas along the route. Add some more water rock decorations and patches
of void.


manuq:
I think this is better! @saspiers you may want to try it as well. Quick playthrough:

https://github.com/user-attachments/assets/4172eee1-66fc-479a-8bcb-a19736eab1f1

manuq (approved):

PR #2214, opened 2026-05-18
Add Phantom Camera addon

### LINKED ISSUE #2175: Add Phantom Camera addon
https://phantom-camera.dev/ seems generic enough, and would make it very easy to:
- Transition from framing the player to frame another object, while describing it (eg. "this is the jar you need to fill")
- Frame a group of entities.
- And much more.

After discussing briefly with @wjt the pros/cons of using an addon for camera, we think  that this one is worth it.

manuq:
Prepare pre-commit and REUSE files for Phantom Camera addon.

Import Phantom Camera addon. Import latest from the Godot Asset Library.

project.godot: Activate Phantom Camera addon.

Resolve https://github.com/endlessm/threadbare/issues/2175


wjt (approved):
Let's try it!

PR #2215, opened 2026-05-18
Grappling hook: Add hook indicators

### LINKED ISSUE #1503: Grappling hook: Explore "hookable" indicator
From [playtest feedback](https://github.com/endlessm/threadbare/discussions/1428#discussioncomment-14778192):

> ### Suggestion: Visual Indicator for Next Stitching Needle
> To improve clarity during the stitching sequence, it would be helpful to add a visual cue for the next needle to interact with. This could be implemented in two ways:
> 
> * A small directional triangle pointing to the next needle
> 
> image
> * A highlight or glow effect on the target needle
> 
> image
> As mentioned earlier in [discussion #1425](https://github.com/endlessm/threadbare/discussions/1425#discussioncomment-14767394), this section can be confusing for players — especially in areas where the interaction range is only reachable from a specific position. For example, in the second part of Stage 2, one of the needles is only accessible from the middle of the river, which feels unintuitive without guidance.

manuq:
## UI indicators: Add specific asset for interacting

Rename the assets to how they are used in the game. Duplicate the previously named "indicator_down.png" and modify this asset so it looks like a yellow arrow. The red one (which matches the aim indicator around the player) will be used for indicating when something can be hooked.

## Grappling hook: Add hook indicators

Maintain an array of observers in HookableArea. In the same way as InteractArea.
But these are the HookControl nodes that are pointing at this area.

Add a marker for the indicator in HookableBox. The fallback that automatically
creates the marker 32 pixels above the anchor point seems fine for pins and
needles.

## Grappling hook: Pause aiming also when hooking

The last hook control was set to the AIMING_PAUSED state only when hitting a
wall or the air. It should pause too while hooking. As before, if the hooked
area has a hook control (like the needles), disable this control to "transfer"
the aiming to the next one.

## Hook control: Fix bug in _set_state

The setter for state wasn't checking if the node was ready.

Resolve https://github.com/endlessm/threadbare/issues/1503


manuq:
I recommend testing this in the following scenes:
- https://play.threadbare.game/branches/endlessm/indicators/#quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_needles
- https://play.threadbare.game/branches/endlessm/indicators/#game_elements/components/custom_hookable_objects_test

[recording.webm](https://github.com/user-attachments/assets/7e532c87-737a-4ce0-98a3-60393f1327c1)

It seems to work. The problem I'm still trying to solve is that the indicator stays after throwing. This is noticeable when collecting buttons:

[recording.webm](https://github.com/user-attachments/assets/ec41da5d-365c-43a1-9edb-92442d223735)

The HookControl has a AIMING_PAUSED state and I'm trying to use it, but that state is obviously not being set.

manuq:
OK I've found and fixed the issue above. It was indeed a case that was missing setting the control to "aiming paused".

scenes/game_elements/props/hookable_area/components/hookable_indicator.gd
wjt, line 1: This is a copy-paste of `interact_indicator.gd`. I guess we may want to change the behaviour of one or the other at some point in the future, so, OK!

scenes/game_elements/props/hookable_area/components/hookable_area.gd
```diff
@@ -67,6 +92,20 @@ func _ready() -> void:
...
+		indicator_point = Marker2D.new()
+		indicator_point.name = "IndicatorMarker"
+		add_child(indicator_point)
+		# Vaguely sensible default position
+		indicator_point.global_position = get_anchor_position() + Vector2(0, -32)
+
+	_indicator = HOOKABLE_INDICATOR.instantiate()
+	if indicator_point:
+		indicator_point.add_child(_indicator)
+		# _indicator.bouncing = true
```
wjt: Leftover commented-out code?

scenes/game_elements/props/hook_control/components/hook_control.gd
```diff
@@ -183,23 +187,40 @@ func release() -> void:
...
+		if area not in exclude_areas:
+			clear_pointing_at_area()
+			area.add_observer(self)
+			_pointing_at_area = area
 
 
 func _set_state(new_state: State) -> void:
 	state = new_state
-	if not ready:
+	if not is_node_ready():
```
wjt, line 214: What's the difference between property `ready` and function `is_node_ready`? Oh, the former is a signal so is always truthy! Good catch.

scenes/game_elements/props/hook_control/components/hook_control.gd
```diff
@@ -183,23 +187,40 @@ func release() -> void:
 		hooked_to = null
 
 
-func _can_connect() -> bool:
+## If an area was being pointed at, clear it.
+func clear_pointing_at_area() -> void:
+	if _pointing_at_area:
+		_pointing_at_area.remove_observer(self)
+		_pointing_at_area = null
```
wjt, line 194: Another way to spell this would be with a setter on the property above:

```gdscript
var _pointing_at_area: HookableArea:
	set = _set_pointing_at_area

func _set_pointing_at_area(new_value: HookableArea) -> void:
	if _pointing_at_area == new_value:
		return
	if _pointing_at_area:
		_pointing_at_area.remove_observer(self)
	_pointing_at_area = new_value
	if _pointing_at_area:
		_pointing_at_area.add_observer(self)
```

Then you would just assign to `_pointing_at_area` in `_update_pointing_at_area`. Not sure if it's clearer!

wjt: OK, it's more complicated than I thought because of the `exclude_areas`. Ignore my suggestion!

wjt (approved):
Looks great! I like that the new interact indicator is a different shape as well as colour.

scenes/game_elements/props/hookable_area/components/hookable_area.gd
manuq: Oh good catch!

scenes/game_elements/props/hook_control/components/hook_control.gd
manuq: Yes I forgot to mention this fix in the commit message, fixing it now.

scenes/game_elements/props/hookable_area/components/hookable_indicator.gd
manuq: The other option I handled was to move the interact_indicator.gd to just indicator.gd and reuse it. But yes unless we want consistency between all the indicators... I think the missing one is the arrow pointing at things off camera.

scenes/game_elements/props/hook_control/components/hook_control.gd
manuq: The setter was an interesting idea, though.

PR #2218, opened 2026-05-19
sheep: Shake during interaction

wjt:
Make the sheep wobble a little when you interact with it, as a visual counterpart to the “baa” sound. Particularly if the game is muted, this makes annoying the sheep have a slightly more tangible effect.

Helps #2220


wjt:
This is a silly idea to help #2220

manuq:
> This is a silly idea to help #2220

I like it! And is better than nothing. Here is how it looks:


https://github.com/user-attachments/assets/818f34e1-bcd5-4786-b9cf-fbfde69bf31b

wjt:
Alright! I also thought it was fun.

(I don't think it's the final story so I won't mark #2220 as closed by this.)

wjt:
I missed that this makes the controller vibrate just as much as it does when you are hit by an ink blob. But it's funny so I'll leave it.

PR #2222, opened 2026-05-19
tutorial ruined village: Replace walkable water with normal water

wjt:
There used to be a bridge up to the one island that remains in this
level. I removed the bridge but left the alternative walkable water
tiles in place.


wjt:
image

PR #2223, opened 2026-05-19
tutorial ink combat: Add river

wjt:
tutorial ink combat: Add river

During playtesting, Sarah remarked that her instinct (if she did not
already know how combat works) would be to walk up to the InkDrinkers
and try using the Repel action against them directly.

Add a body of water dividing StoryWeaver from the InkDrinkers.

Move the top wall back slightly. I found a position I could stand in and
have one of the ink blobs ricochet into the pot without me repelling it.

Add dirt patches on the floor indicating roughly the distance the ink
blobs can travel before popping (if not repelled first), and where you
might want to walk and repel.


manuq:
Nice! I like the zig-zag patterns.

image

manuq (approved):

wjt:
The zig-zag is because the inkdrinkers are also staggered, but I can't remember why I did that!

PR #2226, opened 2026-05-19
Make it a bit easier to interact with buttons collector

### LINKED ISSUE #1858: Too easy to not talk to button collector
In the void quest, we have yellow buttons which you collect by moving through them, and purple ones you can hook. Collecting them has no in-game function.

In [the scene that introduces buttons](https://play.threadbare.game/branches/endlessm/main/#quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_pins), we have an NPC with dialogue that just directly tells you this:

Image

However if you have moved due right by hooking all the pins, you are too far down for the interact prompt to trigger:

Image

There is an invisible area around him which temporarily switches the player from hooking mode to cozy mode so that the player can interact (but not hook), but this doesn't quite reach down far enough to trigger:

Image

and if we make it go too much further down, there will be a strange region where for no apparent reason you cannot hook, even if you might want to hook the pin below.

manuq:
buttons_collector.tscn: Make the collision shapes bigger: the one for the interact area, and the CharacterBody2D collision capsule. So the player will always get the "Talk" hint and the arrow indicator, passing in front or behind the NPC.

grappling_hook_pins.tscn: Expand the corridor at the end, and center the buttons collector in it. Also move the player a bit, so when playing with the keyboard the aim ray points exactly to the pins anchor points. In the best case, the player will end up colliding with the NPC.

Resolve https://github.com/endlessm/threadbare/issues/1858


manuq:
Playable at https://play.threadbare.game/branches/endlessm/move-buttons-collector/#quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_pins

This is how it looks:

Captura desde 2026-05-19 15-19-03

PR #2227, opened 2026-05-19
Replace custom camera behavior with Phantom Camera for hooking

manuq:
## Grappling hook scenes: use Phantom Camera

In scenes:
- tutorial_grappling.tscn
- grappling_hook_needles.tscn
- grappling_hook_powerup.tscn
- void_grappling.tscn
- void_grappling_round_2.tscn

Add the PhantomCameraHost node to the Camera2D in order to control it.

Add a PhantomCamera2D node to the scene as well. Set priority to 10, so it will be active when the one at the tip
of the string isn't. Set follow target to the Player node. Also enable follow
damping which is similar to the previous position smoothing of Camera2D.

Use the Void TileMapLayer in each scene (except one) as the limit target of the
PhantomCamera2D node. Paint the tilemap to match the existing limit. In the
cases in which the void tiles were used to block the player going out of the
camera/world bounderies, use water instead.

Zoom out each PhantomCamera2D when grabbing the longer thread powerup.

## PlayerHook: Add PhantomCamera2D at the tip of the string

Set follow mode to Framed and set dead zone to 0.5 width and height. Also enable
follow dumping. Otherwise the transitions between cameras is smooth, but the
transition when the tip of the string changes to another needle or other
hookable object isn't.

And change its priority to:
- Zero when the camera is not supposed to follow the tip
- 20 when the camera should follow the tip

Add a function to use the same limit target for the camera that's at
the tip of the string.

## Remove FrameCameraBehavior node

Most scenes that had the grappling hook ability active, had this node as child
of the camera for controlling it. With the target set to the HookEnding marker
of PlayerHook in player.

This will be in favour of using Phanton Camera in next commit.


manuq:
The main issue I'm finding is:

I added a PhantomCamera2D to each scene for following the player, instead of one to the player scene, because of the camera limits. But when switching to the PhantomCamera2D at the tip of the string (in PlayerHook), it doesn't have those limits so the undecorated world appears. This is noticeable, for example, in the tutorial grapping:

https://github.com/user-attachments/assets/96006e45-4bf7-4c81-a8ee-b666f1bb6250

Also, something I need to fix: when grabbing the longer thread we tween the camera zoom to zoom out, and that has to be done with Phantom Camera now.

wjt:
We could simply declare that the level designer needs to dress enough of the level around each hook point :)

manuq:
> We could simply declare that the level designer needs to dress enough of the level around each hook point :)

I made it work by using a tilemap layer as the limit target. I think I like that the world bounderies is now done by painting tiles rather than by numbers (which I always tried to make them match multiplying by 64 in the input field).

Then in the player hook script, I added this method that is called before the camera at the tip of the string is activated (by changing its priority). To assign the limit target to this other camera:

```
## Return the absolute path to the limit target of the current active camera, if it has one.
func _get_phantom_camera_limit_target() -> NodePath:
	var phantom_camera_hosts := PhantomCameraManager.phantom_camera_hosts
	if not phantom_camera_hosts:
		return ""
	var active_pcam := phantom_camera_hosts[0].get_active_pcam() as PhantomCamera2D
	if not active_pcam:
		return ""
	if not active_pcam.limit_target:
		return ""
	if active_pcam.limit_target.is_absolute():
		return active_pcam.limit_target
	var absolute_target_path := NodePath(
		String(active_pcam.get_path()) + "/" + String(active_pcam.limit_target)
	)
	return absolute_target_path
```

I repainted the Void tilemap to be the bounderies. Then I realized that in some scenes, the void tilemap was also used to limit the player to go back (by painting it outside boundaries). So I used water instead. Check the vertical water line at the west side:

Captura desde 2026-05-20 14-54-19

But! Now I realize that the enemies are painting void tiles, so they are enlarging the boundaries. Ouch. So I'm redoing this, by using the water tilemap as the limit target.

manuq:
I fixed the above. I also fixed the zoom out when grabbing the longer thread in the 2 scenes that have it. I did it by zooming out each PhantomCamera2D. This goes a bit against what's indended in the addon, which I guess will have a second already zoomed-out PhantomCamera2D, and alternate between them by priority. But in that case we'll need 2 zoomed out versions, one for the camera following the player and one for the one at the end of the tip. So I'm not sure... should I do that instead?

scenes/game_elements/characters/player/components/player_hook.gd
```diff
@@ -141,6 +144,24 @@ func _new_hook_string() -> Line2D:
...
+	if not active_pcam:
+		return ""
+	if not active_pcam.limit_target:
+		return ""
+	if active_pcam.limit_target.is_absolute():
+		return active_pcam.limit_target
+	var absolute_target_path := NodePath(
+		String(active_pcam.get_path()) + "/" + String(active_pcam.limit_target)
+	)
+	return absolute_target_path
```
wjt:
```suggestion
	var absolute_target_path := active_pcam.get_node(active_pcam.limit_target).get_path()
	return absolute_target_path
```

I think this has the same meaning but I think it'll give slightly simpler paths, e.g. `/root/MyScene/A/B/C/PCam/../../../../TileMapLayers/Void` would be simplified to `/root/MyScene/TileMapLayers/Void`, and this would be done by Godot's standard node path resolution logic.

scenes/game_elements/characters/player/components/player_hook.gd
wjt, line 280: I wonder whether it would be better to reset the priority to recenter the player while they are pulling.

I tried this quick patch:


```suggestion
	character.set_collision_mask_value(Enums.CollisionLayers.NON_WALKABLE_FLOOR, false)
	phantom_camera_2d.priority = 0
```

Before (notice how the camera tracks away from the action while the player is pulling along, then slowly pans back to recenter storyweaver):

https://github.com/user-attachments/assets/2fbc543b-5979-4c96-be76-9bf483091113

After:

https://github.com/user-attachments/assets/9ab15709-a467-4a98-aafc-9de23c6cac03

wjt (approved):
Nice! I like the approach of defining the camera perimeter with a layer. In our case, void or water are normally good candidates for padding out the edges of levels.

scenes/game_elements/characters/player/components/player_hook.gd
wjt: NB. completely untested!

wjt:
> I fixed the above. I also fixed the zoom out when grabbing the longer thread in the 2 scenes that have it. I did it by zooming out each PhantomCamera2D. This goes a bit against what's indended in the addon, which I guess will have a second already zoomed-out PhantomCamera2D, and alternate between them by priority. But in that case we'll need 2 zoomed out versions, one for the camera following the player and one for the one at the end of the tip. So I'm not sure... should I do that instead?

I think what you have here is fine. I think we still need to decide how we indicate "longer thread" - because although the powerup persists past this level, the zoom does not! - so I don't think we should do anything more complicated than just tweaking the zoom for now.

scenes/game_elements/characters/player/components/player_hook.gd
manuq: Oh you showed that recentering the player while pulling is much better!

scenes/game_elements/characters/player/components/player_hook.gd
manuq: You are right! I tested it and gave:
`/root/GrapplingHookPowerup/TileMapLayers/Void`

Previously this was:
`/root/GrapplingHookPowerup/OnTheGround/Player/PhantomCamera2D/../../../TileMapLayers/Void`

PR #2228, opened 2026-05-20
Void runner: Replace Tiny Swords townie with a randomized one

wjt:
Move the custom is_enemy_defeated() method to a new script on the root
node of the scene, which is added to the context of dialogue
automatically. Remove the custom scene & script.


wjt:
| Before | After |
| ------ | ----- |
| image | image |

manuq:
Yes!

manuq (approved):

PR #2232, opened 2026-05-21
stealth_level.tscn: Add camera dolly

manuq:
Use follow mode "Path" of the PhantomCamera, and draw a path along the zig-zag level. The player is still the target, but now the action is kept framed even if the player is next to a cliff.

This was an idea from Will Thompson.


manuq:
I think this is a great idea @wjt . Here is the path I've drawn:

Captura desde 2026-05-21 12-05-54

The level works better now, because when the player gets closer to a cliff the guards are still framed. Except for the area with the crops (circled in red above) which has this "camera jump" issue:

https://github.com/user-attachments/assets/2d592201-3489-44b4-97e6-c27c085608c0

For which we could try making the path simpler (go in diagonal in that place). Or switch to another "fixed" camera in that area.

manuq:
> For which we could try making the path simpler (go in diagonal in that place). Or switch to another "fixed" camera in that area.

Making the path simpler didn't work. But adding a fixed camera to the dead-end and enabling it when the player enters this area worked like a charm! Ready for review.

wjt:
I like how it feels to play this! I took a different route through the last section as a result and discovered that the tower has no physics and does not occlude sight!

image

scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn
```diff
@@ -146,8 +153,21 @@ script = ExtResource("3_idy4y")
 type = 2
 metadata/_custom_type_script = "uid://bgmwplmj3bfls"
 
+[sub_resource type="Curve2D" id="Curve2D_86cpf"]
+_data = {
+"points": PackedVector2Array(0, 0, 0, 0, -96, 224, 0, 0, 0, 0, 672, 232, 0, 0, 0, 0, 1952, 288, 0, 0, 0, 0, 2528, 424, 0, 0, 0, 0, 2528, 920, 0, 0, 0, 0, 2784, 1184, 0, 0, 0, 0, 3360, 1184, 0, 0, 0, 0, 4256, 1120, 0, 0, 0, 0, 5288, 1120, 0, 0, 0, 0, 5600, 1056, 0, 0, 0, 0, 6496, 896, 0, 0, 0, 0, 8448, 896)
```
wjt: I would cut this path a little shorter:


```suggestion
"points": PackedVector2Array(0, 0, 0, 0, -96, 224, 0, 0, 0, 0, 672, 232, 0, 0, 0, 0, 1952, 288, 0, 0, 0, 0, 2528, 424, 0, 0, 0, 0, 2528, 920, 0, 0, 0, 0, 2784, 1184, 0, 0, 0, 0, 3360, 1184, 0, 0, 0, 0, 4256, 1120, 0, 0, 0, 0, 5288, 1120, 0, 0, 0, 0, 5600, 1056, 0, 0, 0, 0, 6496, 896, 0, 0, 0, 0, 7096, 896)
```

but perhaps you made it extend far to the right intentionally!

wjt (approved):

manuq:
@wjt the path being longer or shorter wasn't changing much the gameplay. Let's merge it!

manuq:
> I like how it feels to play this! I took a different route through the last section as a result and discovered that the tower has no physics and does not occlude sight!

Good catch! Is funny how many times we played this level and never noticed it.

PR #2234, opened 2026-05-21
Add stub Mythical Meadows LoreQuest

wjt:
Add stub Mythical Meadows LoreQuest


manuq (approved):
Looks good! At this point I'm not sure if we should continue using numbers for the lore quests, but we can rename the folders later!

scenes/quests/lore_quests/quest_004/quest.tres
```diff
@@ -0,0 +1,11 @@
...
+
+[ext_resource type="Script" uid="uid://dts1hwdy3phin" path="res://scenes/menus/storybook/components/quest.gd" id="1_rp1q0"]
+
+[resource]
+script = ExtResource("1_rp1q0")
+title = "Mythical Meadows"
+first_scene = "uid://bsptn2yxyv22"
+threads_to_collect = 1
+is_lore_quest = true
+metadata/_custom_type_script = "uid://dts1hwdy3phin"
```
manuq, line 11: Status default to "Work in progress", so this is good.

manuq (approved):

wjt:
Yes, I agree about the numbers. The reason I want to upstream this trivial scene is to "reserve" the number 4 :)

PR #2235, opened 2026-05-21
Refactor game state

### LINKED ISSUE #2016: Refactor game state
The game state script became big-enough that the gdlint started complaining about the amount of public methods. But more importantly, it is being used for multiple states, namely:
- State that should be persisted to continue the game in a new play session, after exiting the current session. For example: the completed quests, the current scene.
- State that is transient the current play session.
- State that is transient to the current scene. For example: if the intro dialogue has been shown. If the lights are on or off.

This would welcome a refactor.

wjt:
GameState mixes the actual load/save logic with the state itself, and
also is based on a mixture of directly accessing the contents of a
ConfigFile object and fields of the GameState node.

State API
---------

Factor all the state itself out to a (shallow) tree of Resource subclasses:

- SavedGame
  - GlobalState
    - PlayerState
  - QuestState (nullable)
    - Quest-specific PlayerState instance
  - PerSceneState

Expose the GlobalState, QuestState, and PerSceneState resources on
GameState with short names to save on typing when referring to them. Add
a special GameState.player property which is a proxy for the
quest-specific PlayerState when on a quest and the global PlayerState
otherwise. Update all code in the game to refer to these.

Remove `@tool` from GameState, and instead short-circuit in other
scripts which need to be `@tool` but refer to GameState.

Storage
-------

Replace the ConfigFile-based load and save logic with ResourceLoader and
ResourceSaver. This avoids writing bespoke serialisation code & makes it
relatively easy to add new saved fields – drop an `@export` into the
appropriate state class – at the cost of coupling the savegame format to
these Resource subclasses existing at these exact paths. This is the
approach recommended, among others, by GDQuest.

https://www.gdquest.com/library/save_game_godot4/

The state conceptually needs to refer to Quest resources. Storing these
as resources would mean that the `saved_game.tres` would have an
`ext_resource` header for each quest.tres being referred to, and moving
a quest elsewhere in the source tree (or deleting it) will cause the
whole `saved_game.tres` to fail to load. Mark the Quest and Array[Quest]
as not to be stored, and define stored proxy properties that convert to
and from resource_path strings. This matches how these were previously
stored.

Similarly, the inventory is in terms of InventoryItem resources. Unlike
Quests we don't store these in the game repo but define them inline
wherever they are used. I think this is a bad design - it would be
better to use just the ItemType enum throughout given that we hardcode
exactly 3 items and 2 textures for each - but to avoid this PR getting
even further out of hand, translate these to and from the ItemType enum
member name when storing, as before.

The result of jumping through these hoops is that the `saved_game.tres`
only refers to the scripts that define aspects of the game state. I
think that's acceptable coupling. I am disappointed to have to write
quite a bit of what amounts to custom serialisation code - I was hoping
to avoid that by using ResourceLoader/ResourceSaver - but it's still a
net improvement I think.

I did not make any attempt to migrate the old save game format to the
new one. This game is small, the state has little impact, and this will
also mean more people see the tutorial. :)

Saving
------

Only save the state in a few specific places, rather than whenever parts
of the state change (if we remembered to add a _save() call):

- When a checkpoint is activated (so that if you quit the game with
  Alt+F4 after activating a checkpoint, it is kept);
- When reloading the current scene, typically due to being defeated, to
  record the loss of life;
- When switching to a new scene;
- When you use the debug menu to tweak the completed quests.

Resolves https://github.com/endlessm/threadbare/issues/2016

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -93,7 +93,7 @@ func _on_interacted(player: Player, _from_right: bool) -> void:
 	animation_player.play("collected")
 	await animation_player.animation_finished
 
-	GameState.add_collected_item(item)
+	GameState.global.add_collected_item(item)
 
 	if collected_dialogue:
 		DialogueManager.show_dialogue_balloon(collected_dialogue, dialogue_title, [self, player])
@@ -103,7 +103,10 @@ func _on_interacted(player: Player, _from_right: bool) -> void:
 	queue_free()
 
 	if next_scene:
-		GameState.set_challenge_start_scene(next_scene)
+		if GameState.quest:
+			GameState.quest.challenge_start_scene = next_scene
+		else:
+			push_warning("Collectible collected while not on a quest")
 		switch()
 
 
--- a/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
+++ b/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
@@ -27,22 +27,16 @@ func _ready() -> void:
 	talk_behavior.title = "have_threads" if _have_threads else "no_threads"
 	interact_area.interaction_ended.connect(self._on_interaction_ended)
 
-	if GameState.incorporating_threads:
+	if GameState.quest and GameState.quest.incorporating_threads:
 		if Transitions.is_running():
 			await Transitions.finished
 
-		var elder: Elder
-		if GameState.current_quest:
-			elder = _find_elder(GameState.current_quest)
-		else:
-			push_warning("incorporating_threads was set, but current_quest was null")
-			if elders:
-				# Pick any elder.
-				elder = elders.pick_random()
+		var elder: Elder = _find_elder(GameState.quest.quest)
 		if elder:
 			await elder.congratulate_player()
+		else:
+			push_warning("Could not find elder for %s" % [GameState.quest.quest.resource_path])
 
-		GameState.set_incorporating_threads(false)
 		GameState.mark_quest_completed()
 
 
@@ -59,7 +53,7 @@ func _on_interaction_ended() -> void:
 		if not ProjectSettings.get_setting(ThreadbareProjectSettings.SKIP_SOKOBANS):
 			# Hide interact label during scene transition
 			interact_area.disabled = true
-			GameState.set_incorporating_threads(true)
+			GameState.quest.incorporating_threads = true
 			SceneSwitcher.change_to_file_with_transition(SOKOBANS.pick_random())
 		else:
 			GameState.mark_quest_completed()
@@ -68,12 +62,12 @@ func _on_interaction_ended() -> void:
 func on_offering_succeeded() -> void:
 	loom_offering_animation_player.play(&"loom_offering")
 	await loom_offering_animation_player.animation_finished
-	GameState.clear_inventory()
+	GameState.global.clear_inventory()
 
 
 func is_item_offering_possible() -> bool:
 	return (
-		GameState.current_quest
-		and GameState.current_quest.threads_to_collect > 0
-		and GameState.items_collected().size() >= GameState.current_quest.threads_to_collect
+		GameState.quest
+		and GameState.quest.quest.threads_to_collect > 0
+		and GameState.global.inventory.size() >= GameState.quest.quest.threads_to_collect
 	)
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -1,50 +1,13 @@
-# gdlint: disable=max-public-methods
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
-@tool
 extends Node
 
-## Emitted when a new item is collected, even if it wasn't added to the
-## inventory due to it being already there.
-signal item_collected(item: InventoryItem)
+## Emitted when [member player] changes to a new [PlayerState] object, due to a
+## quest starting or ending. This is used by the [Player] scene to rebind to
+## signals like [signal PlayerState.abilities_changed].
+signal player_changed(old: PlayerState, new: PlayerState)
 
-## Emitted when a item is consumed, causing it to be removed from the
-## [member inventory].
-signal item_consumed(item: InventoryItem)
-
-## Emitted whenever the items in the inventory change, either by collecting
-## or consuming an item.
-signal collected_items_changed(updated_items: Array[InventoryItem])
-
-## Emitted when the player's lives change.
-signal lives_changed(new_lives: int)
-
-## Emitted when it becomes too dark that artificial lights can turn on, or
-## when darkness goes away so artificial lights should turn off.
-signal lights_changed(lights_on: bool, immediate: bool)
-
-## Emitted when a quest is added or removed from [member completed_quests].
-signal completed_quests_changed
-
-## Emitted when lore or StoryQuest player abilities change.
-signal abilities_changed
-
-const GAME_STATE_PATH := "user://game_state.cfg"
-const INVENTORY_SECTION := "inventory"
-const INVENTORY_ITEMS_KEY := "items_collected"
-const QUEST_SECTION := "quest"
-const QUEST_PATH_KEY := "resource_path"
-const QUEST_CHALLENGE_START_KEY := "challenge_start_scene"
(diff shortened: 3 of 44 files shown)


scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -27,7 +27,7 @@ func _ready() -> void:
 	talk_behavior.title = "have_threads" if _have_threads else "no_threads"
 	interact_area.interaction_ended.connect(self._on_interaction_ended)
 
-	if GameState.incorporating_threads:
+	if GameState.state.quest_state.incorporating_threads:
```
wjt: It's clear that this is less clear than the previous one because you have to use `.state.[kind_state]` everywhere. I could make it simpler by proxying the `quest_state` and `player_state` properties out one layer so you could write: `GameState.quest_state.incorporating_threads`

or even:

`GameState.quest.incorporating_threads`

?

scenes/globals/game_state/game_state.gd
```diff
@@ -56,74 +37,31 @@ const DEBUG_PLAYER_ABILITIES := [
...
-			property.hint_string = ",".join(Enums.PlayerAbilities.keys())
+func _ready() -> void:
+	if Engine.is_editor_hint():
+		return
 
+	if ResourceLoader.exists(SAVE_PATH):
+		state = ResourceLoader.load(SAVE_PATH, "", ResourceLoader.CACHE_MODE_IGNORE)
+
+	if state:
+		state = state.duplicate_deep(Resource.DEEP_DUPLICATE_INTERNAL)
```
wjt: TODO: per #2250 this should be deferred, we should only load the saved state once we know for sure we want to do that rather than use transient state.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
manuq: Yes I like it! `GameState.quest` reads as "the state of the quest", :+1: .

wjt: The weird thing about this is that `QuestState` has a `quest` property so one ends up writing `GameState.quest.quest` to access it. Tolerable I think.

scenes/globals/game_state/game_state.gd
wjt: This is surprisingly hard to do and this PR has already got quite big so I'll not work on this further for now.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -103,7 +103,8 @@ func _on_interacted(player: Player, _from_right: bool) -> void:
 	queue_free()
 
 	if next_scene:
-		GameState.set_challenge_start_scene(next_scene)
+		assert(GameState.quest)
```
manuq: This could be written as a conditional, like in other places:

```
if next_scene and GameState.quest:
    ...
```

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -27,22 +27,16 @@ func _ready() -> void:
...
 
-		var elder: Elder
-		if GameState.current_quest:
-			elder = _find_elder(GameState.current_quest)
-		else:
-			push_warning("incorporating_threads was set, but current_quest was null")
-			if elders:
-				# Pick any elder.
-				elder = elders.pick_random()
+		var elder: Elder = _find_elder(GameState.quest.quest)
```
manuq, line 34: This is the `quest.quest` you mentioned. It's fine!

scenes/globals/game_state/game_state.gd
```diff
@@ -1,50 +1,13 @@
...
-const CURRENTSCENE_KEY := "current_scene"
-const SPAWNPOINT_KEY := "current_spawn_point"
-const GAME_PLAYER_ABILITIES_KEY := "game_player_abilities"
-const LIVES_KEY := "current_lives"
-const MAX_LIVES := 2 ** 53
-const DEBUG_LIVES := false
+## Emitted when [member player] changes to a new [PlayerState] object, due to a
+## quest starting or ending. This is used by the [Player] scene to rebind to
+## signals like [signal PlayerState.abilities_changed].
+signal player_changed(old: PlayerState, new: PlayerState)
```
manuq, line 8: 👍

scenes/globals/game_state/game_state.gd
```diff
@@ -56,74 +19,58 @@ const DEBUG_PLAYER_ABILITIES := [
...
 ## with a direct URL hash to a scene in the web build. In the latter cases, this variable is false.
 var persist_progress: bool
 
-var _state := ConfigFile.new()
+## Game-wide state.
+var global: GlobalState:
+	get():
+		return _saved_game.global if _saved_game else null
+	set(new_value):
+		push_error("Do not set GameState.global")
```
manuq, line 34: Interesting way of using the setter to forbid setting the property!

scenes/globals/game_state/game_state.gd
```diff
@@ -137,45 +84,28 @@ func _ready() -> void:
...
-		push_error("Failed to load %s: %s" % [GAME_STATE_PATH, err])
-
-	if DEBUG_LIVES:
-		prints("[LIVES DEBUG] GameState initialized with", current_lives, "lives")
 
+## Sets [member quest], sets up a new [PlayerState] if necessary, and clears the
+## inventory. Note that this does not actually switch to the first scene of [param
+## new_quest].
+func start_quest(new_quest: Quest) -> void:
+	# TODO: this suggests that the inventory should be part of QuestState
```
manuq, line 95: Yes, although this may change in the near future...

scenes/globals/game_state/game_state.gd
```diff
@@ -190,320 +120,62 @@ func guess_quest(scene_path_or_uid: String) -> void:
...
+## Save the game state (if [persist_progress] is [code]true[/code])
+func save() -> void:
 	if not persist_progress:
 		return
-	_state.set_value(GLOBAL_SECTION, COMPLETED_QUESTS_KEY, completed_quests)
-	var err := _state.save(GAME_STATE_PATH)
-	if err != OK:
-		push_error("Failed to save settings to %s: %s" % [GAME_STATE_PATH, err])
+
+	print("Saving game")
```
manuq: I understand this is not a leftover, but an actual print, now that the game is saved less frequently. So 👍

scenes/globals/game_state/per_scene_state.gd
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name PerSceneState
```
manuq, line 3: I see that `PerSceneState` is not a `@tool`. In the pull request description you mention:

> Remove @tool from GameState, and instead short-circuit in other
scripts which need to be @tool but refer to GameState.

So yes, this one doesn't need to be a tool script.

manuq (approved):
This is a huge refactor and you make it look so easy (I know it wasn't). Doing the storage with just resources and ResourceLoader/ResourceSaver rather than ConfigFile for INI files is great.

I like how the player state can be one or another, the global or the quest-specific one.

scenes/globals/game_state/game_state.gd
wjt: Indeed, that's why I didn't think too hard about it :)

scenes/globals/game_state/game_state.gd
wjt: It feels like a bit of a kludge but since there is no such thing at the GDScript level as a readonly property, this is the best I could think of!

scenes/globals/game_state/game_state.gd
wjt: Actually I left it in by mistake. I'm on the fence about it!

scenes/game_elements/props/collectible_item/components/collectible_item.gd
wjt: I asserted here because I think this situation should not occur - in the current model these items should only be found during quests - but maybe it's a bit extreme. I replaced this with:

```diff
        if next_scene:
-               assert(GameState.quest)
-               GameState.quest.challenge_start_scene = next_scene
+               if GameState.quest:
+                       GameState.quest.challenge_start_scene = next_scene
+               else:
+                       push_warning("Collectible collected while not on a quest")
                switch()
```

PR #2238, opened 2026-05-22
Refactor Teleporter to a reusable SceneLink

### LINKED ISSUE #2159: Factor out Teleporter, Cinematic, CollectibleThread scene switching logic
We currently have at least 3 reusable entities in the game which take the player to another scene:

| Node | Behaviour | spawn_point_path | Transition |
| ---- | --------- | ---------------- | ---------- |
| Cinematic | Plays dialogue, then (optionally) switches to another scene | Supported but without a drop-down menu in the inspector | Can't be configured |
| Teleporter | When the player enters the Area2D, teleports to another scene, or within the same scene | Supported, with drop-down menu in inspector | Configurable |
| CollectibleItem | Optionally plays dialogue, then optionally switches to another scene | Not supported | Not configurable |

(We also have some more bespoke components that switch scene, such as the eternal loom, but I'm ignoring that because you can never refactor away all the code in a project!)

In the same vein as https://github.com/endlessm/threadbare/issues/1346, it is annoying that they have different features. For example, right now I want to adjust the tutorial quest so that collecting the final thread takes you to the starting island of Fray's End rather than to the Eternal Loom, but CollectibleItem doesn't let you do that.

I think we should have some generic node that has all the configuration of Teleporter. Then, all three of the nodes above could use it.

wjt:
Previously, Teleporter allowed you to specify the target spawn point and
configure the scene transitions, but CollectibleItem and Cinematic did
not. Annoying – I want the player to enter Fray's End from the south at
the end of the tutorial, but the last action is to collect a thread so I
couldn't.

Reintroduce a teleporter scene. (Before commit f7c68090a1d7d786ffffa1a899c3e3431f68472f, we had a mixture of bare teleporter.gd nodes, and instances of teleporter.tscn. In that
commit I removed the scene, and I now regret it.) Rewrite all scenes to use teleporter.tscn.

Then rename Teleporter to SceneLink. Give it a fresh UID. Change its base
class to Node2D rather than Area2D, and move the Area2D-specific logic
to a helper node inside the Teleporter scene. Use my favourite niche
feature of the Godot type system, as described in a comment.
Use CONNECT_DEFERRED to avoid needing to put a call_deferred inside the
teleport logic.

Then, make Cinematic and CollectibleItem extend SceneLink; delete their
bespoke scene-switching logic.

Add validation of spawn_point_path.

It's a bit weird for this to be a Node2D but it's necessary because
CollectibleItem is a Node2D. (Cinematic has always been a Node2D rather
than a Node – but this error has proven useful!)

Change the default transition effects to FADE. This matches what
Cinematic did by default before. It also doesn't look actively wrong if
misconfigured, unlike the directional effects. Nonetheless, I also
attempted to fix all teleporters to use the correct directions.

I updated the splash screen to use this node directly. It's not really
much help but it doesn't hurt either.

I did not update the Sokoban rule engine or some other places that
switch scenes in code to use this node because it doesn't really help
much there.

Resolves https://github.com/endlessm/threadbare/issues/2159

### Side-quest: flip directional screen-wipe transitions to black

Transitions are implemented as follows. A fully-black ColorRect covers
the whole screen. Its alpha channel is varied by a shader, which samples
a monochrome mask texture and compares its red channel to a cutoff
threshold to determine whether the alpha channel should be 0 or 1 (or a
smoothed value in between).

The cutoff is tweened from 1 to 0 when fading to black, and from 0 to 1
when fading from black.

For the FADE (whole screen fades at once) and RADIAL effects, this works
as expected. For RADIAL, the fade-out ends at a point in the middle of
the screen, and the fade-in begins at that point. Fine.

When a human reads LEFT_TO_RIGHT_WIPE they have a different expectation
in the two directions. When fading out, we expect the blackness to start
at the left and wipe across the screen to the right. When fading in, we
expect the scene to appear at the left of the screen and wipe across the
screen. If you think about it in terms of the mask texture, these are
actually opposites of one another! The LEFT_TO_RIGHT_WIPE texture does
what you would expect for the fade-in, but the opposite for the fade
out.

Invert the sense of directional transitions when fading out. This means
that using LEFT_TO_RIGHT_WIPE for the fade-out has the expected effect
of the black mask wiping across the screen from left to right.

(Confusingly Teleporter refers to the fade-out as "enter" and the
fade-in as "exit", but when thinking in terms of scenes you might
imagine the fade-out is the "exit" and the fade-in is the "enter".)


scenes/globals/scene_switcher/scene_link.gd
```diff
@@ -1,8 +1,12 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 @tool
-class_name Teleporter
-extends Area2D
+class_name SceneLink
```
wjt, line 4: I'm not really sure about this name but it is short and seems about right.

wjt (commented):
I'm also not all that enthusiastic about using inheritance here, but there are 5 properties to expose, and it hardly changes the API/Inspector properties of Cinematic, Teleporter or CollectibleItem at all - it just adds a couple more to Cinematic and CollectibleItem and groups them together.

scenes/globals/scene_switcher/scene_link.gd
manuq: Yes, is good.

wjt:
Pushed one more change to use PROPERTY_HINT_GROUP_ENABLE which I think looks good:

image

image

Before:

image

wjt:
Rebased to fix a minor merge conflict!

scenes/game_elements/characters/player/components/player.gd
```diff
@@ -207,7 +207,7 @@ func defeat(falling: bool = false) -> void:
 	# Check if player has lives remaining
 	if GameState.current_lives > 0:
 		# Still have lives - reload current scene/checkpoint
-		SceneSwitcher.reload_with_transition(Transition.Effect.FADE, Transition.Effect.FADE)
+		SceneSwitcher.reload_with_transition()
```
manuq, line 210: 👍

manuq (approved):
Fantastic refactor. This shows how skilled you are becoming in Godot!

PR #2239, opened 2026-05-22
Fray's End: Small fix for positions and scales

manuq:
Remove scale from foam tilemap layer.

Remove 8 pixels vertical offset from the Node2D that contains all tilemap layers. Adjust OnTheGround nodes accordingly.

### DIFF
--- a/scenes/world_map/frays_end.tscn
+++ b/scenes/world_map/frays_end.tscn
@@ -75,7 +75,6 @@ metadata/_edit_lock_ = true
 stream = ExtResource("2_epsnn")
 
 [node name="TileMapLayers" type="Node2D" parent="." unique_id=1722660259]
-position = Vector2(0, 8)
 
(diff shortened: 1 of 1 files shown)


manuq:
I checked by taking screenshots and comparing with GIMP that the before/after positions are the same.

The reason I had to do it is because I was trying to set the camera limits to the Water tilemap layer (with PhantomCamera2D). And found that it had a 8 pixels offset!

scenes/world_map/frays_end.tscn
```diff
@@ -87,7 +86,6 @@ canvas_item = NodePath("..")
 metadata/_custom_type_script = "uid://ctko5bdas16ah"
 
 [node name="Foam" type="TileMapLayer" parent="TileMapLayers" unique_id=632936150]
-scale = Vector2(1.00113, 1)
```
wjt, line 90: Weird!

wjt (approved):
Good catch!

PR #2240, opened 2026-05-22
Center camera on player when dialogue balloon appears

manuq:


manuq:
This is hardcoded for now, just for testing:

https://github.com/user-attachments/assets/909bc234-93e0-4006-bf9f-c994ba335ae6

What about a bit of zoom?

https://github.com/user-attachments/assets/b7522e50-59d5-4b82-93be-ecfdf8bc39fb

Again this brings the issue of having to pass the world limits between cameras. And I think that if the limit is at the west (case of Lore elder) then the balloon should appear at the top-right corner, and the camera should shift to the other side.

manuq:
This might not be needed after all, https://github.com/endlessm/threadbare/pull/2275 supersedes it.

PR #2245, opened 2026-05-26
Cinematic: Always emit cinematic_finished

### LINKED ISSUE #2244: Cinematic.cinematic_finished not emitted if GameState.intro_dialogue_shown is true
### Threadbare version

0.2.0.5

### Affected scene

e.g. res://scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_2.tscn

### Description

1. Start/Continue the game from the title screen. This bug is in the savegame logic so cannot be reproduced by deep-linking to a specific scene or playing a scene in the editor.
2. Go through the game to a scene which has a Cinematic with `autostart` enabled and some other action connected to `cinematic_finished`. `res://scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_2.tscn` is an example:

```tscn
[connection signal="cinematic_finished" from="Cinematic" to="FillGameLogic" method="start"]
```

3. Finish the introductory dialogue and allow the fill game to begin
4. Press Pause
5. Press "Exit to Title"
6. At the title screen, press Continue

Expected results:

- Either the introductory dialogue plays again, after which the fill game begins;
- Or the fill game begins immediately.

Actual results:

- The introductory dialogue doesn't play;
- but the fill game never begins.

This is because, in cinematic.gd:

```gdscript
func start() -> void:
	if not GameState.intro_dialogue_shown:
		DialogueManager.show_dialogue_balloon(dialogue, "", [self])
		cinematic_finished.emit()
		GameState.intro_dialogue_shown = true
```

The `cinematic_finished` signal is only emitted if `intro_dialogue_shown` is `false`. I think this is wrong: the signal should always be emitted.

It is possible to work around it. For example, I did so it `tutorial_sequence_puzzle.gd`:

```
func _ready() -> void:
	if not GameState.intro_dialogue_shown:
		await cinematic.cinematic_finished

	tutorial_npc.walk_path()
```

but I think this is a nasty footgun and it would be better to emit the signal even if the dialogue is skipped.

(Thanks to @Snowsd for reporting the problem that I tracked down to, among other things, this bug.)

### Checklist

wjt:
Previously, `Cinematic.cinematic_finished` was only emitted if `intro_dialogue_shown` was false.

This makes it very easy to break game scenes: if you connect `Cinematic.cinematic_finished` to `FillGameLogic.start`, the scene works when the player first reaches it, but if they exit to title then continue the fill game never starts and so can't be finished.

Resolves https://github.com/endlessm/threadbare/issues/2244

### DIFF
--- a/scenes/ui_elements/cinematic/cinematic.gd
+++ b/scenes/ui_elements/cinematic/cinematic.gd
@@ -38,8 +38,9 @@ func start() -> void:
 	if not GameState.intro_dialogue_shown:
 		DialogueManager.show_dialogue_balloon(dialogue, "", [self])
 		await DialogueManager.dialogue_ended
-		cinematic_finished.emit()
 		GameState.intro_dialogue_shown = true
+
+	cinematic_finished.emit()
 
 	if next_scene:
 		(
--- a/scenes/quests/lore_quests/quest_000/3_sequence_puzzle/components/tutorial_sequence_puzzle.gd
+++ b/scenes/quests/lore_quests/quest_000/3_sequence_puzzle/components/tutorial_sequence_puzzle.gd
@@ -8,7 +8,6 @@ extends Node2D
 
 
 func _ready() -> void:
-	if not GameState.intro_dialogue_shown:
-		await cinematic.cinematic_finished
+	await cinematic.cinematic_finished
 
 	tutorial_npc.walk_path()


scenes/ui_elements/cinematic/cinematic.gd
```diff
@@ -38,9 +38,10 @@ func start() -> void:
 	if not GameState.intro_dialogue_shown:
 		DialogueManager.show_dialogue_balloon(dialogue, "", [self])
 		await DialogueManager.dialogue_ended
-		cinematic_finished.emit()
 		GameState.intro_dialogue_shown = true
 
+	cinematic_finished.emit()
```
manuq, line 43: 👍

manuq (approved):
Excellent! The issue was very well described in the ticket, thanks.

PR #2247, opened 2026-05-26
ci: Use ubuntu-slim runners for linting

wjt:
Since we don't pay for GitHub Actions this won't save us any money; but
it might save some electrons.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -16,7 +16,7 @@ jobs:
   pre-commit:
     name: Linting and Formatting
     if: ${{ !github.event.repository.fork }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     steps:
     - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


wjt:
As I expected this is slower:

```
STATUS  TITLE                                                             WORKFLOW  BRANCH                                                       EVENT         ID           ELAPSED  AGE
✓       ci: Use ubuntu-slim runners for linting                           Checks    wjt/ci-use-ubuntu-slim-runners-for-linting                   pull_request  26445263491  1m15s    about 2 minutes ago
✓       report_bug_button: Use full git describe --tags version           Checks    wjt/report-bug-button-use-full-git-describe-tags-version     pull_request  26443459468  37s      about 40 minutes ago
✓       Cinematic: Always emit cinematic_finished                         Checks    wjt/cinematic-always-emit-cinematic-finished                 pull_request  26443341014  35s      about 42 minutes ago
```

The question is: does it help anyone? I think in the scheme of things, probably not.

Issue #2186, opened 2026-05-14
Add closing scene with Arun at Fray's End after tutorial quest


wjt:
Currently at the end of the tutorial quest you are unceremoniously dumped into the middle of Fray's End. Arun, who has accompanied you all this way, is nowhere to be seen.

I think we should instead have the player spawn on the bottom island and follow him up to the Eternal Loom, whereupon he tells you about interacting with the loom and how to go on future quests.

PR #2256, opened 2026-05-29
Go to bottom of Fray's End after tutorial

wjt:
The old intro used to take the player to the bottom island of Fray's End
rather than taking them to right outside the loom. This was removed in
the new tutorial because CollectibleItem didn't have a way to specify a
spawn point in the target scene.

Now it does: do so.

Helps https://github.com/endlessm/threadbare/issues/2186


wjt:
As described on https://github.com/endlessm/threadbare/issues/2186 I imagine we should have Arun be in Fray's End as well, but I don't want to change Fray's End while @manuq is working on a big branch touching it - I've already done that more than once and I feel bad :)

manuq (approved):
Yes!

PR #2257, opened 2026-05-29
Move void-spreading enemy out of quest_002/1_void_runner folder

wjt:
Move void-spreading enemy out of quest_002/1_void_runner folder

This enemy living in a subfolder of the first challenge of this quest is
an artifact of how it was developed. But it's now used in other
challenges of this quest, and in the tutorial quest, so I think the time
has come to move it to its own folder at a higher level of the project.
(Actually the new folder path happens to be exactly the same length as
the old one...)

Remove `@experimental` - I don't think this is any more or less
experimental than other bits of the game at this point.

Remove a couple of unused `const`s from the void_spreading_enemy.gd
script itself.


manuq (approved):
Yes, excellent!

PR #2258, opened 2026-05-29
void_runner: Use Phantom Camera for zoom-out effect

wjt:
Previously I manually tweened the Camera2D's zoom to a number which
framed both the player/monk/temple, and the thread at the opposite edge
of the level.

Move the zoom-out effect logic to void_runner.gd, so that the monk
doesn't need a special script, and remove monk.gd. Tweak the dialogue
slightly.

Set up PhantomCameraHost + a PhantomCamera2D for the player. Configure
it to be limited by the Water layer which surrounds the dressed area of
the level.

Add a second camera that frames the temple plus the ruined house to the
left of the thread, with a minimum zoom that allows this.

Then change the script to adjust its priority up and down rather than
manually tweening the camera. I changed the animation to start the
void's fade-out while the camera is still tweening, and to not wait for
the camera to zoom back before continuing the dialogue. I think this
makes it a bit snappier.


scenes/quests/lore_quests/quest_002/1_void_runner/components/void_runner.gd
```diff
@@ -2,8 +2,18 @@
...
 
 func is_enemy_defeated() -> bool:
 	return not is_instance_valid(enemy)
+
+
+func repel_void() -> void:
+	whole_scene_camera.priority += 20
+	await void_layer.uncover_all(3.0)
+	await collectible_thread.reveal()
+	whole_scene_camera.priority -= 20
```
manuq, line 19: 👍

manuq (approved):
It works very well!

PR #2261, opened 2026-05-31
Improve music puzzle

wjt:
Add background music to this scene: initially just a bass line and very
small percussion part, which progresses as you complete each step of the
puzzle. The full melody plays when you have completed the puzzle. Change
the melody on the rocks to match. Include the strudel.cc source code to
the music; the five files are simply slices of a 16-cycle export. The
guitar arpeggios + kalimba + folk harp stack was me trying to imagine
what the musician's weird three-necked instrument might sound like.

The clips for each step of the puzzle are configured to loop from
halfway through the clip, so that the gentle cymbal hit at the start
only plays once per step. This currently doesn't work in the Godot web
target but it sounds OK anyway.

Increase the attenuation and reduce the volume on the ambient sounds so
they're not so overwhelming when you're far away from the forest/river.
Add an AudioListener to the player so that the distance is measured from
the player, rather than the middle of the screen, which makes a
big difference at the start of the level.

The new melody is in A♭ major. I picked this key (and cheated by having
the double bass play a low E♭, which can't be played without downtuning
or having a C extension) because the combat music is in F minor, which
is the relative minor of A♭ major.

The melody incorporates the opening four notes of the theme,
specifically as the third step of the puzzle.

The rock scale starts on G (the 7th note of the scale) rather than A♭.
It's a bit weird but it lets you play the full Threadbare theme melody
(in B♭ dorian) with rocks 5 4 5 3 2 1, 5 4 5 3 5 6. I would like to add
these as 2 further optional steps to the puzzle to unlock a new area of
the level in future.

Make the rock strike duration significantly shorter. This duration
happens to be almost exactly a quaver (8th note) at 80bpm, the speed of
the music. I thought about trying to sync the demo to the start of a
bar, but it's not worth it.

Adjust the musician's hints. I removed my pet peeve – that he would
previously give you partial hints that were wrong:

> Musician: Try tapping the first, and second stones, then check the
> first altar above. If you see flames, we're on to the right melody.

This is a bad hint because if you do what he says, nothing happens!

Instead his clues are now more like:

1. Listen to the altar./Look at the symbol.
2. Explain the symbol.
3. Give you the answer.

For the final step, he tells you exactly what to do, but not the rocks.
I think if someone's got that far they can work it out.


wjt:
https://play.threadbare.game/branches/endlessm/wjt/improve-music-puzzle/#quests/lore_quests/quest_001/1_music_puzzle/music_puzzle

wjt:
Returning to draft to tweak the transitions a little

manuq (approved):
I played it with earphones and whoa, I'm impressed! I love the bass and how it progresses as the player solve the puzzle.

I tried making the "keep the time with the foot" animation of the musician to match the rithm by repeating the first frame multiple times. But it was impossible, so I gave up.

wjt:
Thanks for trying!

I'll rebase this and fix the merge conflicts.

One thing I'm a bit worried about is that most laptop speakers have poor bass response and so the music might be inaudible. Oh well if so, I guess.

PR #2270, opened 2026-06-02
Player: Consider multiple systems controlling it

manuq:
Previously, if two scripts called take_control() of player, then the first who called return_control() would change the player to USER_CONTROLLED, even if the other script was assuming that control was taken.

Do as the observers in InteractArea, keeping an array of nodes controlling it.


manuq:
At one point I had this as part of the townies pull request. I ended up not needing it. But my first implementation of the system-controlled player seems wrong.

wjt (approved):

PR #2271, opened 2026-06-02
Phantom Camera: Reduce unnecessary processing

wjt:
Phantom Camera: Reduce unnecessary processing

Apply the changes I have submitted upstream in
https://github.com/ramokz/phantom-camera/pull/657 to avoid processing
every frame in the editor, even when the open scene does not use
PhantomCamera.

As detailed in that PR and in
https://github.com/ramokz/phantom-camera/issues/649, I haven't figured
out how to avoid the PhantomCameraHost node processing every frame even
when the viewfinder is closed.

And as also detailed there: any inactive PhantomCamera2D still
_process()es every frame, and the PlayerHook has an inactive
PhantomCamera2D, so...

However in most scenes containing the player we have at least one moving
tree or water ripple animation, which causes the scene to render at
60fps anyway. So, oh well!

### DIFF
--- a/addons/phantom_camera/examples/example_scenes/2D/sub_scenes/playable_character_2d.tscn
+++ b/addons/phantom_camera/examples/example_scenes/2D/sub_scenes/playable_character_2d.tscn
@@ -106,6 +106,7 @@ script = ExtResource("4_rloon")
 zoom = Vector2(2, 2)
 frame_preview = false
 tween_resource = ExtResource("5_4iyk1")
+inactive_update_mode = 1
 follow_damping_value = Vector2(0, 0)
 draw_limits = true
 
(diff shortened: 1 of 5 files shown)


manuq (approved):

manuq:
> And as also detailed there: any inactive PhantomCamera2D still _process()es every frame, and the PlayerHook has an inactive PhantomCamera2D, so...

Could this be because of the "Always" default in inactive update mode?

image

wjt:
> Could this be because of the "Always" default in inactive update mode?

Yes. However if you disable that, switching to this camera is janky in-game, because its position is not updated until the moment it is activated. Sorry, I forgot to write that in this PR!

PR #2275, opened 2026-06-02
Redesign dialogue balloon

### LINKED ISSUE #1846: "Next →" button in one-line dialogue is confusing
### Description

We saw a playtester struggle with the opening dialogue in Fray's End:

Image

They clicked "Next" expecting to see something else (I believe). In fact the dialogue balloon disappeared. They repeatedly talked to the NPC again, with the same result.

My theory is that the word "Next" is what is confusing here.

### Checklist

### LINKED ISSUE #2113: Adjust input HUD during dialogue
Now that we always show input hints for player movement or Sokoban controls, it is confusing that they are displayed during dialogue:

Image

Image

We should swap these for dialogue-specific hints during dialogue.

The actions depend on what stage the dialogue is at:

- If the dialogue is being typed, Skip can be used
- If the dialogue is waiting for the player to choose from a number of responses, the up/down arrows and the next/accept/interact button can be used to choose
- Otherwise, dialogue_next can be used to advance to the next line of dialogue (possibly ending the dialogue, c.f. https://github.com/endlessm/threadbare/issues/1846)

### XP Summary (total 450)

| Subdomain | Skill | XP |
| - | - | - |

### LINKED ISSUE #2278: Redesign dialogue balloon
The current balloon UI is a bit cluttered, is a bit narrow with few room for words per line. Also it's anchored to the top left of the screen, while it is more common to have subtitles in games or movies at the bottom or top of the screen.

manuq:
The balloon is now anchored to the screen bottom-center by default. And is anchored
to the screen top-center through scripting if the player position in the
screen is below the 3/4 of the screen height.

The balloon size is now the double as before, with more room for words per line.

The "next" button now has a TextureRect inside with the interact_input.gd script
attached, so it will display the corresponding input to close the dialogue. For
theming the background it has its own theme type variation.

The "next" button is now outside the visible panel, stacked vertically with a
bit of negative margin to overlap it. The visible panel also has margin (through
MarginContainer nodes) to prevent the text to be below the "next" button, and to
make the "next" button appear outside horizontally.

The label for the talking character name with the ribbon is also outside the visible
panel and overlapping it.

For now, the panels of this dialogue use existing textures modulated to be:
lighter and a bit transparent for the main panel. And darker for the "next"
button background.

### Balloon: Add small squash-stretch effect when it appears

This is intended to grab the eye of the player to the dialogue when it appears.
Because now it can appear at the top or at the bottom, conditionally.

### Input HUD: Hide it when displaying dialogue

During dialogue, the input is grabbed by it and the player can't move, run or
trigger other actions. Also, now this HUD may appear below the dialogue when it
is displayed at the bottom.

### Powerup dialogue: Center text in dialogue

It looks better!

Resolve https://github.com/endlessm/threadbare/issues/2278
Fix https://github.com/endlessm/threadbare/issues/1846

### DIFF
--- a/scenes/ui_elements/dialogue/components/balloon.gd
+++ b/scenes/ui_elements/dialogue/components/balloon.gd
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: 2022-present Nathan Hoad and Dialogue Manager contributors.
 # SPDX-License-Identifier: MIT
+@tool
 extends CanvasLayer
 ## A basic dialogue balloon for use with Dialogue Manager.
 
@@ -11,6 +12,9 @@ const NPC_RIBBON_TYPE_VARIATION := &"NPCRibbon"
 
 ## The action to use to skip typing the dialogue
 @export var skip_action: StringName = &"dialogue_skip"
+
+@export_tool_button("Anchor to top") var to_top_editor_button: Callable = anchor_to_top
+@export_tool_button("Anchor to bottom") var to_bottom_editor_button: Callable = anchor_to_bottom
 
 ## The dialogue resource
 var resource: DialogueResource
@@ -75,6 +79,8 @@ var _player_name: String = ""
 
 
 func _ready() -> void:
+	if Engine.is_editor_hint():
+		return
 	balloon.hide()
 	Engine.get_singleton("DialogueManager").mutated.connect(_on_mutated)
 
@@ -104,6 +110,8 @@ func _notification(what: int) -> void:
 		self.dialogue_line = await resource.get_next_dialogue_line(dialogue_line.id)
 		if visible_ratio < 1:
 			dialogue_label.skip_typing()
+	elif what == NOTIFICATION_EDITOR_PRE_SAVE:
+		anchor_to_bottom()
 
 
 ## Start some dialogue
@@ -141,15 +149,9 @@ func apply_dialogue_line() -> void:
 
 	next_button.hide()
 
-	if _is_player_at_top():
-		# Anchor container to bottom:
-		balloon_container.anchor_top = 1
-		balloon_container.anchor_bottom = 1
-		balloon_container.offset_top = 4096
-		balloon_container.offset_bottom = 0
-		balloon_container.grow_vertical = Control.GROW_DIRECTION_BEGIN
-		# And change the order of the next button:
-		balloon_container.move_child(next_button, 0)
+	# Assumes that the balloon scene is anchored to the bottom by default:
+	if _is_player_at_bottom():
+		anchor_to_top()
 
 	# Show our balloon
 	balloon.show()
@@ -189,7 +191,8 @@ func apply_dialogue_line() -> void:
 		next_button.show()
 
 
-func _is_player_at_top() -> bool:
+## True if the player position is at the bottom vertical quarter of the screen.
+func _is_player_at_bottom() -> bool:
 	var player: Node2D = get_tree().get_first_node_in_group("player")
 	if not player:
 		return false
@@ -197,7 +200,25 @@ func _is_player_at_top() -> bool:
 	var viewport := player.get_viewport()
 	var screen_pos: Vector2 = viewport.get_canvas_transform() * player.global_position
 	var viewport_size: Vector2 = viewport.get_visible_rect().size
-	return screen_pos.y <= viewport_size.y / 2.0
+	return screen_pos.y > viewport_size.y * 3 / 4.0
+
+
+## Anchor container to top:
+func anchor_to_top() -> void:
+	balloon_container.anchor_top = 0
+	balloon_container.anchor_bottom = 0
+	balloon_container.offset_top = 0
+	balloon_container.offset_bottom = -4096
+	balloon_container.grow_vertical = Control.GROW_DIRECTION_END
+
+
+## Anchor container to bottom:
+func anchor_to_bottom() -> void:
+	balloon_container.anchor_top = 1
+	balloon_container.anchor_bottom = 1
+	balloon_container.offset_top = 4096
+	balloon_container.offset_bottom = 0
+	balloon_container.grow_vertical = Control.GROW_DIRECTION_BEGIN
 
 
 ## Go to the next line
--- a/scenes/ui_elements/dialogue/components/balloon.gd
+++ b/scenes/ui_elements/dialogue/components/balloon.gd
@@ -53,11 +53,11 @@ var _player_name: String = ""
 ## The base balloon anchor
 @onready var balloon: Control = %Balloon
 
-## The panel holding the label showing the name of the currently-speaking character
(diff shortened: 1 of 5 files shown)


addons/dialogue_manager/dialogue_manager.gd
```diff
@@ -37,6 +37,19 @@ signal bridge_get_line_completed(call_index: int, line: DialogueLine)
...
+	"bottom": "res://scenes/ui_elements/dialogue/balloon_bottom.tscn",
+	"left": "res://scenes/ui_elements/dialogue/balloon_left.tscn",
+}
+
+const BALLOON_SIMPLE_VARIANTS := {
+	"top": "res://scenes/ui_elements/dialogue/balloon_top_simple.tscn",
+	"right": "res://scenes/ui_elements/dialogue/balloon_right_simple.tscn",
+	"bottom": "res://scenes/ui_elements/dialogue/balloon_bottom_simple.tscn",
+	"left": "res://scenes/ui_elements/dialogue/balloon_left_simple.tscn",
+}
```
wjt: This will prevent us from updating Dialogue Manager without reapplying this patch. If we do this, we should disable the update checker.

manuq:
Today I realized that there is no point in centering the dialogue with the camera (which I drafted in https://github.com/endlessm/threadbare/pull/2240) if we don't decide first the position of the dialogue.

So in this branch I'm playing with variations of the existing dialogue. There are balloon scenes for:
- Anchored at top-left (the existing one)
- Anchored at top-right
- Anchored at bottom-center, with double width
- Anchored at top-center, with double width

And also variants with a simpler texture for the panel, which in my opinion reads better because it's less cluttered and needs less margin.

In a separate commit, I hardcoded in the addon (just for testing) how it would look like to conditionally show the dialogue at the top or at the bottom (with the simpler textures) depending on the player position on screen:

image

image

Without that commit, all the variants can be tried by picking one of them in the project settings:

image

wjt:
Oh I really like the variant with the simplified texture. If we also simplify the name header to be bold text, and replace the massive "next" button with a tiny bouncing arrow, we'll save loads of space and we may not even need to recentre the player!

addons/dialogue_manager/dialogue_manager.gd
wjt: OK, now I saw your comment & understand!

manuq:
> Oh I really like the variant with the simplified texture. If we also simplify the name header to be bold text, and replace the massive "next" button with a tiny bouncing arrow, we'll save loads of space and we may not even need to recentre the player!

Here is a test, using the interactive input hint for the button (related to https://github.com/endlessm/threadbare/issues/1846). The good thing is that it changes when the input changes from keyboard to controller. Also having it "floating" (is actually stacked vertically with some offset) makes the dialogue not resize when the button appears.

Screenshot From 2026-06-03 14-50-43

Screenshot From 2026-06-03 14-51-49

manuq:
Oh and forgot to mention: The font we use doesn't have bold or italic for the player/NPC name. I tried underline (converting the label to a RichTextLabel). But it also doesn't look good. So maybe just uppercase? Although I don't like "StoryWeaver" uppercased. With BBCode we can change the color to be (for example) the player name in blue, like we had the blue ribbon before. But it doesn't look very good. I have yet to try with shadows or outlines.

manuq:
Back to draft to address feedback from @JoniCeceri:
- Make the dialogue default to bottom, not top.
- Reintroduce the ribbons.

manuq:
I have updated it according to the feedback above https://github.com/endlessm/threadbare/pull/2275#issuecomment-4623662953

manuq:
Here is how one part of the game looks:

[recording.webm](https://github.com/user-attachments/assets/a0206658-522c-419c-8caa-9c23364aa524)

manuq:
The video above also shows that the code we have to change the ribbon to blue when the player is talking is unreliable. The farmer dialogue has the name as:
```
{{player_name}}: ...
```
While the eternal loom dialogue has "StoryWeaver" hardcoded.

wjt:
If you look at the dialogue script, it actually tries to extract the `player_name` property from a `Player` object in the dialogue context. Then it's a normal string comparison.

The idea was that the same logic needs to work in StoryQuests where the player has a different name.

The `{{player_name}}:` syntax is useful in generic dialogue that might be spoken by a different player. But what gets compared is the expanded speaker name.

But… I don't think it's worth it. It would be more reliable (i.e. nothing to go wrong) to always use one colour.

wjt:
I haven't reviewed the implementation but, wow, that looks so much better. I particularly like the bounce effect.

wjt:
```gdscript
	DialogueManager.show_dialogue_balloon(GameState.quest.quest.retelling, "", [self])
```

If you get a reference to `player` and pass it in along with `self`, the colour changing will work.

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -138,10 +149,22 @@ func apply_dialogue_line() -> void:
...
+
 	# Show our balloon
 	balloon.show()
 	will_hide_balloon = false
 
+	# Add a squash-stretch effect when the dialogue appears.
+	balloon_container.scale = Vector2(1.15, 0.7)
+	var tween := create_tween().set_parallel(true).set_trans(Tween.TRANS_BACK).set_ease(
+		Tween.EASE_OUT
+	)
```
wjt: Perhaps a different workaround for gdlint/gdformat's weird taste in formatting:

```suggestion
	var tween := create_tween().set_parallel(true)
	tween.set_trans(Tween.TRANS_BACK).set_ease(Tween.EASE_OUT)
```

wjt:
It's a particularly big improvement here:

image

Previously the balloon obscured what Arun is talking about.

wjt:
The "is the player at the bottom of the screen?" check doesn't work in the boilerplate StoryQuest opening & closing cutscenes because the "player" there is not a real instance, just a bare sprite:

image

I don't think this is a blocker but we should file an issue to follow up on this. Perhaps we could add API to `player.gd` which makes them walk along a given `Path2D`?

wjt:
image

But in the next scene, again: previously the balloon partially obscured Stella and the Firekeeper. Big improvement. BTW I spy a bug in this screenshot that I'll report separately: https://github.com/endlessm/threadbare/issues/2289.

wjt (approved):
What a massive improvement.

manuq:
> The "is the player at the bottom of the screen?" check doesn't work in the boilerplate StoryQuest opening & closing cutscenes because the "player" there is not a real instance, just a bare sprite:
> 
> image
> I don't think this is a blocker but we should file an issue to follow up on this. Perhaps we could add API to `player.gd` which makes them walk along a given `Path2D`?

Oh good point! Yes one more reason to put the actual player on rails, rather than animating the sprite directly.

PR #2276, opened 2026-06-02
Loom: Save after end-of-quest interaction

wjt:
Previously the Eternal Loom implicitly saved the game state at the end
of the end-of-quest interaction, originally because GameState always
saved itself, and then, after the recent refactor, because it switched
to a Sokoban scene immediately which triggers a save.

Now we need to explicitly save the game state at the end of the
interaction, once the quest has been marked as complete.

In addition, the loom would check when instantiated whether the
end-of-quest interaction is possible, but it cached the result for the
lifetime of the instance. With the removal of the Sokoban transition
this no longer works. We can call is_item_offering_possible() directly
in the dialogue, and eliminate this code to change the dialogue title at
startup. While we're there, set the dialogue resource in the scene not
the script.

### DIFF
--- a/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
+++ b/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
@@ -7,20 +7,11 @@ signal retelling_started
 signal retelling_finished
 signal give_retelling_upgrade(type: InventoryItem.ItemType)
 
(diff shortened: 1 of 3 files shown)


manuq (approved):

manuq:
@wjt sorry I thought I merged this yesterday after trying it, but it turns out I didn't!

PR #2277, opened 2026-06-02
Fray's End: Replace generic NPC with Arun

### LINKED ISSUE #2186: Add closing scene with Arun at Fray's End after tutorial quest

wjt:
Fray's End: Replace generic NPC with Arun

We previously had a generic Tiny Swords NPC at the bottom island, where
StoryWeaver first appears when they first get to Fray's End.

Now that StoryWeaver gets there from the tutorial, it makes more sense
for it to be the same recognisable NPC as in the tutorial. Replace it,
and add some dialogue that branches based on whether the player is on a
quest and, if so, if it's the tutorial.

Resolves https://github.com/endlessm/threadbare/issues/2186

### DIFF
--- a/scenes/world_map/components/tutorial_npc.dialogue
+++ b/scenes/world_map/components/tutorial_npc.dialogue
@@ -1,5 +1,13 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 ~ start
-Villager: Welcome, traveler. Fray's End is just across the bridge.
+if not GameState.quest:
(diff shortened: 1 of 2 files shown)


manuq (approved):
This is a great idea!

PR #2283, opened 2026-06-03
Add a LoreQuest subclass of Quest

wjt:
There are some properties that only make sense on LoreQuests, and we treat them differently in the game state.

Rather than using `_validate_property` magic to hide those properties on quests that don't have the is_lore_quest property set, make a subclass, update all quest resources for lorequests to be LoreQuest instances instead, and adjust code that uses these LoreQuest-specific properties.

Arguably the authors/affiliation properties of Quest only make sense for StoryQuests, so perhaps those should have a subclass, but that's one for another day.


manuq (approved):
Good idea, it makes sense.

PR #2286, opened 2026-06-04
Sign: Add automatic artificial light

wjt:
I am adding signposts to Fray's End and connected scenes that have
time-of-day effects. Putting a light in the signs helps make them
visible even at night.


manuq (approved):
Nice!

PR #2301, opened 2026-06-05
Show dialogue after abandoning a Quest

### LINKED ISSUE #2299: Show brief dialogue after abandoning a quest
Abandoning a quest now takes you back to the hub area where you started it. It would be nice if StoryWeaver spoke to the camera after respawning in the world, just a single line of dialogue, e.g.:

> Maybe I'll try that another time.

> I can always come back later.

etc.

wjt:
We need to pass some temporary state between the point where the quest
is abandoned, and the point where the player has spawned back in the
world. Rather than adding more explicit state to the game, play this
dialogue directly from the pause menu, which is what drives abandoning
quests anyway.

Vary the dialogue based on whether the quest was being played for the
first time, or is being replayed. We could in future vary it based on
whether the quest was a LoreQuest or StoryQuest too.

Resolves https://github.com/endlessm/threadbare/issues/2299

### DIFF
--- a/scenes/globals/pause/pause_overlay.gd
+++ b/scenes/globals/pause/pause_overlay.gd
@@ -4,6 +4,7 @@ extends CanvasLayer
 
 @export_file("*.tscn") var title_scene: String
 @export_file("*.tscn") var frays_end: String
+@export var abandon_dialogue: DialogueResource
 
 @onready var pause_menu: Control = %PauseMenu
 @onready var resume_button: Button = %ResumeButton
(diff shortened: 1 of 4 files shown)


manuq:
> So then the question is: in which component should this logic live? Here's a bad idea: the pause menu! :D

May not be too bad! Just so you know: today I've been looking at references for inventories in gameuidatabase.com and I'm heading towards making the inventory be the pause menu as well.

wjt:
Interesting! OK, I suppose the existing "abandon quest" logic lives in the pause menu already…

wjt:
OK, I reworked this to live in the pause menu.

manuq (approved):

Issue #2280, opened 2026-06-03
Add collected threads to inventory


wjt:
ASCII mockup of a page accessed by pressing some key:

```
INVENTORY

[green thread]  3
[yellow thread] 0
[red thread]    1
```

tracking all threads collected across all quests thus far.

PR #2302, opened 2026-06-05
Refactor abilities names

manuq:
### Add default ability names into a new LoreInfo class

They were set in powerup instances and passed to the dialogue from them. But
these names or verbs will appear in the inventory soon. Also, StoryQuest
creators may want to rename them or modify them.

Also format the "Collect [ability name]" through scripting.

### Add StoryQuest resource with custom ability names

Because StoryQuest creators may want to change these names, or use their own custom abilities.

Update StoryQuest bootstrap automation.

Quest: Use subclass when converting to string for log purposes.

### DIFF
--- a/scenes/game_elements/props/powerup/powerup.tscn
+++ b/scenes/game_elements/props/powerup/powerup.tscn
@@ -113,7 +113,6 @@ unique_name_in_owner = true
 position = Vector2(0, -34)
 sprite_frames = ExtResource("6_ljncq")
 autoplay = "default"
-frame_progress = 0.90992624
 
 [node name="HookableArea" type="Area2D" parent="." unique_id=1984587489 node_paths=PackedStringArray("controlled_entity", "anchor_point")]
 collision_layer = 4096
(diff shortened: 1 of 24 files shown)


manuq:
Another place where these names are used is the input HUD. But in that case we have a mismatch: we do have "Repel" for the repel ability but we have "Throw" for the "Grapple" ability.

scenes/game_elements/props/powerup/powerup.tscn
```diff
@@ -114,6 +113,7 @@ unique_name_in_owner = true
 position = Vector2(0, -34)
 sprite_frames = ExtResource("6_ljncq")
 autoplay = "default"
+frame_progress = 0.90992624
```
wjt:
```suggestion
```

wjt (approved):

wjt (commented):
I think you missed one quest though: NO_EDIt

manuq:
> I think you missed one quest though: NO_EDIt

Oh yes. Also, I need to update the automation to create a StoryQuest. And maybe we should mark the Quest class as abstract? `@abstract class_name Quest`.

PR #2313, opened 2026-06-08
Music puzzle: Demodulate when solved

wjt:
Music puzzle: Demodulate when solved


wjt:
I think this is a bit cheesy - and it doesn't really feel right when the houses are still ruined.

PR #2317, opened 2026-06-08
TalkBehavior: Remove before_dialogue hook

wjt:
TalkBehavior: Remove before_dialogue hook

This was only used by the elder to vary the dialogue based on the game
state. I think it is cleaner to do that directly in response to game
state signals.


manuq (approved):

manuq:
Found a bug and fixed it! https://github.com/endlessm/threadbare/pull/2320

PR #2320, opened 2026-06-08
Fix Elder dialogue title

manuq:
Since ce46b91c _update_dialogue_title() is called too early, before the _storybook is instantiated. So the dialogue is set to stgart at the "no_quests" title, giving the message "I don't have any quests for you". This happens in all 3 elders.

Move the call to update the title, and the signal connections to use it as handler, below on the _ready() method.


wjt:
Oh! I only tested the  change that caused this together with another one that I haven't submitted. Sorry :(

wjt (approved):
Thanks!

PR #2323, opened 2026-06-08
Options menu: Split in multiple setting pages

manuq:
Make the Video, Sound and Language settings have each their own page. This adds
space for expanding these settings.

### Debug Settings: Add title and prevent horizontal overflow

Add ellipsis to quest names and ability names, in case they are too long, so
they don't resize the settings panel.

### Options: Add default page to show

Usually this would be the first element of the TabContainer. We now have
submenus in the pause overlay, so if eg. Video Settings was being displayed and
the player resumes the game, then pauses the game and selects Options, it should
not go to Video Settings (the last one shown) directly.

### Debug Settings: Split in submenus

With submenus introduced, it makes more sense to have the existing 2 debug
setting in their own page, for easier navigation.

### DIFF
--- a/scenes/globals/pause/pause_overlay.tscn
+++ b/scenes/globals/pause/pause_overlay.tscn
@@ -7,7 +7,6 @@
 [ext_resource type="Script" uid="uid://i4urwefxrrdt" path="res://addons/threadbare_git_describe/version_label.gd" id="3_m62bn"]
 [ext_resource type="PackedScene" uid="uid://dkeb0yjgcfi86" path="res://scenes/menus/options/options.tscn" id="3_sd5t1"]
 [ext_resource type="Script" uid="uid://gntvq01vvati" path="res://scenes/globals/pause/report_bug_button.gd" id="4_mjd51"]
-[ext_resource type="PackedScene" uid="uid://56ja3m4283wa" path="res://scenes/menus/debug/debug_settings.tscn" id="5_ai8ue"]
 
 [sub_resource type="StyleBoxEmpty" id="StyleBoxEmpty_1tcw0"]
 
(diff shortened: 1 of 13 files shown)


scenes/menus/debug/components/debug_completed_quests.tscn
```diff
@@ -0,0 +1,59 @@
+[gd_scene format=3 uid="uid://bdbgcqsts7sk2"]
+
+[ext_resource type="Theme" uid="uid://cvitou84ni7qe" path="res://scenes/ui_elements/components/theme.tres" id="1_theme"]
+[ext_resource type="Script" uid="uid://r7cjyiw5log4" path="res://scenes/menus/options/components/options.gd" id="2_back"]
+[ext_resource type="Texture2D" uid="uid://xe25gqovxxpe" path="res://assets/first_party/icons/left_arrow.png" id="3_arrow"]
+[ext_resource type="PackedScene" uid="uid://cie2w455avmwm" path="res://scenes/menus/debug/debug_completed_quest_list.tscn" id="4_list"]
```
manuq, line 6: I believe that these IDs "1_theme", "2_back", "3_arrow", "4_list" come from Claude, which I used to help with the refactoring. Although most of the work was amended by me. I don't know how to make Godot regenerate these IDs, if that's something we should do.

manuq:
This is how it looks like:

Screenshot From 2026-06-08 19-53-04
Screenshot From 2026-06-08 19-52-59
Screenshot From 2026-06-08 19-52-50
Screenshot From 2026-06-08 19-52-45

scenes/globals/pause/pause_overlay.tscn
```diff
@@ -70,6 +70,7 @@ theme = ExtResource("2_sd5t1")
 metadata/_tab_index = 0
 
 [node name="VBoxContainer" type="VBoxContainer" parent="TabContainer/PauseMenu" unique_id=113544093]
+custom_minimum_size = Vector2(640, 0)
```
wjt, line 72: 👍🏻

wjt (approved):
Nice, I think this works.

It does make the nesting a bit deep in the debug menu, but I don't think that's the thing we need to prioritise the usability of.

I also thought at first glance that it was strange to have the debug menu accessible from the title screen - which is conceptually before the saved game state has been loaded - but I guess it could be useful to tweak the saved game prior to restoring it.

PR #2326, opened 2026-06-09
Keep list of quests in the Elder scene

wjt:
Keep list of quests in the Elder scene

Previously, the quest folder path was a property of the elder which
would pass it through to the storybook scene, and rely on the storybook
to see if there are any quests inside.

Instead, give the storybook scene an Array[Quest] property. Set it in
the .tscn to two semi-randomly-selected quests to ease testing.

Then have the elder list the contents of the quest directory and pass
the list of quests into the storybook. Construct the storybook on demand.


scenes/menus/storybook/storybook.tscn
```diff
@@ -268,7 +272,7 @@ grow_vertical = 2
 size_flags_horizontal = 4
 theme = ExtResource("1_m2ghf")
 script = ExtResource("1_gw8hn")
-quest_directory = "res://scenes/quests/story_quests/"
+quests = Array[ExtResource("3_21r2o")]([ExtResource("4_ywp77"), ExtResource("5_ywp77"), ExtResource("5_x5pjn")])
```
manuq, line 275: Looks good!

manuq (approved):

PR #2330, opened 2026-06-09
Elder: Don't crash when setting quest_directory before ready

wjt:
`_update_dialogue_title` requires `talk_behavior` to have been initialised,
which happens via `@onready`. But `quest_directory` is set prior to the
scene becoming ready.


wjt:
@manuq This unbreaks `main`

manuq:
Excellent!

PR #2331, opened 2026-06-09
Add placeholder inventory in Pause overlay

manuq:
### Pause Overlay: Replace background texture with solid color

Use solid black with 30% transparency. The background was too tiled. I tried to
replace the background texture first. But a solid color makes the overlay
visually cleaner, letting the eye focus on the displayed content. Also it makes
more sense to darken the game behind rather than lighten it.

Override the color of the VersionLabel to be white, which matches the Input HUD.
Also add placeholder text to it, to see it in the editor.

### Project: Add input actions for previous_tab and next_tab

To be used for easier navigation in the pause overlay. It could be used in the
storybook as well.

### Input hints: Add more textures for actions

For actions "pause", "previous_tab" and "next_tab"

### Pause Overlay: Use tabs and add shortcuts to switch between them

Rename the Resume button to "Back" with the same arrow icon as others.
Make it grab the focus when the Menu tab becomes visible.

Add input hints as a bottom row, with the hints aligned to the left, keeping the
version label aligned to the right through a separator (an empty control that
expands horizontally).

### Project: Move Pause Overlay autoload below Input Hints

Otherwise _on_input_device_changed() is called too early in the
interact_input.gd script, with the device argument assigned to an empty string.

### Add shortcut to Back buttons for going back in all menus

Add shortcut for "ui_cancel" to reuse it in Back buttons.

### Theme tabs with ribbons, and use them in the Pause overlay

At the same time, remove the titles on each tab, because they are now redundant.
These were PanelContainers using the PlayerRibbon variant.
### Debug Settings: Use 300px as min height for ScrollContainer

Otherwise the Pause overlay overflows.

Ideally this shouldn't be hardcoded. I think it's because of the CenterContainer
that wraps everything.

### Add placeholder Inventory tab in Pause overlay

Just a placeholder view for now.

----


manuq:
Sorry but this pull request became bigger than I was expecting.

This is how it looks:

[recording.webm](https://github.com/user-attachments/assets/a44c4536-625e-40bb-90c6-29433af8b252)

scenes/globals/pause/pause_overlay.gd
```diff
@@ -21,6 +22,16 @@ func _unhandled_input(event: InputEvent) -> void:
...
+	elif event.is_action_pressed("next_tab"):
+		tab_container.current_tab = wrapi(
+			tab_container.current_tab + 1, 0, tab_container.get_tab_count()
+		)
+		tab_container.accept_event()
+	elif event.is_action_pressed("previous_tab"):
+		tab_container.current_tab = wrapi(
+			tab_container.current_tab - 1, 0, tab_container.get_tab_count()
+		)
+		tab_container.accept_event()
```
wjt, line 34: Nice, I was thinking about adding next/prev page bindings for the storybook too. I guess we can consider each spread of the storybook to be a tab.

Also, I have not seen `Control.accept_event` before. Interesting, I wonder if the `get_viewport().set_input_as_handled()` above should also be `self.accept_event()`, and looking at https://docs.godotengine.org/en/stable/tutorials/inputs/inputevent.html whether this should all be in `_gui_input`. (Not blocking this PR, just a thought.)

scenes/menus/options/components/video_settings.tscn
```diff
@@ -57,6 +58,7 @@ unique_name_in_owner = true
 layout_mode = 2
 size_flags_horizontal = 4
 theme_type_variation = &"FlatButton"
+shortcut = ExtResource("4_jbrqx")
```
wjt, line 61: TIL!

wjt (approved):
Looks great. One thought: I don't think the `Menu` tab needs an `Options` option when there is an `Options` tab.

manuq:
> Looks great. One thought: I don't think the `Menu` tab needs an `Options` option when there is an `Options` tab.

I only left the Options option because that tab has a "Back" button that was previously showing the Menu. And I didn't want to remove that button because the Options tab is reused in the title screen. Either I should make the Back button conditional or something. But you gave me an idea: Add a Back button to the inventory too, and rename Resume to be a Back button, and put it at the bottom of the buttons container for consistency. I think this works better.

scenes/globals/pause/pause_overlay.gd
manuq: Yes so basically:
- to consume the event from a function: `Viewport.set_input_as_handled()`
- to consume the event from a Control: `Control.accept_event()`

Now, I didn't know about the `_gui_input()` virtual method! I will read about it.

PR #2332, opened 2026-06-09
Add title screen when entering a LoreQuest

### LINKED ISSUE #2298: Add a title/separator screen when starting a quest
Since #2287, walking across a particular boundary in the world starts a quest.

In https://github.com/endlessm/threadbare/pull/2047#issuecomment-4042398873 @manuq prototyped a little title screen to show at the start of the quest. We should add this to make clear what is happening. It would also be a place to show when the player is replaying an already-completed quest, picking up a quest where they left off (#2281), and perhaps a way to reset the quest.

wjt:
Add title screen when entering a LoreQuest

The player no longer chooses LoreQuests from a book, so there is no
longer any clear demarcation that a quest has begun.

Add a quest title screen, which displays the the title, description, and
art from the current Quest resource, then switches to the first scene
after a short delay. Update QuestTeleporter to switch to this scene
rather than to the first scene of the quest. Remove the exit_transition
property from QuestTeleporter: the title card removes the spatial
relationship between the hub and the starting scene of the quest.

Resolves https://github.com/endlessm/threadbare/issues/2298

Co-authored-by: Will Thompson <wjt@endlessaccess.org>

### DIFF
--- a/scenes/game_elements/characters/npcs/elder/components/elder.gd
+++ b/scenes/game_elements/characters/npcs/elder/components/elder.gd
@@ -22,7 +22,6 @@ const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
 ## The quest chosen by the player from the storybook
 var chosen_quest: Quest
 
-var _dialogue_balloon: CanvasLayer
 var _quests: Array[Quest]
 
 @onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
@@ -75,8 +74,10 @@ func show_storybook() -> void:
 
 	# GDM will hide the balloon after a short pause if the awaitable hasn't resolved, but we want it
 	# to be replaced with the storybook immediately.
-	if _dialogue_balloon:
-		_dialogue_balloon.balloon.hide()
+	if talk_behavior.dialogue_balloon:
+		# Confusingly the DialogueBalloon node (root of our balloon.tscn) has a balloon property;
+		# it is the latter which gets hidden and shown.
+		talk_behavior.dialogue_balloon.balloon.hide()
 
 	_storybook_layer.add_child(storybook)
 	chosen_quest = await storybook.selected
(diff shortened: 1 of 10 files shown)


scenes/game_elements/characters/npcs/elder/components/elder.gd
```diff
@@ -75,8 +74,10 @@ func show_storybook() -> void:
 
 	# GDM will hide the balloon after a short pause if the awaitable hasn't resolved, but we want it
 	# to be replaced with the storybook immediately.
-	if _dialogue_balloon:
-		_dialogue_balloon.balloon.hide()
+	if talk_behavior.dialogue_balloon:
+		# Confusingly the DialogueBalloon node (root of our balloon.tscn) has a balloon property;
+		# it is the latter which gets hidden and shown.
+		talk_behavior.dialogue_balloon.balloon.hide()
```
manuq, line 80: Oh!

scenes/menus/quest_separator/quest_separator.tscn
```diff
@@ -0,0 +1,113 @@
...
+void fragment() {
+	// Called for every pixel the material is visible on.
+}
+
+//void light() {
+//	// Called for every pixel for every light affecting the CanvasItem.
+//	// Uncomment to replace the default light processing function with this one.
+//}
+"
+
```
manuq, line 38: Is this shader doing anything at all?

manuq (approved):
Nice, the quest teleporter makes this easier than when I tried it!

The title screen is ugly, I did it in a rush. But we can land it now and improve it later!

scenes/menus/quest_separator/quest_separator.tscn
wjt: Yes it is - if you remove it you'll only see the radial texture. I think the `render_mode blend_sub` is doing everything but I don't really know!

wjt: Maybe I'll try to recreate this with the visual shader editor. Or just remove the stub entrypoints and document it.

PR #2333, opened 2026-06-09
quest separator: Remove shader

wjt:
quest separator: Remove shader

Use a CanvasItemMaterial and set its blend mode to Substract.


manuq (approved):

manuq (changes requested):
I see that the same could be achieved by using a CanvasItemMaterial and setting the blend mode to "substract".

Image

wjt:
Oh, even better!

PR #2336, opened 2026-06-10
cinematic: Support selecting dialogue title

wjt:
I want to use one dialogue file across three scenes, with one title per
scene.

Add a dialogue_title property to cinematic. Add a small amount of
validation and inspector magic to go with it.

Ideally we would have the same thing on TalkBehavior...


manuq (approved):
Great idea validating that the title exists and displaying a warning in the editor!

PR #2337, opened 2026-06-10
Split placeholder quest in two

### LINKED ISSUE #2268: Split “Placeholder” quest into 3
We need a larger number of quests to be able to really test the improved quest loop.

“Placeholder” has a sequence puzzle + three rounds of ink combat. Let's split it into 3 quests. My suggestion from #2264 is that the split should be:

- Stealth/sequence + Ink Combat 4 (order TBC)
- Ink Combat 5
- Ink Combat 6

wjt:
Separate the stealth/sequence combo from the ink combat trio. Give each
a slightly better title and description.

Add intro dialogue from the musician to each of the ink combat scenes to
explain the new mechanic in each.

Redesign the Song Sanctuary hub to have entrypoints for each.

Adjust Bridget and the LoreQuest Elder's dialogue for the split,
prioritising the Guarded Stones but allowing for the player having
completed the quests in either order.

While we are moving these files around anyway: some of the ink
barrel-related files in this quest were referenced from scenes in
scenes/game_elements/props/filling_barrel. Move these files out to that
folder, so that generic components are not referencing files inside a
quest.

Resolves https://github.com/endlessm/threadbare/issues/2268


scenes/quests/lore_quests/song_sanctuary_3/1_ink_drinker_levels/components/musician.dialogue
```diff
@@ -0,0 +1,16 @@
...
+
+~ round_5
+Musician: The jars here are fragile: the wrong color of ink will crack them.
+Musician: But you only need to fill one up. Good luck!
+=> END
+
+~ round_6
+Musician: You have to fill these jars in order. Look for the one that is sparkling.
+Musician: And watch out for the Mothsache! They have a nasty sting.
+=> END
```
manuq, line 16: Yes! Explaining the mechanics at least throug dialogue is so important.

manuq (approved):

PR #2343, opened 2026-06-10
Theme: Set it as default for the whole project

manuq:
Otherwise, each Control scene has to set the theme to a root node for viewing it correctly in the editor, while the game doesn't need it.

Remove duplicate themes in the StoryBook. Instead, new theme variation or theme overrides should be used.


manuq:
Draft for now because I spotted that the sign panel wasn't using our theme:
Screenshot From 2026-06-10 14-44-07

Also the StoryBook has regressed. It was using 2 custom themes in 2 nodes, that were a duplicate of our theme:

Screenshot From 2026-06-10 14-37-41
Screenshot From 2026-06-10 14-37-50
Screenshot From 2026-06-10 14-38-45
Screenshot From 2026-06-10 14-38-50

scenes/menus/storybook/components/storybook_page.tscn
```diff
@@ -39,6 +39,7 @@ theme_type_variation = &"PlayerRibbon"
 [node name="Title" type="Label" parent="HBoxContainer/VBoxContainer/TitleBox" unique_id=1799767777]
 unique_name_in_owner = true
 layout_mode = 2
+theme_override_font_sizes/font_size = 20
```
wjt, line 42: This is preserving the previous appearance, so we shouldn't block on this, but I am not sure it's right. Each Jersey font only looks crisp at (a multiple of) one particular pixel size, which is not the same as the number in its name (which is the height of a capital letter).

I think I originally figured out the numbers per font by looking at the "charted" variants at https://github.com/scfried/soft-type-jersey/tree/main/fonts/ttf and counting the number of squares from the baseline to the top of the grid. Naturally I didn't write them down. For Jersey20 it's 27px for example, and for the font used for these labels, Jersey15, it's 21 (not 20).

However because it is unusual to play this game at an integer multiple of 1280×720, I wonder if for the UI we should use linear interpolation rather than nearest-neighbour, and use a non-pixel font. Again one for another day!

wjt (approved):
Nice cleanup!

scenes/menus/storybook/components/storybook_page.tscn
manuq: Oh thanks for the information. I will keep it in mind when doing the StoryBook with TabContainer (not urgent).

PR #2347, opened 2026-06-10
Input hints: Display logical keys

### LINKED ISSUE #2185: Show logical labels for keyboard keys
The keyboard bindings for our input actions are bound to "Physical Keycode (Position on US QWERTY Keyboard)". This means that regardless of the keyboard layout in use, pressing the keys in the same physical position as WASD on a US English keyboard moves the player; X throws; Z repels; etc.

On a French [AZERTY](https://en.wikipedia.org/wiki/AZERTY) keyboard the glyphs printed on these keys are different. QWERTZ (used chiefly in German-speaking countries) also has one relevant difference. (Keyboards with Dvorak labels are even rarer than Dvorak typists and I don't think we should go out of our way to support Dvorak users unless it also helps us support AZERTY and other regional layouts that normal people use, but since I do actually type Dvorak I can use myself as a proxy for an AZERTY user so I include Dvorak in the table below :))

| Action | QWERTY | QWERTZ | AZERTY | Dvorak |
| ------ | ------ | ------ | ------ | ------ |
| Up     | W      | W      | Z      | ,      |
| Left   | A      | A      | Q      | A      |
| Down   | S      | S      | S      | O      |
| Right  | D      | D      | D      | E      |
| Throw  | X      | X      | X      | Q      |
| Repel  | Z      | Y      | C      | J      |

I think it is the correct behaviour to bind the physical location rather than the logical key symbols; but we currently show QWERTY labels to the player too:

arrows to aim, x to throw, z to repel

(We don't show WASD anywhere, but annoyingly `Z` to repel is one of the keys that really is differently labelled on the two major non-QWERTY Latin layouts.)

I think we should adjust the displayed symbols to reflect the logical value of the key rather than showing the QWERTY one.

wjt:
Our keybindings are, correctly, configured to use the physical location
of keys on the keyboard. This is because WASD are chosen to be the same
shape as the arrow keys, not because "W" means up; and so other bindings
such as Z and X must also be physical to avoid clashing.

However this means the labels are wrong for QWERTZ and AZERTY keyboards:

| Action | QWERTY | QWERTZ | AZERTY |
| ------ | ------ | ------ | ------ |
| Up     | W      | W      | Z      |
| Left   | A      | A      | Q      |
| Down   | S      | S      | S      |
| Right  | D      | D      | D      |
| Throw  | X      | X      | X      |
| Repel  | Z      | Y      | W      |

We don't actually show WASD in the UI, but we do show Z for Repel (for
example).

Rather than hardcoding that the glyph for Repel is Z on a keyboard and
L1 on a gamepad, instead look these up dynamically.

Update to Kenney Input Prompts 1.5, carrying forward the change from
commit c17ffac3c3837e45d30ce64c79ced6145779498c to remove the solid grey
region in the top-right of the keyboard-and-mouse atlas. Apply the same
change to the joypad spritesheets: the Switch stick glyphs would have
the same problem, but we've never tested them!

Among other things, this update introduces a glyph for the "Pause" key
on a keyboard, which we do actually have a binding for (though it's not
the primary so is not shown). Only check in the five atlases sheets that
we actually use – keyboard, switch, steamdeck, xbox, playstation – along
with the accompanying XML file.

Change the button resources to be based on the `Key`/`JoyButton`/etc.
enums. Rewrite the importer accordingly to generate just one resource
per device type (which contains all the necessary `AtlasTexture`s)
rather than hundreds of separate `AtlasTexture` resources. (It was my
idea to do the latter; I was wrong!)

Then in the input hint controls, determine which key/button the event is
bound to, and display that.

In theory this will mean that we never have to manually look up the
glyphs for a new action again: interact_input.gd will do that for us.

On the web, `DisplayServer.keyboard_get_keycode_from_physical()` is not
implemented: it is not possible to ask the browser for the logical
keycode corresponding to a physical keycode. So on the web platform we
always show 'Z' even on an AZERTY keyboard where it should be 'W'. A bit
sad but not a regression.

There is a web API for this, but it is not widely supported, is blocked
in some browsers for anti-fingerprinting reasons, and Godot doesn't
(yet) know how to use it.

https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/getLayoutMap

I considered working around this limitation, either by calling
`getLayoutMap` directly, or by inferring the keyboard layout based on
the events we actually receive: i.e. if we receive an InputEventKey
whose keycode is 'z' but physical keycode is 'w', we can infer that the
layout is almost certainly AZERTY. This does work – which makes me think
that Brave blocking `getLayoutMap()` is probably not improving user
privacy! – but I don't think it's worth the complexity.

There are some other annoying edge cases to handle, including:

- R2 is analogue so is represented in Godot as an axis that can never be
  negative, but we present it as a button.

- The arrow keys want to be treated like a joystick in the UI.

- We show mouse + keyboard controls for some actions, so we need to keep
  the keyboard vs mouse branching.

- Sony helpfully relabelled the Start/Select buttons from PS3 (and
  older) to something else in PS4 and something else again in PS5. I
  chose the PS5 glyphs because they don't have any text on them. If we
  wanted to fix this in future, we could ask InputHelper for the
  "granular" identifier for the joypad and handle the three PlayStation
  generations separately.

Resolves https://github.com/endlessm/threadbare/issues/2185


wjt:
Needs rebase; conflicts due to #2331.

Also needs testing on the web platform.

scenes/game_elements/props/hint/input_key/interact_input.gd
```diff
@@ -23,10 +24,53 @@ func _ready() -> void:
...
+		set_process(false)
+
+
+func _set_keyboard_texture() -> void:
+	var event := InputHelper.get_keyboard_input_for_action(action_name)
+	if event is not InputEventKey:
+		push_warning("Primary keyboard binding for %s not a key: %s" % [action_name, event])
+		return
+
+	var logical_keycode := DisplayServer.keyboard_get_keycode_from_physical(event.physical_keycode)
```
wjt: TODO: keyboard_get_keycode_from_physical is not supported on the web platform

index.js:467 ERROR: Not supported by this display server.
onPrintError @ index.js:467
index.js:467    at: keyboard_get_keycode_from_physical (servers/display/display_server.cpp:1211)

wjt: Physical bindings work on the web but that's because at the time the keyboard event is delivered, both the physical and logical codes are provided:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#instance_properties

There is an API to do the lookup, but only in Chromium:

https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/getLayoutMap

I wonder if there is a widely used workaround.

wjt: But in the worst case we can just make this call conditional on "is not the web", and fall back to the current behaviour.

wjt:
> Play this branch at https://play.threadbare.game/branches/endlessm/wjt/rewrite-input-hud-again/.

Note that as explained in the commit message, the physical-to-logical mapping **does not work on the web**. So on the web we can only test for regressions.

On desktop, one must reconfigure one's keyboard to e.g. QWERTZ, AZERTY, Dvorak. I will attach some screenshots of the latter...

wjt:
image

Well I'm glad I took a screenshot - that's a nasty black line in the mouse button glyph!

wjt:
Sokoban on Dvorak: 
image

wjt:
Ah I remember now - in commit c17ffac3c3837e45d30ce64c79ced6145779498c the keyboard and mouse atlas was edited to remove the solid grey region directly above that mouse glyph.

I'm going to see if I can fix it in the shader. Update: I couldn't. It's easier to carry forward the change to the atlas!

wjt:
Dvorak pause menu:

image

scenes/game_elements/props/hint/input_key/interact_input.gd
```diff
@@ -23,10 +24,65 @@ func _ready() -> void:
...
+	var event := InputHelper.get_keyboard_input_for_action(action_name)
+	if event is not InputEventKey:
+		push_warning("Primary keyboard binding for %s not a key: %s" % [action_name, event])
+		return
+
+	if event.physical_keycode:
+		# Try to show the logical label for physical mappings; i.e. on AZERTY
+		# show W when the physical binding is for Z. As of Godot 4.6, this API
+		# is only implemented on X11/Wayland/Mac/Windows; notably it is not
+		# available on the Web port.
```
manuq, line 61: 👍

scenes/game_elements/props/hint/resources/joypad_axis_textures.gd
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name JoypadAxisTextures
+extends Resource
+
+@export var neutral: Texture2D
+@export var negative: Texture2D
+@export var positive: Texture2D
```
manuq, line 8: 👍

scenes/game_elements/props/hint/resources/joypad_textures.gd
```diff
@@ -0,0 +1,15 @@
...
+## Digital buttons which can be either pressed or unpressed: face, shoulders,
+## and theoretically paddles.
+@export var buttons: Dictionary[JoyButton, Texture2D]
+
+@export var dpad: DirectionalInputTextures
+@export var left_stick: DirectionalInputTextures
+@export var right_stick: DirectionalInputTextures
+
+## L2/R2 (aka LT/RT or ZL/ZR) are technically axes, but we treat them as buttons.
+@export var triggers: Dictionary[JoyAxis, Texture2D]
```
manuq, line 15: 👍

manuq (approved):
This is impressive, how you managed to overcome the corner cases. I thought this needed more work on the addon side, but I only see https://github.com/nathanhoad/godot_input_helper/pull/88 upstreamed.

Unfortunately I don't have a keyboard to test it, but I can confirm that it doesn't regress in my case. The screen captures you've posted are amazing.

I don't have anything to comment this time on the scripting side.

Issue #1275, opened 2025-10-01
Storybook: Organize quests by status


manuq:
Follow up of https://github.com/endlessm/threadbare/issues/673

Since https://github.com/endlessm/threadbare/issues/750 the quests have a completion status: work in progress, complete, and broken.

Filter out the broken ones from the book. And organize the rest into two chapters: all complete quests first, and all work-in-progress quests second.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 150 |
| Game Design: Game Feel & UX | Technical Literacy | 150 |

manuq:
Should we add bookmarks for the chapters?

manuq:
Language organization is added in https://github.com/endlessm/threadbare/issues/2548

And completion is handled in https://github.com/endlessm/threadbare/issues/1274

So closing this one.

PR #2351, opened 2026-06-11
Mark Tutorial and Void lore quests as completed

manuq:


wjt (approved):
Oops, good catch!

wjt:
I wonder if we should show a “⚠️ In Development ⚠️” banner on the LoreQuest title screen, by analogy to #1275

PR #2361, opened 2026-06-12
Phantom Camera: Filter exported files

wjt:
Phantom Camera: Filter exported files

Most of this plugin is not needed in exported builds. In particular, it
includes some example scenes that trigger warnings during export, so
excluding these feeds two birds with one scone: reduced export warnings
& smaller .pck.

Annoyingly I have learnt that EditorExportPlugin doesn't get the chance
to exclude .gd files, so scripts from the example scenes are included.
At least the .gdc files are relatively small.

The alternatives are:

1. Filter out unwanted files in each export preset. I don't like this
   because you have to configure this for each export preset separately.
   I think it is error-prone.
2. Delete the files from the source tree. This would allow us to fully
   exclude the examples, including scripts, but would mean taking
   special care when we update the plugin, and also doesn't let us
   filter out unwanted scenes.

Helps https://github.com/endlessm/threadbare/issues/2353

### DIFF
--- a/addons/phantom_camera/export_plugin.gd
+++ b/addons/phantom_camera/export_plugin.gd
@@ -1,0 +1,26 @@
+extends EditorExportPlugin
+## Excludes Phantom Camera files that are not needed in exported game
+##
+## Annoyingly EditorExportPlugin._export_file() is not called for .gd files, so
+## (for example) example scripts are included in the export.
+
+const IGNORED_PATHS := [
+	"/assets",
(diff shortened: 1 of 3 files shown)


manuq (approved):
Of the alternatives, I wonder if "2. Delete the files from the source tree." would be easier for maintenance than keeping the list of excluded files + the patch to `plugin.gd`. But you already did the work this time, so let's go with this!

PR #2365, opened 2026-06-12
Elders: Hide balloon from dialogue

wjt:
Elders: Hide balloon from dialogue

I have just noticed that the balloon script adds itself to the context
that is available to dialogue. This means there is a cleaner way to hide
the dialogue balloon before showing the storybook than what I
implemented in commit dfa9e69bc3011695f41967fd95532a9fb67bbae4: call
`balloon.hide()` directly from the dialogue, right before calling
`show_storybook()` on the Elder (which talk_behavior.gd passes as
context).

Implement this. Remove the plumbing that I added to make the balloon
node accessible from the elder script.


manuq (approved):
Oh much better!

Issue #1208, opened 2025-09-17
Support checkpoints in expanding void levels


wjt:
Suppose you have a level which has a series of islands:

<img width="1358" height="323" alt="Image" src="https://github.com/user-attachments/assets/7af3999d-0175-46e7-8bbb-57a67de81a7b" />

(Imagine that the template enemies are void-spreading enemies which chase you, but can't cross the bridges.)

You successfully escape the first one, and get to the checkpoint:

<img width="1358" height="323" alt="Image" src="https://github.com/user-attachments/assets/3c4f05b9-225e-4fd0-8dae-02a3a705c687" />

Then you get caught on the second island:

<img width="1358" height="323" alt="Image" src="https://github.com/user-attachments/assets/d201da31-a937-48f7-b662-29e5adbc5170" />

Currently, you would respawn at the checkpoint, but the void covering the first island would be gone:

<img width="1358" height="323" alt="Image" src="https://github.com/user-attachments/assets/76b4c4e5-1021-4c82-85de-7f117eeedd3e" />

This is because we only save a very small amount of state: current scene and current checkpoint.

We can work around this through level design, like we do in ink combat scenes: break the level into 3 scenes so that changing scene acts as a checkpoint. But it would be nice for this to work properly. To do so we'd probably need a more general state-saving mechanism that different elements of the scene could plug into.

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 150 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 150 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 150 |

wjt:
I think it would make sense to wait on the resolution of:

- https://github.com/endlessm/threadbare/issues/784

before tackling this, because both will involve fleshing out how state is saved and loaded.

manuq:
As the description says, the easiest way to work-around this is by splitting the chasing in 2 scenes.

wjt:
Now that we have the `facts` dictionary we could do this. But what's tricky is that we need to control exactly when the state is persisted: if you haven't made it to the checkpoint, the void that caught you shouldn't be persisted; and if you make it to the checkpoint, try the next leg of the quest, and backtrack to the checkpoint (reactivating it), again the state should not be saved.

manuq:
[...]

PR #2366, opened 2026-06-12
void: Add checkpoints to void runner levels

wjt:
In the Loom Lounge yesterday, Stephen noticed that when I was caught in
void_runner, there was no Knitwitch when I respawned.

I don't think we strictly need to have Knitwitches at the start of every
level where you might be defeated. If you are caught by the very first
guard in the musician quest stealth level, you just respawn at the start
of the level, no witch.

However, perhaps it's reassuring to have one at the start of
void_runner. Having done that, maybe we can also put one after the
tricky three-needle crossing in void_grappling_round_2, and then also at
the start of that scene for good measure.

There is no good place (other than the very start) to put one in
void_grappling round 1.


wjt:
Draft because I'm not sure this is necessary.

manuq:
I played through it and it certainly feels better! I remember we discussed about it before, for persisting the Void tiles. Having played it, I don't think that is necessary.

manuq:
I particularly like the nitwitcher as it appears after the pin.

wjt:
Unlike the situation in https://github.com/endlessm/threadbare/issues/1208 these are all placed before (not after) sections where the void chases you. So there is no need to persist tiles that the chasing-void has eaten!

PR #2369, opened 2026-06-12
Do not assume class of first node in group player

### LINKED ISSUE #2306: New dialogue balloon placement obscures "player" in quest intro/outro cutscenes
Since #2275, just looking at 2 randomly chosen storyquests:

Image
Image

Basically all storyquests have this problem because they inherit it from the template:

Image

The balloon has code that detects when the player is in the bottom ¼ of the screen and moves itself to the top of the screen. It does not work in these cutscenes because the "player" is not actually an instance of the player scene or in the `player` group - it's just an animated sprite.

For the existing quests we could try adding the AnimatedSprite2D node representing the player to the `player` group.

manuq:
So cinematics can assign "player" group to the mock player sprite.

This makes the dialogue balloon to appear at the top of the screen when the
sprite is at the bottom.

### Do not assume class of first node in group player

Certain scripts could treat player as a CharacterBody2D for physics,
as a Node2D for its position, or as a CanvasItem for visual stuff, etc.

Input HUD: It considers the player a CharacterBody2D and checks for its
components using `Object.get(COMPONENT_NAME) as COMPONENT_CLASS`. Previously it
was assuming that the output of `get_first_node_in_group()` was a
CharacterBody2D. The assignment would fail if the first node in group "player"
is of another type. This doesn't allow mock players, for example an
AnimatedSprite2D in a cinematic.

Throwing enemy: All this enemy needs is the global_position of player for
throwing. This makes it closer to the ThrowProjectileBehavior which has `@export
var target: Node2D`. Also get the player node from group "player" only once.

player_follower.gd: All this script needs is the global_position of player
for calculating the offset.

----

Fix https://github.com/endlessm/threadbare/issues/2306

### DIFF
--- a/scenes/game_elements/props/player_follower/player_follower.gd
+++ b/scenes/game_elements/props/player_follower/player_follower.gd
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 extends PathFollow2D
 
-@onready var player: Player = get_tree().get_first_node_in_group("player")
+@onready var player := get_tree().get_first_node_in_group("player") as Node2D
 
 
 func _process(_delta: float) -> void:
--- a/scenes/game_elements/props/spawn_point/components/spawn_point.gd
+++ b/scenes/game_elements/props/spawn_point/components/spawn_point.gd
@@ -26,7 +26,7 @@ func _ready() -> void:
 
 
 func move_player_to_self_position(smooth_camera: bool = false) -> void:
-	var player: Node2D = get_tree().get_first_node_in_group("player")
+	var player := get_tree().get_first_node_in_group("player") as Node2D
 
 	if is_instance_valid(player):
 		player.teleport_to(self.global_position, smooth_camera, look_at_side_on_spawn)
--- a/scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
+++ b/scenes/game_elements/characters/enemies/throwing_enemy/components/throwing_enemy.gd
@@ -129,6 +129,7 @@ var allowed_labels: Array[String] = ["???"]
 ## color-matching game.
 var color_per_label: Dictionary[String, Color]
 
+var _player: Node2D
 var _initial_position: Vector2
 var _target_position: Vector2
 var _is_attacking: bool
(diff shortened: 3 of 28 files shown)


manuq:
I only made the change in Renya intro for testing. But if we agree on this solution, at least for the existing cinematics, then I can update them all. Link to watch the test scene:

https://play.threadbare.game/branches/endlessm/improve-getting-player-node/#quests/story_quests/renya_beyond_sorrow/0_intro/renya_intro

manuq:
I updated all the cinematic scenes now.

scenes/game_elements/props/player_follower/player_follower.gd
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 extends PathFollow2D
 
-@onready var player: Player = get_tree().get_first_node_in_group("player")
+@onready var player := get_tree().get_first_node_in_group("player") as Node2D
```
wjt, line 5: This makes sense; it also reveals that this script is much more generic than its name suggests!

scenes/game_elements/props/spawn_point/components/spawn_point.gd
```diff
@@ -26,7 +26,7 @@ func _ready() -> void:
 
 
 func move_player_to_self_position(smooth_camera: bool = false) -> void:
-	var player: Node2D = get_tree().get_first_node_in_group("player")
+	var player := get_tree().get_first_node_in_group("player") as Node2D
 
 	if is_instance_valid(player):
 		player.teleport_to(self.global_position, smooth_camera, look_at_side_on_spawn)
```
wjt, line 32: I was about to complain that this is wrong because `Node2D` does not define `teleport_to()` (that's on `player.gd`) but the existing code was equivalent!

wjt (approved):

scenes/game_elements/props/spawn_point/components/spawn_point.gd
manuq: You are right. I think the complaint is valid for the existing code. teleport_to() is defined in Player.

PR #2373, opened 2026-06-12
Add health component

### LINKED ISSUE #1747: Implement character health
Implement a health component for the player. Health is lost when the player is hit by a projectile. When health reaches zero, the `Player.defeat()` method should be called.

It should be possible to use the health component from `fragile_barrel.gd`, that was factored out in https://github.com/endlessm/threadbare/pull/2373, for the player.

### XP Summary (total 325)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 100 |

Litraxx:
Added a generic health component that can be reused in different scenes.
It exports a max health property and provides various properties regarding health status.
It also provides two signals
- `health_depleted` emitted when a change to the current health pushes it to zero or below
- `health_changed` emitted when the health is changed and the change does not cause the health to deplete; emits the current health as an argument

Refactored the fragile barrel to use the new health component

Resolves https://github.com/endlessm/threadbare/issues/1747

### DIFF
--- a/scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
+++ b/scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
@@ -30,10 +30,9 @@ func hit_by_droplet(droplet_label: String) -> void:
 		health_component.damage(1)
 
 
-func take_damage(_current_health: int, has_depleted_health: bool) -> void:
-	if not has_depleted_health:
-		crack_sound.play()
-		update_cracks()
+func take_damage() -> void:
+	crack_sound.play()
+	update_cracks()
 
 
 func update_cracks() -> void:
--- a/scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
+++ b/scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
@@ -42,7 +42,7 @@ func update_cracks() -> void:
 	crack_overlay_node.frame = frame_index
 
 
-func break_barrel() -> void:
+func _on_health_component_health_depleted() -> void:
 	crack_overlay_node.visible = false
 
 	shatter_sound.play()
--- a/scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
+++ b/scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
@@ -7,50 +7,39 @@ extends FillingBarrel
 ## Emitted when [member current_health] reaches 0.
 signal barrel_destroyed(barrel_instance: FragileBarrel)
 
-## Maximum hits the barrel can take before breaking.
-@export_range(1, 100) var max_health: int = 4
-
-var current_health: int
-
 @onready var crack_overlay_node: AnimatedSprite2D = %CrackOverlay
 @onready var crack_sound: AudioStreamPlayer2D = %CrackSound
 @onready var shatter_sound: AudioStreamPlayer2D = %ShatterSound
+@onready var health_component: HealthComponent = %HealthComponent
 
 
 func _ready() -> void:
 	super._ready()
-	current_health = max_health
 	crack_overlay_node.visible = false
 
 
 # Logic called by Projectile when it hits this object
 func hit_by_droplet(droplet_label: String) -> void:
 	# Ignore if already full or destroyed
-	if _amount >= needed_amount or current_health <= 0:
+	if _amount >= needed_amount or health_component.has_depleted_health:
 		return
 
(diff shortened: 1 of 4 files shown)


scenes/game_logic/health_component.gd
```diff
@@ -0,0 +1,40 @@
...
+
+signal health_depleted
+signal health_changed(current_health: int)
+
+@export_range(1, 100) var max_health: int = 4
+
+var current_health: int = max_health:
+	set(value):
+		if value == current_health:
+			pass
```
wjt: I think you meant this:


```suggestion
		if value == current_health:
			return
```

`pass` means "do nothing and continue".

scenes/game_logic/health_component.gd
```diff
@@ -0,0 +1,40 @@
...
+		else:
+			health_changed.emit(current_health)
+
+var damage_taken: int:
+	get:
+		return max_health - current_health
+
+var damage_taken_percentage: float:
+	get:
+		return float(damage_taken) / max_health
```
wjt: This value isn't a percentage (value between 0 and 100). It is a value between 0.0 and 1.0. The pattern used in Godot APIs (such as `PathFollow2D`) is to call this a "ratio".


```suggestion
var damage_taken_ratio: float:
	get:
		return float(damage_taken) / max_health
```

scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
```diff
@@ -7,50 +7,38 @@ extends FillingBarrel
...
 		return
 
 	if droplet_label == self.label:
 		increment()
 	else:
-		take_damage()
+		health_component.damage(1)
 
 
 func take_damage() -> void:
```
wjt: I think this method should be renamed. It no longer implements "taking damage". Instead it responds to the barrel's health changing.

scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
```diff
@@ -7,50 +7,38 @@ extends FillingBarrel
 ## Emitted when [member current_health] reaches 0.
```
wjt: The `current_health` member no longer exists. Please update this docstring.

scenes/game_logic/health_component.gd
```diff
@@ -0,0 +1,40 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name HealthComponent
+extends Node2D
+
+signal health_depleted
+signal health_changed(current_health: int)
```
wjt, line 11: Please document when these are emitted. In particular `health_changed` is not emitted every time `current_health` changes: if the change causes health to fall below `0`, then `health_depleted` is emitted instead.

scenes/game_logic/health_component.gd
wjt: Actually it's not strictly a value between `0.0` and `1.0` because if `current_health` is `1` and I call `damage(2)` then `current_health` will be `-1`. I think `current_health` should be clamped (in its setter) so that `0 <= current_health <= max_health`.

wjt (changes requested):
Great start, thank you for working on this!

scenes/game_logic/health_component.gd
Litraxx: Changed to `return`

scenes/game_logic/health_component.gd
Litraxx: Mathematically speaking this would be a correct representation, changed it to ratio to be consistent with godot conventions. Also added a comment to `current_health` that reflects its range

scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
Litraxx: Changed to the default generated name when connecting signal

scenes/game_elements/props/filling_barrel/components/fragile_barrel.gd
Litraxx: Update to refer to barrel's health instead

scenes/game_logic/health_component.gd
Litraxx: Documented behavior of signals

wjt (approved):
Thanks! I added a couple of tiny changes for things I only spotted while re-reviewing.

Issue #2339, opened 2026-06-10
StoryBook: Improve navigation


manuq:
Make it similar to the Pause overlay:
- Use a TabContainer and make each page a tab. Use "previous_tab", "next_tab" input actions for navigating the pages. Althogh this one should have "tabs visible" disabled.

Done:
- Have a row with input hints at the bottom.
- Remove the custom arrow buttons from the sides.

manuq:
Here is useful information about font sizes in the StoryBook: https://github.com/endlessm/threadbare/pull/2343#discussion_r3394734427 . To consider while tweaking its user interface.

PR #2375, opened 2026-06-13
Update StoryBook layout

manuq:
Make the StoryBook scene a CanvasLayer.

Similar to the pause overlay. Use the same ColorRect for the background
transparency instead of a Panel node.

Replace left/right buttons with input helpers.

Add a bottom row similar to the pause overlay. Use next-tab and previous-tab
input actions for switching pages. Change the _input() function to
_unhandled_input().

Move all inside PanelContainer, which have hardcoded positions, 64 pixels
vertically. To adjust to the new row.

Helps https://github.com/endlessm/threadbare/issues/2339


manuq:
I couldn't make it use a TabContainer yet, but this is a step toward that.

scenes/menus/storybook/storybook.tscn
```diff
@@ -8,119 +8,101 @@
...
 [ext_resource type="Resource" uid="uid://tphgv15oanfs" path="res://scenes/quests/story_quests/el_ojo_revelador/quest.tres" id="5_x5pjn"]
 [ext_resource type="Resource" uid="uid://ddxn14xw66ud8" path="res://scenes/quests/template_quests/NO_EDIT/quest.tres" id="5_ywp77"]
-[ext_resource type="Texture2D" uid="uid://b6qlxbgr1eq5v" path="res://scenes/menus/storybook/Textures/Arrow_Button_Left_Pressed.png" id="14_jl72s"]
-[ext_resource type="Texture2D" uid="uid://beg3lp7xpkpf2" path="res://scenes/menus/storybook/Textures/Arrow_Button_Left_Normal.png" id="14_m2ghf"]
-[ext_resource type="Texture2D" uid="uid://qs4jy8b1curv" path="res://scenes/menus/storybook/Textures/Arrow_Button_Left_Hover.png" id="15_jl72s"]
-[ext_resource type="Texture2D" uid="uid://c7ey8lnnb4ugg" path="res://scenes/menus/storybook/Textures/Arrow_Button_Right_Normal.png" id="17_m2ghf"]
-[ext_resource type="Texture2D" uid="uid://c24put1gr05cu" path="res://scenes/menus/storybook/Textures/Arrow_Button_Right_Hover.png" id="18_jl72s"]
-[ext_resource type="Texture2D" uid="uid://bgcp50dpx5gd3" path="res://scenes/menus/storybook/Textures/Arrow_Button_Right_Pressed.png" id="18_m2ghf"]
```
wjt, line 16: Do we want to also remove these now-unused arrow textures?

wjt (approved):
Nice improvement!

scenes/menus/storybook/storybook.tscn
manuq: Yes indeed. I'll remove them before I continue.

PR #2388, opened 2026-06-16
Move inventory to QuestState

wjt:
Move inventory to QuestState

Previously the inventory was stored in GlobalState, but was cleared when
starting or abandoning a quest, because the only collectible items are
the threads of Memory, Imagination, and Spirit that are found during
quests.

Move the inventory to the QuestState. Represent it as a separate
resource. I anticipate that we may want a global inventory as well in
future, with items moving from the QuestState to the GlobalState. But I
also think it makes the code easier to read in both places.


manuq (approved):
Nice cleanup. Yes it makes more sense that the current threads are part of the quest state.

Issue #2281, opened 2026-06-03
Persist progress within quests


wjt:
If the player abandons a quest, they should be able to return to it later and pick up where they left off. The persisted state here should include:

- Threads collected so far
- Current scene
- Current spawn point
- etc.

In effect: the PerSceneState + QuestState at the point where it was abandoned.

wjt:
Started looking at this.

Design question: Do we want to allow the player to start over from the beginning of a quest? Or will we always pick up the quest where they left off?

If we want to give them the choice, where should this choice be presented? For StoryQuests it's obvious: when they press "Play" in the storybook. For LoreQuests it's not obvious. It could be in the Quest Teleporter, when they cross the threshold; or it could be in the quest title screen.

wjt:
More design questions as to where you should resume from:

1. The exact scene + spawn point where you abandoned?
2. The start of that scene?
3. The start of the "challenge" (i.e. just after the last thread you collected)?

Take the Void quest for example. Imagine there was a KnitWitch here in the last round:

<img width="1366" height="610" alt="Image" src="https://github.com/user-attachments/assets/2bdfd720-a821-474e-9bc3-c4a06adc02c9" />

Imagine you had activated that knitwitch, then abandoned. In this case the 3 options are:

1. At the knitwitch
2. At the start of the level (where the player is in the screenshot)
3. At the start of the previous scene (below)

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/5c0b24b3-4acd-4699-bed6-19c79e97e03f" />

manuq:
> Started looking at this.
> 
> Design question: Do we want to allow the player to start over from the beginning of a quest? Or will we always pick up the quest where they left off?
> 
> If we want to give them the choice, where should this choice be presented? For StoryQuests it's obvious: when they press "Play" in the storybook. For LoreQuests it's not obvious. It could be in the Quest Teleporter, when they cross the threshold; or it could be in the quest title screen.

Yes I think for LoreQuests the title screen is the place for a "Restart" or "Start from the beginning". With a confirm message of losing progress, maybe?

And for StoryQuests, the StoryBook could have it next to the "Play" button.

For LoreQuests, I personally would like to display the "Play" button as well in a menu. Have it focused and automatically pressed after 5 seconds or so. If the player use input for navigating the menu (except mouse movement) then the timer is cancelled.

That way both kind of quests could have:
- Play/Resume (default, text depending on quest ever started or completed)
[...]

PR #2389, opened 2026-06-16
Quest Separator: Add Play & Back buttons

wjt:
This screen previously automatically advanced to the first scene of the
quest after a few seconds, unconditionally.

Increase that timeout slightly. Add a “Play” button. Show a progress bar
over the top of it that follows the timer. Stop the timer on any input
event except a mouse movement.

Add a “Back” button which abandons the quest before it has even started.

Add some margins to the scene as a whole.

Ignore mouse input in the TextureRect which covers the whole scene, so
that the buttons can be clicked.

https://github.com/endlessm/threadbare/issues/2281#issuecomment-4692428165

### DIFF
--- a/scenes/globals/pause/pause_overlay.gd
+++ b/scenes/globals/pause/pause_overlay.gd
@@ -47,6 +47,13 @@ func toggle_pause() -> void:
 
 func _on_abandon_quest_pressed() -> void:
 	toggle_pause()
+	abandon_quest()
+
+
+func abandon_quest() -> void:
+	if not GameState.quest:
+		push_warning("No quest to abandon")
+		return
 
 	var quest := GameState.quest.quest
 	var abandon_scene := GameState.quest.abandon_scene_path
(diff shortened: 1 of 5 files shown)


wjt:
@manuq I think this is roughly what you had in mind?

(Still to do is to switch this to "Continue" / "Restart". And should there be a "Back" to abandon the quest before it has even started?)

manuq:
> @manuq I think this is roughly what you had in mind?
> 
> (Still to do is to switch this to "Continue" / "Restart". And should there be a "Back" to abandon the quest before it has even started?)

Works great! The button is not clickable but I can use the Interact input action.

https://github.com/user-attachments/assets/76281db8-4f6f-4b61-99c5-c8c9052d70cb

wjt:
> Works great! The button is not clickable but I can use the Interact input action.

Good catch! It's the TextureRect vignette sitting atop the whole scene swallowing mouse input. I fixed that in the latest revision which also adds a back button and puts the progress bar on top of the Back button. (It looks a bit ugly still but it's more like what I'm going for: "button that will press itself" is a familiar pattern on mobile.)

manuq:
Great! It can't be uglier than the quest separator that I prototyped. This is adding useful buttons to it.

PR #2390, opened 2026-06-16
Pause Overlay: Don't handle next/previous_tab while unpaused

wjt:
Change the process mode for the whole Control hierarchy of the Pause
menu to "WHEN_PAUSED", which was previously set on what is now just one
page of it.

Move the handling of the next_tab and previous_tab events into _input()
in a script attached to the TabContainer, rather than in
_unhandled_input() on the PauseOverlay itself.

This fixes an issue I was seeing where pressing A/D/X/Z (the keys bound
to these two actions) in a menu I am making for the quest separator
would cause focus to be taken away from all visible controls.

### DIFF
--- a/scenes/globals/pause/pause_overlay.gd
+++ b/scenes/globals/pause/pause_overlay.gd
@@ -22,16 +22,6 @@ func _unhandled_input(event: InputEvent) -> void:
 	if event.is_action_pressed(&"pause"):
 		toggle_pause()
 		get_viewport().set_input_as_handled()
-	elif event.is_action_pressed("next_tab"):
-		tab_container.current_tab = wrapi(
-			tab_container.current_tab + 1, 0, tab_container.get_tab_count()
-		)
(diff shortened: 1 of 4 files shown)


manuq (approved):
Ouch I missed this, thanks!

wjt:
I couldn't find any tool in the editor's debug panel to see what was happening to the GUI focus but it was easy enough to drop this line into an appropriate place:

```gdscript
print(get_viewport().gui_get_focus_owner())

manuq:
> I couldn't find any tool in the editor's debug panel to see what was happening to the GUI focus but it was easy enough to drop this line into an appropriate place:
> 
> ```gdscript
> print(get_viewport().gui_get_focus_owner())
> ```

Ha I did that too while implementing the new tabs. In a `_process()` it was a bit too much output but it really helped. I was expecting something in the Remote dock.

PR #2394, opened 2026-06-17
After the Tremor: Remove now-incorrect control hints from dialogue

wjt:
After the Tremor: Remove now-incorrect control hints from dialogue

This predates the change to the control scheme. Space no longer throws
the grappling hook.

I don't think we need to say what the buttons are in dialogue because
it's shown on-screen.


manuq (approved):
Nice! Your Spanish gets better and better!

PR #2404, opened 2026-06-18
Bind skipping dialogue to space/Xbox A

wjt:
Previously Tab (or Xbox B) skips typing-out of the current line of
dialogue and shows it all at once. I chose a different key to the "next
line"/"accept choice" button (Space / A) because I didn't want people to
skip dialogue by accident.

But watching people at Serious Play testing the game (as I am doing
now), everyone expects to be able to use the same key that advances to
the next line of dialogue, to skip the typing out of the dialogue.

Change the binding.

Credit the playtester who specifically suggested this!

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -172,8 +172,9 @@ dialogue_next={
 }
 dialogue_skip={
 "deadzone": 0.5,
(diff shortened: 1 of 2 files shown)


wjt:
I haven't tested the gamepad binding because I do not have a gamepad with me.

manuq (approved):
Yes, I have verified with a gamepad and it works. This is what I would expect too for skipping.

PR #2405, opened 2026-06-18
Sign post: Add unshaded material to label

manuq:
This label is not supposed to be part of the world. Add a new CanvasItemMaterial to the PanelContainer and set it to light mode: unshaded.


manuq:
This is subtle but is more evident now that sign posts have a light source. Before:
Screenshot From 2026-06-18 19-05-48

After:
Screenshot From 2026-06-18 19-05-53

wjt (approved):
I have a lot to learn about how lighting works!

PR #2412, opened 2026-06-20
Retelling: Show spools in responses

wjt:
Previously the text "(Memory available!)" was shown next to the
retelling dialogue branch that related to the thread of Memory (resp.
other threads). The text was not shown if no Memory thread had been
collected.

Instead, show the icon for the thread of Memory. Use Dialogue Manager's
tags feature to get this information from the dialogue file through to
the balloon. Add a script to the response button that has a `response`
property, which causes Dialogue Manager to assign the DialogueResponse
resource to this property rather than setting a `text` property.

Position the icon to the right of the text. Configure the button to
shrink to the centre so that the icon is aligned to the right edge of
the label rather than all 3 icons being right-justified.

It is not possible (as far as I can tell) to conditionally apply the
tag. Unconditionally apply the tags. We could in future consider
disabling the retelling branches for threads the player did not collect,
but currently all the threads are mandatory.

Fix the `has_imagination()` check in two places to be `has_spirit()` as
intended, and move “Tell us a story about your last adventure!” to the
right place in the Void retelling.

### DIFF
--- a/scenes/quests/lore_quests/quest_001/quest_001_retelling.dialogue
+++ b/scenes/quests/lore_quests/quest_001/quest_001_retelling.dialogue
@@ -3,9 +3,9 @@
 ~ start
 Townies: Tell us a story about your last adventure!
 StoryWeaver: I went on a quest with the musician and I...
-- Learned a forgotten song {{ memory_text() }} => retell_memory
-- Slipped past the guards {{ imagination_text() }} => retell_imagination
-- Fought off some InkDrinkers {{ spirit_text() }} => retell_spirit
+- Learned a forgotten song [#thread=memory] => retell_memory
+- Slipped past the guards [#thread=imagination] => retell_imagination
+- Fought off some InkDrinkers [#thread=spirit] => retell_spirit
 
 ~ retell_memory
 StoryWeaver: Let me whistle it to you...
@@ -23,7 +23,7 @@ if has_imagination()
 
 ~ retell_spirit
 StoryWeaver: Because they had stolen the ink!
-if has_imagination()
+if has_spirit()
 	do give_spirit_upgrade()
 	Townie: Ha! I feel brave now!
 => END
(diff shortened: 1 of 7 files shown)


wjt:
Screenshot From 2026-06-20 13-23-02

manuq:
I like it! Is much more clear with the spool icon.

wjt:
What do you think about disabling the dialogue choices for spools that the player didn't collect?

manuq:
> What do you think about disabling the dialogue choices for spools that the player didn't collect?

I think the retelling should happen regardless of the spools collected.

But also I think this really depends on the spools being mandatory or not. Technically the quests can have zero to 6 threads upon completion. Current lore quests are always finished with 3, one of each kind (memory, imagination, spirit). And they are mandatory because only collecting them moves the quest forward. So currently for lore quests, all 3 dialogues will be shown.

StoryQuests were designed with 3 threads in mind for a 3-act narrative.

Also in lore, the goal of the StoryWeaver is to collect the magical threads and return them to the Eternal Loom, doing that the world is rewove.

Thinking about this now, maybe reusing the magical threads for this is a bit confusing. Proposal: what about adding a separate, optional item that can be collected and is exclusive of the townies helper system? Then the retelling is always about memory, imagination, spirit (3 dialogue branches). But if the player has the item, the picked option touches a townie and becomes the next helper of that kind.

@JoniCeceri

wjt:
image

scenes/quests/lore_quests/quest_001/quest_001_retelling.dialogue
```diff
@@ -23,7 +23,7 @@ if has_imagination()
 
 ~ retell_spirit
 StoryWeaver: Because they had stolen the ink!
-if has_imagination()
+if has_spirit()
```
manuq, line 26: Ouch! Thanks for this fix.

scenes/ui_elements/dialogue/balloon.tscn
```diff
@@ -123,8 +125,12 @@ response_template = NodePath("ResponseExample")
 
 [node name="ResponseExample" type="Button" parent="Balloon/BalloonContainer/MarginContainer/MarginContainer/PanelContainer/MarginContainer/VBoxContainer/ResponsesMenu" unique_id=957333658]
 layout_mode = 2
+size_flags_horizontal = 4
 theme_type_variation = &"FlatButton"
 text = "Response example"
+icon = ExtResource("5_oh0dh")
+icon_alignment = 2
+script = ExtResource("5_u3aeq")
```
manuq, line 133: 👍

manuq (approved):

PR #2413, opened 2026-06-20
tutorial: Add retelling

wjt:
The branches do not quite match where the threads were discovered, but
it is better than nothing.


manuq (approved):
Nice! Yes certainly the tutorial can also show the first retelling. Well written dialogue.

PR #2421, opened 2026-06-22
credits: Add playtesters from Serious Play in Mainz

wjt:
credits: Add playtesters from Serious Play in Mainz

We had Threadbare in an arcade cabinet at this event last week. I took
the opportunity to watch a number of people playing the game and made
extensive notes.

Thanks to all these individuals whose names I noted for their time spent
playing the game and discussing it with me. (I believe I checked with
each person named here that they were happy to be credited.) Thanks also
to those whose names I did not record, and to those who declined to be
credited.


manuq (approved):
Wonderful!

PR #2438, opened 2026-06-23
Don't translate placeholder text & node names for hidden tabs

### LINKED ISSUE #2437: Node names extracted for translation due to being (hidden) tab titles
### Threadbare version

_No response_

### Affected scene

scenes/menus/options/options.tscn

### Description

When generating the translation template (which we commit as `locale/threadbare.pot`) some nodes' names are extracted for translation. For example:

```
#: scenes/menus/options/options.tscn
msgid "SoundSettings"
msgstr ""
```

```
#: scenes/menus/debug/debug_settings.tscn
msgid "CompletedQuests"
msgstr ""
```

I believe the reason that these are extracted for translation is that they are the titles of tabs.

Image

But we don't actually show the tab bar on these controls!

### Checklist

wjt:
Don't translate placeholder text & node names for hidden tabs

Previously, the placeholder “0123456789” text used in the version label
on the title and pause screens was extracted for translation. Set the
translation mode so that this does not happen.

In TabContainer, the name of each child node is used as the tab title,
and extracted for translation even if the tabs are hidden. Disable
translations on the TabContainer and re-enable translations on each
child.

Add a couple more scenes to those that are to be translated. In the
quest separator, don't auto-translate the placeholder title and
description.

Regenerate threadbare.pot, and update the the existing translations with:

    msgmerge -U locale/es.po locale/threadbare.pot
    msgmerge -U locale/fr.po locale/threadbare.pot

Resolves https://github.com/endlessm/threadbare/issues/2437


manuq (approved):
Ouch my bad. Thanks!

wjt:
I noticed it a while ago and simply couldn't figure it out… then today inspiration struck!

PR #2439, opened 2026-06-23
Persist abandoned quests to resume later

### LINKED ISSUE #2281: Persist progress within quests

wjt:
Capture the QuestState and PerSceneState at the point where the quest is
being abandoned. Restore them later. For LoreQuests, add any new
abilities the player picked up on other quests in the meantime.

Add a "restart" button to the quest separator scene (for LoreQuests) and
storybook (for StoryQuests) that erases the suspended state rather than
restoring it.

Eliminate the special case for NO_EDIT/quest.tres. The quest duplicator
already replaces the title and description with whatever was entered
into the dialogue. This also means that StorybookPage can be given the
real Quest resource, which it needs in order to know whether it is in
progress.

Resolves https://github.com/endlessm/threadbare/issues/2281


scenes/globals/game_state/suspended_quest_state.gd
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name SuspendedQuestState
```
wjt, line 3: I'm not wild about this name. Maybe I should name it `AbandonedQuestState` like the verb shown to the player and used in the internal APIs?

scenes/globals/game_state/game_state.gd
```diff
@@ -88,6 +88,37 @@ func _ready() -> void:
...
+	var suspended_quest := global.suspended_quests[key]
+	global.suspended_quests.erase(key)
+
+	quest = suspended_quest.quest
+	scene = suspended_quest.scene
+
+	if quest.quest is LoreQuest:
+		# Add any lore abilities that the player gained since suspending this
+		# quest.
+		quest.player.abilities |= global.player.abilities
```
wjt, line 113: This API is a bit weird: you have to first call `start_quest()` which sets up fresh `QuestState` and `PerSceneState`, and then later call `restore_quest` to restore them.

It's done like this because the quest separator is just a normal scene, and so we need some way to pass the selected quest through to it: and the hammer we have for that is the `QuestState`.

manuq: Maybe `start_quest()` can be renamed to `setup_quest()` or even `set_quest()`? The documentation string says: "Sets [member quest], setting up a new [PlayerState] if necessary."

scenes/globals/game_state/game_state.gd
```diff
@@ -152,7 +183,18 @@ func mark_quest_completed() -> void:
 
 
 ## Abandon the current [member quest] without marking it as completed.
-func abandon_quest() -> void:
+## If [param suspend] is [code]true[/code], progress in the quest will be saved
+## so that it can be resumed later.
+func abandon_quest(suspend: bool = true) -> void:
```
manuq, line 181: In which cases would this be called without suspending? Oh I see in quest separator `PauseOverlay.abandon_quest(false)`. I guess this is only to skip a redundant save.

scenes/menus/quest_separator/components/quest_separator.gd
```diff
@@ -34,14 +45,35 @@ func stop_timer() -> void:
...
+
+	GameState.restore_quest()
+	# Now GameState.scene has been restored, jump to the scene & spawn point it
+	# describes
+	SceneSwitcher.change_to_file_with_transition(
+		GameState.scene.path,
+		GameState.scene.spawn_point,
+		Transition.Effect.RADIAL,
+		Transition.Effect.FADE
+	)
```
manuq, line 66: 👍

scenes/globals/game_state/suspended_quest_state.gd
manuq: I'm personally happy with it. Looking at this from another side I was considering "Attempted" or "Initiated". But "Suspended" is better.

manuq (approved):
This works well as specified in the ticket.

I'm not really sure about adding the quest separator to StoryQuests. Mainly because we have the StoryBook already for showing all the information. And the info is displayed twice in a row. The "Continue" button can be in the book too, similar to the existing Play but without the autoplay in the book case.

scenes/globals/game_state/game_state.gd
wjt: Indeed. It's a bit awkward but I couldn't find a better way to structure it when the separator is a scene of its own.

wjt:
Good idea, I'll move the "continue/restart" to the storybook.

wjt:
I reworked this to put the "restart / continue" buttons in the Storybook. In doing so I realised we can remove the special handling of `NO_EDIT/quest.tres`.

I also renamed `start_quest` to `set_quest`.

manuq:
Nice I played half of "Renya" and was able to continue directly from the StoryBook!
image

scenes/game_elements/characters/npcs/elder/components/elder.gd
```diff
@@ -74,7 +77,12 @@ func show_storybook() -> void:
 	storybook.quests = _quests
 
 	_storybook_layer.add_child(storybook)
-	chosen_quest = await storybook.selected
+
+	# GDScript doesn't support "x, y = await ...". Alas!
+	var args: Array = await storybook.selected
+	chosen_quest = args[0]
+	restart = args[1]
```
manuq, line 84: 👍 the other option is to use a Dictionary with named arguments.

scenes/globals/game_state/game_state.gd
```diff
@@ -88,10 +88,34 @@ func _ready() -> void:
...
+
+	var suspended_quest := global.suspended_quests[key]
+	global.suspended_quests.erase(key)
+
+	quest = suspended_quest.quest
+	scene = suspended_quest.scene
+
+	if quest.quest is LoreQuest:
+		# Add any lore abilities that the player gained since suspending this
+		# quest.
```
manuq, line 112: Is this comment new? I like it!

manuq (approved):

scenes/game_elements/characters/npcs/elder/components/elder.gd
wjt: When you `await` a signal which has more than one argument, you get back an untyped `Array`.

True, we could change the signal signature to:

```gdscript
signal selected(Dictionary[String, Variant])
```

and have its keys be `quest` and `restart`. But I think this is clearer.

PR #2441, opened 2026-06-23
Ink combat round 3: remove ink follows player

### LINKED ISSUE #2440: Ink combat round 3: explain ink follows player
The ThrowingEnemy instances in scene `res://scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn` have "Follows Player" enabled (under Projectile in the Inspector). This makes the ink blobs to not go in a straight line, making the game hard to target and thus unfair. Also there is no indication that this changes from round 2 to 3. Alternatively, we should explain that now ink behaves differently, at least with dialogue.

### XP Summary (total 350)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Game Design: Narrative & Storytelling | Problem Solving | 50 |
| Game Design: Narrative & Storytelling | Creativity & Design Thinking | 75 |
| Game Design: Narrative & Storytelling | Technical Literacy | 50 |

manuq:
This makes the ink blobs to go in a straight line, as expected.

Fix https://github.com/endlessm/threadbare/issues/2440

### DIFF
--- a/scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
+++ b/scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
@@ -135,7 +135,6 @@ attack_sound_stream = ExtResource("11_7kldn")
 projectile_scene = ExtResource("12_ms350")
 projectile_speed = 20.0
 projectile_duration = 10.0
-projectile_follows_player = true
 walking_range = 0.0
 walking_speed = 20.0
 
@@ -149,7 +148,6 @@ idle_sound_stream = ExtResource("10_uyhkg")
 attack_sound_stream = ExtResource("11_7kldn")
 projectile_scene = ExtResource("12_ms350")
 projectile_duration = 10.0
-projectile_follows_player = true
 walking_range = 0.0
 walking_speed = 20.0
 


felixwalberg:
I noticed that the `projectile_follows_player` is enabled for the ink combat in the song_sanctuary_3 folder. Is this a change that should be extended to those levels, or does it only disrupt the flow of round 3 in this quest?

felixwalberg:
> I noticed that the `projectile_follows_player` is enabled for the ink combat in the song_sanctuary_3 folder. Is this a change that should be extended to those levels, or does it only disrupt the flow of round 3 in this quest?

I will say that none of these levels have Filling Barrels hidden behind walls which is part of the reason the round 3 level is so tricky!

scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
```diff
@@ -135,7 +135,6 @@ attack_sound_stream = ExtResource("11_7kldn")
 projectile_scene = ExtResource("12_ms350")
 projectile_speed = 20.0
```
wjt, line 136: I had never noticed before that the two enemies have different `projectile_speed` configured!

wjt: I wonder if this was intentional.

wjt (commented):
I'm on the fence about this. It makes the scene a bit easier - you can simply hold down "repel" without moving and the level will be solved. At least before in my experience you had to move to a different location and stand still there for the level to solve itself!

felixwalberg:
> At least before in my experience you had to move to a different location and stand still there for the level to solve itself!

I am noticing that too, particularly with the blue Filling Barrel behind the walls. Since the projectiles lose so much momentum by readjusting to the player's position, it is difficult to get a lucky bounce up there. I was still able to achieve it by standing still, but it seems to take longer, and you end up with a screen full of ink blobs before you get lucky enough.

I feel like part of what makes these levels a little tricky for me is the rate at which you are dealing with new (and different colored) ink blobs. It is hard to focus on just one when you are constantly getting hit by others, and I appreciate that aspect of the challenge. 

That is all to say that I didn't feel like whether or not the ink blobs tracked me was make-or-break for how challenging the level is.

scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
manuq: Oh I haven't noticed this before!

manuq:
@felixwalberg @wjt thanks for the feedback! Then the alternative solution is to communicate to the player that the ink behaves differently in this particular level. I guess the easiest is with a Cinematic node to start with dialogue, like in round 1.

PR #2448, opened 2026-06-23
Warning on title screen if the game is opened on a touchscreen device

### LINKED ISSUE #2358: Show a warning on the title screen if device has no hardware keyboard or gamepad
Threadbare is not currently playable with a touchscreen. (#806)

I suggest we could show a note/warning to that effect on the title screen if we can detect that the player has neither a keyboard nor a gamepad. Specifically we could check some combination of the following:

- ~~If `DisplayServer.has_hardware_keyboard()` returns `true`.~~ Unfortunately this is only implemented for native iOS and Android builds.
- If the `web_android` or `web_ios` [feature tag](
https://docs.godotengine.org/en/stable/tutorials/export/feature_tags.html) is present. It is of course possible to connect a hardware keyboard to an Android or iOS device but we can probably assume they aren't.
- If `DisplayServer.is_touchscreen_available()` is `true`.
- If `Input.get_connected_joypads()` is empty.

Based on some combination of these, show a message on the title screen to the effect of:

> You need a gamepad or a keyboard to play Threadbare. You can't play Threadbare with just a touchscreen.

Make it not block the player entering the game in case we have detected wrong.

### XP Summary (total 500)
| Subdomain | Skill | XP |
| - | - | - |

alcole2:
Threadbare is not playable on a mobile device that lacks a keyboard and mouse/trackpad. Detect this case, and show a warning on the title screen if so.

Resolves https://github.com/endlessm/threadbare/issues/2358


wjt:
See https://github.com/endlessm/threadbare/wiki/Contributing#pull-request-descriptions for some notes on pull request descriptions. In particular you should include the issue on a line of its own at the end, in the form:

```
Resolves https://github.com/endlessm/threadbare/issues/2358
```

wjt:
If you set up `pre-commit` on your computer https://github.com/endlessm/threadbare/wiki/Contributing#coding-style then the coding style checks can be run on your computer, and many problems fixed automatically!

scenes/menus/title/components/title_screen.gd
```diff
@@ -15,6 +19,10 @@ func _ready() -> void:
 			_on_main_menu_continue_pressed()
 		else:
 			_on_start_pressed()
+			
+	$TouchscreenWarning/VBoxContainer/DismissButton.pressed.connect(_on_dismiss_pressed)
```
wjt: I suggest connecting this signal in the editor instead:

Image

scenes/menus/title/components/title_screen.gd
```diff
@@ -15,6 +19,10 @@ func _ready() -> void:
...
+	if _should_show_touchscreen_warning():
+		touchscreen_warning.show()
```
wjt: You either need to hide the node in the scene editor; or do something like this instead:


```suggestion
	touchscreen_warning.visible = _should_show_touchscreen_warning()
```

alcole2:
@wjt Great! I'll do that. Will I need to redo the pull request?

scenes/menus/title/components/title_screen.gd
```diff
@@ -57,3 +65,24 @@ func _on_credits_back() -> void:
...
+		return false
+
+	var has_touch: bool = DisplayServer.is_touchscreen_available()
+	
+	var has_keyboard: bool = false
+	
+	if OS.has_feature("web_android") or OS.has_feature("web_ios"):
+		has_keyboard = false
+	else:
+		has_keyboard = DisplayServer.has_hardware_keyboard()
```
wjt: This method is documented to not return the correct result on any platform other than `ios` or `android`:

> Note: This method is implemented on Android and iOS. On other platforms, this method always returns true.

scenes/menus/title/components/title_screen.gd
```diff
@@ -57,3 +65,24 @@ func _on_credits_back() -> void:
...
+
+	if has_touch and not has_keyboard:
+		return true
+		
+	return false
```
wjt: I think you could simplify this logic:

```suggestion
func _should_show_touchscreen_warning() -> bool:
	var has_touch: bool = DisplayServer.is_touchscreen_available()
	var has_keyboard: bool
	if OS.has_feature("web_android") or OS.has_feature("web_ios"):
		# DisplayServer.has_hardware_keyboard() is not implemented on the web.
		# Assume mobile web browsers have no hardware keyboard.
		has_keyboard = false
	else:
		has_keyboard = DisplayServer.has_hardware_keyboard()

	return has_touch and not has_keyboard
```

scenes/menus/title/title_screen.tscn
```diff
@@ -65,6 +65,34 @@ layout_mode = 2
...
+anchor_left = 0.5
+anchor_top = 0.5
+anchor_right = 0.5
+anchor_bottom = 0.5
+offset_left = -599.0
+offset_top = -248.0
+offset_right = 599.0
+offset_bottom = 227.0
+grow_horizontal = 2
+grow_vertical = 2
```
wjt: I would perhaps make this a little narrower, and configure the label to wrap.

Image

wjt (changes requested):
The touchscreen warning does not prevent interacting with the menu underneath it.

Here's an idea: Why not make it another page under the "Pages" node? Then when it is dismissed, show the main menu.

scenes/menus/title/components/title_screen.gd
wjt: (In theory, you could have a Windows or Linux device that does not have a hardware keyboard. Now as it happens `DisplayServer.has_hardware_keyboard()` currently always returns `true` on those platforms but perhaps in the future someone will implement it properly!)

wjt:
>  Will I need to redo the pull request?

No: make more commits on the same branch, and push the branch. The PR will be updated.

scenes/menus/title/title_screen.tscn
wjt: Ah sorry, I screenshotted the wrong window here!

I stand by this point because the warning was too wide for my smartphone screen. Screenshot to follow. I have attempted to fix this by implementing what I suggested here: making the warning narrower, with word-wrapping.

scenes/menus/title/components/title_screen.gd
wjt: I made this change in the commit I just pushed.

We document this convention in our contribution guidelines: https://github.com/endlessm/threadbare/wiki/Contributing#connect-signals-to-callbacks-through-the-editor-interface

I have to say I don't always agree with it - sometimes it is clearer to connect in code - but in this case connecting it in the editor seemed an improvement.

scenes/menus/title/title_screen.tscn
wjt: Screenshot_20260625-141654

scenes/menus/title/components/title_screen.gd
manuq: Ouch I forgot to remove that from the wiki after our last discussion! I agree with you that this is not an "always do" situation. Removing now.

wjt (approved):
Looks good! I think this design of having it be a separate page works better.

I had a number of improvements to suggest which were easier to describe by making the change, so I pushed a commit that makes those changes. The commit message describes what I did.

scenes/menus/title/components/title_screen.gd
wjt: I think it's a good rule of thumb though!

scenes/menus/title/title_screen.tscn
wjt: And with my change; 
Screenshot_20260625-143644

scenes/menus/title/components/title_screen.gd
manuq: I reworded it as a how-to and moved it to https://github.com/endlessm/threadbare/wiki/Building-the-Game#godot-tips--how-tos

Issue #2279, opened 2026-06-03
More generic persistent game state


wjt:
We currently have a small amount of game state that persists. We periodically want to save new facts about the world; e.g. I want to turn the musician scene into part of the world map (rather than part of a quest) #2264 , and have solving the puzzle unlock paths to other scenes. We already have state for "quests that have been completed" but no way to save the state of the puzzle. Rather than adding more ad-hoc fields for this case, we may want to add a more generic database of facts about the world.

(How this interacts with being able to replay quests remains to be scene. I anticipate that quest scenes would not save state.)

wjt:
My vague idea here is that we should have:

```gdscript
# in global_state.gd
var facts: Dictionary[String, Variant]

# and also in quest_state.gd
var facts: Dictionary[String, Variant]
```

Then we can refer to this easily in dialogue (pretending that we have some sugar to default missing keys to null/false):

```
# in the musician's dialogue for the song sanctuary:
if GameState.quest.facts.spoke_to_musician:
    => hint_flow

set GameState.quest.facts.spoke_to_musician = true
...
```

and then delete the special-case in the musician's script for this:

https://github.com/endlessm/threadbare/blob/ba13b57c93d617c5d1a78f598907a78f97ed98f7/scenes/game_elements/characters/npcs/musician/components/sequence_puzzle_assistant.gd#L18-L20

And similarly remove the "hint level" special case:

https://github.com/endlessm/threadbare/blob/ba13b57c93d617c5d1a78f598907a78f97ed98f7/scenes/game_elements/characters/npcs/musician/components/sequence_puzzle_assistant.gd#L31-L42

We could add a component to the sheep like this:

```gdscript
# fact_counter.gd
@export var fact_name: String = "sheep_annoyed"

# connect this to interaction_started
func increment() -> void:
    GameState.global.facts[fact_name] += 1
```

and now we can implement achievements like "Annoyed 100 sheep".

PR #2449, opened 2026-06-23
GameState: Add global and per-quest generic facts

manuq:
And use the per-quest for the musical puzzle.

### GameState: Add global and per-quest generic facts

A generic database of facts about the world. Quests also have their own, for
per-quest facts.

### Musician: Use generic game state fact for checking first conversation

Remove the variable from the script and use a fact instead, making this persist
when the game is saved.

In lore quests quest_001 and song_sancturary_2. Note that these are 2 separate
facts, because they are for different quests.

### Lore quest_001: Persist progress in music puzzle

Add a setter to SequencePuzzle for it. And also a separate signal for when the
progress (the step) changes. It has to be a separate signal because we want to
persist state when step_solved is emitted, and only do the side effects when
progress_changed is emitted.

----

Helps https://github.com/endlessm/threadbare/issues/2279

Co-authored-by: Will Thompson <wjt@endlessaccess.org>

### DIFF
--- a/scenes/globals/game_state/quest_state.gd
+++ b/scenes/globals/game_state/quest_state.gd
@@ -5,7 +5,8 @@ extends Resource
 
 @export var quest: Quest
 
-## Generic game state facts about the current quest. Outside quests, use [member GlobalState.facts] instead.
+## Generic game state facts about the current quest. Outside quests, use [member GlobalState.facts]
+## instead.
 @export var facts: Dictionary[String, Variant]
 
 ## Resource path of [member quest], which is what is actually saved to disk.
--- a/scenes/globals/game_state/quest_state.gd
+++ b/scenes/globals/game_state/quest_state.gd
@@ -5,7 +5,7 @@ extends Resource
 
 @export var quest: Quest
 
-## Generic game state facts. Outside quests, use [member GlobalState.facts] instead.
+## Generic game state facts about the current quest. Outside quests, use [member GlobalState.facts] instead.
 @export var facts: Dictionary[String, Variant]
 
 ## Resource path of [member quest], which is what is actually saved to disk.
--- a/scenes/globals/game_state/quest_state.gd
+++ b/scenes/globals/game_state/quest_state.gd
@@ -5,6 +5,9 @@ extends Resource
 
 @export var quest: Quest
 
+## Generic game state facts. Outside quests, use [member GlobalState.facts] instead.
+@export var facts: Dictionary[String, Variant]
+
 ## Resource path of [member quest], which is what is actually saved to disk.
 ## Use [member quest] instead.
 @export_storage var quest_path: String:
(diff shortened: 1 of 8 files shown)


manuq:
To remember if player talked to the musician: This is trivial, and the DialogueManager doesn't complain if the key doesn't exist in the facts Dictionary. Is also nice to play with it in the Inspector from the Remote dock:


https://github.com/user-attachments/assets/82ce572b-9f82-4f81-ad9f-6d795e5164cc

manuq:
To remember the steps in the music puzzle: I added `set_progress()` to the SequencePuzzle and also a separate `progress_changed` signal, for the side effects. Now I realize that `step_solved` can be replaced by this new one, with a signal argument "persist" or something.

See below a video resuming the game from the puzzle already solved. Although changing the music doesn't work, maybe is too early for `MusicPlayer`?  When `switch_to_clip()` is called, the following warning is pushed to the output: "Switching clips is only supported with AudioStreamInteractive".

[recording.webm](https://github.com/user-attachments/assets/b3767579-b8ce-42b5-8c96-5ae7278bec55)

I could also save/restore state for the thread being given by the musician. Or not! We should decide what to persist and what is not relevant.

scenes/globals/game_state/quest_state.gd
```diff
@@ -5,6 +5,9 @@ extends Resource
 
 @export var quest: Quest
 
+## Generic game state facts. Outside quests, use [member GlobalState.facts] instead.
```
wjt:
```suggestion
## Generic game state facts about the current quest. Outside quests, use [member GlobalState.facts] instead.
```

wjt (approved):
I have some lingering worries about the "namespace" of random facts, but this seems better than having to add bespoke fields to the state resources for facts in random scenes.

manuq:
> I have some lingering worries about the "namespace" of random facts

Me too! And also we need to decide when to use this and when not. For example for the quest log in the inventory, for quests attempted I could either add global facts, or add an `attempted_quests` that matches the existing `completed_quests`:

```
## [Quest]s which the player has previously completed. Modify this with
## [method set_quest_completed_state].
@export var completed_quests: Array[Quest]
```

wjt:
Can we combine the "attempted quests" with the changes on #2439 ?

manuq:
> Can we combine the "attempted quests" with the changes on #2439 ?

Yes I'm about to review it!

manuq:
> Although changing the music doesn't work, maybe is too early for `MusicPlayer`? When `switch_to_clip()` is called, the following warning is pushed to the output: "Switching clips is only supported with AudioStreamInteractive".

I fixed this. It turns that when restoring `MusicPlayer.play_stream()` should be called instead.

manuq:
The usage in this PR is only for the music puzzle in quest 001. But I need to check others like the one in the tutorial.

I have 2 more things to try regarding these generic facts:

- Achievements like "Annoyed 100 sheep"
- Persist an array of many of the same things, like the coi.. buttons collected in several levels

But since this is approved, I'll merge and leave the task open.

Issue #1757, opened 2025-12-16
Storybook: List of stories does not fit on the page


wjt:
We have a wonderful new problem: we have too many StoryQuests to be listed on one page of the storybook!

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/56cb7f89-92d4-426a-b98c-c1acf6367480" />

Currently the storybook gains a scrollbar in this case. However if you navigate with the arrow keys/controller D-pad, you can focus a button that is not visible.

A minimal fix would be to scroll the container to ensure the focused item is always visible. But even better would be to spread the list of stories across as many pages as are needed.

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 75 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 75 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 75 |
| Game Design: Game Feel & UX | Technical Literacy | 75 |

alcole2:
I am going to take the second approach: moving what doesn't fit on one page onto the next.

alcole2:
I duplicated the node hierarchy from the left page to make one for the right page as well. I got a error on the `ScrollContainer` called `QuestContainer` because it can only have one child node. I placed both page nodes inside an `HBoxContainer` called `BookSpread` to resolve that and hopefully give the side by side look like a book spread. 

<img width="258" height="194" alt="Image" src="https://github.com/user-attachments/assets/9b1e4b7c-526c-4158-9fc3-0df9c735a460" />

alcole2:
Idea:

- I'm going to divide the StoryQuests between `LeftQuestList` and `RightQuestList`.

- It looks like 7 StoryQuests can fit on one page, so once the 7 spots on the left page are filled up, the rest goes to the right page. This means you can see 14 StoryQuests at once, and if there's more, you'll have to press `Next` to flip the page and see more.

- After pressing `Next`:
        -  If there's more than 14 StoryQuests, the page flipping animation will play and the remaining StoryQuests will be shown.
        - If there aren't anymore StoryQuests, it will continue it's current behavior of showing each individual StoryQuest page

manuq:
> Idea:
> 
[...]

PR #2461, opened 2026-06-24
Storybook: Scroll to focused button

wjt:
Previously, if you scrolled down in the storybook using the down arrow, you would eventually focus a button that is scrolled out of the visible area of the scroll container.

Scroll the container when a button is focused so that that button is in view.

Helps https://github.com/endlessm/threadbare/issues/1757


wjt:
@alcole2 @felixwalberg FYI - I didn't know the [`ensure_control_visible`](https://docs.godotengine.org/en/stable/classes/class_scrollcontainer.html#class-scrollcontainer-method-ensure-control-visible) method existed until about 10 minutes before opening this PR! :)

felixwalberg (approved):
`ensure_control_visible` for the win!

PR #2462, opened 2026-06-24
tutorial grappling: Guide player more clearly to first crossing

wjt:
In playtesting, people were often deceived by the pin & needle at the
top-left of the island. Those are there just so that you can get back to
the start - and yet they're the most obviously grapple-able thing on the
screen from the starting position.

I want to leave them in because people want to be able to go back and
collect buttons they missed. So:

- Add a couple of buttons to the patch of void in the middle of the
  starting island, which already has pins around it. This should serve
  as an already-visible hint that you should try to fling yourself
  across the void.

- Add a couple more buttons leading the way to the first grappling site
  at the start of the island.

- Reduce the distance of the first grapple by adding 1 more tile of path
  jutting out into the water, and more path leading to it, which should
  make it really very clear what you are meant to do. :)


wjt:
| Before | After |
| ------ | ----- |
| image | Screenshot From 2026-06-24 15-35-56 |
| image | image |

manuq:
@felixwalberg @alcole2 would you like to review?

felixwalberg:
> @felixwalberg @alcole2 would you like to review?

As was talked about on the call this morning, GitHub notifications are fickle! I will start to look at this now...

felixwalberg (commented):
I am noticing that the bottom pin in the loop on the right side of Arun is unreachable via the keyboard when going from above. It doesn't seem to be quite 45 degrees. This doesn't really break the level at all, but it might be nice to be able to complete the loop this way.

Image
<br>
<br>

Same way with the previously unreachable pin. I don't think I can reach it with just the keyboard. With the mouse it has to be almost pixel perfect!

Image

felixwalberg:
I realize now that this was me thinking about your message in Slack about the unreachable pin and not something addressed here, oops!

felixwalberg (approved):
I think these changes make it much clearer that you are meant to explore to the right and down. As a gamer who loves collecting every item before progressing, I think having buttons in the middle of the void encourages you to attempt to get them before getting sidetracked by the unreachable area to the left.

I played around with connecting the sand path all the way down but I don't know if it adds anything extra.

wjt:
The story I am trying to tell by having the paths not be connected to one another is that (before we fade in) Arun & StoryWeaver must have walked in through the closed door, and followed the path to the bridge. But then the bridge collapsed! So StoryWeaver must explore a little to find the alternative route, going off the beaten track.

PR #2464, opened 2026-06-24
Hide inconsistent input HUD hints on After The Tremor combat level

### LINKED ISSUE #2364: after_the_tremor_combat shows incorrect controls
The controls are meant to be different for this scene: ⬅️ should cause the left-hand player to repel, while ➡️ should cause the right-hand player to repel. Both players are fixed in place. However now that we show the Input HUD at all times, incorrect controls are shown on screen.

Image

### XP Summary (total 475)
| Subdomain | Skill | XP |
| - | - | - |

felixwalberg:
Input HUD hints are shown throughout the game whenever there is a `player.tscn` node in the scene tree. Some scenes, like the combat level for After The Tremor, have a `player.tscn` node that cannot perform the same actions as suggested by the hints. In this case, the player cannot move around (WASD or arrow keys) but instead uses A and D or left and right arrows to repel projectiles from the boss. The input hints are now incorrect and confusing. This PR addresses this issue by hiding the input hints during the combat scene.

Charlie (player on left) is now a deconstructed "character" rather than a full `player.tscn` node and does not alert the `input_hud.gd` script. Bryan (player on right) is still a `player.tscn` node and is removed from the `player` group on `_ready()` and then added back when the boss is defeated. For the input hints to reappear, we needed to "refresh" the visibility of the hints and that is done with `refresh_scene_status()` (but could be changed to reference the `_on_scene_changed()` method directly if we wanted to remove the leading underscore).

Additionally, the player position was modified slightly, as a change in the repel mechanic seems to have made the previous angle nearly impossible to hit the boss!

Possible future work:
- Input hints that indicate you can press the left arrow for Charlie's repel and the right arrow for Bryan's repel
- Improve visibility of collectable (since we use input hints rather than text floating on screen above it there is less visual guidance to picking it up)

Resolves https://github.com/endlessm/threadbare/issues/2364


project.godot
```diff
@@ -35,7 +35,7 @@ MusicPlayer="*res://scenes/globals/music_player/music_player.tscn"
 CameraShake="*res://scenes/globals/camera_shaker/camera_shake.tscn"
 InputHelper="*res://addons/input_helper/input_helper.gd"
 PauseOverlay="*res://scenes/globals/pause/pause_overlay.tscn"
-InputHud="res://scenes/ui_elements/input_hints/input_hud.tscn"
+InputHud="*res://scenes/ui_elements/input_hints/input_hud.tscn"
```
felixwalberg, line 38: Is there a way to indicate to `input_hud.gd` that the scene has changed without adding this?

felixwalberg:
| Before | After |
| ------ | ----- |
| image | Screenshot 2026-06-24 at 12 28 58 PM |




<br>
Currently has no input hints and might benefit from some indication of the controls. This could also be something more general that might be nice to have. As seen in some other PRs, some learner teams are creating custom combat or other mechanics that either override existing buttons or are new ones. Maybe it would be possible to pass the Input Hud a list of input key -> action + image of hint for each relevant input the level looks for. If none are provided, it would default to movement plus standard mechanics used in the current level.
<br>
Screenshot 2026-06-24 at 12 30 34 PM
<br>
Not completely hidden, as the sign indicates something is going on here.

manuq:
> Maybe it would be possible to pass the Input Hud a list of input key -> action + image of hint for each relevant input the level looks for. If none are provided, it would default to movement plus standard mechanics used in the current level.

Yes. In https://github.com/endlessm/threadbare/pull/2302 I added a way to override the ability names in a StoryQuest:

image

This is reflected in the powerups. But I didn't update the HUD https://github.com/endlessm/threadbare/pull/2302#issuecomment-4635752266 . We have "Throw" in the HUD and the ability is called "Grapple".  The input action is called "throw". So maybe we need to introduce another mapping (a Dictionary) from input action name to the text displayed in the UI?

To consider besides displayed text: should we bring the posibility use different input actions for custom abilities? An ability may need multiple inputs. For example the grapple ability has the `aim` actions plus the `throw` action. But certainly we don't want learners to add new input mapping, because that's a change for the whole project. The existing `champ_blink` and `champ_walk_on_water` should be removed, in my opinion.

project.godot
manuq: I think this is correct. It enables the global variable to the autoload.

manuq (approved):
This relies on adding/removing the player from the "player" group. And this was the solution discussed in the ticket, so approving. Added an alternative in the ticket: https://github.com/endlessm/threadbare/issues/2364#issuecomment-4794843079

PR #2465, opened 2026-06-24
Generic state facts

manuq:
### Pet interaction: Add optional shaker

Pass the shaker to this component for shaking on interaction. Add it to the
sheep and keep it blank for the cat.

Also rename miaow_player to sound_effect_player and adjust it in both sheep and
cat.

### Pet interaction: Move script to npcs/components

Move it out of the cat folder.

### Add fact counter for petting cats and annoying sheeps

Add a FactCounter node with a prefix that can persist a counter as
global fact.

Use it in cat and sheep scenes.

Note that this will count multiple times for the same animal, as many as the
player interacts. For counting if each specific animal was petted, a different
approach (more similar to the upcoming button collected fact) is needed.

Helps https://github.com/endlessm/threadbare/issues/2279


scenes/game_logic/achievement_counter/achievement_counter.gd
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+class_name AchievementCounter
```
manuq: @wjt re-reading your comment https://github.com/endlessm/threadbare/issues/2279#issuecomment-4691661509, you called this `fact_counter`. I can rename it if you wish.

manuq:
This is how these achievements appear in the Remote dock / Inspector (no UI for them yet):
Screenshot From 2026-06-24 14-24-35

scenes/game_logic/achievement_counter/achievement_counter.gd
wjt: I think I narrowly prefer "FactCounter" because these could be used for things other than achievements... For example one might want to count how many times the player has interacted with a specific NPC so that the dialogue can change after 10 interactions.

scenes/game_elements/props/decoration/sheep/sheep.tscn
```diff
@@ -58,13 +59,17 @@ position = Vector2(5, -28)
...
+achievement_counter = NodePath("../AchievementCounter")
 
 [node name="AudioStreamPlayer2D" type="AudioStreamPlayer2D" parent="PetInteraction" unique_id=1930789457]
 stream = SubResource("AudioStreamRandomizer_3vki4")
 bus = &"SFX"
 
-[connection signal="interaction_started" from="InteractArea" to="AnimatedSprite2D/Shaker" method="shake" unbinds=2]
+[node name="AchievementCounter" type="Node" parent="." unique_id=2116673480]
+script = ExtResource("7_1k1ft")
+prefix = "sheeps_annoyed"
```
wjt, line 75:
```suggestion
prefix = "sheep_annoyed"
```

English is a strange language.

wjt (approved):
Nice design! I think `fact_counter` would make clearer that no achievements are (yet) involved.

> Note that this will count multiple times for the same animal, as many as the
> player interacts. For counting if each specific animal was petted, a different
> approach (more similar to the upcoming button collected fact) is needed.

I think this is fine for this kind of simple, just-for-fun counter.

scenes/game_elements/props/decoration/sheep/sheep.tscn
manuq: Ah thank you for noticing!

scenes/game_logic/achievement_counter/achievement_counter.gd
manuq: You are right. Amending...

PR #2479, opened 2026-06-25
Pause overlay: Save game state on Exit to Title

manuq:
I believe this was like this before and it has regressed lately.

### DIFF
--- a/scenes/globals/pause/pause_overlay.gd
+++ b/scenes/globals/pause/pause_overlay.gd
@@ -95,6 +95,7 @@ func _on_back_button_pressed() -> void:
 
 
 func _on_title_screen_button_pressed() -> void:
+	GameState.save()
 	toggle_pause()
 	SceneSwitcher.change_to_file_with_transition(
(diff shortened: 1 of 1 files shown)


manuq:
I will have to bisect to confirm.

scenes/globals/pause/pause_overlay.gd
wjt, line 102: This shouldn't be necessary because SceneSwitcher (implicitly) saves the game when it changes scene:

https://github.com/endlessm/threadbare/blob/a8e99051cfae8bdb5893fb55a688879959d973cd/scenes/globals/scene_switcher/scene_switcher.gd#L173-L174

wjt: Oh, no, I'm wrong:

https://github.com/endlessm/threadbare/blob/a8e99051cfae8bdb5893fb55a688879959d973cd/scenes/globals/game_state/game_state.gd#L153-L156

It will have been me who introduced this.

wjt (approved):

PR #2487, opened 2026-06-26
Add CollectibleFact and use it in buttons

### LINKED ISSUE #2279: More generic persistent game state

manuq:
The exported property `collectible` is any Node2D that has a "collected" signal.

Persist one Array per scene file, using the current scene file path as key for the fact. And use the collectible node path inside the scene for identifying each collectible.

Adapt hookable button item to be freed when ready, if the wrapped button item is freed too (or queued for deletion).

Also add a FactCounter named "buttons_collected" to buttons.

Resolve https://github.com/endlessm/threadbare/issues/2279

### DIFF
--- a/scenes/game_elements/props/button_item/components/button_item.gd
+++ b/scenes/game_elements/props/button_item/components/button_item.gd
@@ -45,10 +45,3 @@ func _on_area_2d_area_entered(area: Area2D) -> void:
 		collected.emit()
 		_emit_shine_particles()
 		queue_free()
-
-
-## Called from [CollectibleFact] when the game state already has this button
-## marked as collected.
-func start_collected() -> void:
-	visible = false
-	queue_free()
--- a/scenes/game_elements/props/button_item/components/button_item.gd
+++ b/scenes/game_elements/props/button_item/components/button_item.gd
@@ -40,7 +40,15 @@ func _emit_shine_particles() -> void:
 
 func _on_area_2d_area_entered(area: Area2D) -> void:
 	# TODO: This is not added to an inventory or anything, is just cosmetic.
+	# Although now is persisted in game state.
 	if area.owner.is_in_group(&"player"):
 		collected.emit()
 		_emit_shine_particles()
 		queue_free()
+
+
+## Called from [CollectibleFact] when the game state already has this button
+## marked as collected.
+func start_collected() -> void:
+	visible = false
+	queue_free()
--- a/scenes/game_logic/collectible_fact/collectible_fact.gd
+++ b/scenes/game_logic/collectible_fact/collectible_fact.gd
@@ -64,7 +64,12 @@ func _on_collectible_ready() -> void:
 		if collectible.has_method("start_collected"):
 			collectible.start_collected()
 		else:
+			# Fallback to just free the collectible:
 			collectible.queue_free()
+		# If the game state indicates that the collectible was collected, we can safely free this
+		# node after applying the effects (calling start_collected() on them or its fallback).
+		# In the future, we may want to revisit this, if we add a way to bring back the collected
+		# items while the game is running.
 		queue_free()
 	else:
 		collectible.connect(COLLECTED_SIGNAL_NAME, _on_collectible_collected)
--- a/scenes/game_logic/collectible_fact/collectible_fact.gd
+++ b/scenes/game_logic/collectible_fact/collectible_fact.gd
@@ -12,7 +12,7 @@ extends Node
 
 const COLLECTED_SIGNAL_NAME := "collected"
 
-## A node with a signal named [member COLLECTED_SIGNAL_NAME]. Also if the node has a method named
+## A node with a signal named [constant COLLECTED_SIGNAL_NAME]. Also if the node has a method named
 ## "start_collected", it may be called when restoring the game state.
 @export var collectible: Node2D:
 	set = _set_collectible
--- a/scenes/game_logic/collectible_fact/collectible_fact.gd
+++ b/scenes/game_logic/collectible_fact/collectible_fact.gd
@@ -1,0 +1,82 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name CollectibleFact
+extends Node
+## Persist and restore the collected state of a collectible. For example, a
+## [ButtonItem].
+##
+## Persist one Array per scene file, using the current scene file path as key for
+## the fact. And use the collectible node path inside the scene for identifying
+## each collectible.
+
+const COLLECTED_SIGNAL_NAME := "collected"
+
+## A node with a signal named [member COLLECTED_SIGNAL_NAME]. Also if the node has a method named
+## "start_collected", it may be called when restoring the game state.
+@export var collectible: Node2D:
+	set = _set_collectible
+
+var _fact_name: String
+var _fact_value: String
+var _facts: Dictionary
+
+
+func _enter_tree() -> void:
+	if not collectible and get_parent() is Node2D:
+		collectible = get_parent()
+
+
+func _set_collectible(new_collectible: Node2D) -> void:
+	collectible = new_collectible
+	if not Engine.is_editor_hint():
+		if collectible.is_node_ready():
+			_on_collectible_ready()
+		else:
+			collectible.ready.connect(_on_collectible_ready)
+	update_configuration_warnings()
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+	if not collectible:
+		warnings.append("Collectible must be set.")
+	elif not collectible.has_signal(COLLECTED_SIGNAL_NAME):
+		warnings.append("Collectible must have a signal named %s." % COLLECTED_SIGNAL_NAME)
+	return warnings
+
+
+func _get_fact_name() -> String:
+	return "collected_in_%s" % get_tree().current_scene.scene_file_path
+
+
+func _get_fact_dict() -> Dictionary:
+	return GameState.quest.facts if GameState.quest else GameState.global.facts
+
+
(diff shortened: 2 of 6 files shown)


manuq:
With this change, the collected buttons won't appear again in levels, unless the game is restarted.

@felixwalberg @alcole2 would you like to review?

alcole2:
Yes!

scenes/game_elements/props/button_item/components/button_item.gd
```diff
@@ -40,7 +40,15 @@ func _emit_shine_particles() -> void:
...
 	if area.owner.is_in_group(&"player"):
 		collected.emit()
 		_emit_shine_particles()
 		queue_free()
+
+
+## Called from [CollectibleFact] when the game state already has this button
+## marked as collected.
+func start_collected() -> void:
+	visible = false
```
wjt: Interesting, without this is it visible for a single frame?

scenes/game_logic/collectible_fact/collectible_fact.gd
```diff
@@ -0,0 +1,82 @@
...
+## Persist and restore the collected state of a collectible. For example, a
+## [ButtonItem].
+##
+## Persist one Array per scene file, using the current scene file path as key for
+## the fact. And use the collectible node path inside the scene for identifying
+## each collectible.
+
+const COLLECTED_SIGNAL_NAME := "collected"
+
+## A node with a signal named [member COLLECTED_SIGNAL_NAME]. Also if the node has a method named
```
wjt:
```suggestion
## A node with a signal named [constant COLLECTED_SIGNAL_NAME]. Also if the node has a method named
```

scenes/game_logic/collectible_fact/collectible_fact.gd
```diff
@@ -0,0 +1,82 @@
...
+## "start_collected", it may be called when restoring the game state.
```
wjt:
```suggestion
## [code]start_collected[/code], it may be called when restoring the game state.
```

scenes/game_elements/props/button_item/components/button_item.gd
```diff
@@ -40,7 +40,15 @@ func _emit_shine_particles() -> void:
...
+	queue_free()
```
wjt: If I understand the flow correctly, this method is not actually needed because the fallback path when the node doesn't implement `start_collected()` is the same: hide and queue free.

scenes/globals/game_state/game_state.gd
```diff
@@ -129,8 +129,8 @@ func set_quest(new_quest: Quest) -> void:
 	quest = QuestState.new(new_quest, quest_player_state)
 
 
-## Guess which quest the given scene is part of, and set [member current_quest]
-## accordingly. If the quest cannot be determined, unset [member current_quest].
+## Guess which quest the given scene is part of, and set [member quest]
+## accordingly. If the quest cannot be determined, unset [member quest].
```
wjt, line 133: Good catch.

scenes/game_logic/collectible_fact/collectible_fact.gd
```diff
@@ -0,0 +1,82 @@
...
+	_facts = _get_fact_dict()
+
+	if _fact_value in _facts.get(_fact_name, []):
+		# Restore from saved state:
+		if collectible.has_method("start_collected"):
+			collectible.start_collected()
+		else:
+			collectible.visible = false
+			collectible.queue_free()
+		queue_free()
```
wjt, line 73: Why do you need to free the `CollectibleFact` node here?

scenes/game_logic/collectible_fact/collectible_fact.gd
```diff
@@ -0,0 +1,82 @@
...
+		else:
+			collectible.ready.connect(_on_collectible_ready)
+	update_configuration_warnings()
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+	if not collectible:
+		warnings.append("Collectible must be set.")
+	elif not collectible.has_signal(COLLECTED_SIGNAL_NAME):
```
wjt, line 44: I understand now why a slice of the community is so adamant that Godot should have traits...

scenes/game_logic/collectible_fact/collectible_fact.gd
```diff
@@ -0,0 +1,82 @@
...
+
+
+func _get_fact_dict() -> Dictionary:
+	return GameState.quest.facts if GameState.quest else GameState.global.facts
+
+
+func _on_collectible_ready() -> void:
+	if not GameState.persist_progress:
+		queue_free()
+		return
```
wjt: I'm not sure this is necessary - if not persisting progress, we can do whatever we like to the game state - and I think it will have surprising results when (for example) testing a pair of scenes linked with teleporters in the editor, in which case `persist_progress` will be false, but I think you'd still expect the coins collected in scene A to be gone after going to scene B and back to scene A.

wjt (approved):
Looks good, just a few notes.

scenes/game_logic/collectible_fact/collectible_fact.gd
```diff
@@ -0,0 +1,82 @@
...
+
+	_fact_name = _get_fact_name()
+	_fact_value = collectible.get_path()
```
wjt: Do you not want the path relative to the current scene, rather than relative to the root node?

scenes/game_elements/props/button_item/components/button_item.gd
manuq: Hmm let me check because I don't remember!

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: Thanks!

scenes/game_elements/props/button_item/components/button_item.gd
manuq: Yeah I was thinking in adding this as example, but is not worth it. The magical threads (CollectibleItem) can do it.

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: My thinking was: if the fact is in the game state (the item was collected), it was read and applied (button removed, thread disabled), then it doesn't make sense to keep it in the scene tree. Unless we add a way to revert it while the game is running.

But yeah it doesn't save much to free this node. If you want, I will remove the `queue_free()`. Otherwise I can add a comment with the above.

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: This exists but I never tried... https://github.com/Earewien/godot-traits

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: Good point!

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: I use both! The way I ended up organizing it is one Array per scene file. For example here I grabbed 3 buttons in `grappling_hook_start.tscn` and 22 buttons in `grappling_hook_pins.tscn`:
image

scenes/game_logic/collectible_fact/collectible_fact.gd
wjt: There is a PR to add traits to the engine proper, which keeps stalling because it's a massive change to the type system

scenes/game_logic/collectible_fact/collectible_fact.gd
wjt: I was just trying to understand if it is necessary. I think you're right to have it, but yes, let's comment why. I think in general we would expect the fact node to be a child of `collectible` so it will be freed anyway..

scenes/game_logic/collectible_fact/collectible_fact.gd
wjt: My point is the `/root/` in those paths. If you instead `get_tree().current_scene.get_path_to(collectible)` it'll just be `GrapplingHookPins/OnTheGround/ButtonsA/ButtonItem9`.

scenes/game_elements/props/button_item/components/button_item.gd
manuq: Nope, I commented the transitions and removed this line from the fallback of `start_collected()`. I also played the game in slow motion from the editor. I couldn't see the buttons. So I'm leaving `collectible.queue_free()` as the fallback.

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: Oh right, thanks, I'll use the path relative to the current scene. That's what I really wanted.

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: Will be freed anyway.. true!

scenes/game_logic/collectible_fact/collectible_fact.gd
manuq: Tested and yes, `get_tree().current_scene.get_path_to(collectible)` is the correct path relative to the scene. Thanks for the review @wjt. Merging.

PR #2498, opened 2026-06-30
Spawn in front of elder after abandoning a StoryQuest

### LINKED ISSUE #2496: Spawn in front of elder after abandoning a StoryQuest
When you abandon a LoreQuest, you now respawn just before the teleporter that starts the quest. When you abandon a StoryQuest, you instead respawn in front of the Eternal Loom in Fray's End (the default position in that scene).

It would be better to spawn in front of the elder who sent you on the quest - it makes some spatial sense, and also makes it possible to move those elders to another scene (#2265).

wjt:
This is roughly the point where StoryWeaver must have been to have
interacted with the elder and start the quest in the first place, unlike
the area in front of the Loom which is the default spawn point after
abandoning otherwise.

It also means that if we move the elder to another scene, the abandon
point will move with them.

Resolves https://github.com/endlessm/threadbare/issues/2496

### DIFF
--- a/scenes/game_elements/characters/npcs/elder/components/elder.gd
+++ b/scenes/game_elements/characters/npcs/elder/components/elder.gd
@@ -19,6 +19,14 @@ const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
 		eternal_loom = new_value
 		update_configuration_warnings()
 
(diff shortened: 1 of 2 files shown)


manuq (approved):
I thought "in front of the loom" was enough, but this better, smart thinking!

PR #2504, opened 2026-07-03
SceneLink: Don't crash when inspected in running game

wjt:
Previously this code asumed get_tree().edited_scene_root would be
non-null in the case where SceneLink.next_scene is set and the node is
inside the tree.

This is not the case (at least) when running the game and inspecting a
SceneLink(-derived) node in the remote scene tree:
get_tree().edited_scene_root is documented to be blank outside the
editor.

Handle that case.


manuq (approved):

wjt:
Ugh I introduced a new error with this change:

```
  ERROR: scene/main/node.h:559 - Parameter "data.tree" is null.
  ERROR: res://scenes/globals/scene_switcher/scene_link.gd:136 - Invalid access to property or key 'edited_scene_root' on a base object of type 'null instance'.
```

PR #2505, opened 2026-07-03
Stella combat: Improve intro dialogue

wjt:
Previously, the autoplayed dialogue at the start of this scene included
the full conversation with the fisherman. But if you later interacted
with him, he would repeat the full conversation, including the
narration.

Split the dialogue script in two. Play just the opening line of
narration automatically. Play the rest of the dialogue when interacting
with the fisherman. Add a script that starts the fill game logic at the
end of the first interaction with the fisherman.

Move the fisherman to the OnTheGround node to y-sort him relative to
Stella.

### DIFF
--- a/scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_combat.dialogue
+++ b/scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_combat.dialogue
@@ -1,11 +1,13 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
-
-~ start
-Stella finds herself following the ocean breeze to the quiet shores of Northern Australia, where she meets a Yolŋu fisherman mending his net.
+~ fisherman
 Stella: Do you know about my missing star? Is it from the hunter? From a hearth?
 Fisherman: Hunter? Hearth? No, no. That’s Djulpan, the canoe that sails the sky.
 Fisherman: One of the canoe’s stars was a fish that leapt overboard, plunging into the dark sea. Without all three stars, the canoe is incomplete.
+=> END
+
+~ intro_narration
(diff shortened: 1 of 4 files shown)


scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_combat.gd
```diff
@@ -0,0 +1,16 @@
...
+
+
+func _ready() -> void:
+	fisherman.interact_area.interaction_ended.connect(
+		_on_fisherman_interaction_ended, CONNECT_ONE_SHOT
+	)
+
+
+func _on_fisherman_interaction_ended() -> void:
+	fill_game_logic.start()
```
wjt, line 16: It could also be this (untested)

```suggestion
		fill_game_logic.start, CONNECT_ONE_SHOT
	)
```

This can't be done in the scene file because the interaction signals are not exposed on the root node, and I don't like editable children.

scenes/quests/story_quests/stella/2_stella_combat/stella_combat_components/stella_combat.dialogue
```diff
@@ -1,13 +1,15 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
-
-~ start
-Stella finds herself following the ocean breeze to the quiet shores of Northern Australia, where she meets a Yolŋu fisherman mending his net.
+~ fisherman
```
manuq, line 3: Yeah I was unsure because `start` is usually the first title, but I checked and "start" is just an example. DialogueResource has a `first_title` and it can even be empty, in which case the dialogue starts from the first line. I remember that the smallest dialogue is a single line like "hello world", not needing a title or an `=> END`.

manuq (approved):
Nice improvements for Stella.

PR #2428, opened 2026-06-22
Primer PR de StoryQuest "El Hilo Que No Teji"

Nox871:
Creación de la storyquest "El Hilo Que No Teji"
- Intro
- Miniquest_01
- Miniquest_02
- Miniquest_03
- Outro

Este juego contiene nuevos:
- Assets
- Scripts
- Shaders
- HUDs

Falta:
- Canciones / SFX
- Animaciones
- Assets de sombras
- Assets de Boss
- Modificar Miniquest_03

PR #2506, opened 2026-07-03
renya: Set music as looping

wjt:
All these tracks are designed to loop, but the loop flag was not set in
their import settings, so the music would simply stop at the end of the
track rather than looping.

Set the loop flag. While we're here, set the beats per bar and bpm (both
of which seem to be regular throughout each piece) in case someone wants
to do something fancy with them in future.


wjt:
I noticed this while reviewing #2428 which reuses one of these tracks.

manuq (approved):
Good to know about the bpm and bar_beats!

wjt:
> Good to know about the bpm and bar_beats!

That's how the switch from Void Still to Void Creep is synced to the first beat of a bar of Still :)

PR #2508, opened 2026-07-06
Quest separator: Keep music playing

wjt:
Quest separator: Keep music playing

Previously the separator screen was silent. I don't like silence!

Add a `BackgroundMusic` node with no `stream` configured. In a child
node, copy the currently-playing AudioStream (which may be `null`) to
its `stream`. In practice this means that the bed from Fray's End keeps
playing uninterrupted until the quest begins. For now it would be
simpler to set the same music manually in the separator. But you could
imagine in future having different background music for different parts
of the world; this way whatever music happens to be playing will
continue playing.

### DIFF
--- a/scenes/menus/quest_separator/quest_separator.tscn
+++ b/scenes/menus/quest_separator/quest_separator.tscn
@@ -7,6 +7,8 @@
 [ext_resource type="Shortcut" uid="uid://c5pwgp1w8p3y1" path="res://scenes/ui_elements/shortcuts/back_shortcut.tres" id="5_wq0a1"]
 [ext_resource type="Script" uid="uid://b78blxbleps14" path="res://scenes/menus/quest_separator/components/timer_progress_bar.gd" id="6_amr4g"]
 [ext_resource type="Texture2D" uid="uid://dcwmaoqgu5t84" path="res://scenes/globals/scene_switcher/transitions/Radial.png" id="6_wq0a1"]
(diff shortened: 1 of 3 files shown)


scenes/menus/quest_separator/quest_separator.tscn
```diff
@@ -7,6 +7,8 @@
 [ext_resource type="Shortcut" uid="uid://c5pwgp1w8p3y1" path="res://scenes/ui_elements/shortcuts/back_shortcut.tres" id="5_wq0a1"]
 [ext_resource type="Script" uid="uid://b78blxbleps14" path="res://scenes/menus/quest_separator/components/timer_progress_bar.gd" id="6_amr4g"]
 [ext_resource type="Texture2D" uid="uid://dcwmaoqgu5t84" path="res://scenes/globals/scene_switcher/transitions/Radial.png" id="6_wq0a1"]
+[ext_resource type="Script" uid="uid://c8405c212rbn6" path="res://scenes/game_elements/props/background_music/components/background_music.gd" id="8_mwkko"]
+[ext_resource type="Script" uid="uid://bgffgm5flql38" path="res://scenes/menus/quest_separator/components/keep_that_groove_going.gd" id="9_1wjdo"]
```
manuq, line 11: Fun filename! :)

manuq (approved):
Yes I think for now is best to keep the previous music going, good point!

This reminds me how bad does this separator looks visually, is a placeholder.

wjt:
This would actually be much easier to implement (or rather - would work naturally) if the separator were part of the pause overlay... :)

PR #2509, opened 2026-07-06
Stealth/sequence combo: Save puzzle progress

wjt:
This is similar code to music_puzzle but with a few tweaks to handle
revealing the collectible. This is necessary because the `solved` signal
is not emitted when the puzzle's progress is changed in code rather than
through player interaction.

I think this makes the level fairer because you no longer have to
repeatedly solve the early steps before you can make another attempt at
the more difficult third step, or (worse!) solve all three steps again
if you are caught while running towards the exit!

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -69,7 +69,6 @@ func _ready() -> void:
 
 	_set_item(item)
 	_update_based_on_revealed()
-	sprite_2d.modulate = Color.WHITE if revealed else Color.TRANSPARENT
 
 	if Engine.is_editor_hint():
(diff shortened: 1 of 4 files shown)


wjt:
I was originally planning to do this by moving the fact save/load logic directly into `sequence_puzzle.gd`. Then I thought I'd try making it a separate component - which turned out to be more trouble than it was worth. So then I just open-coded it again.

Having written this I do actually think it can live in the sequence puzzle, but it may need to be optional because the surrounding scene does need to be designed to take account of the fact that the puzzle may start out in the solved state, so it'll break existing scenes and maybe not everyone will care to design the scene with that in mind.

felixwalberg (approved):
I think this looks good and definitely makes the combo level more manageable!

PR #2512, opened 2026-07-07
NO_EDIT: Add BackgroundMusic nodes

### LINKED ISSUE #2136: Add BackgroundMusic nodes in NO_EDIT templates
Currently the templates have no background music. This means that we don't show learners that we have a BackgroundMusic node, and so often we find that StoryQuests have bare AudioStreamPlayer(2D) nodes, often assigned to the wrong audio bus, which do not have the fading-between-scenes behaviour of BackgroundMusic.

I suggest we add a BackgroundMusic node to each scene of the NO_EDIT template, with one of the short loops of the Threadbare theme as the stream. e.g.:

intro: Threadbare Loop_Main_Intro_01.ogg
scene 1: Threadbare Loop_Main_Bridge_01.ogg
scene 2: Threadbare Loop_Main_Bridge_02.ogg
sequence puzzle: silent (no BackgroundMusic node)
outro: Threadbare Loop_Main_Cresendo_01.ogg

### XP Summary (total 175)

| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 50 |
| Game Design: Core Gameplay & Mechanics | Creativity & Design Thinking | 50 |

wjt:
NO_EDIT: Add BackgroundMusic nodes

This will help guide learners to use these rather than hand-configured
AudioStreamPlayer nodes (which don't cross-fade between scenes and tend
not to be assigned to the correct bus).

I picked different clips of the main title theme as placeholder music
for each. For the sequence puzzle I left the stream unset.

Resolves https://github.com/endlessm/threadbare/issues/2136


manuq (approved):
I think this is good, it will also mean that some StoryQuests will arrive with these placeholder clips, in the same way they arrive with the placeholder projectile, for example. I'm starting to feed bad about seeing the gray placeholder assets in submissions :/

manuq (approved):

PR #2513, opened 2026-07-07
Move SPAWN_POINT_GROUP_NAME constant to SpawnPoint

wjt:
Previously `scene_link.gd` defined a `const SPAWN_POINT_GROUP_NAME`, but
it did not use it consistently, and `spawn_point.gd` also hardcoded the
group name.

Move the `const` to `spawn_point.gd` and use it consistently in code.


felixwalberg (approved):
Definitely makes more sense to have the const in SpawnPoint class and to use that rather than hardcoded strings.

PR #2515, opened 2026-07-07
ci: Remove Discord Notification workflow

### LINKED ISSUE #1089: Discord merge notification action failed

wjt:
This is supposed to post on our Discord whenever a PR is merged.

It used to run on the `pull_request` event, which only worked when PRs
from our own team were merged: useless for highlighting community
merges.

Then I changed it to the `pull_request_target` event, which doesn't work
at all.

Just remove it. I deleted the channel on the other end.

Resolves https://github.com/endlessm/threadbare/issues/1089


wjt:
I'm merging this without review because:

1. This workflow has done nothing for close to a year
2. I deleted the target channel so [the workflow fails on every merge to main](https://github.com/endlessm/threadbare/actions/workflows/discord.yml)

PR #2518, opened 2026-07-08
CollectibleItem: Don't flash thread for 1 frame while revealing

### LINKED ISSUE #2517: 1 frame of fully-visible thread at start of reveal animation
### Threadbare version

v0.2.1-11-g9eb3c1fb9

### Affected scene

res://scenes/quests/lore_quests/quest_000/2_stealth/tutorial_stealth.tscn

### Description

In commit cf47b8a90fcb274e0def1c03060fbaf9b2c76b98 I changed `collectible_item.gd` so that setting the `CollectibleItem.revealed` property to `true` directly (rather than using `reveal()`) would adjust the thread sprite's `modulate` property to full opacity, rather than leaving it transparent. (`reveal()` animates this property.) In my testing this was fine because the animation would immediately set it back to transparent in the 0th frame of the animation.

However I have since found that has regressed in some scenes that `reveal()` the thread from dialogue, e.g. the tutorial stealth:

https://github.com/user-attachments/assets/5f84300c-8488-4664-b8de-a42f0d8489e5

I tried a simple test scene that calls `reveal()` from an `Area2D.body_entered` callback and the problem does not manifest there.

The dialogue file in question does this:

```
Arun: As you go through the world, you'll come across threads of Memory, Imagination, and Spirit.
do wait(0.5)
do memory_thread.reveal()
```

If I comment out the `do wait(0.5)` line then _the bug goes away_.

I think the difference is:

- With the `wait`, the `reveal()` call is made in the context of a `timeout` signal:

```gdscript
await Engine.get_main_loop().create_timer(float(args[0])).timeout
```

- Without it, the `reveal()` call is made in an `_input()` callback because it is triggered by me advancing the previous line of dialogue.

So these must be at different points in the mainloop, meaning that in one case the first frame of the animation is processed before the next frame is drawn, and in the other it is deferred to the following frame (leading to the flash of opaque thread).

### Checklist

wjt:
Commit cf47b8a90fcb274e0def1c03060fbaf9b2c76b98 caused a regression
where calling `reveal()` would sometimes show the fully-opaque thread
for 1 frame before fading in from transparent. This depended
on what stage of the mainloop `reveal()` was called from: calling it
from an `_input` callback worked correctly, but calling it from a
`Timer.timeout` signal handler did not.

Move the assigment `revealed = true` to after the animation is complete.
The animation already sets `visible` on the `Sprite2D` to `true`; and
it's fine for the thread to be non-interactive and non-collidable for
the 0.5s duration of the reveal animation.

Document `reveal()` and make it a no-op in the case where the thread is
already revealed or is being revealed.

Resolves https://github.com/endlessm/threadbare/issues/2517

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -76,12 +76,22 @@ func _ready() -> void:
 	interact_area.interaction_started.connect(self._on_interacted)
 
 
-## Make the collectible appear
+## Make the collectible appear with an animation, ultimately setting [member revealed].
+##
+## This does nothing if [member revealed] is [code]true[/code], or if the collectible is already
+## being revealed.
+##
+## To reveal the collectible immediately without any animation or sound, set [member revealed]
+## directly.
 func reveal() -> void:
-	revealed = true
+	if revealed or animation_player.current_animation == &"reveal":
+		return
+
 	appear_sound.play()
 	animation_player.play("reveal")
 	await animation_player.animation_finished
+
+	revealed = true
 
 
 ## When interacted with, the collectible will display a brief animation
@@ -114,8 +124,6 @@ func _update_based_on_revealed() -> void:
 		interact_area.disabled = not revealed
 	if sprite_2d:
 		sprite_2d.visible = revealed
-		# If we are playing the reveal animation, this will be set back to
-		# TRANSPARENT immediately then tweened to white.
 		sprite_2d.modulate = Color.WHITE if revealed else Color.TRANSPARENT
 	if physical_collider:
 		physical_collider.disabled = not revealed


wjt:
I find this a bit spaghetti-ish and I would prefer to do this, eliminating `_update_based_on_revealed`:

    @export var revealed: bool = false:
        set = set_revealed

    func set_revealed(value: bool, animate: bool = false) -> void:
        if value == revealed:
            return

        revealed = value

        if not is_node_ready():
            return

  		interact_area.disabled = not revealed
  		sprite_2d.visible = revealed
        physical_collider.disabled = not revealed

        if revealed and animate:
            animation_player.play("reveal")
        elif revealed:
			sprite_2d.modulate = Color.WHITE
		else:
			sprite_2d.modulate = Color.TRANSPARENT

    func reveal() -> void:
        set_revealed(true, true)
       	await animation_player.animation_finished

Unfortunately this is (intentionally) illegal in Godot's object system/GDScript: setters (and getters) cannot take optional parameters.

Another option is to replace the `var revealed: bool` with an enum with 3 states HIDDEN/REVEALING/REVEALED, rather than having two flags which can represent 4 states even though we only have 3 states. However `revealed` is part of the API of this script and it is used in many scenes so not worth breaking.

I would really welcome suggestions for a better API!

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -78,11 +81,18 @@ func _ready() -> void:
 
 ## Make the collectible appear
 func reveal() -> void:
+	if revealed or _revealing:
+		return
+
+	_revealing = true
 	revealed = true
```
manuq: I agree I'm not a fan of having `_revealing` and `revealed` being true at the same time!

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -78,11 +81,18 @@ func _ready() -> void:
...
+
 	appear_sound.play()
 	animation_player.play("reveal")
 	await animation_player.animation_finished
 
+	_revealing = false
```
manuq: I think there is a way to avoid the internal flag by checking the AnimationPlayer node: `animation_player.current_animation == "reveal"`

wjt: Ah interesting idea!

wjt: For that to work we would have to start the animation before setting `revealed` to `true`...

wjt: and that turns out to be much simpler!

wjt (commented):
Here's a different approach. Thanks for encouraging me to find a better way.

scenes/game_elements/props/collectible_item/components/collectible_item.gd
```diff
@@ -76,13 +76,23 @@ func _ready() -> void:
...
 
 
-## Make the collectible appear
+## Make the collectible appear with an animation, ultimately setting [member revealed].
+##
+## This does nothing if [member revealed] is [code]true[/code], or if the collectible is already
+## being revealed.
+##
+## To reveal the collectible immediately without any animation or sound, set [member revealed]
+## directly.
```
manuq, line 85: Great explanation!

manuq (approved):

PR #2544, opened 2026-07-14
GameState: Set scene when not persisting progress

### LINKED ISSUE #2543: Game state bug: Invalid access to property or key 'lights_on' on a base object of type 'Nil'
### Threadbare version

main

### Affected scene

Any with a ArtificialLightBehavior. For instance:
res://scenes/game_elements/components/custom_hookable_objects_test.tscn

### Description

Way to reproduce it:
- In the editor, go to menu Project -> Open User Data Folder
- Remove, if it exists, `saved_game.tres` (warning, you'll lose your progress)
- Then try to run a scene with lights, for instance: `res://scenes/game_elements/components/custom_hookable_objects_test.tscn`

The game crashes, the following error is reported in `ArtificialLightBehavior` `_ready()` method:
```
E 0:00:00:793   ArtificialLightBehavior._ready: Invalid access to property or key 'lights_on' on a base object of type 'Nil'.
  <GDScript Fuente>artificial_light_behavior.gd:51 @ ArtificialLightBehavior._ready()
  <Rastreo de Pila>artificial_light_behavior.gd:51 @ _ready()
```

Thanks @jgbourque for reporting it.

### Checklist

manuq:
When progress is not persisted, that is when the game runs from a specific scene that is not the main one, GameState.scene should be set to a new PerSceneState resource. Otherwise the game may fail to access its properties like lights_on.

Fix https://github.com/endlessm/threadbare/issues/2543

### DIFF
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -82,6 +82,7 @@ func _ready() -> void:
 	if not persist_progress:
 		if current_scene:
 			guess_quest(current_scene.scene_file_path)
+			set_scene(current_scene.scene_file_path)
 		# Grant all debug player abilities:
 		for ability: Enums.PlayerAbilities in DEBUG_PLAYER_ABILITIES:
 			player.set_ability(ability, true)


felixwalberg (approved):
This makes sense. Otherwise, the game would have no current scene and no way to reference information related to it, as you mentioned.

alcole2 (approved):

Issue #2533, opened 2026-07-13
Dev Archipelago: Add all existing test/demo/gym scenes to Elder's Storybook


felixwalberg:
- Every "toy" example scene we have for testing out individual mechanics or systems in Threadbare needs to be accessible via Dev Archipelago
- Add them all to the Storybook as quests that can be selected and started from the hub scene

PR #2545, opened 2026-07-14
Add hub for Dev Archipelago

### LINKED ISSUE #2531: Dev Archipelago: Design the Hub scene
- This is the scene that the access point teleporter from `frays_end.tscn` will bring you to 
- Design tilemap attempting the vibe of Archipelago, unique collections of game assets in one place (possibly a sign or simple dialogue explaining what this place is, or is it better to be discovered organically?)
- Existing Elder from the lore should be present with a storybook connected to them
- Pathways to QuestTeleporters with some test scenes (so we can evaluate the storybook vs physical distance approach)

felixwalberg:
This PR introduces a scene `dev_archipelago.tscn`, which will eventually be connected to Fray's End. The scene consists of a few islands, some of which are themed to correspond with mechanics in the game. Each island will have its own elder and set of quests leading to test scenes, gyms, and other playable documentation. The bridge behind the initial spawn point teleports the player back to `frays_end.tscn`.

Resolves #2531


felixwalberg:
Still need:
- QuestTeleporters for some example test scenes from each island
- Finish deciding themes for other islands
- Update Elder's storybooks (#2533)

felixwalberg:
This doesn't address all the needs of the Dev Archipelago hub, but I wanted to open it up to review before it gets too big!

scenes/game_elements/characters/npcs/elder/components/elder.gd
```diff
@@ -27,6 +27,11 @@ const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
 		abandon_point = new_value
 		update_configuration_warnings()
 
+## Optional dialogue passed to `TalkBehavior`. Used when
+## instantiating an elder in a scene that doesn't use the standard
+## eternal loom dialogue flow.
+@export var dialogue: Resource
+
```
wjt, line 34: The 3 elders in Fray's End actually already have different dialogues: they are each an instance of a different scene, so we can just set the TalkBehavior node's dialogue directly in each one.

Though if there are many elders in the archipelago I can see why you want to avoid continuing that trend. I think it's fine to do this. I wish Godot had a built-in way to export specific properties of child nodes from the root of the owning scene!

scenes/world_map/components/dev_archipelago_repel_elder.dialogue
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+# TODO: Name is up for debate
```
wjt: Personally I love it. I laughed out loud when I got the joke.

scenes/world_map/components/sheep_volleyball.gd
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+extends Node2D
+@onready var repel_timer: Timer = %RepelTimer
+@onready var repel_timer_2: Timer = $Sheep2/PlayerRepel2/RepelTimer2
```
wjt: I would suggest either consistently using `%` unique names, or using `@export`s. But...

scenes/world_map/components/sheep_volleyball.gd
```diff
@@ -0,0 +1,12 @@
...
+
+
+## Called when the node enters the scene tree for the first time.
+func _ready() -> void:
+	if repel_timer_2.is_node_ready():
```
wjt: ...this is guaranteed by the scene structure. A node becomes ready after all its children are ready. But...

scenes/world_map/components/sheep_volleyball.gd
```diff
@@ -0,0 +1,12 @@
...
+		await get_tree().create_timer(.5).timeout
+		repel_timer.start()
```
wjt: Since the other two timers are nodes in the scene, I would personally structure this with a third timer, with autostart set to True, timeout set to 0.5, oneshot set to true, and connect its `timeout` to `start` on `RepelTimer`. I think that makes the whole setup more consistent: it's all done in the scene tree.

scenes/world_map/components/dev_archipelago_basics_elder.dialogue
```diff
@@ -0,0 +1,17 @@
...
+Basics Elder: While I can't explain all of what happens here, I invite you to learn about the basics of our unraveling world!
+do balloon.hide()
+do show_storybook()
+if chosen_quest == null:
+	% Basics Elder: Feel free to explore these islands, however these quests might be a helpful way to start!
+	% Basics Elder: Explore and have fun! I hope to see you again if you need to relearn the basics.
+	% Basics Elder: The people of Threadbare are counting on you. Your understanding of our world is necessary for its survival!
+else:
+	% Basics Elder: Excellent choice, your training will surely help with your world-saving efforts.
+	% Basics Elder: Ah, those! Capable of such inginuity. To do what? I can't quite remember...
```
wjt:
```suggestion
	% Basics Elder: Ah, those! Capable of such ingenuity. To do what? I can't quite remember...
```

scenes/world_map/components/sheep_tug_of_war.gd
```diff
@@ -0,0 +1,45 @@
...
+	hook_control_2.state = 0
+
+
+## Called every frame. 'delta' is the elapsed time since the previous frame.
+func _process(_delta: float) -> void:
+	# TODO: Better way to avoid getting input from player aim/controls?
+	if hook_control_1.state == 1:
+		hook_control_1.state = 0
+	if hook_control_2.state == 1:
+		hook_control_2.state = 0
```
wjt: Could you not use `HookControl.State.AIMING` rather than `1` here?

scenes/world_map/components/sheep_tug_of_war.gd
```diff
@@ -0,0 +1,45 @@
...
+
+
+## Called when the node enters the scene tree for the first time.
+func _ready() -> void:
+	player_hook.remove_from_group("hook_listener")
+	player_hook_2.remove_from_group("hook_listener")
+	hook_control_1 = player_hook.find_child("HookControl")
+	hook_control_2 = player_hook_2.find_child("HookControl")
+	hook_control_1.state = 0
+	hook_control_2.state = 0
```
wjt: Interesting. That component really wasn't designed for non-interactive use! I think this is fine as a workaround but rather than the boilerplate "called when..." comments, I think it would be worth documenting here why these lines are needed (as below for the state).

scenes/world_map/components/sheep_tug_of_war.gd
```diff
@@ -0,0 +1,45 @@
...
+
+func _on_timer_timeout() -> void:
+	var sheep_options := [player_hook, player_hook_2]
+	sheep_options.shuffle()
+	var first_sheep: PlayerHook = sheep_options.pop_front()
+	var second_sheep: PlayerHook = sheep_options.pop_front()
+
+	first_sheep.hooked(hookable_box.find_child("HookableArea"), true)
+
+	await get_tree().create_timer(max(randf() * MAX_DELAY, MIN_DELAY)).timeout
```
wjt:
```suggestion
	await get_tree().create_timer(randf_range(MIN_DELAY, MAX_DELAY)).timeout
```

scenes/world_map/dev_archipelago.tscn
wjt, line 1: I love how this scene is taking shape.

Some minor nitpicking about foam:

Image

Red circle: as documented on https://github.com/endlessm/threadbare/wiki/TileMapLayers-and-TileSets#foam the fake perspective means that the top edge of a patch of grass raised "one level" above the water should not have foam.

Blue circle: if you look closely you can see the foam through the cliff, which is because there is no cliff underneath that patch of grass:

Image

wjt (approved):

wjt:
Feel free to fix the suggestions either in this PR before merging, or later.

scenes/game_elements/characters/npcs/elder/components/elder.gd
felixwalberg: I am happy to go back and change this if we end up making different scenes for each category's Elder. I was noticing it would be nice to have the ability to export child node properties as well. I'm not sure how difficult it would be, but I wonder if this would be worth me (on my own time) creating a request in the Godot repository...

scenes/world_map/components/sheep_volleyball.gd
felixwalberg: This makes sense! Removes a script with the single purpose of connecting signals.

scenes/game_elements/characters/npcs/elder/components/elder.gd
wjt: There are lots of existing proposals around this idea, e.g. from a quick search:

- https://github.com/godotengine/godot-proposals/issues/7803
- https://github.com/godotengine/godot-proposals/issues/14223
- https://github.com/godotengine/godot-proposals/issues/9536
- https://github.com/godotengine/godot-proposals/issues/8266

felixwalberg:
> Feel free to fix the suggestions either in this PR before merging, or later.

Thank you for this feedback! It's been helpful to look through.

Issue #2563, opened 2026-07-16
Create smaller project bundle for StoryQuest creation


wjt:
Our learning team would like a smaller Godot project for learners creating StoryQuests, both to reduce the mental burden of navigating the full project, and to improve performance with Backstitch (and reduce the burden on the server).

Here is my suggestion:

1. Delete all StoryQuests. We know that this can be done without breaking the surrounding game because I did this for the LoreJam a few months ago.
2. Delete large unused assets – for example, we have a full copy of the non-CC0 Tiny Swords free pack in the repo (so that we can record the special permission we have been given to redistribute it) but it is not used (except for 4 rocks in `champ`).
3. Delete the world map (i.e. Fray's End and connected scenes) and all LoreQuests. Keep Dev Archipelago.
4. Change the title screen to launch the NO_EDIT template quest rather than the now-deleted tutorial. Change the "Abandon Quest" and end-of-quest target scene to be the Dev Archipelago hub scene.
5. Rename "NO_EDIT" to "PLEASE_EDIT" (name tbd!).
6. Add the backstitch launcher.
7. Put this in a .zip file

Now learners would use this as the basis for their quest. They use backstitch not git. They don't duplicate the template, they just edit it in place. The quest is a standalone game. Later we tackle how to launch them dynamically; but more likely, we just import the good ones back into the git tree (changing the UIDs in the quest scenes/resources when we do so because they will be the same as the NO_EDIT quest).

We would not prepare this bundle for every commit. Instead we would generate it for each release and attach it to the release. Learners would be pointed to this. This would avoid the current problem of learners getting a random snapshot of whatever happens to be on git `main` at the time their fork is created.

jgbourque:
@hydrolet @wjt This sounds great. The one thing we want to be OK with here is that the learners making quests from here on is that the learners will be working separately from the full game. They can of course still access, play, and download the full game but their quest designs happen separately. 

Is the intention to try and keep character functionality, scripts, and game systems used in the LoreQuests preserved inside the Dev Islands or at least in the smaller SQ bundle so that learners will not be without them? Is that possible?

wjt:
> Is the intention to try and keep character functionality, scripts, and game systems used in the LoreQuests preserved inside the Dev Islands or at least in the smaller SQ bundle so that learners will not be without them?

Yes. Without these the game would not work. These already live outside the quests & world map.

wjt:
[...]

PR #2566, opened 2026-07-20
Sokoban: Use SceneLink

wjt:
Previously the Sokoban `RuleEngine` had its own `next_scene` property and
manually triggered a switch to that scene.

Add a `SceneLink` to each level (and the template), configured to match
`RuleEngine.next_scene`. Use the same left-to-right wipe when switching
round-to-round; use a radial transition when returning to Fray's End.

Connect `RuleEngine.goal_reached` to `SceneLink.switch()`. Emit that
signal when skipping as well as when the goal is actually reached.
Remove `RuleEngine.next_scene`.

### DIFF
--- a/scenes/eternal_loom_sokoban/components/system/rules/rule_engine.gd
+++ b/scenes/eternal_loom_sokoban/components/system/rules/rule_engine.gd
@@ -25,9 +25,6 @@ const RULE_REPEAT_LIMIT := 99
 ## Every tag being used by pieces need to be set here to be recognized as tags.
 @export var tags_legend: Array[StringName] = ["controllable"]
 
-## When all goals are met, the RuleEngine will attempt to load this scene next.
-@export_file("*.tscn") var next_scene: String
-
 ## Offer skipping to the next scene after this amount of seconds.
 @export var seconds_to_offer_skip: int = 120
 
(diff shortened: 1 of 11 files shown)


wjt:
The Sokobans aren't actually accessible in-game at the moment of course.

This is a prelude to removing direct references to Fray's End by adding a special "go home" mode to SceneLink as described in https://github.com/endlessm/threadbare/issues/2563#issuecomment-5023272416

scenes/eternal_loom_sokoban/components/system/rules/rule_engine.gd
```diff
@@ -97,9 +94,8 @@ func _reset() -> void:
 
 
 func _skip() -> void:
-	if not next_scene:
-		return
-	SceneSwitcher.change_to_file_with_transition(next_scene)
+	goals_reached.emit()
+	directional_input.enabled = false
```
manuq, line 98: Oh yes, emitting the signal is better than using the SceneSwitcher global directly.

scenes/eternal_loom_sokoban/levels/eternal_loom_template.tscn
```diff
@@ -57,7 +58,15 @@ tile_set = ExtResource("5_mdri7")
...
 stream = ExtResource("9_mdri7")
+
+[node name="SceneLink" type="Node2D" parent="." unique_id=1011300014]
+script = ExtResource("9_5hoht")
+next_scene = "uid://c80o1bys5radq"
+enter_transition = 1
+exit_transition = 1
+metadata/_custom_type_script = "uid://d2vp62mjrf6gn"
+
+[connection signal="goals_reached" from="EternalLoomRuleset" to="SceneLink" method="switch"]
```
manuq, line 72: 👍

manuq (approved):

Issue #2250, opened 2026-05-26
“Change the main scene to your quest's first scene” doesn't work quite right with GameState


wjt:
A common thing we suggest to people making a StoryQuest is that they should set the main scene in their fork to the first scene of their quest, so that running the game launches directly into their StoryQuest.

@Snowsd reported an interesting problem in a fork of Threadbare that boiled down to a combination of https://github.com/endlessm/threadbare/issues/2244 plus some weird behaviour in the GameState code when someone follows this advice. Try running [this branch](https://github.com/Snowsd/threadbareVRmontCup2026/tree/bill-the-pill) (browser or in editor, it doesn't matter), and exit the game once some introductory dialogue has played but before finishing that scene. Then you will run into #2244, but you will also be unable to break out of it because there is no title screen to choose "new game" and erase the game state.

What I couldn't understand at first is why the game state was being saved at all, and then why the flag wasn't being cleared when changing scenes. Here's what I figured out...

----

The `intro_dialogue_shown` flag is cleared here:

```gdscript
## Set the scene path and [member current_spawn_point] without triggering a save.
func _do_set_scene(scene_path: String, spawn_point: NodePath = ^"") -> void:
	if get_scene_to_restore() != scene_path:
		intro_dialogue_shown = false
	# ...
```

`get_scene_to_restore()` is defined as:

```gdscript
## If there is a scene to restore, return it.
func get_scene_to_restore() -> String:
	return _state.get_value(QUEST_SECTION, QUEST_CURRENTSCENE_KEY, "")
```

`_state` is initialised in `GameState._ready`:

```gdscript
func _ready() -> void:
	var current_scene := get_tree().current_scene
	var initial_scene_uid := (
		ResourceLoader.get_resource_uid(current_scene.scene_file_path) if current_scene else -1
	)
	var main_scene_uid := ResourceLoader.get_resource_uid(
		ProjectSettings.get_setting("application/run/main_scene")
	)
	persist_progress = initial_scene_uid == main_scene_uid
	if not persist_progress:
		if current_scene:
			guess_quest(current_scene.scene_file_path)
		return

	var err := _state.load(GAME_STATE_PATH)
	if err != OK and err != ERR_FILE_NOT_FOUND:
		push_error("Failed to load %s: %s" % [GAME_STATE_PATH, err])
	# ...
```

[...]

PR #2568, opened 2026-07-20
Store initial quest & home scene in project settings

wjt:
Previously the initial quest (i.e. the tutorial) was configured as a
property of the title screen scene; and the path to what I am now
calling the "home scene" (i.e. Fray's End) was configured as a property
of the pause menu.

I want these to be stored in a more neutral location, so that they can
more easily be modified in alternative builds of Threadbare.

Add them as custom project settings. Add a helper function to fetch
them, since default-valued settings are not stored so
`ProjectSettings.get()` returns `null` in that case. Update the pause &
title screens to fetch these paths from settings.

Helps https://github.com/endlessm/threadbare/issues/2250
Helps https://github.com/endlessm/threadbare/issues/2563

### DIFF
--- a/addons/threadbare_project_settings/threadbare_project_settings.gd
+++ b/addons/threadbare_project_settings/threadbare_project_settings.gd
@@ -4,6 +4,14 @@
 class_name ThreadbareProjectSettings
 extends Node
 
+## The scene to return to when a quest is completed, skipped, or abandoned.
+## (This is Fray's End in normal builds of Threadbare.)
+const HOME_SCENE = "threadbare/general/home_scene"
+
+## The [Quest] to start from the “New Game” title screen item. (This is the
+## tutorial in normal builds of Threadbare.)
+const OPENING_QUEST = "threadbare/general/opening_quest"
+
 ## Display a letterbox overlay in the game, to debug aspect ratio
 const DEBUG_ASPECT_RATIO = "threadbare/debugging/debug_aspect_ratio"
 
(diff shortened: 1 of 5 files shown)


addons/threadbare_project_settings/threadbare_project_settings.gd
```diff
@@ -42,3 +64,11 @@ static func setup_threadbare_settings() -> void:
...
 		ProjectSettings.set_as_basic(setting_name, not setting_config.get("is_advanced", false))
 		ProjectSettings.set_as_internal(setting_name, setting_config.get("is_hidden", false))
+
+
+## Gets a Threadbare-specific setting, returning its default value if unset in
+## the project settings. [param key] should be one of the constants defined in
+## [ThreadbareProjectSettings].
+static func get_setting(key: String) -> Variant:
+	assert(key in settings_configuration)
+	return ProjectSettings.get_setting(key, settings_configuration[key].value)
```
manuq, line 74: Good, we can use this in the previous Threadbare-specific settings like in `aspect_ratio_debugger.gd` (after merge is fine!).

manuq (approved):

manuq:
@wjt this is great! Are you considering a special-case property in SceneLink for StoryQuests to all go back to the home scene in project settings? Although not needed for #2563.

image

wjt:
Precisely

wjt:
It actually is needed for #2563 because I want to keep the NO_EDIT template (whose last scene points to Fray's End) but delete Fray's End.

addons/threadbare_project_settings/threadbare_project_settings.gd
wjt: We could do. I didn't make that change because the default being `null` is fine for those boolean properties - it's false-y.

PR #2569, opened 2026-07-20
SceneLink: Add a special case for "go home"

wjt:
Previously, the last scene of every quest (and Sokoban trio) hardcoded
the path (or rather UID) to Fray's End.

I want to make a stripped-back version of the repo that does not contain
Fray's End or any StoryQuests or LoreQuests as a StoryQuest kit; yet the
last scene of the template quest hardcodes the UID to Fray's End. I want
to make it possible to change the scene that follows a quest being
completed without modifying every quest in the game.

Add a flag to SceneLink (represented with an `enum` with two values for
future extensibility and self-documentation reasons) which, if set,
means that the link points to the "home scene" configured in the project
settings rather than a specific path/UID.

Update the final scene of all quests, and all Sokoban trios, to use this
new flag rather than hardcoding the Fray's End UID, with one exception.
I kept the explicit reference to Fray's End in
`scenes/quests/lore_quests/quest_000/6_grappling/tutorial_grappling.tscn`
because it also sets a particular spawn point, and doing that doesn't
make sense with a generic "home scene" link.

Helps https://github.com/endlessm/threadbare/issues/2563

### DIFF
--- a/scenes/globals/scene_switcher/scene_link.gd
+++ b/scenes/globals/scene_switcher/scene_link.gd
@@ -8,10 +8,29 @@ extends Node2D
 ## This can be used directly, but acts primarily a base class for other
 ## components: [Cinematic], [CollectibleItem], and the Teleporter scene.
 
+enum Target {
+	## The scene specified in [member SceneLink.next_scene].
+	SPECIFIED_SCENE,
+	## The scene from [constant ThreadbareProjectSettings.HOME_SCENE], typically
+	## Fray's End. Use this at the end of quests.
+	HOME_SCENE,
+}
+
+@export var target: Target = Target.SPECIFIED_SCENE:
+	set(new_value):
+		target = new_value
+		_update_available_spawn_points()
+
 ## Scene to switch to when the player enters this teleport. If empty, the player
 ## will teleport within the current scene, to the position specified by [member
 ## spawn_point_path].
 @export_file("*.tscn") var next_scene: String:
+	get():
+		if target == Target.HOME_SCENE:
+			return ThreadbareProjectSettings.get_setting(ThreadbareProjectSettings.HOME_SCENE)
+
+		return next_scene
+
 	set(new_value):
 		next_scene = new_value
 		_update_available_spawn_points()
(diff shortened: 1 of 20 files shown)


scenes/globals/scene_switcher/scene_link.gd
```diff
@@ -8,10 +8,29 @@ extends Node2D
 ## This can be used directly, but acts primarily a base class for other
 ## components: [Cinematic], [CollectibleItem], and the Teleporter scene.
 
+enum Target {
+	## The scene specified in [member SceneLink.next_scene].
+	SPECIFIED_SCENE,
+	## The scene from [const ThreadbareProjectSettings.HOME_SCENE], typically
+	## Fray's End. Use this at the end of quests.
+	HOME_SCENE,
+}
```
wjt, line 17: Ugly names. Alternative names welcome.

Image

Image

manuq: I was considering just a "go home" boolean to override the specific scene, but this is more future-proof, it covers potential new special cases. So 👍

manuq (approved):
Looks great. The only caveat I see is that by using the home scene there is no way to specify the spawn point.

Image

Image

In the description I read:
> I kept the explicit reference to Fray's End in
scenes/quests/lore_quests/quest_000/6_grappling/tutorial_grappling.tscn
because it also sets a particular spawn point, and doing that doesn't
make sense with a generic "home scene" link.

Seems a conscious design decision, so 👍

wjt:
Ah... Actually that's not entirely intentional, I thought it worked for the spawn points in Fray's End. But perhaps I should disable the spawn point property entirely in "go home" mode.

wjt:
Works for me.

image

I think you were testing a version of this branch without #2567 - which fixed exactly that pre-existing problem of the spawn point path choices not getting listed for Cinematic.

manuq:
> Works for me.
> 
> image
> I think you were testing a version of this branch without #2567 - which fixed exactly that pre-existing problem of the spawn point path choices not getting listed for Cinematic.

That is great! Yes probably I did `gh pr checkout` too soon.

PR #2574, opened 2026-07-21
NO_EDIT combat: Switch music when combat finishes

wjt:
This is the same AudioStreamInteractive configuration as used in the
LoreQuest combat scenes, but inlined into the scene. It seems useful to
have an example of how to do this.


wjt:
Unrelated to the task I was working on, but I realised this was a good opportunity to demo this component. :)

manuq (approved):
Nice! I personally welcome this demo :)

PR #2576, opened 2026-07-21
Add scripts to make a pruned “StoryQuest Kit” bundle

### LINKED ISSUE #2563: Create smaller project bundle for StoryQuest creation

wjt:
Rather than having Core: Threadbare participants start from a fork of
the entire repo, we want to try giving participants a stripped-down copy
of the project - specifically with all LoreQuests and other StoryQuests
removed. The goals are manifold:

- Reduce the cognitive burden of navigating a large project.

- Reduce the need to take special steps to deep link to a StoryQuest for
  testing: the team's StoryQuest should be launched from the title
  screen.

- Reduce the number and size of the files in the project, saving
  electrons and helping with performance in Backstitch.

- Move us towards a model where rather than expecting that all
  storyquests will be merged upstream (requiring learners to take
  special care to only modify files in their folder, increasing
  complexity; and hitting human scalability cliffs in the form of there
  only being two full-time maintainers of the game reviewing StoryQuest
  submissions) we instead treat them more as stand-alone games. For the
  ones we choose to merge upstream, we can do the necessary work to
  confine them to their own folder at that point; and you could imagine
  adding a way to launch a separately-downloaded .pck for a StoryQuest.

As well as removing the lore_quests and story_quests folders, we also
have to remove the world_map folder (i.e. Fray's End and connected
scenes) because these scenes now link directly to LoreQuests.

I also decided to remove the non-CC0 Tiny Swords free pack, which we
have special permission to redistribute but which is not used outside of
one StoryQuest. This significantly reduces the number of files in the
project, which is helpful for Backstitch.

Having removed these files, we adjust the project settings so that the
quest launched from the title screen is the NO_EDIT template, and the
scene to return to when a quest is completed or abandoned is Dev
Archipelago.

In Dev Archipelago, add the Template Elder so that the NO_EDIT template
can be accessed. Add the StoryQuest Elder too, because otherwise the
team's SQ can't be played once completed or abandoned without erasing
the game state, but remove him in normal versions. Add an alternative
broken version of the bridge back to Fray's End which blocks access to
that teleporter. Reconfigure that teleporter to point to the home scene
rather than explicitly to Fray's End so that if it is reached it is a
no-op.

One approach would be to have learners modify the NO_EDIT template
in-place, in which case we would want to rename all those files and
delete the bootstrap tool. For now I just updated the StoryQuest
bootstrap tool to change the initial quest to the newly-duplicated one.

I checked for references into these folders from elsewhere in the game
as follows:

    git grep -E \
        '(lore_quests|story_quests|world_map|tiny-swords-non-cc0|uid://cufkthb25mpxy)' \
        ':!scenes/quests/lore_quests' \
        ':!scenes/quests/story_quests' \
        ':!scenes/world_map' \
        ':!assets/third_party/tiny-swords-non-cc0'

In a perfect world there would be no output from this command. There are
some lingering references: for example, the elders' dialogue references
the paths to various `quest.tres` resources. However there are (as far
as I can tell) no references that will cause a scene or resource to fail
to load. (This command is imperfect because, for example, teleporters
refer to other scenes by their UID stored as a string; that's why Fray's
End's uid is in the command above.)

Creating the bundle is implemented with 2-3 scripts:

1. A GDScript script which deletes files and adjusts project settings,
   which is most easily done using Godot's own APIs (and reuses some
   preexisting logic).

2. A shell script which drives the GDScript script, then does some
   additional stuff (downloading the backstitch launcher, running Git
   commands) which is more easily accomplished in shell.

3. A GitHub Actions workflow which drives the shell script and uploads
   the artifact.

I couldn't resist calling the script StoryQuest Construction Kit
Construction Kit (SQCKCK for short) but this is a placeholder name!

Resolves https://github.com/endlessm/threadbare/issues/2563

### DIFF
--- a/.github/workflows/make-storyquest-kit.yml
+++ b/.github/workflows/make-storyquest-kit.yml
@@ -1,0 +1,70 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+name: "Make StoryQuest Kit"
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'tools/sqckck.*'
+      - '.github/workflows/make-storyquest-kit.yml'
+  release:
+    types:
+      - published
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+env:
+  GODOT_VERSION: "4.7.1"
+
+jobs:
+  make-storyquest-kit:
+    name: Make StoryQuest Kit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          # Fetch full history of commits and tags so that git describe --tags works
+          fetch-depth: 0
+
+      - uses: ./.github/actions/lfs-cached-checkout
+
+      - name: Install Godot
+        uses: endlessm/godot-export-action/download-godot@v3
+        with:
+          godot_version: ${{ env.GODOT_VERSION }}
+
+      - name: Restore .godot directory cache
+        uses: actions/cache/restore@v6
+        with:
+          path: |
+            .godot
+          key: ${{ runner.os }}-dotgodot-${{ inputs.godot_version }}.${{ inputs.godot_flavor }}-${{ hashFiles('**/*.import') }}
+          restore-keys: |
+            ${{ runner.os }}-dotgodot-${{ inputs.godot_version }}.${{ inputs.godot_flavor }}-
+
+      - name: Make StoryQuest Kit
+        id: sqckck
+        run: ./tools/sqckck.sh
+        env:
+          GODOT: build/godot
+          GIT_AUTHOR_NAME: Threadbare StoryQuest Kit Maker
+          GIT_COMMITTER_NAME: Threadbare StoryQuest Kit Maker
+          GIT_AUTHOR_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+          GIT_COMMITTER_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+
+      - name: Upload kit artifact
+        uses: actions/upload-artifact@v7
+        with:
+          path: ${{ steps.sqckck.outputs.zip_file }}
+          archive: false
+
+      - name: Attach to release
+        if: ${{ github.event_name == 'release' }}
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          set -ex
+          gh release upload '${{ github.ref_name }}' "${{ steps.sqckck.outputs.zip_file }}" --repo '${{ github.repository }}'
--- a/addons/storyquest_bootstrap/copier.gd
+++ b/addons/storyquest_bootstrap/copier.gd
@@ -263,5 +263,14 @@ func create_storyquest() -> void:
 	copy_tilesets()
 
 	var quest: StoryQuest = load(TEMPLATE_PATH.path_join(QUEST_FILENAME))
-	await copy_resource(quest)
+	var copied: StoryQuest = await copy_resource(quest)
 	EditorInterface.save_all_scenes()
+
+	var opening_quest := ThreadbareProjectSettings.get_setting(
+		ThreadbareProjectSettings.OPENING_QUEST
+	)
+	if ResourceUID.ensure_path(opening_quest) == ResourceUID.ensure_path(quest.resource_path):
+		ProjectSettings.set(
+			ThreadbareProjectSettings.OPENING_QUEST, ResourceUID.path_to_uid(copied.resource_path)
+		)
+		ProjectSettings.save()
--- a/addons/storyquest_bootstrap/copier.gd
+++ b/addons/storyquest_bootstrap/copier.gd
@@ -266,6 +266,8 @@ func create_storyquest() -> void:
 	var copied: StoryQuest = await copy_resource(quest)
 	EditorInterface.save_all_scenes()
 
+	# If the NO_EDIT template is configured as the opening quest, make this copy
+	# the opening quest instead.
 	var opening_quest := ThreadbareProjectSettings.get_setting(
 		ThreadbareProjectSettings.OPENING_QUEST
 	)
--- a/scenes/dev/dev_archipelago.tscn
+++ b/scenes/dev/dev_archipelago.tscn
@@ -1,23 +1,29 @@
 [gd_scene format=4 uid="uid://bhdpwut7lgfnm"]
 
+[ext_resource type="Script" uid="uid://db0hjoby7su6n" path="res://scenes/dev/components/dev_archipelago.gd" id="1_kt7u4"]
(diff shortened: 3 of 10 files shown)


wjt:
@jgbourque **Still a work in progress** but please have a read of the description above then take a look at https://drive.google.com/file/d/1-6nJkNDaWB57XZlTyY2YZhawQKTvSNQ_/view?usp=sharing

wjt:
The point of committing the pruned build to Git is:

1. If we keep that commit somewhere, then if we do want to import a quest built from this kit into the upstream repo, it will be much easier to diff!
2. I want to keep open the option of using Git to make a StoryQuest. What I have in mind is that our magic repo-forking script would not base the new repo off the upstream `main` branch but off a branch containing the latest storyquest kit. This new repo would be detached from `endlessm/threadbare` (implementation TBD) so would operate as its own separate universe.

.github/workflows/make-storyquest-kit.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          # Fetch full history of commits and tags so that git describe --tags works
+          fetch-depth: 0
+
+      - uses: ./.github/actions/lfs-cached-checkout
+
+      - name: Install Godot
+        uses: endlessm/godot-export-action/install-godot@wjt/split-setup-steps
```
wjt: Need to merge this change over in that repo...

jgbourque:
Just tried out the new StoryQuest Kit and it worked really well with Backstitch. I was able to create all my own files for an SQ using the "create SQ from template" script and there were no hiccups or delays in syncing those changes or any of the small changes I made afterwards in my testing.

jgbourque (approved):

scenes/dev/dev_archipelago.tscn
```diff
@@ -102,6 +121,41 @@ text = "Fray's End"
 [node name="Elders" type="Node2D" parent="OnTheGround" unique_id=452918502]
 y_sort_enabled = true
 
+[node name="SQ" type="Node2D" parent="OnTheGround/Elders" unique_id=2119212521]
```
wjt, line 124: Adding 2 more elders to that first island makes it more confusing. Maybe I need to redesign the level more. I am tempted though to keep it as-is and then iterate in a subsequent PR

tools/sqckck.gd
```diff
@@ -0,0 +1,50 @@
...
+
+
+func rm_rf(path: String) -> void:
+	# You can only delete empty folders, so using DirAccess.remove() means
+	# traversing the whole folder tree and deleting each leaf before its parent
+	# folder. But you can trash a non-empty folder in a single step!
+	var global_path := ProjectSettings.globalize_path(path)
+	var ret := OS.move_to_trash(global_path)
+	if ret != OK:
+		push_error("Failed to trash ", global_path, ": ", error_string(ret))
```
wjt, line 31: Could also do this in the shell script, might be easier...

wjt (commented):
Some notes to self.

manuq:
> Add the StoryQuest Elder too, because otherwise the
> team's SQ can't be played once completed or abandoned without erasing
> the game state, but remove him in normal versions.

Makes sense, I haven't thought about this!

manuq:
> I couldn't resist calling the script StoryQuest Construction Kit Construction Kit (SQCKCK for short) but this is a placeholder name!

I love the name and its short version SQCKCK, please keep it.

addons/storyquest_bootstrap/copier.gd
```diff
@@ -263,5 +263,14 @@ func create_storyquest() -> void:
...
 
 	var quest: StoryQuest = load(TEMPLATE_PATH.path_join(QUEST_FILENAME))
-	await copy_resource(quest)
+	var copied: StoryQuest = await copy_resource(quest)
 	EditorInterface.save_all_scenes()
+
+	var opening_quest := ThreadbareProjectSettings.get_setting(
+		ThreadbareProjectSettings.OPENING_QUEST
+	)
+	if ResourceUID.ensure_path(opening_quest) == ResourceUID.ensure_path(quest.resource_path):
```
manuq, line 274: This would welcome an explanation:

```suggestion
	# If the NO_EDIT template is the opening quest in Project Settings, make this copy the opening quest. 
	if ResourceUID.ensure_path(opening_quest) == ResourceUID.ensure_path(quest.resource_path):
```

tools/sqckck.gd
```diff
@@ -0,0 +1,50 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends MainLoop
+## StoryQuest Construction Kit Construction Kit
```
manuq, line 5: 💯

tools/sqckck.gd
```diff
@@ -0,0 +1,50 @@
...
+	"res://scenes/world_map",
+]
+
+
+func rm_rf(path: String) -> void:
+	# You can only delete empty folders, so using DirAccess.remove() means
+	# traversing the whole folder tree and deleting each leaf before its parent
+	# folder. But you can trash a non-empty folder in a single step!
+	var global_path := ProjectSettings.globalize_path(path)
+	var ret := OS.move_to_trash(global_path)
```
manuq, line 29: I didn't know about `OS.move_to_trash` and how it's a single command without need of traversing, interesting!

tools/sqckck.gd
manuq: Yes an actual  `rm -rf`. But this works!

manuq (approved):
Excellent! Great work.

tools/sqckck.gd
wjt: Now that I have a shell script wrapping the gdscript, it would be easier to use `rm -rf`, but this is here and works, so..

PR #2580, opened 2026-07-21
Change PlayerState abilities for Dev Archipelago quests

felixwalberg:
Quests accessible from Dev Archipelago should start the player with `DEBUG_PLAYER_ABILITIES` (basic repel and grapple). Before this change, entering a quest in Dev Archipelago would reset the `PlayerState`, removing all abilities from the player. 

Makes #2533 easier

### DIFF
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -118,7 +118,7 @@ func restore_quest() -> void:
 ## Note that this does not switch scenes into the quest.
 func set_quest(new_quest: Quest) -> void:
 	var quest_player_state: PlayerState
-	if new_quest is LoreQuest:
+	if new_quest is not StoryQuest:
 		# Duplicate the current global player state. If the quest is completed,
 		# it will be copied back; if abandoned, it will be discarded.
 		quest_player_state = global.player.duplicate()
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -112,7 +112,6 @@ func restore_quest() -> void:
 		# Add any lore abilities that the player gained since suspending this
 		# quest.
 		quest.player.abilities |= global.player.abilities
-	# TODO: There might be an edge case for resuming a dev quest that contains modifiers
 	elif quest.quest is not StoryQuest:
 		for ability: Enums.PlayerAbilities in DEBUG_PLAYER_ABILITIES:
 			quest.player.set_ability(ability, true)
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -112,6 +112,10 @@ func restore_quest() -> void:
 		# Add any lore abilities that the player gained since suspending this
 		# quest.
 		quest.player.abilities |= global.player.abilities
+	# TODO: There might be an edge case for resuming a dev quest that contains modifiers
+	elif quest.quest is not StoryQuest:
+		for ability: Enums.PlayerAbilities in DEBUG_PLAYER_ABILITIES:
+			player.set_ability(ability, true)
 
 
 ## Sets [member quest], setting up a new [PlayerState] if necessary.
@@ -125,9 +129,10 @@ func set_quest(new_quest: Quest) -> void:
 		quest_player_state.reset_lives()
 	elif new_quest is not StoryQuest:
 		# For quests that are not LoreQuests or StoryQuests (e.g. dev archipelago quests)
-		# grant all debug player abilities.
+		# use a fresh player state and grant all debug player abilities.
+		quest_player_state = PlayerState.new()
 		for ability: Enums.PlayerAbilities in DEBUG_PLAYER_ABILITIES:
-			player.set_ability(ability, true)
+			quest_player_state.set_ability(ability, true)
 	else:
 		# Use a fresh player state for StoryQuests
 		quest_player_state = PlayerState.new()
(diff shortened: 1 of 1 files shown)


manuq:
> Quests accessible from Dev Archipelago should be able to copy the player state rather than getting grouped in with `StoryQuest` to get reset to a fresh instance. This allows the player to maintain all abilities when entering any Dev Archipelago Quest.
> 
> Helps #2533

I'm not entirely sure about this! It is for making Dev Archipielago feel part, an extension of, the world map? Because I guess each island that demos an ability would like to set the player in that ability. How the upcoming "grappling hook gym" would demo the longer thread, if the player (in the game lore) hasn't completed the quest that provides that ability modifier?

felixwalberg:
> How the upcoming "grappling hook gym" would demo the longer thread, if the player (in the game lore) hasn't completed the quest that provides that ability modifier?

This is a good point. I was thinking that upon entering `dev_archipelago.tscn` all of the abilities would be "unlocked" including the modifiers so that the user can instantly use each of them. 

Would it make more sense to have a more nuanced approach where we associate a subset of abilities with each "dev quest" that modify the `player_state` when it is selected? This might be useful if there are certain testing scenes where we *don't* want the user to have certain abilities or modifiers (until they possibly interact with them in the level).

felixwalberg:
>  if there are certain testing scenes where we _don't_ want the user to have certain abilities or modifiers (until they possibly interact with them in the level).

I feel like this is the bigger question. It would certainly be possible for me to add quest-specific abilities for each one so that we don't have to change `game_state.gd` as a whole. I think it definitely makes sense that the available abilities in quests accessible from the Archipelago should not be associated with how much progress the player has made on the LoreQuests

manuq:
> This is a good point. I was thinking that upon entering `dev_archipelago.tscn` all of the abilities would be "unlocked" including the modifiers so that the user can instantly use each of them.

The problem is that by unlocking all modifiers the player wouldn't be able to test the ability without it. The only case today is Grapple. How far can you grapple normally should also be tested in the grappling hook gym.

> Would it make more sense to have a more nuanced approach where we associate a subset of abilities with each "dev quest" that modify the `player_state` when it is selected? This might be useful if there are certain testing scenes where we _don't_ want the user to have certain abilities or modifiers (until they possibly interact with them in the level).

Precisely. I wouldn't mind the dev islands cheating. For instance, the `unify_abilities_test.tscn` grants the 2 abilities with powerups:

image

Now, since then we made the game grant these 2 abilities when not running the game from the main scene (check `DEBUG_PLAYER_ABILITIES` in `game_state.gd`).

The abilities can also be granted or removed with scripting (`GameState.player.set_ability(ability, true/false)`) but I think adding powerups to the scene better exemplifies what should be done in levels.

manuq:
@felixwalberg just to be clear: I think you should assume that ABILITY_A (Repel) and ABILITY_B (Grapple) are already obtained, and grant ABILITY_B_MODIFIER_1 when necessary in a dev island, with a powerup.

manuq:
> @felixwalberg just to be clear: I think you should assume that ABILITY_A (Repel) and ABILITY_B (Grapple) are already obtained, and grant ABILITY_B_MODIFIER_1 when necessary in a dev island, with a powerup.

And to make this happen, maybe you should change this pull request to be:
```
	if quest.quest is LoreQuest:
		# Add any lore abilities that the player gained since suspending this
		# quest.
		quest.player.abilities |= global.player.abilities
	elif quest.quest is not StoryQuest:
		# Not a LoreQuest and not a StoryQuest, grant all debug player abilities:
		for ability: Enums.PlayerAbilities in DEBUG_PLAYER_ABILITIES:
			player.set_ability(ability, true)

```

felixwalberg:
I will change it to something like what you suggested. 

I realize that the reason `DEBUG_PLAYER_ABILITIES` was not being used is that I was testing from `dev_archipelago.tscn` and then using the StoryBook to start a quest. Your fix will happen each time we start a quest, which should solve it!

manuq:
> I will change it to something like what you suggested.
> 
> I realize that the reason `DEBUG_PLAYER_ABILITIES` was not being used is that I was testing from `dev_archipelago.tscn` and then using the StoryBook to start a quest. Your fix will happen each time we start a quest, which should solve it!

:+1: for the sake of completeness I see 2 options here:
- Either grant DEBUG_PLAYER_ABILITIES which also happen to match the abilities obtained after completing the tutorial quest, or granted after skipping it. Add a powerup for each extra ability or modifier as needed in each dev island.
- Or start without any ability (like StoryQuests) and add a powerup to the dev island, when the ability is needed.

I'm inclined to the former, but is up to discuss.

image

manuq:
> Remaining work: resuming a Dev Archipelago quest that involves acquiring an ability modifier will currently not save.

Do you see this as an issue? Dev Archipelago quests are demos, so I wouldn't expect them to persist any progress, unless the demo is precisely about game state.

scenes/globals/game_state/game_state.gd
```diff
@@ -112,6 +112,10 @@ func restore_quest() -> void:
 		# Add any lore abilities that the player gained since suspending this
 		# quest.
 		quest.player.abilities |= global.player.abilities
+	# TODO: There might be an edge case for resuming a dev quest that contains modifiers
```
manuq: It looks to me that this comment and the one below is assuming upcoming work. It would be better to have dev quests first, before doing these considerations.

manuq (approved):

felixwalberg:
> Do you see this as an issue? Dev Archipelago quests are demos, so I wouldn't expect them to persist any progress, unless the demo is precisely about game state.

Fair point, I agree that persisting progress isn't really essential to any of the demos. In that case, this should cover the first option you mentioned:

> grant DEBUG_PLAYER_ABILITIES which also happen to match the abilities obtained after completing the tutorial quest, or granted after skipping it. Add a powerup for each extra ability or modifier as needed in each dev island.

PR #2581, opened 2026-07-22
Main menu: Restart logo animation when displayed

wjt:
Previously the logo was animated once, then stayed on its last frame.
I think it's nice to reshow the animation when moving back from the
credits or options screens.

Add the necessary API to AnimatedTextureRect to make this work.

### DIFF
--- a/scenes/menus/storybook/components/animated_texture_rect.gd
+++ b/scenes/menus/storybook/components/animated_texture_rect.gd
@@ -6,9 +6,9 @@ extends TextureRect
 ## A [TextureRect] which shows a single animation from a [SpriteFrames].
 ##
 ## This is useful because the only built-in way to display an animation in a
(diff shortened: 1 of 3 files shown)


wjt:
I'm happy to be disagreed with on this one! I have just had versions of this patch lying around in my local clone for months, so decided to finish it to see if others agree that this is good.

manuq:
> I'm happy to be disagreed with on this one! I have just had versions of this patch lying around in my local clone for months, so decided to finish it to see if others agree that this is good.

I find it a bit too much to also play the animation when going back from Options or Credits, to be honest. But is only a matter of taste. Feel free to merge it!

wjt:
Ok, I trust your judgement! In that case I will revise this PR to keep the new features in AnimatedTextureRect but preserve the old behaviour

wjt:
Superseded by #2607

PR #2585, opened 2026-07-22
Add teleporter demo scene for Dev Archipelago

### LINKED ISSUE #2534: Dev Archipelago: Teleporter
- Design a pair of scenes that highlight the behavior of the `teleporter.tscn`
- 1: A teleporter that moves the player from one scene to another
- 2: A teleporter that moves the player to another position within the same scene
- If there are no issues with circular teleporting, this could just exist as one quest where the two scenes point to each other, and the user can `Abandon Quest` to return to Dev Archipelago

felixwalberg:
Add two scenes, `teleporter_test_1.tscn` and `teleporter_test_2.tscn`, which demonstrate how to use the `teleporter.tscn` to teleport from one scene to another *and* from one point to another within the same scene. This will be accessible via Dev Archipelago.

Note: These scenes will soon reside in `scenes/dev_archipelago/quests/...` rather than within the `scenes/world_map/components` directory so that they will survive the reduced bundle.

Resolves #2534


felixwalberg:
Screenshot 2026-07-22 at 2 38 13 PM
Screenshot 2026-07-22 at 2 38 27 PM

I try using environmental clues to connect the teleporter to the spawn point, but let me know if this is too much clutter!

felixwalberg:
I feel like `teleporter_test_2.tscn` feels a little cramped, but I wanted to make sure you could see both the teleporter area and spawn point area within the same view of the camera. I wonder if it would be worth spreading things out a little bit more or thinning out some of the assets so it doesn't feel as overwhelming.

felixwalberg:
> I feel like `teleporter_test_2.tscn` feels a little cramped

I also tried removing the transition on the point-to-point teleporters, which makes it feel more separated from the teleporter used to go from one scene to another. Not sure if that's what it needs though...

manuq:
> I try using environmental clues to connect the teleporter to the spawn point, but let me know if this is too much clutter!

I think this looks great!

scenes/world_map/components/teleporter_test_1.tscn
```diff
@@ -0,0 +1,125 @@
...
+
+[node name="SpawnPoint" type="Marker2D" parent="." unique_id=141313087 groups=["spawn_point"]]
+position = Vector2(1147, 94)
+script = ExtResource("7_ru1dy")
+metadata/_custom_type_script = "uid://0enyu5v4ra34"
+
+[node name="Teleporter" parent="." unique_id=1779636661 instance=ExtResource("8_b4723")]
+position = Vector2(1395, 97)
+next_scene = "uid://m0f3uc2tme4l"
+spawn_point_path = NodePath("SpawnPoints/StartingSpawnPoint")
```
manuq, line 53: This transition from test_1 to test_2 should be:

Image

To match the rest of the game. From test_2 to test_1:

Image

I used `transition_test.tscn` to check.

manuq (approved):
Looks great!

scenes/world_map/components/teleporter_test_1.tscn
manuq: Maybe this is something to explain as well in the demo? How to match the transitions according to the player direction.

felixwalberg: Thank you, this tool is helpful! I'll try to add this concept to the hints.

felixwalberg: This made me realize I hadn't set the `look_at_sign_on_spawn` property when returning from scene 2 to scene 1!

manuq: Is `transition_test.tscn` going to be accessible from dev archipielago?

felixwalberg: That's a good question. Would it be jarring for the player to go from controlling StoryWeaver around all the other quests to one that exists as just `Control` nodes?

We would be able to have them Abandon Quest in the same way to return. It is a helpful tool. I think it would be a good idea to show examples of things that we try to maintain consistently throughout Threadbare.

felixwalberg: For now, I am going to merge this. I will add `transition_test.tscn` to the list of demos to consider.

PR #2590, opened 2026-07-23
Move quests and dev archipelago out of world_map folder

felixwalberg:
Create quest directories for basic, repel, and grapple-related demo scenes, linking them to their respective elders.

Turns existing `custom_repellable_objects.tscn`, `custom_hookable_objects.tscn`, and `teleporter_test_1.tscn`/`teleporter_test_2.tscn` scenes into quests so that they can be referenced by the Storybook.

If it is alright, the original custom repel/hook scenes can be deleted as they exist within the `scenes/dev_archipelago/dev_quests/...` directories now.

The goal is to give an example of the file structure of the dev quests, with one in each. It will get expanded as more scenes are moved in and quests are created.


scenes/dev_archipelago/dev_quests/basics/teleporter/quest.tres
```diff
@@ -0,0 +1,14 @@
+[gd_resource type="Resource" script_class="Quest" format=3 uid="uid://ueveeummkq5"]
+
+[ext_resource type="Script" uid="uid://dts1hwdy3phin" path="res://scenes/menus/storybook/components/quest.gd" id="1_gc3bi"]
+
+[resource]
+script = ExtResource("1_gc3bi")
+status = 1
+title = "Teleporters"
+description = "Learn how to use teleporters to go from scene to scene or between two points in the same scene."
+authors = Array[String](["Felix Walberg"])
```
felixwalberg: Is it necessary to add myself as an author for the test scenes I design? Would learner submissions to the archipelago quests be something we would want to encourage in the future? In that case, it would be nice to credit contributors. Otherwise, I'm not sure how much we have to treat each like a full-on quest creation. 

I can go more in depth in a description and possibly add art/visuals once I have a better idea on how the format should look across all dev quests.

manuq: We don't fill authors or affiliation in lore quests. I think the purpose is more for StoryQuests. Although there is a "side" lore quest in the works, and a lore quest that comes from a game jam which may have authors / affiliation added.

For general attribution, we already do it in the Credits (`scenes/menus/credits/data/`).

manuq (changes requested):
This duplicates scenes and scripts from `scenes/game_elements` to `scenes/dev_archipielago`:

Image

For better maintenance, we should avoid these duplicates. Was it decided if all the files of Dev Archipielago are going to be under the same folder? Otherwise the dev quests can point to the tests in `scenes/game_elements`. I think I prefer the latter, keeping the tests inside a `components/` subfolder next to the thing they are testing.

manuq:
> For better maintenance, we should avoid these duplicates. Was it decided if all the files of Dev Archipielago are going to be under the same folder? Otherwise the dev quests can point to the tests in `scenes/game_elements`. I think I prefer the latter, keeping the tests inside a `components/` subfolder next to the thing they are testing.

Oh and I don't think this would be affected by https://github.com/endlessm/threadbare/issues/2563 because both `scenes/game_elements` and `scenes/dev_archipielago` will be kept in the smaller bundle.

felixwalberg:
I agree that it makes more sense to keep the tests in the project closer to the element they actually test. My one concern is how we would distinguish between which quest belongs to which themed island. For example, `custom_hookable_objects.tscn` and `custom_repellable_objects.tscn` are both in the `scenes/game_elements` directory, but with the current Dev Archipelago setup, they would exist in two different storybooks.

manuq:
> I agree that it makes more sense to keep the tests in the project closer to the element they actually test. My one concern is how we would distinguish between which quest belongs to which themed island. For example, `custom_hookable_objects.tscn` and `custom_repellable_objects.tscn` are both in the `scenes/game_elements` directory, but with the current Dev Archipelago setup, they would exist in two different storybooks.

This is a good topic to discuss with Will next week! How to structure the dev quests. Move all existing tests into it or not? Do you have a list of the existing tests that are going to be part of the archipielago? Another examples are:

```
scenes/game_elements/props/area_filler/area_filler.gd
scenes/game_elements/props/area_filler/components/area_filler_test.tscn
```

And:

```
scenes/game_elements/fx/time_and_weather/time_and_weather.tscn
scenes/game_elements/fx/time_and_weather/components/2_time_and_weather_museum.gd
scenes/game_elements/fx/time_and_weather/components/1_time_and_weather_museum.tscn
```

felixwalberg:
> Do you have a list of the existing tests that are going to be part of the archipielago?

The prototype tab on this document has been a running list of ideas we have had so far, including the comments that add more.
https://docs.google.com/document/d/19V3y4Lrin6bS_-px0uOytw3PPQ9GBOV-DyY9Lia5LdQ/edit?usp=sharing


However, it is a bit all over the place. I can make a simple and to the point list with all existing scenes and their file paths so that we can all discuss the best approach next week.

felixwalberg:
@wjt I am working on making a list of all the scenes we want to have as "dev quests" connected to the elders in Dev Archipelago. In the meantime, I wanted to bring this up, since Manuel and I were discussing the best way to reference them and we know there are scenes that connect to different themes within the same parent directory. 

If we keep the themed islands where each elder corresponds to a subset of testing scenes, we wouldn't be able to point at the `scenes/game_elements` directory for each elder without overlapping the quests. The question we have now is:
> How to structure the dev quests. Move all existing tests into it or not?

Do we move all the scenes into the `scenes/dev_archipelago` directory where they are separated by theme, or do we leave them to exist in the file system closest to the element that they test/demonstrate and find another way to point to them?

wjt:
(I agree that these scenes shouldn't be duplicated but I don't think that's what's being suggested!)

There is no **technical** reason why a `Quest` resource has to be stored near the scene(s) it points to. We could have `scenes/dev_archipelago/dev_quests/grapple/custom_hookable_objects/quest.tres` with its `first_scene` set to `scenes/game_elements/components/custom_hookable_objects_test.tscn` and it would work correctly. (With the exception that guessing the quest when you deep-link to the scene or run it with <kbd>F6</kbd> would not work.)

Storing the test scenes alongside the component they test makes some conceptual sense,  **but** I think it's one reason that no-one finds them: there's no way to know which components have test scenes or not. We're building Dev Archipelago in part to make these scenes easier to discover. Also we will have scenes that are demos of more than one component - so these wouldn't have a natural home alongside one or another component. So actually I think we should probably make the folder structure reflect the in-game structure and put the test scenes into the same folder structure as dev archipelago.

On folder structure, personally I might go for one less layer of nesting, removing the `dev_quests/`:

- scenes/ (I have come to really dislike this top-level folder but we're kind of stuck with it
  - dev_archipelago/ (could even call this just dev/)
    - dev_archipelago.tscn
    - components/
      - sheep_tug_of_war.gd
    - basics/
      - teleporter
        - quest.tres
        - teleporter_test_1.tscn
        - teleporter_test_2.tscn

scenes/dev_archipelago/dev_quests/basics/teleporter/quest.tres
wjt: I wonder if we should push these fields down into `StoryQuest` (in a separate PR)?

manuq: > I wonder if we should push these fields down into `StoryQuest` (in a separate PR)?

Yeah, I wondered the same, but I was cautious because of the lore jam quest and the new side quest.

felixwalberg:
> On folder structure, personally I might go for one less layer of nesting, removing the `dev_quests/`

This sounds good to me! I have added changes that move the themed quest folders up one directory and removed the duplicate test scenes (so the only reference is within `dev/`.

manuq (approved):

PR #2591, opened 2026-07-23
Check if mushroom AudioStreamPlayers are in scene tree before reveal

### LINKED ISSUE #2589: Abandoning quest when in range of `hiding_mushroom.tscn` throws error
### Threadbare version

v0.2.1-29-g2b10bbdd

### Affected scene

res://scenes/dev_archipelago/dev_quests/basics/teleporter/teleporter_test_1.tscn

### Description

When `Player` exits the `CollisionShape2D` of the `hiding_mushroom.tscn`, `_reveal()` is called. If the `Player` "exits" by abandoning the quest, the body of `_reveal()` references nodes that are no longer in the scene tree. Specifically, `reveal_player`, which throws the error:
```
hiding_mushroom.gd:22 @ _reveal(): Playback can only happen when a node is inside the scene tree
```

This can be replicated by going to `dev_archipelago.tscn`, interacting with the first elder and choosing the Teleporters quest, and then immediately abandoning it without moving around.

### Checklist

felixwalberg:
Check that the nodes associated with the `_reveal()` function for the hiding mushroom exist in the scene tree before calling it. Gets rid of the error when using Abandon Quest while `Player` is within range of the mushroom.

Resolves #2589

### DIFF
--- a/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
+++ b/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
@@ -31,5 +31,5 @@ func _on_player_detector_body_entered(body: Node2D) -> void:
 
 
 func _on_player_detector_body_exited(body: Node2D) -> void:
-	if body.is_in_group(&"player"):
+	if body.is_in_group(&"player") and reveal_player.is_inside_tree():
 		_reveal()
--- a/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
+++ b/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
@@ -19,7 +19,9 @@ func _hide() -> void:
 func _reveal() -> void:
 	animated_sprite_2d.visible = true
 	animated_sprite_2d.play(&"reveal")
-	reveal_player.play()
+	# A body may call body_exit while simultaneously switching scenes, causing this to be freed
+	if reveal_player.is_inside_tree():
+		reveal_player.play()
 	await animated_sprite_2d.animation_finished
 	if animated_sprite_2d.animation == &"reveal":
 		animated_sprite_2d.play(&"idle")
--- a/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
+++ b/scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
@@ -33,5 +33,5 @@ func _on_player_detector_body_entered(body: Node2D) -> void:
 
 
 func _on_player_detector_body_exited(body: Node2D) -> void:
-	if body.is_in_group(&"player") and reveal_player.is_inside_tree():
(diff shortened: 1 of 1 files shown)


felixwalberg:
It also references the `animated_sprite_2d` node before `reveal_player`, but that doesn't seem to matter if it exists within the scene tree. I am not sure if just checking `reveal_player` on its own is best practice.

scenes/game_elements/props/decoration/hiding_mushroom/hiding_mushroom.gd
```diff
@@ -31,5 +31,5 @@ func _on_player_detector_body_entered(body: Node2D) -> void:
 
 
 func _on_player_detector_body_exited(body: Node2D) -> void:
-	if body.is_in_group(&"player"):
+	if body.is_in_group(&"player") and reveal_player.is_inside_tree():
```
manuq: Interesting bug! If the whole mushroom is freed, then I guess the following check would be better:
```suggestion
	if body.is_in_group(&"player") and is_inside_tree():
```

Or maybe:
```suggestion
	if body.is_in_group(&"player") and is_instance_valid(self):
```

Untested! And I would also add a comment about this because, at least to me, is not obvious:
```
	## The body may call this body_exit callback at the same time that this object is freed, when switching scenes:
	if body.is_in_group(&"player") and is_inside_tree():
```

felixwalberg: I'll test this out, thanks!

felixwalberg: Neither of these actually work. I am trying to figure out why... but when the quest is abandoned, `is_instance_valid(self)` still returns `true`

felixwalberg: My theory: I assume that `queue_free()` is sort of implicitly (or maybe explicitly, I don't know exactly where) called when the player abandons the quest, but this will only remove the node from the tree after the current frame is done. The script continues to execute (namely the `_reveal()` function) before the frame can end, so at the point of checking on line 34, `self` is valid and exists in the scene tree.

felixwalberg: `is_queued_for_deletion()` returns false however, so this might not hold up.

Seems to be because `_exit_tree()` is not called until after `_reveal()`

manuq: @felixwalberg thanks for the investigation! In that case, can you move the `reveal_player.is_inside_tree()` check  closer to its usage? That is:

```
if reveal_player.is_inside_tree():
    reveal_player.play()
```

felixwalberg: Done! Still confuses me why `animated_sprite_2d` doesn't have the same issues...

manuq (approved):

PR #2592, opened 2026-07-24
Storybook: Eliminate scrollbar by moving what doesn't fit on the left page to the right page.

### LINKED ISSUE #1757: Storybook: List of stories does not fit on the page

alcole2:
Currently, all StoryQuests are on the left page of the Storybook, using a scrollbar to see and navigate through them all. I have removed the scrollbar by having what doesn't fit in a no-scrollbar view of the left page moved to the right page. When there isn't enough space on either page, the use can flip the page and the remaining StoryQuests will be there.

Resolves #1757

### DIFF
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 class_name Storybook
 extends CanvasLayer
+
 ## Offers a choice of quests by scanning a given [member quest_directory].
 
 ## Emitted when the player chooses a quest from the storybook, with
@@ -13,11 +14,21 @@ signal selected(quest: Quest, restart: bool)
 
 ## Quests to show in the storybook.
 @export var quests: Array[Quest]
+@export var quests_per_page: int = 8:
+	set(value):
+		quests_per_page = value
+		quests_per_spread = value * 2
+
+var quests_per_spread: int = 16
 
 var _current_spread_index: int = -1
 var _navigation_locked: bool = false
+var _current_list_page: int = 0
 
-@onready var quest_list: VBoxContainer = %QuestList
+# Node References
+@onready var left_quest_list: VBoxContainer = %LeftQuestList
+@onready var right_quest_list: VBoxContainer = %RightQuestList
+
 @onready var quest_container: ScrollContainer = %QuestContainer
 @onready var storybook_page: StorybookPage = %StorybookPage
 @onready var back_button: Button = %BackButton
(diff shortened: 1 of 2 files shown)


scenes/menus/storybook/components/storybook.gd
```diff
@@ -13,11 +14,21 @@ signal selected(quest: Quest, restart: bool)
 
 ## Quests to show in the storybook.
 @export var quests: Array[Quest]
+@export var quests_per_page: int = 8:
+	set(value):
+		quests_per_page = value
+		quests_per_spread = value * 2
```
manuq, line 20: The `quests_per_spread` variable is not needed, because it will always be twice the `quests_per_page`. So  I think you should remove it and just do `quests_per_page * 2` whenever you need it.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -27,42 +38,72 @@ var _navigation_locked: bool = false
...
+## Clears and regenerates the quest buttons based on the current page view
+func _populate_quest_lists() -> void:
+	#Clear out any existing buttons from previous views
+	for child in left_quest_list.get_children():
+		child.queue_free()
+	for child in right_quest_list.get_children():
+		child.queue_free()
+
+	#Calculate the quest slices for this specific book spread
+	var left_start: int = _current_list_page * quests_per_spread
```
manuq:
```suggestion
	var left_start: int = _current_list_page * quests_per_page * 2
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -137,25 +175,45 @@ func _switch_to_page(spread_index: int) -> void:
...
 	_switch_to_page(_current_spread_index - 1)
 
 
 func _on_right_button_pressed() -> void:
 	if _navigation_locked:
 		return
+
+	# If we are on the main index, check if there are more quests to reveal on a new page
+	if _current_spread_index == 0:
+		var max_visible_so_far: int = (_current_list_page + 1) * quests_per_spread
```
manuq:
```suggestion
		var max_visible_so_far: int = (_current_list_page + 1) * quests_per_page * 2
```

manuq (changes requested):
This works very well! Have you tried this with dozens of entries? If so, how? It would be nice to have a way to mock a lot of entries for testing the book.

I left a suggestion about an unneeded variable, it's just a detail.

Please:
- Use imperative for the title of the pull request. Something like "Storybook: Eliminate scrollbar..."
- Add a description to the pull request
- In the description, add "Resolves... " as last line with a link to the resolved issue.

Check https://github.com/endlessm/threadbare/wiki/Contributing#pull-request-descriptions for the convention details of the request above!

scenes/menus/storybook/components/storybook.gd
```diff
@@ -13,11 +14,21 @@ signal selected(quest: Quest, restart: bool)
...
+		quests_per_page = value
+		quests_per_spread = value * 2
+
+var quests_per_spread: int = 16
 
 var _current_spread_index: int = -1
 var _navigation_locked: bool = false
+var _current_list_page: int = 0
+
+# Node References
```
manuq: This is commonly done in all scripts, so the comment is not needed:
```suggestion
```

manuq:
@alcole2 I tested this with the Template Elder which has a single quest, and this is what I see:

https://github.com/user-attachments/assets/6c6d362d-233e-4ba5-806c-f7b1766a6702

alcole2:
Interesting, I think I know how to fix that!

manuq:
> Interesting, I think I know how to fix that!

Great, I have added a test scene with mock titles: https://github.com/endlessm/threadbare/pull/2593

Please review it,  as it will be useful for this pull request. This is how your change looks like with 30 mock titles:


https://github.com/user-attachments/assets/906e84b0-2a93-46e2-971e-c7f93e5a15a0

And with six:

image

alcole2:
I fixed it! Initially, I had moved the MarginContainer to the center of the book so everything would align properly when both sides of the book were needed, not realizing that it placed the StoryQuests in the middle of the book when there was 8 or less items. If the right page isn't needed, it wouldn't put any buttons on it and therefore it kind of didn't exist. I just changed it so that if there is 8 StoryQuests or less, it will basically fill up the right side with space so it still correctly aligns with the book.

I need to fix the rest of your changes and then update the pull request.

Question: Are the pages for each StoryQuest in the scene with mock titles blank simply because there isn't actually any information attached to them?

manuq:
> I fixed it! Initially, I had moved the MarginContainer to the center of the book so everything would align properly when both sides of the book were needed, not realizing that it placed the StoryQuests in the middle of the book when there was 8 or less items. If the right page isn't needed, it wouldn't put any buttons on it and therefore it kind of didn't exist. I just changed it so that if there is 8 StoryQuests or less, it will basically fill up the right side with space so it still correctly aligns with the book.
> 
> I need to fix the rest of your changes and then update the pull request.
> 
> Question: Are the pages for each StoryQuest in the scene with mock titles blank simply because there isn't actually any information attached to them?

Yes, check the script in https://github.com/endlessm/threadbare/blob/cf8d301f78122f0c01468c771879ca2919df21d5/scenes/menus/storybook/components/storybook_test.gd

scenes/menus/storybook/components/storybook.gd
```diff
@@ -27,42 +37,77 @@ var _navigation_locked: bool = false
...
 func _update_page_visibility() -> void:
 	if _current_spread_index == 0:
 		quest_container.visible = true
 		storybook_page.visible = false
 
-		if quest_list.get_child_count() > 0:
-			var first_button: Button = quest_list.get_child(0)
+		# Grab focus on the first visible item of the left page
+		if left_quest_list.get_child_count() > 0:
+			var first_button: Button = left_quest_list.get_child(0)
```
manuq, line 110: Is this really working? When I press the "Next" action and switch to the next 2 pages, the focus is lost. We don't have enough storyquests for testing it in the actual game, but it can be reproduced using the test scene:

https://github.com/user-attachments/assets/d4481bb3-f12e-4de9-a123-e0645d10593b

So you can't use the keyboard to select the quest anymore.

manuq (approved):
This is better than before so I think we can merge it and improve. Things to improve as follow ups:
- The focus is lost
- The layout is hardcoded to make the existing StoryBook titles look right, but if the titles are shorter or longer, the pages aren't centered anymore.

Image

Image

Issue #2547, opened 2026-07-15
Add a simple cutscene to the threadlings conversation that shows the world being rebuilt.


saspiers:

manuq:
One thing that I would like to test for this is a shader, to make a sprite or a tilemap dissolve using a textile texture. Which would try to fit the game thematically.

This could also be used in the void runner level when Moss The Monk restores the area affected by the Void, instead of the current fade-in effect.

And is what I had in mind for https://github.com/endlessm/threadbare/issues/2282

https://godotshaders.com/ has references, searching for "dissolve" or "burn" CanvasItem shaders. They usually use a noise texture. I wonder if we can use a textile texture instead.

Issue #2282, opened 2026-06-03
Townie helpers: Add generic animation for things appearing


manuq:
Currently the townie helpers added in https://github.com/endlessm/threadbare/issues/2119 can:
- make objects appear (the needles at the Void quest "needles" level)
- make tiles change (fixing the bridge at the stealth level)
- make objects change (filling the inkwell at the third combat level).

It would be nice to add a generic animation when that happens. Idea: let the townie unravel and jump over the level to the position of things to change, emit some particles to cover the change, jump back and rebuild themselves. In the Void floor eater enemy we have something similar (particles covering the change).

PR #2594, opened 2026-07-24
Add cutscenes of the world being rewoven

manuq:
## Fray's End and Eternal Loom changes

Show a cutscene in the Eternal Loom interaction, the order of actions is now as follows:
1. The Eternal Loom welcomes the StoryWeaver (in dialogue).
2. The townies walk in.
3. The StoryWeaver tells them a story of the last quest (if the quest has retelling)
4. The townies walk out. In parallel, the thread spools animate, floating in the air and incorporated into the loom.
5. Fade to: the cutscene of a part of the world being rewoven.
6. Fade to: back in Fray's End, the LoreQuest Elder thanks the StoryWeaver for healing the fabric of the world.

## Cutscene effects

This is achieved with a couple effects:
- Two layers, the broken world and the fixed world.
- A shader that makes the fixed world layer appear using a textile pattern and with a "burn" effect.
- Sparkle particles as it appears.
- Modulate on water and clouds.
- Camera shaker.
- AnimationPlayer to control all of it.

For the sparkles, I reused the assets done by learners for collecting buttons.
I did this with existing assets. I realized that we lack assets for creating a "broken world". The tiles with tears are not enough, and they can't be used in borders. No animated tiles or props can be used, because everything feels so alive (I had to duplicate the foam and rocks just to slow down the animation). I made the broken world gray with the same shader.

I used clouds from tiny-swords to improve the composition. To avoid empty areas with a plane color (water).

Note: The shader effect depends on the size of the CanvasItem to which it's applied. So centering the radial effect, for instance, is done by updating the texture.

Two initial cutscenes are provided, and they are picked randomly.

Note: The Godot scenes that contain the cutscenes should have all collisions disabled. Otherwise the StoryWeaver can be displaced by them.

These visual effects would really welcome sound effects!

Helps https://github.com/endlessm/threadbare/issues/2547


felixwalberg:
This looks super cool! I like how the transition is not bound by 64x64 tiles, if that makes sense. So that, in situations where you have something like a rock, sheep, or other decoration that is not square, it can still smoothly reveal it.

Having the before and after views use separate tilemaps could even allow us to incorporate the swapping ability that the lore jam introduced. I don't know how this fares in a "live" scene with the player moving around and dynamically updating the before/after.

I am going to have to spend some time learning about shaders before I totally understand how this works and can give more helpful feedback... but this is great.

manuq:
> Play this branch at https://play.threadbare.game/branches/endlessm/world-reweaven/.
> 

Direct links:
- https://play.threadbare.game/branches/endlessm/world-reweaven/#game_elements/fx/world_reweaven/components/world_reweaven_test
- https://play.threadbare.game/branches/endlessm/world-reweaven/#game_elements/fx/world_reweaven/components/world_reweaven_test2

manuq:
Thanks for your feedback @felixwalberg !

> This looks super cool! I like how the transition is not bound by 64x64 tiles, if that makes sense. So that, in situations where you have something like a rock, sheep, or other decoration that is not square, it can still smoothly reveal it.

Yeah, I understand what you mean by not being bound to tiles.

> Having the before and after views use separate tilemaps could even allow us to incorporate the swapping ability that the lore jam introduced. I don't know how this fares in a "live" scene with the player moving around and dynamically updating the before/after.

True! I haven't thought consciously about that, but certainly that effect of Team 11 might have been in the back of my mind! And it was probably done in a similar way.

wjt:
I haven't looked at the implementation at all but, visually, I love it.

scenes/game_elements/fx/world_reweaven/components/world_reweaven.gdshader
```diff
@@ -0,0 +1,79 @@
...
+uniform float progress : hint_range(0.0, 1.0) = 0.0;
+
+// Controls how smooth the edge of the fade is
+uniform float smoothness : hint_range(0.0, 1.0) = 0.1;
+
+// Burn effect settings
+uniform float burn_size : hint_range(0.0, 1.0) = 0.15;
+uniform vec4 burn_color : source_color = vec4(1.0, 1.0, 0.58, 1.0);
+
+uniform float fabric_scale : hint_range(0.0, 10.0) = 4.0;
```
wjt: I was surprised that making this number bigger made the fabric effect in the burn *smaller* - it's the inverse of what I expected. (Not a problem as such, just a documentation thing.)

scenes/game_elements/fx/world_rewoven/components/world_rewoven.gdshader
```diff
@@ -0,0 +1,79 @@
...
+	textile_uv.x = UV.x * fabric_scale;
+	textile_uv.y = UV.y * fabric_scale;
+}
+
+void fragment() {
+	// Sample the main texture color of the node
+	vec4 main_color = textureLod(screen_texture, SCREEN_UV, 0.0);
+	if (main_color.a > 0.0001) {
+		main_color.rgb /= main_color.a;
+	}
```
wjt, line 68: Interesting, I see that this (and `render_mode unshaded`) comes directly from the `CanvasGroup` documentation.

wjt (approved):
I like the design of having two completely independent branches of the scene tree. Your examples show clearly why it is needed, rather than having some special knowledge of each prop that might need to be restored with a shader on each one or something.

I guess the downside is that it involves keeping two copies of the level in sync. An interesting exercise would be to try replacing the (bad!) animation in the void runner with this. (Doesn't block this cutscene work though!)

scenes/game_elements/fx/world_reweaven/components/world_reweaven.gdshader
manuq: Yeah I actually been experimenting so far, I didn't care much about the code. This is a parameter I made for scaling the texture for now, because it is in screen coordinates, but it should be in world coordinates, should zoom with the camera/viewport.  Right now is not:

image

scenes/game_elements/fx/world_rewoven/components/world_rewoven.gdshader
manuq: Yes! Otherwise I wouldn't be able to apply the shader to the CanvasGroup.

manuq:
> I like the design of having two completely independent branches of the scene tree. Your examples show clearly why it is needed, rather than having some special knowledge of each prop that might need to be restored with a shader on each one or something.
>
> I guess the downside is that it involves keeping two copies of the level in sync. An interesting exercise would be to try replacing the (bad!) animation in the void runner with this. (Doesn't block this cutscene work though!)

Yes honestly I focused on this as a cutscene. And because we don't have enough "sad world broken" assets, I had to fake a lot, making the "broken world" grayscale, unanimated, etc. I definitely see the downside, adding an effect like this to the world can be a lot of setup. So yes, next I will need to see how would this work for actual playable levels, and for https://github.com/endlessm/threadbare/issues/2282.

manuq:
In the last commit I have integrated the cutscene with the eternal loom interaction. It is instantiated in Fray's End inside a CanvasLayer. But to do it that way, I had to disable collisions in all the tilemap layers of the cutscene, otherwise the StoryWeaver would be displaced!

Here is game state to try it: 

[saved_game_loom.zip](https://github.com/user-attachments/files/30482946/saved_game_loom.zip)

manuq:
> I didn't spend too much time on the shader. At the current state it is a bit inflexible: the textile texture doesn't scale with the camera zoom, and it doesn't follow the camera (a cutscene with camera movement won't be possible). 

This (also shown in https://github.com/endlessm/threadbare/pull/2594#discussion_r3661033101) is now fixed! I had a nice session with Claude in which I've learned about the CANVAS_MATRIX for just applying the camera zoom to a texture, and the MODEL_MATRIX for applying both zoom and panning. And also learned about the world "texel". I found the session really impressive, so I left it verbatim here: https://gist.github.com/manuq/ecb2c3df75c5006ca28f40c3535ea773

scenes/world_map/components/cutscene_layer.gd
```diff
@@ -0,0 +1,65 @@
...
+
+
+func _on_eternal_loom_play_cinematic() -> void:
+	if _load_error != OK:
+		eternal_loom.on_cinematic_finished()
+		return
+
+	# Note: this is a hacky way of hidding the input HUD during the cutscene:
+	var player: Node = get_tree().get_first_node_in_group(&"player")
+	player.remove_from_group(&"player")
```
manuq, line 49: Since I'm already hacking the player, maybe I can remove it from the collision layers and re-add it to the existing collision layers, instead of disabling collisions in the cutscenes.

scenes/game_elements/fx/curly_trail/components/curly_trail.gd
```diff
@@ -0,0 +1,79 @@
...
+	tip.position.x = radius
+
+
+func _draw() -> void:
+	if debug:
+		draw_circle(Vector2.ZERO, tip.position.x, Color(1.0, 0.0, 0.0, 0.486))
+		draw_line(Vector2.ZERO, tip.position.rotated(wheel.rotation), Color(1.0, 1.0, 0.0, 0.486))
+
+
+func _process(_delta: float) -> void:
```
manuq, line 54: Not using delta here means that the trail effect is different in high refresh rate monitors. The effect actually looks better in high refresh. Maybe I should "bake" the trail path in the editor, and just play it without calculations in the game. But I really wanted to make it dynamic to be reused in game, not just in a cutscene.

scenes/world_map/components/cutscene_layer.gd
```diff
@@ -0,0 +1,65 @@
...
+	Transitions.do_transition(do_add, Transition.Effect.FADE, Transition.Effect.FADE)
+
+	await cutscene.animation_finished
+
+	var do_remove: Callable = func() -> void: remove_child(cutscene)
+	Transitions.do_transition(do_remove, Transition.Effect.FADE, Transition.Effect.FADE)
+
+	player.add_to_group(&"player")
+
+	eternal_loom.on_cinematic_finished()
```
manuq, line 65: Similar to the townies (`retelling_manager.gd`) I did the communication one-way: this script knows about the loom and connects to its signals, and calls a method on it. But the loom doesn't know about this script.

scenes/world_map/components/cutscene_layer.gd
```diff
@@ -0,0 +1,65 @@
...
+##
+## Note: for this to work, the cutscenes root node must be [AnimationPlayer] and they must have
+## an animation set to [member AnimationPlayer.autoplay].
+##
+## Warning: If the cutscenes have collision shapes, they may interfere with the physic objects
+## below!
+
+## Paths to scenes that will be shown as overlay. Their root node must be an AnimationPlayer.
+@export var cutscene_paths: Array[String] = [
+	"res://scenes/game_elements/fx/world_reweaven/components/world_reweaven_cutscene_1.tscn",
```
wjt: It's just a filename but "(re)woven" is the past participle of "(re)weave". English is a weird language.

scenes/world_map/components/cutscene_layer.gd
```diff
@@ -0,0 +1,65 @@
...
+	"res://scenes/game_elements/fx/world_reweaven/components/world_reweaven_cutscene_2.tscn",
+]
+
+## The picked cutscene. Used to preload it.
+var _cutscene_path: String
+var _load_error: Error
+
+## The Eternal Loom, for listening to signals and calling
+## [member EternalLoom.on_cinematic_finished()].
```
wjt:
```suggestion
## [member EternalLoom.on_cinematic_finished].
```

scenes/world_map/components/cutscene_layer.gd
wjt: There are other physics bodies in Fray's End though... I think this is OK for now.

scenes/world_map/components/cutscene_layer.gd
```diff
@@ -0,0 +1,65 @@
...
+	# Note: this is a hacky way of hidding the input HUD during the cutscene:
+	var player: Node = get_tree().get_first_node_in_group(&"player")
+	player.remove_from_group(&"player")
+	InputHud.refresh_scene_status()
+
+	var cutscene_packed: PackedScene = ResourceLoader.load_threaded_get(_cutscene_path)
+	var cutscene: AnimationPlayer = cutscene_packed.instantiate()
+
+	var do_add: Callable = func() -> void: add_child(cutscene)
+	Transitions.do_transition(do_add, Transition.Effect.FADE, Transition.Effect.FADE)
```
wjt, line 56: Perhaps we can revisit this later and actually change the current scene between Fray's End and the cutscene. We used to do this for sokobans - there was a flag in the game state that triggered the end of the elders' dialogue ("thank you, storyweaver!") once the player returned from the sokobans.

scenes/game_elements/fx/stitch_trail/components/stitch_trail.gd
```diff
@@ -0,0 +1,71 @@
...
+func _process(delta: float) -> void:
+	if _last_stitch_seconds < stitches_per_second:
+		_last_stitch_seconds += delta
+		return
+	_last_stitch_seconds = 0
+
+	var direction := Vector2.ZERO if not _last_position else _last_position - global_position
+	var orthogonal := direction.orthogonal().normalized()
+	if on_floor:
+		orthogonal *= Vector2(1, 0.5)
```
wjt, line 60: Very clever!

scenes/game_elements/fx/world_rewoven/components/world_rewoven_cutscene_1.tscn
```diff
@@ -0,0 +1,788 @@
+[gd_scene format=4 uid="uid://d02db8ykyyos0"]
+
+[ext_resource type="TileSet" uid="uid://oynx002hv8tl" path="res://tiles/water.tres" id="2_yewlk"]
+[ext_resource type="Texture2D" uid="uid://cexg7otw5enpu" path="res://assets/third_party/tiny-swords/Terrain/Water/Foam/Foam.png" id="3_3d5w8"]
+[ext_resource type="Texture2D" uid="uid://8mupkacbuxpr" path="res://assets/third_party/tiny-swords-non-cc0/Terrain/Decorations/Clouds/Clouds_01.png" id="3_3mcre"]
```
wjt, line 5: You don't need to change this! but: this scene won't load in the SQCK because I delete this entire asset pack to remove 378 unused assets from the tree... Maybe we can delete everything except the clouds :) or do a smarter prune that checks if assets are referenced before  deleting them. I'll file a followup for that.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -74,10 +76,16 @@ func give_spirit_upgrade() -> void:
...
+	load_cinematic.emit()
+
 	loom_offering_animation_player.play(&"loom_offering")
 	await loom_offering_animation_player.animation_finished
 	GameState.quest.inventory.clear_inventory()
 
+	play_cinematic.emit()
+
+
+func on_cinematic_finished() -> void:
```
wjt, line 88: I think this coupling is backwards. Rather than `cutscene_layer.gd` having special knowledge of the Eternal Loom, it should instead implement an async method that plays the cutscene. Then the loom would have a reference to the cutscene layer with something like this:

```suggestion
func on_offering_succeeded() -> void:
	cutscene_layer.load()

    loom_offering_animation_player.play(&"loom_offering")
    await loom_offering_animation_player.animation_finished
    GameState.quest.inventory.clear_inventory()

	await cutscene_layer.play()
```

but I don't think you should change that in this PR. Instead, in a future change, we should probably actually change scene to the cut scene, and change back to Fray's End at the end.

wjt (approved):
LGTM. Nothing to change now, some suggestions for follow-up.

scenes/world_map/components/cutscene_layer.gd
manuq: Oh thanks for correcting me! Google translator deceived me. I will do the renames.

scenes/world_map/components/cutscene_layer.gd
manuq: Yes, and I really want these places of the world to be actual levels.

scenes/game_elements/fx/world_rewoven/components/world_rewoven_cutscene_1.tscn
manuq: Oh ouch! I can remove the clouds for now. We probably want a fabric themed version. They looked good in marketing images and in talk slides and they weren't in the game, that I wanted to use them. Also for composing the image, without them, the empty areas are all water so they look a bit empty.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
manuq: Good point, I haven't thought about doing the communication in the other direction. I handled another option: adding a manager script to Fray's End.

scenes/world_map/components/cutscene_layer.gd
wjt: Yes! Maybe we can have a "cutscene mode" for scenes like 
image
 where the player doesn't spawn and instead we focus the camera on the middle island, fix the house and the bridge.

scenes/game_elements/fx/world_rewoven/components/world_rewoven_cutscene_1.tscn
wjt: Keep them in, I can make the prune smarter.

scenes/world_map/components/cutscene_layer.gd
manuq: Good idea!

PR #2610, opened 2026-07-28
Shine particles effect: Make it faster

manuq:
When collecting buttons, this effect was a bit slow.

Reduce lifetime to 0.4. Enlarge initial velocity to vary between 40 and 80. Remove unneeded scale curve for velocity (was at the default).


manuq:
This is how it looks in game:

https://github.com/user-attachments/assets/c1b01eb5-316a-419e-809a-1ec3734c9c5b

wjt (approved):

PR #2480, opened 2026-06-25
Import LoreJam Team 13's project from Backstitch

wjt:
This is https://github.com/wjt/threadbare-lorejam-mythical-meadows/tree/equipo-13 manually rebased onto current main. I discarded some changes that are not intended, but there is more to discard!

PR #2612, opened 2026-07-28
Make mythical-meadows team co-owners of that quest

manuq:
This team comprises Gian Piero Cano and Leopoldo Brito Ruiz.


manuq:
@wjt only if you agree with it. There may be other ways to resolve https://github.com/endlessm/threadbare/pull/2480#issuecomment-5108050522

wjt:
This won't help with the issue described there. It may still be a good idea when this quest is on `main` though. I'll look at the groups tomorrow.

PR #2618, opened 2026-07-29
stealth_sequence_combo: Add music

wjt:
Once again I composed this in Strudel - source code included - and then
chopped it up with Ardour. It was a bit more inconvenient to chop up
than the first music puzzle soundtrack.

Each section of this melody begins on the upbeat, so each clip begins on
beat 4 rather than beat 1. I originally had a 3-beat intro clip with
just percussion, but it wasn't worth the complexity. It sound OK to just
jump straight in.

I also wanted to have a bar-long cymbal swell leading into the full-band
loop. This meant I needed one extra bar of the third melody with the
swell. In theory this could be part of the "full" clip, with loop_offset
set so that the first 4 beats are skipped when looping. However that
doesn't work on the web. So the workaround is to introduce an extra clip
that auto-advances to the loop. The problem with *that* is that if you
die after the music has advanced to the loop, when you respawn the music
jumps back to that lead-in bar. I left a long comment in the code about
this.


wjt:
https://play.threadbare.game/branches/endlessm/wjt/stealth-sequence-combo-add-music/#res://scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/stealth_sequence_combo.tscn

wjt:
Here's how my Ardour session looks, for the curious.

image

The Strudel loop is 24 bars and I don't need all that so I deselected some of the clips during export:

image

An important thing I got wrong at first is that I had to edit the Ogg Vorbis export format settings and disable "Normalize" because to my surprise the normalization is per-range. The full loop is deliberately louder than the per-step clips (just by adding more instruments to the arrangement) but this per-clip normalization undid that!

image

manuq:
I couldn't try it yet but this is great news! Once I finish with my own task I'll be happy to playtest and review (with my very limited music understanding).

wjt:
@saspiers says: the music needs more crows

manuq:
For now I can just say.. this is impressive!!

scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/stealth_sequence_combo.gd
```diff
@@ -25,3 +28,14 @@ func _on_sequence_puzzle_step_solved(step_index: int) -> void:
...
 	collectible_item.reveal()
+
+
+func _on_sequence_puzzle_progress_changed() -> void:
+	var step_index := sequence_puzzle.get_progress()
+	var stream := background_music.stream as AudioStreamInteractive
+	# This assumes that the clip indexes correspond to the puzzle indexes.
+	# TODO: This retriggers clip 3 (which is a transition to clip 4) after respawning if clip 4 was
+	# playing. If the web platform supported loop_offset then the lead-in could be part of the same
+	# clip as the final loop.
```
wjt: Tried to fix this by using the "filler clip" feature of AudioStreamInteractive but found another godot bug in the process :(

manuq:
@wjt I'm curious about your workflow.

How do you go from the Strudel source file to Ardour? Do you paste the source into https://strudel.cc/ and record the browser audio output?

What do you use Ardour for? Do you use it for recording the audio coming from the browser tab? For cropping the audio streams to the beat?

manuq (approved):

wjt:
> How do you go from the Strudel source file to Ardour? Do you paste the source into https://strudel.cc/ and record the browser audio output?

No, I use the export tab that was added since I wrote https://willthompson.co.uk/posts/exporting-stems-from-strudel-to-godot/ to export 24 bars - the full length of the Strudel source before it repeats.

image

Then I import the audio file into Ardour. (It's just a single stereo file with all instruments mixed together - simpler than in that blog post.)

> What do you use Ardour for? Do you use it for recording the audio coming from the browser tab? For cropping the audio streams to the beat?

The latter - I'm doing https://willthompson.co.uk/posts/exporting-stems-from-strudel-to-godot/#3-define-labels but with Ardour rather than Audacity because Audacity works really badly on my current computer for reasons unknown, and Ardour's interface for arrangement markers, exporting clips, etc. is much better.

In this case I offset the WAV file it by 1 beat because I want each clip to start on beat 4, not 1, because the melody starts on the upbeat. (I set the BPM in Ardour to match Strudel.)

scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/stealth_sequence_combo.gd
```diff
@@ -25,3 +28,34 @@ func _on_sequence_puzzle_step_solved(step_index: int) -> void:
...
+		#
+		# I tried two other approaches:
+		# - Tacking the transition onto the start of FullLoop then setting the
+		#   loop offset. This is not supported on the web platform.
+		# - Configuring the Step3 to FullLoop transition to use the LeadIn
+		#   clip as a filler. This almost works, except that Godot adds a 1-beat
+		#   fade-in at the start of FullLoop even though I configure "No Fade".
+		#   I think this is another Godot bug... I left this configured in the
+		#   resource, so one can hear it in the Step3 -> FullLoop edge case
+		#   described above.
```
wjt, line 56: I found the bug!

https://github.com/godotengine/godot/pull/121952

scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/music/GuardedStones.strudel.js
```diff
@@ -0,0 +1,89 @@
...
+    [~ x ~ [x x@3]]
+    [~ x [~ x? ~ x] x]
+  >`),
+  s("bongo:3").struct("x*4"),
+  s("fingercymbal").struct("x"),
+
+  s("sus_cymbal").struct("<~@8 ~@7 x ~@3 x ~@3 x>"),
+  s("clap:4").gain(0.2).struct("<x(5,16,2)!3 x(7,16,9)>").mask("<~ ~ x>/8".early(1/4)),
+  s("clash2").struct("<x ~@3>").mask("<~ ~ x>/8").gain(0.3),
+  s("crow").struct("<~@4 x ~@3>").mask("<~ ~ x>/8").gain(0.2).echo(10, 1/12, .6).pan(rand),
```
wjt, line 20: 🐦‍⬛

PR #2623, opened 2026-07-29
ToggleableTileMapLayer: Don't store target if it is self

wjt:
ToggleableTileMapLayer: Don't store target if it is self

This script can either be attached directly to a TileMapLayer, or
reference one, but in practice I always attach it directly (using my
favourite quirk of the type system to attach it to a TML despite it
extending Toggleable).

There's no point in storing this property in that case. Clear the
STORAGE flag and update the scenes that use the script.


scenes/world_map/frays_end.tscn
```diff
@@ -221,30 +221,27 @@ tile_set = ExtResource("6_5j6h7")
...
 tile_set = ExtResource("7_r1ek1")
 
-[node name="WestPathBarrierVoid" type="TileMapLayer" parent="TileMapLayers" unique_id=1715727017 node_paths=PackedStringArray("target")]
+[node name="WestPathBarrierVoid" type="TileMapLayer" parent="TileMapLayers" unique_id=1715727017]
 y_sort_enabled = true
 material = ExtResource("11_wymun")
 tile_map_data = PackedByteArray("AAAAAAsADQACAAEAAAD//wsADQAAAAEAAAD//woADQAAAAEAAAAAAAoADQACAAEAAAAAAAkADQACAAEAAAD//wkADQAAAAEAAAD//wgADQAAAAEAAAAAAAgADQACAAEAAAAAAAcADQACAAEAAAD//wcADQAAAAEAAAD//wYADQAAAAEAAAAAAAYADQACAAEAAAAAAAUADQABAAAAAAABAAUADQACAAMAAAD//wUADQAAAAAAAAAAAAwADQACAAEAAAD//wwADQAAAAEAAAAAAA0ADQACAAEAAAAAAA4ADQACAAIAAAD//w4ADQAAAAIAAAD//w0ADQAAAAEAAAA=")
 tile_set = ExtResource("12_f3823")
 script = ExtResource("14_f3823")
-target = NodePath(".")
```
manuq, line 230: Yep, no need for it.

manuq (approved):

PR #2627, opened 2026-07-30
quest_004: Fix intro book not starting at the start of Mythical Meadows

GianPierooo:
The intro book ("The Last Petal") at the start of Mythical Meadows was not showing any story text, because the scene instance overrode `titles`/`pages` with empty arrays. This is now fixed, and the book's content is defined directly in the script so it can't accidentally be emptied out from the scene again.

While fixing this, we also reused the page-turn animation and title-ribbon art already used by the storybook quest picker in Fray's End (`scenes/menus/storybook/`), instead of a static image and a plain label, so the intro book now uses very similar logic/art to the one already working well in Fray's End.

Resolves the intro book issue in the reported bug list.


GianPierooo:
Hi @manuq, this is ready for review and merging. I’ve fixed the issue where the introductory book wasn't displaying text at the start of Mythical Meadows :)

manuq (approved):

PR #2629, opened 2026-07-30
Shaker: Allow shaking position without rotation

manuq:
For allowing the same effect than a Camera2D would have without a camera, by shaking instead the root node. For instance, in a cutscene that is played as overlay inside a Control node.

Keep the default as before.


manuq:
Needed to simplify https://github.com/endlessm/threadbare/pull/2594 .

wjt (approved):

PR #2630, opened 2026-07-30
after the tremor: Add CC0 license for boss asset

wjt:
I happened to find the original asset that this one is derived from
while browsing on opengameart.org!


manuq (approved):
Good finding!

PR #2638, opened 2026-07-31
teleporter_test_2: Use blank next_scene to mean "same scene"

wjt:
Previously these explicitly specified the path/uid to the current scene.
This works, but the documentation says to use a blank value to mean
"same scene".

Doing this matches the documentation, and also means that an internal
teleporter copy-pasted into another scene will remain an internal
teleporter.

Follow-up to https://github.com/endlessm/threadbare/pull/2636.


felixwalberg (approved):
>  an internal teleporter copy-pasted into another scene will remain an internal teleporter.

Perfect! This eases the process of finding something you like and wanting to integrate it to your own level.

PR #2640, opened 2026-07-31
GameState: Fix "Guessed quest ..." message when running in editor

wjt:
For reasons unknown (to me), `quest.resource_path` is blank in this code
path when I run a scene in the editor with F6:

> Guessed quest  from scene res://scenes/dev/basics/teleporter/teleporter_test_2.tscn

(Note the double space.)

We already know the quest's path because we loaded it by path after
guessing it. Print that instead. Move the print to before the
`ResourceLoader.load()` call: if that call fails, then the messages will
be in a more logical order.

### DIFF
--- a/scenes/globals/game_state/game_state.gd
+++ b/scenes/globals/game_state/game_state.gd
@@ -151,10 +151,10 @@ func guess_quest(scene_path_or_uid: String) -> void:
 	while dir_path != "res://":
 		var quest_path := dir_path.path_join("quest.tres")
 		if ResourceLoader.exists(quest_path, "Resource"):
+			prints("Guessed quest", quest_path, "from scene", scene_path)
 			var q := ResourceLoader.load(quest_path) as Quest
 			quest = QuestState.new(q, PlayerState.new())
 			quest.challenge_start_scene = scene_path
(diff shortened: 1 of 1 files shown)


felixwalberg (approved):
Makes sense, if the quest path is already known from the path we used to guess it there is no need to try and get it another way.

manuq (approved):

manuq:
> Makes sense, if the quest path is already known from the path we used to guess it there is no need to try and get it another way.

The thing is, before the output was:

> Guessed quest scenes/dev/basics/teleporter/quest.tres from scene res://scenes/dev/basics/teleporter/teleporter_test_2.tscn

But now that `quest.resource_path` is null or an empty string.

> For reasons unknown (to me), `quest.resource_path` is blank in this code
> path when I run a scene in the editor with F6

This is the same in any other scene I tried. For instance:
```
Guessed quest  from scene res://scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
Guessed quest  from scene res://scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn
```

manuq (approved):

PR #2644, opened 2026-08-03
Push StoryQuest kit as Git tag

wjt:
The goal here is that, if we want to import a StoryQuest from Backstitch
back into Git, we can do so as follows:

1. Look in the project settings which Git version the StoryQuest kit
   came from;
2. Check out the tag for that kit;
3. Replace all files in the directory (except `.git`) with the
   Backstitch project;
4. Commit.
5. Now we have a clean commit for "add new StoryQuest" which we can
   rebase onto upstream main.

This would also be a useful thing to have if we want to continue to
support running Core: Threadbare in Git(Hub). Two ideas that would both
be better than what we have today:

1. Automatically sync these tags to the `main` branch of a separate
   `threadbare-storyquest-kit` Git repo, and adjust the setup script to
   fork that. Configure a workflow to close all pull requests on that
   separate repo with the "wrong target" message.

2. Rewrite the setup script so that rather than forking our repo, it
   instead creates a new one (not in a fork relationship with
   endlessm/threadbare) and pushes the latest such tag as its `main`.

I have a fancier version of this patch which synthesises a merge commit
between the commit that the kit is based on and the most recent commit
on a `storyquest-kit` branch but I decided that the complexity isn't
worth it unless we decide we want to go down one of those routes.

I also tacked on a change to move the `rm -r` from GDScript to shell. I think it's an improvement.

### DIFF
--- a/tools/sqckck.sh
+++ b/tools/sqckck.sh
@@ -2,8 +2,9 @@
 # SPDX-FileCopyrightText: The Threadbare Authors
 # SPDX-License-Identifier: MPL-2.0
 set -ex
-: ${GODOT:=godot}
-pushd $(dirname "$(dirname "$0")")
+: "${GODOT:=godot}"
+: "${GITHUB_OUTPUT:=/dev/fd/1}"
+pushd "$(readlink -f "$(dirname "$0")/..")"
 
 VERSION=$(git describe --tags)
 declare -a PRUNE_FOLDERS=(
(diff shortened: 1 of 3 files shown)


tools/sqckck.sh
```diff
@@ -6,12 +6,19 @@ set -ex
...
+	"scenes/quests/story_quests"
+	"scenes/world_map"
+)
 
 echo "::group::Importing project"
 $GODOT --headless --import
 echo "::endgroup::"
 
 echo "::group::Pruning project"
+rm -r "${PRUNE_FOLDERS[@]}"
```
wjt, line 22: This is so much shorter in bash...

manuq (approved):
All good! Nothing to comment.

PR #2647, opened 2026-08-03
Storybook: Page content fades in and out

### LINKED ISSUE #2645: Storybook: fade page content in and out during page flips
Alternative to https://github.com/endlessm/threadbare/issues/1816

Currently when you turn the page in the storybook, the pages go blank.

https://private-user-images.githubusercontent.com/86760/534683602-1347fab8-7536-47cd-8feb-5f32c622b0e1.webm?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODU3NjU5ODEsIm5iZiI6MTc4NTc2NTY4MSwicGF0aCI6Ii84Njc2MC81MzQ2ODM2MDItMTM0N2ZhYjgtNzUzNi00N2NkLThmZWItNWYzMmM2MjJiMGUxLndlYm0_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwODAzJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDgwM1QxNDAxMjFaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0xNjQxZjYwMTEzOTRkMzQyMjU3Yjk2MTZiZWMwMzY5MjkwYWYyMTEzYWYzMThjNDQ1MjkyNDdmM2I5N2ZhYmFhJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9dmlkZW8lMkZ3ZWJtIn0.mifwLFOzOqRBmUcDNULBLB8g0Uosl9Ysek5CMlfpbBU

A good idea would be to fade the page content out, have the page turn, and then fade the new content in.

alcole2:
When you flip a page in the Storybook, what's currently on the page will fade out, the page flipping animation will play, and then the new page content will fade in.

Resolves #2645

### DIFF
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -36,10 +36,10 @@ var _current_list_page: int = 0
 @onready var ui_container: Control = %StoryBookContent
 
 
-func _fade_out_ui():
+func _fade_out_ui() -> void:
+	ui_container.visible = false
 	var tween := create_tween()
 	tween.tween_property(ui_container, "modulate:a", 0.0, fade_duration)
-	return tween.finished
 
 
 func _fade_in_ui() -> void:
@@ -180,7 +180,6 @@ func _switch_to_page(spread_index: int) -> void:
 
 	if old_index != -1:
 		await _fade_out_ui()
-		ui_container.visible = false
 
 		if spread_index > old_index or (spread_index == 0 and old_index == total_spreads - 1):
 			animated_book.play("book_right")
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -18,6 +18,7 @@ signal selected(quest: Quest, restart: bool)
 	set(value):
 		quests_per_page = value
 		quests_per_spread = value * 2
+@export var fade_duration: float = 0.15
 
 var quests_per_spread: int = 16
 
@@ -33,6 +34,18 @@ var _current_list_page: int = 0
 @onready var back_button: Button = %BackButton
 @onready var animated_book: AnimatedSprite2D = %AnimatedSprite2D
 @onready var ui_container: Control = %StoryBookContent
+
+
+func _fade_out_ui():
+	var tween := create_tween()
+	tween.tween_property(ui_container, "modulate:a", 0.0, fade_duration)
+	return tween.finished
+
+
+func _fade_in_ui() -> void:
+	ui_container.visible = true
+	var tween := create_tween()
+	tween.tween_property(ui_container, "modulate:a", 1.0, fade_duration)
 
 
 func _ready() -> void:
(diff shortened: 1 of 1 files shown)


scenes/menus/storybook/components/storybook.gd
```diff
@@ -35,6 +36,18 @@ var _current_list_page: int = 0
 @onready var ui_container: Control = %StoryBookContent
 
 
+func _fade_out_ui():
```
manuq: There is an error reported on this line. You probably see it as well in the script editor in yellow:

image

I think you can fix it by:
- Adding "void" as return type, same as in the `_fade_in_ui()` function.
- Awaiting for the `tween.finished` signal instead of returning it.

Check [Awaiting signals or coroutines](https://docs.godotengine.org/en/4.4/tutorials/scripting/gdscript/gdscript_basics.html#awaiting-signals-or-coroutines):
> Using the `await` keyword with a signal or a call to a function that is also a coroutine will immediately return the control to the caller. When the signal is emitted (or the called coroutine finishes), it will resume execution from the point on where it stopped.

scenes/menus/storybook/components/storybook.gd
```diff
@@ -166,6 +179,9 @@ func _switch_to_page(spread_index: int) -> void:
 	_current_spread_index = spread_index
 
 	if old_index != -1:
+		await _fade_out_ui()
+		ui_container.visible = false
```
manuq: I would make the added functions `_fade_out_ui()` and `_fade_in_ui()` more consistent, by moving this immediate `ui_container.visible = false` to the `_fade_out_ui()`. Because `_fade_in_ui()` already does `ui_container.visible = true` before tweening.

manuq (changes requested):
The Storybook is looking good with the fade out / fade in! There is one error to fix.

manuq (changes requested):
@alcole2 sorry, have you tested it? Now the fade-in doesn't happen because of (1) the container is set to invisible before the tweening happens, not after. And there is a new error now (2) because there is no `await tween.finished` at the end of the function.

Image

alcole2:
I tested it, I couldn't tell if the fade out was the same or not but assumed it was because it didn't look that different from before. I'll fix it, sorry about that!

manuq:
> I tested it, I couldn't tell if the fade out was the same or not but assumed it was because it didn't look that different from before. I'll fix it, sorry about that!

no problem!

manuq (approved):
Excellent! Works nice now and without any errors.

manuq:
Could be that this works only with the a single 2-pages table of contents? If I try the `storybook_test.tscn` I see no fading when turning to the second 2-pages ToC: https://play.threadbare.game/branches/endlessm/main/#menus/storybook/components/storybook_test

[recording.webm](https://github.com/user-attachments/assets/3bf8e54b-5b5d-467c-96e9-587f16062680)

@alcole2 can you open a followup task about this?

alcole2:
@manuq Sure! I made a new task and pull request for it, #2652

PR #2643, opened 2026-08-01
Create visuals demo quest for tilesets and y-sorting

felixwalberg:
Add a quest to Dev Archipelago basics island that explains the importance of y-sorting (a common issue in StoryQuests), a few examples of using different tilesets and some key practices (foam, creating elevation, use of shadows, etc.)

Resolves #2642

PR #2649, opened 2026-08-03
Add Interactive Y-Sort Example

felixwalberg:
Adds a visual of y sorting between two sprites with `InteractArea`s used to interactively enable/disable y sorting in game. This also addresses some of the inconsistencies with elevation and bridge shadows as mentioned in the review of #2643. 

My last bridge goes from an island with no elevation to an island elevated one level up. I left the shadows in for this one, but the perspective is a little off.


wjt:
image

I love this detail

wjt (approved):

manuq:
> My last bridge goes from an island with no elevation to an island elevated one level up. I left the shadows in for this one, but the perspective is a little off.

Yes:

image

One way to sort this is to add a signpost explaining this exact problem :)

PR #2655, opened 2026-08-04
Add Ink Combat hub to Dev Archipelago

### LINKED ISSUE #2539: Dev Archipelago: Repelling
- Include `custom_repellable_objects.tscn`
- Design Ink Combat to show off all variations (these should be connected via teleporters)
  - Targets that are cracked by incorrect colors
  - Targets that must be filled in a set order
  - Increased Projectile speed/duration/other behaviors to show they can be changed
  - General Ink barrel filling combat
- If Repel gets an upgrade, this would be shown here as well

felixwalberg:
Add a scene `ink_combat_test.tscn` that walks the player through the different variations of ink combat in Threadbare. Each section of the scene is "replayable" via the repellable levers. Only one section can run at once.

Resolves #2539


felixwalberg:
Also wondering if the scene I made should be modified so that all the barrels are visible in the same frame... I mentioned this in a signpost, but it might add unnecessary difficulty.

felixwalberg:
@wjt @manuq @jgbourque Going to leave this as a draft for now, but I have made some changes (currently in `ink_combat_test.tscn`) to address some of the feedback from last week's build review. It isn't completely the same as the proposed "arena" where everything is firing at once, since some of the examples might get messed up running simultaneously (e.g. the fragile barrels were getting broken before the Player can make it to that section). It is all reachable within the same scene with a little less running. However, I would be open to moving things closer together if that is preferred. I could also look at adding some logic to levers to activate certain throwing enemies and certain combinations of the barrels.

https://play.threadbare.game/branches/endlessm/dev-archipelago-ink-combat/#dev/repel/ink_combat/ink_combat_test

There are still things to do, but wanted to continue this discussion on how best to display the ink combat mechanics.
- Add script to activate the fragile barrel throwing enemy based either on completion of previous section or player entering an area2d (which I think is a better option to avoid hiding stuff behind progress gates)
- Add a similar concept for the sequential completion
- Decorate/make the borders look nicer
- We could possibly add levers in each section that could "reset" the enemy/barrels for repeated practice

jgbourque:
I really like the idea of levers for an in game reset of the mechanic/challenge etc. That way there can be extended playtesting of an idea without having to restart the scene necessarily. 

I also like the idea of the player entering an area as an option to activate whatever components are set up to interact with there rather than progress gating!

felixwalberg:
Some major changes have been made here since the last time this was open for review! Now the majority of variations of ink combat are showcased within one scene. The user can activate a section to experiment with one mechanic at a time. Not complete "sandbox" freedom, but more inclusive than the last version!

felixwalberg:
I am open to suggestions on a better way to handle the "sections"! A lot of the logic is deferred due to issues with replacing `ThrowingEnemy` nodes mid attack animation. It follows the general idea of only include the nodes in groups if they are within the chosen section. Rather than adding a "stop" or idle logic, I just reset them (sometimes you might want to reset mid combat vs after completely filling the barrels).

scenes/dev/repel/ink_combat/ink_combat_test.gd
```diff
@@ -0,0 +1,75 @@
...
+## Use index of section to add and remove ThrowingEnemy and
+## FillingBarrel from their groups before starting the
+## ink combat sequence.
+func set_section(selection: int) -> void:
+	_current_section = sections.get_child(selection)
+	for section in sections.get_children():
+		section.call_deferred("reset")
+
+	# Cooldown between reset and start
+	await get_tree().create_timer(2).timeout
```
felixwalberg, line 34: This is still an issue that I would love to get another pair of eyes on! Without a timeout delay, there is some call happening too early, resulting in enemies that don't activate their attack. I assume there is a function call later that should be deferred, but I was having trouble resolving this!

This also makes the initial delay 2 + `ThrowingEnemy.throwing_period` seconds, which can feel like a long time! Maybe we need an indication that the section is starting.

felixwalberg:
After the build review today (Aug 13), the following changes have been made:
- Add comments to top of scripts (`ink_combat_test.gd`, `ink_combat_section.gd`, and `ink_combat_fragile_section.gd`) indicating this is not necessary to make ink combat levels
- Reset `current_target_index` and enable randomization of the `BarrelUnlockSequence` section so each reset gets a fresh (and working) sequence
- Modify the `vase_destruction_tracker.gd` logic for the `FragileBarrel` section (now breaking all barrels defeats the StoryWeaver and filling one barrel is enough to win)
  - Add checkpoint to this section for respawning after defeat
- Darkness now follows player direction in teleporter transition

Some other notes from the review that might be relevant:
- We could (in a later change) modify the InkDrinker to start firing projectiles immediately, since a delay can be implemented with a timer node but not the other way around
- We want a way to spawn InkDrinkers, which would simplify this scene. Each section could trigger a spawner of a set of enemies, clearing the rest of the scene
- The fragile barrels have collisions enabled even after breaking. They exist on the player and the projectile collision layers

scenes/dev/repel/ink_combat/ink_combat_test.gd
wjt: The two states of the levers are visibly different: there is a lit-up "on" state and a darker "off" state. So, just an idea: you could use the "on" state as the visible indication. When one lever is switched to the "on" state, set all the others back to "off". When the fill game logic reports that the game is "completed", set the currently-"on" lever to "off".

(I'll look at the logic in more detail.)

scenes/dev/repel/ink_combat/ink_combat_fragile_section.gd
```diff
@@ -0,0 +1,57 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+
+## This logic is *not* needed to add Ink Combat to your level. Simply
+## grab the ThrowingEnemy and FillingBarrel nodes and copy them to
+## your scene. This logic helps with sectioning off each part of Ink Combat.
+## This is a simplification of vase_destruction_tracker.gd [VaseDestructionTracker]
```
wjt, line 7: The code really is almost identical and IMO the changes are a nice improvement. I don't think you need to do this in this PR but I suggest we should file an issue to track merging these again:

1. Move the code that adjusts `fill_game_logic.barrels_to_win = 1` to `ink_combat_test.gd` just before calling `setup()` on this node
2. Replace `vase_destruction_tracker.gd` in `res://scenes/quests/lore_quests/song_sanctuary_3/1_ink_drinker_levels/ink_combat_round_5.tscn` with this script (under a different name probably). Adjust the scene to call `setup()` on this script before calling `start()` on the game.
3. Remove the unnecessary extra `Vases` group from the nodes in that scene.
4. That's it I think?

scenes/dev/repel/ink_combat/ink_combat_section.gd
```diff
@@ -0,0 +1,59 @@
...
+
+func _ready() -> void:
+	_update_barrels()
+	_update_enemies()
+
+
+## Store a copy of the FillinBarrel nodes before
+## any changes occur during ink combat.
+func _update_barrels() -> void:
+	for barrel: FillingBarrel in get_child(1).get_children():
```
wjt: I don't like relying on the order of children, particularly since you also rely on the order of children of this node in `ink_combat_test.gd`.


I suggest that you define something like:

```gdscript
@onready var enemy_container = $Enemies
@onready var barrel_container = $Barrels
```

(or if you prefer use `@export var enemy_container` but in this case I think the relative lookup by path is fine).

scenes/dev/repel/ink_combat/ink_combat_section.gd
```diff
@@ -0,0 +1,59 @@
...
+		_enemies.append(enemy_copy)
+
+
+## Add fresh duplicate nodes to the scene tree and
+## corresponding groups to replace the ones depleted
+## in ink combat.
+func reset() -> void:
+	for enemy: ThrowingEnemy in get_child(0).get_children():
+		if is_instance_valid(enemy):
+			enemy.call_deferred("free")
```
wjt: I think the reason you are using `enemy.call_deferred("free")` here rather than `enemy.queue_free()` is that `queue_free()` happens after all `call_deferred` calls, but you want these enemies to be freed sooner, before the deferred `_update_enemies` call happens.

I think you should be more explicit, something like:


```suggestion
	var enemy_holder := get_child(0)
	for enemy: ThrowingEnemy in enemy_holder.get_children():
		if is_instance_valid(enemy):
			enemy_holder.remove_child(enemy)
			enemy.queue_free()
```

scenes/dev/repel/ink_combat/ink_combat_section.gd
```diff
@@ -0,0 +1,59 @@
...
+
+	for barrel: FillingBarrel in _barrels:
+		get_child(1).add_child(barrel)
+		barrel.add_to_group("filling_barrels")
+
+	_barrels = []
+	_update_barrels()
+	# Enemies must be deferred, to avoid errors
+	# with the attack animation flushing.
+	call_deferred("_update_enemies")
```
wjt: I guess that what's happening here is that every time you `reset()` the section, you do this:

1. Add all the nodes from `_enemies` back to the scene
2. Then `duplicate()` them all back into `_enemies`

I think you should do this differently (resp. barrels):

1. In `_ready()`, snapshot `duplicate()`s of all the enemies into `_enemies` (resp. barrels). Do not modify `_enemies` after this point.
2. When you need to replace the enemies, duplicate each element of `_enemies` and add the duplicate back to the enemy container node

scenes/dev/repel/ink_combat/ink_combat_test.gd
wjt: As discussed in the build review last week I think we should change the enemy logic. Currently once started, the enemies wait for `throwing_period` (or `throwing_period / 2` if odd shoot is enabled) before their first throw. I think instead they should throw immediately (or wait `throwing_period / 2` if odd shoot is enabled).

This would fix the feeling like a long time, both here and in quest scenes that use these enemies.

scenes/dev/repel/ink_combat/ink_combat_test.gd
```diff
@@ -0,0 +1,86 @@
...
+	if _current_section == sections.get_child(3):
+		barrel_unlock_sequence.barrels = []
+		for barrel in get_tree().get_nodes_in_group("filling_barrels"):
+			barrel_unlock_sequence.barrels.append(barrel)
+		barrel_unlock_sequence.current_target_index = 0
+		barrel_unlock_sequence.start_sequence()
+
+	# Start up the throwing enemies remaining in the group
+	for enemy in get_tree().get_nodes_in_group("throwing_enemy"):
+		enemy.start()
```
wjt: I think you could use:

```suggestion
	get_tree().call_group("throwing_enemy", "start")
```

as FillGameLogic.start() itself does. But why not call that rather than calling `fill_game_logic.call_deferred("_update_allowed_colors")` above?

wjt (approved):
There's some polish that could be done here - suggestions below - but some/all of it could be deferred to follow-ups. And I think  rather than sinking masses of time into diagnosing the "delay 2 seconds here to fix bug" type points, the time might be better spent thinking about how we could generalise/improve the fill game logic so that it can be more easily used for this kind of purpose.

felixwalberg:
I am going to leave the remaining suggestions for a future change, so that this can be included in my demo today.

Issue #2443, opened 2026-06-23
Townie got stuck on the spot when walking away from retelling


wjt:
### Threadbare version

v0.2.0-77-g2a3f4a571 - commit 2a3f4a5

### Affected scene

res://scenes/world_map/frays_end.tscn

### Description

After an end-of-quest retelling, this happened:

https://github.com/user-attachments/assets/b701d5cf-3f3b-491e-9832-2818328a614f

I saw this exactly once, at Serious Play.

I think what's happening is that the PathWalkBehavior got stuck in the "is at end of path" state.

### Checklist

- [x] I've reviewed and agree to follow the code of conduct

PR #2659, opened 2026-08-05
Frays End: Fix kink in retelling townie path

wjt:
The control points on one of the points on the path were the wrong way
round, causing the townie to briefly turn back and back again when it
hit that point on the path in each direction.

Smooth out the path.


wjt:
This video shows the problem:

https://github.com/user-attachments/assets/2327af80-4371-4bf8-ba5f-d82c8277294b

wjt:
| Before | After |
| -- | -- |
| Screenshot From 2026-08-05 13-53-53 | Screenshot From 2026-08-05 13-54-02 |

wjt:
I hoped this was the same point as shown in https://github.com/endlessm/threadbare/issues/2443 but it's not :(

wjt:
Annoyingly the only way to test this is to finish a quest that has a retelling. The shortest such quest is the tutorial.

manuq (approved):
Thanks!

Issue #2542, opened 2026-07-13
Dev Archipelago: Level Design


felixwalberg:
- Group together some of the existing scenes that showcase level design tools during runtime and general tilemap elements and add them to Dev Archipelago Elder
- Ex: `1_time_and_weather.tscn`, `bridge_gym.tscn`, background music as a hint in sequence puzzles, checkpoint demo
- Showcase the area filler tool on a smaller scale where it can be manually triggered during runtime

Issue #2633, opened 2026-07-30
Explain or add way to open the dev scene in the editor


manuq:
Let's imagine a learner is going through the dev archipielago and wants to tweak the scene that's currently playing, or check how it's currently made, as the scene is proposing (in signposts, for instance).

The scene name can be found in the Remote dock. Still this requires that you hover the mouse in the node, and use the Filesystem dock to search for it. This could be easier if automated.

https://github.com/user-attachments/assets/05685651-59d0-460e-98bb-6844618a38d4

So maybe we can have a button in the editor to "Edit the scene being played". If not at least we should tell the learner in the Dev Archipielago how to find and edit the scene.

wjt:
Not only can you find the scene name there, but you can click the "open scene" button on it:

<img width="552" height="548" alt="Image" src="https://github.com/user-attachments/assets/72ee5dcc-bff4-40dd-832f-d13d1c41d50e" />

I was curious whether it's possible to automatically show the Remote tab while the game is running. Yes, it is, but the option's tooltip explains why it is not the default:

<img width="941" height="442" alt="Image" src="https://github.com/user-attachments/assets/0a8e56be-7382-4a4c-9584-da9ed2ba97ee" />

> [!WARNING]
> Enabling this setting can cause stuttering when running a project with a large amount of nodes (typically a few thousands of nodes or more), even if the editor window isn't focused. This is due to the remote scene tree being updated every second regardless of whether the editor is focused.

wjt:
I tried to write an addon to add an "edit this scene" button when the game is running but it seems to be more work than I expected.

It's easy to write an `EditorDebuggerPlugin` to put a giant button into the debugger panel, but there is no built-in way to access the running game's scene tree from an `EditorDebuggerPlugin` as far as I can tell. So you have to send messages between the debug plugin and code in the running game as documented in https://docs.godotengine.org/en/stable/classes/class_editordebuggerplugin.html#class-editordebuggerplugin.

manuq:
> Not only can you find the scene name there, but you can click the "open scene" button on it:

Oh I didn't see the button before! That's it then!

> I tried to write an addon to add an "edit this scene" button when the game is running but it seems to be more work than I expected.
> 
[...]

PR #2664, opened 2026-08-05
Create interactive area filler demo for Dev Archipelago

felixwalberg:
Add scene to highlight `AreaFiller` mechanic with interactive rocks to "refill" field with `flower.tscn` and `hiding_mushroom.tscn`.

Update `area_filler.gd` script to flag for usage during runtime. The current tool uses "EditorPlugin" which only works in the editor but not while the game is running. The fill method now has a conditional branch for in game usage.

Part of #2542


scenes/game_elements/props/area_filler/area_filler.gd
```diff
@@ -35,6 +35,9 @@ extends Node
 ## this value.
 @export_range(16.0, 256.0, 1.0, "suffix:px", "or_more") var minimum_separation: float = 64.0
 
+## Enable this to modify the fillable area while the game is running.
+@export var in_game_filler: bool = false
```
wjt, line 44:
```suggestion
## Enable this to modify the fillable area while the game is running. If not enabled, this node will remove itself from the tree at runtime.
@export var in_game_filler: bool = false
```

scenes/game_elements/props/area_filler/area_filler.gd
wjt: I think you could write this in a less confusing way:


```suggestion
if Engine.is_editor_hint():
    var plugin: Node = ClassDB.instantiate("EditorPlugin")
    _undoredo = plugin.get_undo_redo()
    plugin.queue_free()
elif not in_game_filler:
    self.queue_free()
```

scenes/game_elements/props/area_filler/area_filler.gd
```diff
@@ -232,3 +237,31 @@ func fill() -> void:
...
 
 	_undoredo.commit_action()
+
+
+## Similar to [method AreaFiller.fill], but works during runtime.
+## Clears [member area] (except for this node and any collision shapes),
+## generate a new set of points according to the current
+## parameters, and fill [member area] with instances of [member scenes]
+## at those points.
+func in_game_fill() -> void:
```
wjt: I don't think this should be a separate function. I think you should have an `if Engine.is_editor_hint()` branch in `fill()`, after the common code. You can move the `_undoredo.create_action("Refill area", UndoRedo.MergeMode.MERGE_DISABLE, scene, false)` into the editor branch.

scenes/game_elements/props/area_filler/area_filler.gd
```diff
@@ -232,3 +237,31 @@ func fill() -> void:
...
+	var scene := get_tree().edited_scene_root
```
wjt: `SceneTree.edited_scene_root` is always `null` in exported builds.

We don't really care about the `owner` at runtime, so `null` is fine, or you could set it to the `owner` of the AreaFiller node itself. (It is important in the editor because this is what causes the nodes to actually be saved back to the scene that's being edited.)

scenes/game_elements/props/area_filler/area_filler.gd
```diff
@@ -232,3 +237,31 @@ func fill() -> void:
...
+
+	for child: Node in _area.get_children():
+		if child != self and child is not CollisionPolygon2D and child is not CollisionShape2D:
+			old_children.append(child)
+
+	var points := _generate_points()
+	for point in points:
+		new_children.append(_prepare_child(point))
+
+	# When performing the action in either direction, we want to remove, then add.
```
wjt: This comment doesn't really make sense ("either direction") in the gameplay-time implementation because there is only one direction.

scenes/game_elements/props/area_filler/area_filler.gd
```diff
@@ -232,3 +237,31 @@ func fill() -> void:
...
+	for child in old_children:
+		_area.remove_child(child)
+
+	for child in new_children:
+		_area.add_child(child, true)
```
wjt: The second argument to `Node.add_child()` is `force_readable_name`, and the [documentation](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-add-child) says:

> This operation is very slow. As such, it is recommended leaving this to false, which assigns a dummy name featuring @ in both situations.

In the editor we want the trees to have readable names because the generated names are ugly. In-game we don't really care because no-one's looking at the names. I would leave this as the default `false` in the in-game codepath.

wjt (changes requested):

scenes/dev/level_design/interactive_area_filler_test.tscn
```diff
@@ -0,0 +1,307 @@
...
+collision_layer = 32
+collision_mask = 0
+script = ExtResource("9_51qfj")
+metadata/_custom_type_script = "uid://du8wfijr35r35"
+
+[node name="CollisionShape2D" type="CollisionShape2D" parent="OnTheGround/FabricRock/InteractArea" unique_id=1977446241]
+shape = SubResource("CircleShape2D_pp1gq")
+
+[node name="Woods" type="Area2D" parent="OnTheGround" unique_id=1080322007]
+y_sort_enabled = true
```
wjt: Conventionally for dense forests like this, we use a `StaticBody2D` to prevent the player from being able to walk through it at all. (See res://scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn for an example: the sparse areas of trees are `Area2D` while the dense areas are `StaticBody2D`.)

I think it's important to demonstrate both. I suggest you change this one to a `StaticBody2D`, with no `collision_mask` bits set and the `collision_layer` `5` ("walls") set. You could have another sign that says e.g.

> This dense forest is a StaticBody2D, which prevents the player from walking in between the trees even if they would fit.

The phrasing in area_filler_test is:

> This Forest is a boundary for the level.
> It is a StaticBody2D with a
> CollisionPolygon2D that blocks the player.

Of course in this new scene the water is the boundary!

scenes/dev/level_design/interactive_area_filler_test.tscn
```diff
@@ -0,0 +1,307 @@
...
+sprite_frames = ExtResource("3_8o87r")
+
+[node name="Camera2D" type="Camera2D" parent="OnTheGround/Player" unique_id=2136918817]
+limit_left = -100
+limit_top = -950
+limit_right = 2025
+limit_bottom = 1000
+
+[node name="FlowerBed" type="Area2D" parent="OnTheGround" unique_id=2026112211]
+y_sort_enabled = true
```
wjt: I generally set `monitoring` to `false`, `monitorable` to `false`, and clear the collision layers & mask in `Area2D`s that are being used purely to provide shapes to `AreaFiller`. My belief is that the first two reduce the amount of work that Godot's physics simulation has to do; and that the latter two is good practice because by default both are set to the "players" physics layer.

scenes/dev/level_design/interactive_area_filler_test.tscn
```diff
@@ -0,0 +1,307 @@
...
+[node name="Sign" parent="OnTheGround/Signs" unique_id=1956680058 instance=ExtResource("4_vilsq")]
+position = Vector2(345, 651)
+direction = 1
+text = "Sometimes you want to decorate a large area
+without placing each scene by hand."
+
+[node name="Sign2" parent="OnTheGround/Signs" unique_id=1009085323 instance=ExtResource("4_vilsq")]
+position = Vector2(587, 563)
+direction = 1
+text = "Threadbare uses a tool called AreaFiller,
```
wjt:
```suggestion
text = "Threadbare includes a tool called AreaFiller,
```

scenes/dev/level_design/interactive_area_filler_test.tscn
```diff
@@ -0,0 +1,307 @@
...
+which can automatically populate an Area2D's shapes
```
wjt:
```suggestion
which can automatically populate an Area2D or StaticBody2D's shapes
```

scenes/dev/level_design/interactive_area_filler_test.tscn
wjt: and the fact that the Area2D is not used for physics/detection in any way - it's only (ab)used to have an easy way to edit shapes - could be documented in a sign. The phrasing I used in `area_filler_test` is:

> The Copse Area2D is only used
> to hold the trees. It is set to not
> collide with anything.

scenes/game_elements/props/area_filler/area_filler.gd
wjt:
```suggestion
## [member scenes]. These children are saved to the owning scene.
##
## By default, this node is only used in the editor and frees itself when the game is running. Enabling [member in_game_filler] allows you to use it in-game. But beware: the whole game freezes while the area is being filled, and it can be slow!
```

wjt (approved):
Nice!

I guess you are planning to link this to the archipelago or put it in a devquest in a follow-up change.

I have a few polish items now that I've looked more closely at the scene itself.

I think we could probably delete res://scenes/game_elements/props/area_filler/components/area_filler_test.tscn after these changes.

-----

A thought I had while reviewing this: I don't think we want to duplicate all the documentation we already have into Dev Archipelago. I wonder if we can tell the player, in some dialogue, about the various ways they can access documentation in the editor (e.g. right-clicking a node, or its properties in the inspector, and picking "Open Documentation" from that menu).

felixwalberg:
> A thought I had while reviewing this: I don't think we want to duplicate all the documentation we already have into Dev Archipelago. I wonder if we can tell the player, in some dialogue, about the various ways they can access documentation in the editor (e.g. right-clicking a node, or its properties in the inspector, and picking "Open Documentation" from that menu).

Good point. It would definitely be a useful skill to know how to check documentation in the editor. Justin had given feedback that the `teleporter_test_2.tscn` not being clear about the spawn point made it tricky to replicate. I think that was part of the reason I was being so explicit with the signposting. Being able to read documentation in the editor goes along with #2633 so that there is a straightforward way for learners to get to the nodes they want to right-click.

This could be added to the dialogue of the NPC we plan to have on the starting island.

felixwalberg:
Based on the conversation in today's  build review (more exploration and less rigid sign instructions), I removed two signs:
> For each decoration you want, add its scene to
 the "scenes" property of the AreaFiller.

and

>Use the "sprite frames" property to randomly
 assign each scene a sprite frames value.

These both address things you would find in the documentation for the `scenes` and `sprite_frames` members when looking in the editor. Having them wouldn't really change how "inspiring" this level is, since it is clearly visible that different scenes + sprite frames can be combined. And as Will said, it might be nice to encourage people to use the "Open Documentation" button in the editor and avoid duplicating it in every dev scene!

wjt (approved):

PR #2670, opened 2026-08-05
Storybook: list of StoryQuests aligns properly regardless of title lengths

### LINKED ISSUE #2617: Storybook table of contents: pages layout depends on titles length
Follow up of https://github.com/endlessm/threadbare/issues/1757

Check: https://github.com/endlessm/threadbare/pull/2592#pullrequestreview-4788859595

It looks like the Control of pages need `Control.minimum_size` that fits the background artwork.

alcole2:
I updated my previous storybook work to ensure that the proper page layout isn't dependent on the StoryQuest title lengths.

Resolves #2617

### DIFF
--- a/scenes/menus/storybook/storybook.tscn
+++ b/scenes/menus/storybook/storybook.tscn
@@ -23,7 +23,6 @@ atlas = ExtResource("10_p0vqm")
 region = Rect2(576, 576, 64, 64)
 
 [node name="Storybook" type="CanvasLayer" unique_id=438542807]
-layer = 10
 script = ExtResource("1_gw8hn")
 quests = Array[ExtResource("3_21r2o")]([ExtResource("5_ywp77")])
 
@@ -103,7 +102,6 @@ layout_mode = 2
 
 [node name="LeftQuestList" type="VBoxContainer" parent="VBoxContainer/CenterContainer/StoryBookContent/MarginContainer/QuestContainer/BookSpread/LeftPage/VBoxContainer" unique_id=887510218]
 unique_name_in_owner = true
-custom_maximum_size = Vector2(450, -1)
 layout_mode = 2
 size_flags_horizontal = 4
 size_flags_vertical = 6
@@ -119,7 +117,6 @@ layout_mode = 2
 
 [node name="RightQuestList" type="VBoxContainer" parent="VBoxContainer/CenterContainer/StoryBookContent/MarginContainer/QuestContainer/BookSpread/RightPage/VBoxContainer" unique_id=1110923687]
 unique_name_in_owner = true
-custom_maximum_size = Vector2(450, -1)
 layout_mode = 2
 size_flags_horizontal = 4
 size_flags_vertical = 4
--- a/scenes/menus/storybook/storybook.tscn
+++ b/scenes/menus/storybook/storybook.tscn
@@ -23,6 +23,7 @@ atlas = ExtResource("10_p0vqm")
 region = Rect2(576, 576, 64, 64)
 
 [node name="Storybook" type="CanvasLayer" unique_id=438542807]
+layer = 10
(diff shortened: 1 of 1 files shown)


scenes/menus/storybook/storybook.tscn
```diff
@@ -23,6 +23,7 @@ atlas = ExtResource("10_p0vqm")
 region = Rect2(576, 576, 64, 64)
 
 [node name="Storybook" type="CanvasLayer" unique_id=438542807]
+layer = 10
```
manuq: Is this change intended? Is not explained in the pull request description. This could be used to put the storybook on top of other things. But I reverted it and don't see any effect when interacting with the elder and opening the book.

scenes/menus/storybook/storybook.tscn
```diff
@@ -88,32 +89,40 @@ horizontal_scroll_mode = 0
 [node name="BookSpread" type="HBoxContainer" parent="VBoxContainer/CenterContainer/StoryBookContent/MarginContainer/QuestContainer" unique_id=1606272031]
 layout_mode = 2
 size_flags_horizontal = 3
-theme_override_constants/separation = 55
+theme_override_constants/separation = 10
```
manuq, line 91: Just for reference, in `storybook_page.tscn` we have a similar book spread and its HBoxContainer has 45 as separation. Ideally we would remove all this duplication! But not a requirement for this particular pull request.

scenes/menus/storybook/storybook.tscn
```diff
@@ -88,32 +89,40 @@ horizontal_scroll_mode = 0
...
 size_flags_horizontal = 3
 theme_override_constants/margin_left = 40
+theme_override_constants/margin_right = 0
 
 [node name="VBoxContainer" type="VBoxContainer" parent="VBoxContainer/CenterContainer/StoryBookContent/MarginContainer/QuestContainer/BookSpread/LeftPage" unique_id=434752580]
 layout_mode = 2
 
 [node name="LeftQuestList" type="VBoxContainer" parent="VBoxContainer/CenterContainer/StoryBookContent/MarginContainer/QuestContainer/BookSpread/LeftPage/VBoxContainer" unique_id=887510218]
 unique_name_in_owner = true
+custom_maximum_size = Vector2(450, -1)
```
manuq: I don't think this custom maximum size is needed. I tried reverting this (and the one in the other page) and I don't see any issue, but please let me know if this was intended. The custom minimum size, instead, is what makes it work properly.

manuq (changes requested):
I can see how this is very hard to setup in the editor, because there aren't any placeholder controls inside LeftQuestList or RightQuestList:

https://github.com/user-attachments/assets/92777450-8e5b-4afb-a021-0fff8324180f

Ideally the book spread would be a single control, not special-cased for the quest list and the StorybookPage. One proposal here https://github.com/endlessm/threadbare/issues/2339 is to use TabContainer.

Feel free to merge after removing the extra changes that aren't needed (custom maximum sizes and layer).

scenes/menus/storybook/storybook.tscn
alcole2: I saw on Google that it could be a potential solution to an unrelated problem I was having. I didn't mean to leave it on there.

scenes/menus/storybook/storybook.tscn
alcole2: I did it on purpose when testing a bunch of different ideas because I wasn't quite sure what the problem was at first. I reverted it and I agree that it's not necessary.

manuq: > I did it on purpose when testing a bunch of different ideas because I wasn't quite sure what the problem was at first. I reverted it and I agree that it's not necessary.

Yes, no problem!

PR #2671, opened 2026-08-05
Update hint sign for spawnpoints in teleporter test

felixwalberg:
Explain that a `spawn_point.tscn` scene must be added to your scene separately and then pointed to by the teleporter. Based on feedback from Justin, it was unclear that spawn points are not part of the teleporter logic and exist as their own scene/script.

Also, reword the hint sign regarding transitions for teleporters with spawn points out of visible range.


manuq (approved):
Better!

PR #2674, opened 2026-08-06
Debug Settings: Add button to open current scene in editor

### LINKED ISSUE #2633: Explain or add way to open the dev scene in the editor

manuq:
And for that, add and register a new plugin that contains an editor debugger plugin. Use the "threadbare_debugger:open_scene_file" message for passing it.

Resolve https://github.com/endlessm/threadbare/issues/2633

### DIFF
--- a/addons/threadbare_editor/threadbare_debugger.gd
+++ b/addons/threadbare_editor/threadbare_debugger.gd
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends EditorDebuggerPlugin
+## Handles Threadbare specific messages sent by the running game through [EngineDebugger].
+
+## Prefix of the messages handled by this plugin.
+const CAPTURE_PREFIX := "threadbare_debugger"
+
+
+func _has_capture(capture: String) -> bool:
+	return capture == CAPTURE_PREFIX
+
+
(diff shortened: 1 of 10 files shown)


wjt (approved):
This is great.

I pushed a patch that also switches to the 2D tab after opening the currently-playing scene. Feel free to keep or drop it as you prefer.

addons/threadbare_editor/threadbare_debugger.gd
```diff
@@ -22,5 +22,6 @@ func _capture(message: String, data: Array, _session_id: int) -> bool:
 			var file_path := data[0] as String
 			if file_path and file_path.ends_with(".tscn"):
 				EditorInterface.open_scene_from_path(file_path)
+				EditorInterface.set_main_screen_editor("2D")
```
manuq, line 25: I looked for this quickly yesterday and couldn't find it. Thanks!

PR #2676, opened 2026-08-07
AreaFiller: Split fill() into clear() + update()

wjt:
In practice I have found it annoying that, if I tweak the area to be
filled, the whole area is refilled from scratch.

Adapt PoissonDiscSampler to handle multiple polygons at once. This
addresses an undocumented issue where if you had two shapes that do not
overlap but are very close to one another, the old implementation could
place items right on the edge of each, not respecting the minimum
separation.

Rewrite fill() to update(), which preserves any existing children that
fall within the current area, removes any children that fall outside it,
and fills in any new regions. (In fact sometimes new trees will be
planted within the unchanged region, because sometimes the sampler finds
new valid unfilled points. That's how randomness works.)

Add a separate clear() function that erases the area in case you
actually want to refill it.

Factor out the code that applies the changes via UndoRedo at edit time
and directly at runtime.

Update the interactive area filler test scene to use clear() + update().

### DIFF
--- a/scenes/dev/level_design/area_filler/interactive_area_filler_test.gd
+++ b/scenes/dev/level_design/area_filler/interactive_area_filler_test.gd
@@ -7,7 +7,8 @@ extends Node2D
 
 ## Refill the FlowerBed on interaction with the FabricRock.
 func _on_interact_area_interaction_started(area: InteractArea) -> void:
-	area_filler.fill()
+	area_filler.clear()
+	area_filler.update()
 	area.end_interaction()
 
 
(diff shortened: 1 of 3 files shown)


wjt:
@felixwalberg perhaps of interest to you!

wjt:
https://github.com/user-attachments/assets/bd5e6537-ee7c-41fd-9bac-72a733cc8068

felixwalberg (approved):
Very nice! I like this change. I was wishing I could change the CollisionPolygon2Ds in the interactive test without resetting the entire area!

PR #2677, opened 2026-08-07
Add intro island to dev archipelago

### LINKED ISSUE #2584: Dev Archipelago: Intro to the islands
Build an intro island with a spawn point connected from Fray's End. This will contain an NPC with dialogue telling players to use Abandon Quest to return to the Archipelago, explain how to "Open Documentation" in the editor + any other useful information about how the Archipelago works as it comes up

felixwalberg:
Add an island before the "basics" island with an NPC talker who gives advice about the islands applicable to all dev quests. Move bridges on toggleable layers and Fray's End teleporter accordingly. 

Now the direction the player leaves Fray's End and enters Dev Archipelago is also consistent (however, I initially built the archipelago envisioning it going to the north (in retrospect this doesn't make sense for tropical islands), so all additional islands are to the northeast of the starting one).

Add shadows to all the cliffs and horizontal bridges. 

Resolves #2584


scenes/dev/components/dev_archipelago_intro_islander.dialogue
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+~ start
+Imagine Islander: Hi StoryWeaver!
+Imagine Islander: Let me give you some advice for exploring the archipelago.
+Imagine Islander: Whenever you get lost trying out some new mechanics, just open up the game menu and click "Abandon Quest". It will bring you right back to the island.
+Imagine Islander: If you stumble across something you like, open the debug settings from the options menu and click "Open Current Scene In Editor" to take a closer look at how the scene works.
```
felixwalberg, line 7: There might be better phrasing for describing how to access the game's various menus/settings pages.

manuq: Ideally we would make it appear and just show it / circle it. But that's hard!

manuq (changes requested):
Nice!

The NPC could be a random townie rather than the Talker which is deprecated:

```
class_name Talker
extends NPC
## @deprecated: Instead of instantiating this class, use a [TalkBehavior] node instead.
```

Image

Not required for this iteration, but maybe in the future if we add more advices, the dialogue can be branching with options, and the user picks what to know about. Or with a bit of state, the dialogue can be linear as-is the first time, and a list of options (as reminders) the following times the user interacts with this NPC.

felixwalberg:
Oops! I realize that sometimes I drag scenes in from the file system rather than searching for them as a node. The deprecated tag was not as visible.

I like the idea of branching dialogue as well. I was thinking it might be annoying to get the same thing every time, but I didn't want to fully remove it after the first time in case users wanted to refer to a specific hint. Having the options would be a good way to solve it. Thanks for the review! I will update the talker to a random townie.

manuq:
> Oops! I realize that sometimes I drag scenes in from the file system rather than searching for them as a node. The deprecated tag was not as visible.

Don't worry, this is useful feedback for finding a better way of surfacing the deprecations.

manuq (approved):

PR #2678, opened 2026-08-10
InteractArea: Set debug color on shapes

wjt:
Use a consistent yellow color for all collision shapes that define an InteractArea (that is, their CollisionShape2D immediate children), so they are easy to identify in the editor.

### DIFF
--- a/scenes/game_elements/props/interact_area/interact_area.gd
+++ b/scenes/game_elements/props/interact_area/interact_area.gd
@@ -20,6 +20,7 @@ signal interaction_ended
 ## Emitted when characters start or stop seeing this area for interaction.
 signal observers_changed
 
+const SHAPE_DEBUG_COLOR := Color(0.6, 0.545, 0.0, 0.42)
 const INDICATOR_SCENE := preload("uid://d252j2mhya0kq")
 
 ## Position at which to show an arrow when this area is being observed. This
@@ -74,6 +75,18 @@ func remove_observer(character_sight: CharacterSight) -> void:
 
 func _get_is_being_observed() -> bool:
 	return bool(_observers.size())
+
+
+func _recolor_shapes() -> void:
+	for child: Node in get_children():
+		if child is CollisionShape2D:
+			child.debug_color = SHAPE_DEBUG_COLOR
+
+
+func _notification(what: int) -> void:
+	match what:
+		NOTIFICATION_CHILD_ORDER_CHANGED when Engine.is_editor_hint():
+			_recolor_shapes()
 
 
 func _ready() -> void:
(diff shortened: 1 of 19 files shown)


scenes/game_elements/props/interact_area/interact_area.gd
```diff
@@ -76,12 +77,27 @@ func _get_is_being_observed() -> bool:
...
 
+func _recolor_shapes() -> void:
+	for child: Node in get_children():
+		if child is CollisionShape2D:
+			child.debug_color = SHAPE_DEBUG_COLOR
+
+
+func _notification(what: int) -> void:
+	match what:
+		NOTIFICATION_CHILD_ORDER_CHANGED when Engine.is_editor_hint() and is_node_ready():
```
manuq, line 88: Oh first time I see `when` in GDScript! I didn't know `match` had it, I would have written this as nested if conditionals.

scenes/game_elements/props/interact_area/interact_area.gd
```diff
@@ -20,6 +20,7 @@ signal interaction_ended
 ## Emitted when characters start or stop seeing this area for interaction.
 signal observers_changed
 
+const SHAPE_DEBUG_COLOR := Color(0.6, 0.545, 0.0, 0.42)
```
manuq, line 23: Not for this PR, but if we add more color codes (like purple for hookable areas) we may want to have them all in a single file. Maybe in a `const SHAPE_DEBUG_COLORS: Dictionary[StringName, Color]`.

manuq (approved):
Personally I love it!

wjt:
I haven't resaved *all* scenes with this change applied, hence draft.

I actually have a script lying around in my inventory branch to rewrite all scenes in the project that use a particular scene. I should be able to do something similar for scenes that use a particular script to do this programmatically. Maybe I can generalise it...

scenes/game_elements/props/interact_area/interact_area.gd
wjt: I have to look up the `match` syntax every time I use it, so I see this form in the documentation every time as well. (Also I first came across pattern matching in Haskell where pattern guards like this are very common.)

manuq:
> I haven't resaved _all_ scenes with this change applied, hence draft.
> 
> I actually have a script lying around in my inventory branch to rewrite all scenes in the project that use a particular scene. I should be able to do something similar for scenes that use a particular script to do this programmatically. Maybe I can generalise it...

Oh it would be great to have it under `tools/`!

manuq (approved):
Confirmed that `NOTIFICATION_CHILD_ORDER_CHANGED` works like a charm to update the color in a `@tool` script.

PR #2679, opened 2026-08-10
Eternal Loom: Pick last elder if no matching one found

### LINKED ISSUE #2662: Test & fix end of NO_EDIT quest
I have a feeling that when, in #2651 I removed the template elder from Fray's End, I broke the end-of-quest sequence. When you complete a quest and return to Fray's End, the code in the eternal loom tries to find the matching elder for the quest so they can thank you.

I want to test this and potentially fix it.

(Actually I think that finishing a quest from Dev Archipelago should take you back to Dev Archipelago, just as in the StoryQuest Kit.)

wjt:
We try to have the correct elder congratulate the player at the end of a
quest. However since I moved the template elder to Dev Archipelago, it
is no longer guaranteed that there is a matching elder for the quest in
Fray's End.

Pick the last one in the scene.

Helps https://github.com/endlessm/threadbare/issues/2662

### DIFF
--- a/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
+++ b/scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
@@ -21,7 +21,8 @@ func _find_elder(quest: Quest) -> Elder:
 		if quest.resource_path.begins_with(elder.quest_directory):
 			return elder
 
-	return null
+	# No matching elder. This was probably a quest launched from Dev Island.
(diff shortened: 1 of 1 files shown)


manuq (approved):

manuq:
Sorry, I didn't spot https://github.com/endlessm/threadbare/issues/2662 when doing the smoke test for the release because I picked a StoryQuest that was too hard and abandoned it, ouch.

PR #2680, opened 2026-08-10
Tweak quest ending in StoryQuest Kit

### LINKED ISSUE #2662: Test & fix end of NO_EDIT quest

wjt:
Remove the reference to "Fray's End" and "Eternal Loom" in the last line
of NO_EDIT_outro's dialogue. Those entities are not in the StoryQuest
kit. Just speak more generically about taking the threads home.

We don't have a loom in Dev Archipelago, so we don't run the whole
end-of-quest sequence. Instead of silently marking it as finished, play
the normal "thank you, StoryWeaver!" dialogue. Have the StoryQuest Elder
say it because we have a reference to him! Do this even in the normal
project, rather than only if we detect we are in a StoryQuest Kit. I
don't think this will happen in practise right now, but you could
imagine that some of the dev "quests" will deliberately take you back to
Dev Archipelago rather than having no way out except to abandon the
quest.

Helps https://github.com/endlessm/threadbare/issues/2662

### DIFF
--- a/scenes/dev/components/dev_archipelago.gd
+++ b/scenes/dev/components/dev_archipelago.gd
@@ -19,9 +19,10 @@ func _ready() -> void:
 	bridge_to_frays_end.set_toggled(is_storyquest_kit)
 	bridge_to_frays_end_broken.set_toggled(not is_storyquest_kit)
 
-	if is_storyquest_kit and GameState.quest:
-		# The player has returned here after completing a quest. Silently mark
-		# it as completed.
+	if GameState.quest:
+		# The player has returned here after completing a quest. There is no
+		# loom to interact with: just have an elder congratulate the player.
+		await story_quest_elder.congratulate_player()
 		GameState.mark_quest_completed()
 		GameState.save()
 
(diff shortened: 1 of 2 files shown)


felixwalberg:
This seems good. To clarify, is there any version of the game where the template quest elder would still exist in Fray's End? I see that the template outro scene points to `frays_end.tscn` as the next scene rather than returning to Dev Archipelago, and I wasn't sure if this switches specifically in the StoryQuest kit.

wjt:
Are you sure it points specifically to Frays End? Or does it point to the "home scene" (option above the path)?

felixwalberg:
> Are you sure it points specifically to Frays End? Or does it point to the "home scene" (option above the path)?

Yes, you're right it is the home scene, which would be changed in the kit. So in the full game, we want the player to come back to Fray's End from the template quest to deposit the threads in the Eternal Loom, but in the kit it wouldn't exist. Got it!

felixwalberg (approved):

PR #2681, opened 2026-08-10
Cap music volume at half of SFX

wjt:
Previously we defaulted the music volume (linear) in-game to 0.5 out of
a total of 1.0. (In the linear_to_db() scale, 0.5 corresponds to
approximately -6.0 dB. A 6.0 dB reduction in volume sounds like a
halving of the volume.) Empirically, this is never too quiet. I have
never seen anyone turn it up, and I often see people turn it further
down.

We also did not reflect this reduced music volume in the default audio
bus layout, so music played much louder in-editor than in-game.

Update the bus layout to reduce the default volume for the Music bus.
Use the values from the default layout as the maximums for each bus in
the settings code. Update the volume sliders to read the maximum from
settings, and adjust the step so that there are always 10 notches. (I
considered doing the scaling in the settings code instead but this
worked out fine.) Clamp any values that are outside the accepted range
for a bus (e.g. because someone's settings file has the Music bus set to
0.7 which is now out of range).

You can hear the result of moving the music slider instantly, but not so
for the sfx slider. Trigger an audio stream each time the SFX slider
moves. Use the "pop" sounds from dialogue, with a miaow at very low
probability as an easter egg.


wjt:
Obviously this is not a priority but I've had a version of this patch lying around for ages and was inspired to finish it by watching yet another recording of a Threadbare playtest where someone had to fiddle with the music volume.

scenes/globals/settings/default_bus_layout.tres
```diff
@@ -5,7 +5,7 @@ bus/1/name = &"Music"
 bus/1/solo = false
 bus/1/mute = false
 bus/1/bypass_fx = false
-bus/1/volume_db = 0.0
+bus/1/volume_db = -6.0
```
wjt, line 8: BTW I set this value by editing the file by hand. If you try to set -6.0 dB in the editor, you get something that is within ±0.05 dB of that value, and I found these aesthetically displeasing.

wjt:
If you listen closely while moving the music slider around, you will hear clicks/pops. Instead we should tween the volume over a very small interval. But.. preexisting problem and I'm sure most people won't notice.

scenes/menus/options/components/sound_settings.tscn
```diff
@@ -65,3 +65,10 @@ shortcut = ExtResource("4_k547c")
 text = "Back"
 icon = ExtResource("5_arrow")
 flat = true
+
+[node name="TestSFXPlayer" type="AudioStreamPlayer" parent="." unique_id=1300034183]
+stream = ExtResource("5_k547c")
+max_polyphony = 10
+bus = &"SFX"
+
+[connection signal="value_changed" from="VBoxContainer/GridContainer/SFXSlider" to="TestSFXPlayer" method="play" flags=3 unbinds=1]
```
manuq, line 74: Nice! Actual feedback as the user changes the slider.

scenes/menus/options/components/test_sfx.tres
```diff
@@ -0,0 +1,22 @@
...
+
+[ext_resource type="AudioStream" uid="uid://b6jipqg6p5pah" path="res://scenes/ui_elements/dialogue/components/sounds/839822__tommylistens__bubble_0.wav" id="1_d3cjg"]
+[ext_resource type="AudioStream" uid="uid://ctx1gif17l0t8" path="res://scenes/ui_elements/dialogue/components/sounds/839823__tommylistens__bubble_1.wav" id="2_jnwge"]
+[ext_resource type="AudioStream" uid="uid://btjeqyv3j0ig1" path="res://scenes/ui_elements/dialogue/components/sounds/839824__tommylistens__bubble_2.wav" id="3_ffpye"]
+[ext_resource type="AudioStream" uid="uid://dy0hk40s7ve8t" path="res://scenes/ui_elements/dialogue/components/sounds/839825__tommylistens__bubble_3.wav" id="4_yynw7"]
+[ext_resource type="AudioStream" uid="uid://debf2ti8718n1" path="res://scenes/ui_elements/dialogue/components/sounds/839826__tommylistens__bubble_4.wav" id="5_vq0kl"]
+[ext_resource type="AudioStream" uid="uid://cp8c37rnfsumh" path="res://scenes/ui_elements/dialogue/components/sounds/839827__tommylistens__bubble_5.wav" id="6_mr5y4"]
+[ext_resource type="AudioStream" uid="uid://di2b2rgca1udv" path="res://scenes/game_elements/characters/npcs/cat/components/412017__skymary__cat-meow-short.wav" id="7_ncifa"]
```
manuq, line 9: I tried it and I wonder if having a single bubble would allow me to more reliably select the SFX level I want. I hear some of them more "muted" than others.

I love the sporadic cat meoww :)

manuq (approved):

scenes/menus/options/components/test_sfx.tres
wjt: Yes, you're probably right. I was prioritising having an excuse to hide a "miaow" in there.

PR #2685, opened 2026-08-10
Use consistent "pop" in volume settings

wjt:
Use consistent "pop" in volume settings

Manuel pointed out in review that it would be better to consistently use
the same "pop" for every adjustment to more clearly hear the difference,
and he's right: some pops are considerably quieter than others.

Remove all except the loudest pop. Remove the pitch randomization which
also made it harder to compare different volume levels. Finally remove
the miaow, which I thought was funny but I can't really justify when
removing it allows also removing the randomizer entirely.

Now that we don't sometimes have the (long) cat noise playing, we can
also reset the max polyphony on the player to 1: it is desirable to cut
off the older pop to play a newer one each time the slider moves.


manuq (approved):
More accurate and more boring... sorry!

PR #2687, opened 2026-08-10
Temple Room 2 Camera Limits Modified and Layout Added

rencarle:
Temple room should be done now. Fixed issues a past merge caused (removed an outdated music node, fixed player camera limits, and added player to bottom of the scene tree) and added the room layout.


franko276 (commented):
Everything looks good on my end!

Issue #2675, opened 2026-08-07
Rewoven transition: Apply to actual world


manuq:
1. On loom interaction, change the cutscene to use SceneSwitcher, actually switching from Fray's End back and forth, not displayed in an overlay.

2. On loom interaction, instead of showing a cutscene, reweave the actual world when conditions are met:
- After completing the Void quest, remove the Void barrier at the west (ideally without transition but camera panning) and the Void barrier at the Linenville path (switching back and forth that scene).
- After completing the Musician quest, transition to Song Sanctuary Path and see the corresponding bridge being fixed (check MusicianQuestUnlocker). 
- After completing the Guarded Stones quest, transition to Song Sanctuary Path and see the corresponding bridge being fixed (check GuardedStonesUnlocker). 
- After completing the "Oops! All InkDrinkers" quest, transition to Song Sanctuary Path and see the corresponding bridge being fixed (check OopsAllInkDrinkersUnlocker).

PR #2688, opened 2026-08-10
Rewoven cutscene: Actually switch to the scene

### LINKED ISSUE #2700: Player interaction is still enabled during reweaving cutscene
### Threadbare version

v0.2.2-25-g9c35feb4

### Affected scene

res://scenes/world_map/frays_end.tscn

### Description

If the spacebar is pressed during the cutscene animation, the Eternal Loom can be interacted with. This prompts a new sequence of dialogue to start on top of the reweaving scene. 
Screenshot 2026-08-12 at 3 33 45 PM

I found that if the dialogue is open when the cutscene ends, it stacks on top of the "Thank you StoryWeaver" dialogue that comes after it.
Screenshot 2026-08-12 at 3 26 56 PM

### Checklist

manuq:
On loom interaction, change the cutscene to use SceneSwitcher, actually switching from Fray's End back and forth, not displayed in an overlay.

This is part one of https://github.com/endlessm/threadbare/issues/2675

Helps https://github.com/endlessm/threadbare/issues/2675
Fix https://github.com/endlessm/threadbare/issues/2700


manuq:
Opening a draft for now with step 1 of https://github.com/endlessm/threadbare/issues/2675 which is the easy part. I'm still figuring the best way to build an array of cutscenes to show in actual levels from the QuestProgressUnlockers. Maybe there is no other way than centralizing the dependency information.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
```diff
@@ -77,16 +82,29 @@ func give_spirit_upgrade() -> void:
 
 
 func on_offering_succeeded() -> void:
-	load_cinematic.emit()
+	var cutscene_path: String = cutscene_paths.pick_random()
+	var load_error: Error = ResourceLoader.load_threaded_request(cutscene_path)
+
+	GameState.global.facts.rewoven_cutscenes = [cutscene_path]
```
manuq, line 88: Note: this generic fact is placeholder, while I figure out a way to generate an array of cutscenes to display using actual quest unlockers.

manuq:
With the transition effects already applied in the actual world https://github.com/endlessm/threadbare/pull/2699 (done, pending review), I still trying to figure out how to show them in sequence upon loom interaction.. Should the quest be marked completed / incompleted multiple times? Should a "fake" transition happen? I guess I'll need a centralized list of quest dependencies -> scenes + camera + quest unlocker within scene.

scenes/game_elements/props/eternal_loom/components/eternal_loom.gd
manuq: I'm marking this PR as ready for review to do the first part of the ticket (actually switch to the scene existing cutscenes, not new ones in the world). Should I leave this as an Array?

scenes/game_elements/fx/world_rewoven/components/world_rewoven_cutscene.gd
```diff
@@ -0,0 +1,17 @@
...
+
+
+func _on_animation_finished(_anim_name: StringName) -> void:
+	# Go back home. Usually Fray's End and next to the Eternal Loom:
+	var home_scene: String = ThreadbareProjectSettings.get_setting(
+		ThreadbareProjectSettings.HOME_SCENE
+	)
+	SceneSwitcher.change_to_file_with_transition(
+		home_scene, "", Transitions.Effect.FADE, Transitions.Effect.FADE
+	)
```
wjt, line 17: Another way to do this would be to put a `SceneLink` into each cutscene, pointing at HOME, and connect `finished` to its `switch`. I think this is fine though, particularly since we'll surely iterate on this to e.g. play multiple cutscenes.

wjt (approved):
Looks good, I think this is much simpler.

I pushed an extra patch to keep the background music playing, same music as Fray's End. I think what we actually want is to change to a different, more triumphant clip, e.g. `res://assets/first_party/music/Threadbare Loop_Main_Bridge_02.ogg` or `res://assets/first_party/music/Threadbare Loop_Main_Cresendo_01.ogg`, and then pick up where we left off in `res://assets/first_party/music/Threadbare_Bed.ogg`.

But... I think that will need some special attention with the future change to have the cutscenes be embedded in real scenes, so I think that can wait for now.

scenes/game_elements/fx/world_rewoven/components/world_rewoven_cutscene.gd
manuq: Oh I like that more! I'll open a follow up PR.

manuq:
> Looks good, I think this is much simpler.
> 
> I pushed an extra patch to keep the background music playing, same music as Fray's End. I think what we actually want is to change to a different, more triumphant clip, e.g. `res://assets/first_party/music/Threadbare Loop_Main_Bridge_02.ogg` or `res://assets/first_party/music/Threadbare Loop_Main_Cresendo_01.ogg`, and then pick up where we left off in `res://assets/first_party/music/Threadbare_Bed.ogg`.
> 
> But... I think that will need some special attention with the future change to have the cutscenes be embedded in real scenes, so I think that can wait for now.

Good point on letting the music continue! Thanks for the patch. Yeah we can revisit how a world being rewoven should sound like in the future.

PR #2689, opened 2026-08-11
Store InventoryItems as separate resource files

wjt:
Previously we defined InventoryItem resources inline in every scene that
uses them; and we hardcoded the world texture, HUD texture, and colour
for each in the script.

Turn these into properties of the resource. Create 3 on-disk resources
for the memory, imagination, and spirit threads. Add migration code to
collectible_item.gd which updates the property so that loading & saving
a scene using this script does the migration. Leave the migration code
in-place so that as out-of-tree quests are merged we can easily update
them.

Change the inventory storage to refer to those, rather than saving just
the item type name and reconstructing resources on load.

I think this will make it easier to add collectibles which are not
threads in future, and also to remember which threads have been
collected when loading or replaying a quest, though I haven't attempted
to do that here.

### DIFF
--- a/scenes/game_elements/props/collectible_item/components/collectible_item.gd
+++ b/scenes/game_elements/props/collectible_item/components/collectible_item.gd
@@ -17,6 +17,7 @@ extends SceneLink
 
 ## [InventoryItem] provided by this collectible when interacted with.
 @export var item: InventoryItem:
+	get = _get_item,
 	set = _set_item
 
 @export_category("Dialogue")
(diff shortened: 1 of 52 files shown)


tools/bulk_resave_scenes.gd
wjt, line 1: @manuq This is the script I mentioned on #2678

wjt:
Rebased and re-rewrote all the scenes.

I think this change makes sense. It does increase the chance of someone accidentally editing the text of one of the three threads rather than making a bespoke one. However we don't currently support collecting anything except these three items. And actually I think this is a useful step towards more generic collectibles.

wjt:
Personally I would recommend tagging a release before merging this and #2693!

PR #2695, opened 2026-08-11
add Beach parrot and pirate hat quest

franko276:
Added the parrot and pirate hat interaction. The player can grapple the parrot to drop the hat, pick it up through dialogue, and return it to the skeleton pirate.


rencarle:
I see that you changed scenes/ui_elements/cinematic/cinematic_placeholder.dialogue, which is part of the main game. Could you revert the change and make a separate dialogue file?

PR #2696, opened 2026-08-11
Skeleton scene complete

rencarle:
We can now add the skeleton and its dialogue for each scene after the beach. Look at temple room 1 to see how it is set up and let me know if you need help with setting up the NPC.


lynette023 (approved):
I think I see what's happening here. I'll work on stuff tomorrow and we can check in when you're back on Monday!

rencarle:
@lynette023 Sounds good! If you need help tomorrow, I may be available over Discord for quick messages. I'm not 100% sure on my availability since I'm busy tomorrow on certain times, but still message me if you need help.

PR #2697, opened 2026-08-11
Treasure room and ending

rencarle:
The dialogue and book asset are the only things missing. The treasure room has been laid out and the thread is set up. The retelling also needs to be finished.


lynette023 (approved):
Looks good to me!

Issue #2548, opened 2026-07-15
Tabs, bookmarks, or some other form of organization for the Storyquests in the book


saspiers:

wjt:
At least strongly related to #1275 if not a duplicate

alcole2:
What do we want to organize them based off of? I've seen language and completion status as suggestions so far.

manuq:
@alcole2 I think that ultimately we want both. List by language and by completion status. The StoryQuest resource doesn't have a language field yet. The current status is:

Complete:
- El Abrigo
- El Juguete Perdido
- Stella and the Missing Star

Work in progress:
- After The Tremor
- Evan and the Search for Champ
- Eldrune
- El ojo revelador
- Renya: Beyond Sorrow
- Sueños nocturnos
- Nora and the Impossible Homework

But this is really outdated. I will open a pull request to mark all of them as completed.

Please start by adding a language field to the StoryQuest resource with options Spanish and English, and updating their quest.tres files accordingly.

alcole2:
Okay sounds good.

alcole2:
@manuq I added the new language field and updated all the quest.tres files! I was envisioning something like when you first open the StoryBook you can choose between "English StoryQuests" and "Spanish StoryQuests" and then the corresponding list would come up for you. Did you have something different in mind?

manuq:
> [@manuq](https://github.com/manuq) I added the new language field and updated all the quest.tres files!

Nice can you send a pull request with that? We usually add `Helps #123` instead of `Fixes #123` as last line for those pull requests that don't fully complete the task, but they help with it.

> I was envisioning something like when you first open the StoryBook you can choose between "English StoryQuests" and "Spanish StoryQuests" and then the corresponding list would come up for you. Did you have something different in mind?

As discussed in the planning meeting today, I would avoid a regular user interface, and instead rely on adding bookmarks to the book. But maybe you can come up with a mockup image first, before starting the implementation?

PR #2702, opened 2026-08-13
Add Language enum to StoryQuest resources

alcole2:
There is a Language enum (ENGLISH, SPANISH) added to quest.gd with an exported variable to categorize each quest by language. All of the existing quest .tres files are updated to their appropriate language setting.

Helps #2548

### DIFF
--- a/scenes/menus/storybook/components/quest.gd
+++ b/scenes/menus/storybook/components/quest.gd
@@ -15,8 +15,17 @@ enum Status {
 	BROKEN = 2,
 }
 
+##The language of a quest.
+enum Language {
+	ENGLISH,
+	SPANISH,
+}
+
 ## [Quest] resources must have this filename to be found by the game.
 const FILENAME := "quest.tres"
+
+## The language of this quest.
+@export var language: Language = Language.ENGLISH
 
 ## The development status of this quest.
 @export var status: Status = Status.WORK_IN_PROGRESS
--- a/scenes/quests/story_quests/el_abrigo/quest.tres
+++ b/scenes/quests/story_quests/el_abrigo/quest.tres
@@ -4,6 +4,7 @@
 
 [resource]
 script = ExtResource("1_4uyxg")
+language = 1
 status = 1
 title = "El Abrigo"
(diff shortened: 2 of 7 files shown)


wjt:
I don't think it should be an enum. I think it should be a String holding a 2-letter ISO language code (es or en for the current quests). If we want it to be a drop-down in the inspector I think it should be possible to use `@export_enum var language: String` to do that.

wjt:
(sorry if I'm contradicting you @manuq !)

wjt:
Or, rather than a drop-down in the inspector: people set these up with the custom duplicator tool so we can add a drop-down to that widget, and have the property be a normal string.

I think this is a better idea because Godot also represents languages with ISO codes so "matches the current language" would otherwise mean converting between string and enum; and because if someone wants to write a storyquest in a language not in the enum they will have to patch the game. By using a string that by convention is drawn from the standard taxonomy of language codes, we don't have to change the types in our game for each new language. (User interface is of course another matter!)

alcole2:
@wjt So I should do the ISO language code property ("en", "es") on the Quest resource for now?

manuq:
> (sorry if I'm contradicting you @manuq !)

Did I said enum? Sorry I meant to say "add language field", not implying a format.

wjt:
Yes, please remove the enum and use a String property instead

wjt:
Superseded by #2703. For future reference, @alcole2, you could have pushed more patches to this branch to change the approach.

alcole2:
@wjt Yes that would've been ideal. I tried, but for some reason I couldn't get my changes onto this branch without it resetting everything in my script editor to the original commit.  `enum Language {
ENGLISH,
SPANISH,
}` kept coming back even though I kept deleting it

I did everything I usually do to push commits to the branch but it wouldn't let me this time. It ended up being easier to just do a new pull request. Do you know why that happened?

wjt:
I can't say what was happening from the description. If you like, next time you are in a similar situation, ping me and we can troubleshoot on a video call?

PR #2703, opened 2026-08-13
Quest: Add language field

### LINKED ISSUE #2548: Tabs, bookmarks, or some other form of organization for the Storyquests in the book

alcole2:
Add a field to the `Quest` resource, `language`, which holds the ISO 639 alpha-2 code for the language of the text in the quest.

Update all existing quests with the appropriate language.

Helps #2548

### DIFF
--- a/scenes/menus/storybook/components/quest.gd
+++ b/scenes/menus/storybook/components/quest.gd
@@ -22,7 +22,7 @@ const FILENAME := "quest.tres"
 @export var status: Status = Status.WORK_IN_PROGRESS
 
 ## The 2-letter ISO language code for the quest.
-@export var language: String = "en"
+@export_enum("en", "es") var language: String = "en"
 
 ## The quest's title. This should be short, like the title of a novel.
 @export var title: String
--- a/scenes/menus/storybook/components/quest.gd
+++ b/scenes/menus/storybook/components/quest.gd
@@ -21,7 +21,7 @@ const FILENAME := "quest.tres"
 ## The development status of this quest.
 @export var status: Status = Status.WORK_IN_PROGRESS
 
-## The 2-letter ISO language code for the quest.
+## The 2-letter ISO 639 language code for the quest.
 @export_enum("en", "es") var language: String = "en"
 
 ## The quest's title. This should be short, like the title of a novel.
--- a/scenes/menus/storybook/components/quest.gd
+++ b/scenes/menus/storybook/components/quest.gd
@@ -21,6 +21,9 @@ const FILENAME := "quest.tres"
 ## The development status of this quest.
 @export var status: Status = Status.WORK_IN_PROGRESS
 
+## The 2-letter ISO language code for the quest.
+@export var language: String = "en"
+
 ## The quest's title. This should be short, like the title of a novel.
 @export var title: String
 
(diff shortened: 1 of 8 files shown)


scenes/menus/storybook/components/quest.gd
```diff
@@ -21,6 +21,9 @@ const FILENAME := "quest.tres"
 ## The development status of this quest.
 @export var status: Status = Status.WORK_IN_PROGRESS
 
+## The 2-letter ISO language code for the quest.
+@export var language: String = "en"
```
manuq: This could be
```suggestion
@export_enum("en", "es") var language: String = "en"
```

For having the both worlds, a dropdown and stored as string:

Image

The advantage is that it makes it less prone to typos. The caveat is that it makes it harder to set a different language. Hmm @wjt and @alcole2 what do you think?

alcole2: I like that idea! I did that at first but I wasn't sure if you guys wanted to make it so any language could be added in the future. If just English and Spanish work for you guys I'll change it.

wjt: The caveat reintroduces the problem that supporting additional languages requires changing the `Quest` datatype. I think we could instead introduce a warning if the value is not one of the elements of `TranslationServer.get_all_languages()`?

alcole2: Should I just do Manuel's suggestion for now so I can continue with the bookmark idea? I could go back later and update it with the warning.

wjt: I had forgotten that `_get_configuration_warnings()` is a method on `Node` not `Resource` so can't be used to add validation for resource properties.

So I guess `@export_enum` is fine for now. Later on we can use `_validate_property()` to generate the full list of possible languages, with `en` and `es` sorted to the top. I'm happy to write that, I like writing property validation code 😁

manuq: I'm with @wjt about adding a warning after validating with the result of TranslationServer.

@alcole2 please do as @wjt said!

alcole2: Okay, just to confirm I'm understanding correctly:

I'll stick with ```@export_enum("en", "es") var language: String = "en"``` for now so we have the inspector dropdown for English and Spanish without risking typos.
I'll leave the advanced validation/dynamic list logic to @wjt for later.
I'll make sure the .tres files are updated with "en" / "es" and move forward with the bookmarks feature.

Thank you guys for helping me and clarifying things!

scenes/menus/storybook/components/quest.gd
```diff
@@ -21,6 +21,9 @@ const FILENAME := "quest.tres"
 ## The development status of this quest.
 @export var status: Status = Status.WORK_IN_PROGRESS
 
+## The 2-letter ISO language code for the quest.
```
wjt:
```suggestion
## The 2-letter ISO 639 language code for the quest.
```

Could also include a link to https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes - check https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_documentation_comments.html to see how to include hyperlinks in docstrings

wjt (approved):

PR #2705, opened 2026-08-13
Tile-o-matic

wjt:
This adds a prototype of a script that can be used to automatically
update water, foam, shadows (cliff and bridge), and cliff collisions
based on the other layers in the scene. It's halfway (or maybe 1/10 of
the way) to a custom level editor.

It supports multiple layers of bridges but only one layer of elevations
or (elevation or bridge) shadows.

The intended workflow is to paint floors, cliffs, and bridges, then
click Update to fix the others.

One thing I discovered while writing this is that we didn't have
collisionless alternatives for every cliff-edge tile, only for two of
them, and those alternatives were missing their occlusion (physics and
light) polygons and possibly some other properties. I wrote a script to
give every cliff-edge tile an alternative that does not have wall
collisions.

However this is still not quite correct. Bottom edges include include
the cliff-face below. That face needs to be walkable when the bridge is
coming from the bottom; but when coming from the side, the bottom should
block the player. So we need two alts for that. Similarly, if you have a
bridge reaching a top-right corner from its right edge, and land at a
lower height above its top edge, you shouldn't be able to walk off the
top. You can see the problem in this test scene. I did not fix this
family of issues.

This is also all painfully hard-coded to the specific tilesets in the
main game...


manuq:
This is a kind of magic! The only thing I can see missing is undo-redo. I was going to propose applying the elevation shadows to the elevation PNGs, because the only case I can think of not having shadows is if the elevations are floating (in the void?). But even with that we can't get rid of the shadows tileset (because of bridges). And anyways for this is much better!

https://github.com/user-attachments/assets/6729aa27-4b68-4df4-9e9c-885d30212dda

wjt:
> the only case I can think of not having shadows is if the elevations are floating (in the void?).

I have argued that elevations should always have shadows, even if they are atop foam. But if you look closely at the banner that Pixel Frog made, it's actually a little inconsistent.

image

In the rightmost box I've drawn on this image, the cliff casts a shadow over the foam as well as the ground. In the central one (and all others), it does not. In the central one, it casts a shadow onto the ground. I guess that he's picking and choosing depending on what looks best.

The other case where the shadow may look wrong is if you have two levels of elevation stacked directly on top of each other. It doesn't really make sense to have a shadow cast below the top layer & above the bottom layer in that case.

This prototype doesn't automate that but the human level designer can tweak it.

> But even with that we can't get rid of the shadows tileset (because of bridges).

Bridge shadows are actually part of the bridge tileset!

wjt:
Since I will not have any more time to work on this, I tried to get it into a mergeable state where it could be experimented with in future and then either improved, or deleted!

Issue #2578, opened 2026-07-21
Dev Archipelago: Add new themed islands


felixwalberg:
In addition to the basics, grapple, and repel themed islands, we should have ones for:
- stealth levels
- level design (area filler, weather/time)

PR #2706, opened 2026-08-13
Add enemy interaction island to Dev Archipelago

felixwalberg:
Add elder + dialogue and some decorations to the Enemy Interaction island on Dev Archipelago. I tried this originally with the NO_EDIT template guard sprite frames which looked like a more menacing pursuer of the sheep, but it seemed better to use a townie. 

I try to allude to the 6th quest in the narrative document with the dialogue, however there are no visible indications of the void in the archipelago. (Will there be in the new narrative? @saspiers ).

Move `gym_area_test.tscn` to the new `dev/enemy_interaction/` section of dev quest folders. 

Enemy Interaction island can be home to:
- gym area (and any other scenes that show off enemies in this way)
- stealth level demo (guard paths, whether you can outrun (or even outwalk) guards)
- void demo
- possibly demos of path walk and other behavior nodes

Makes progress on #2578


scenes/dev/components/sheep_chase.gd
```diff
@@ -0,0 +1,31 @@
...
+
+var sheep_sprite: AnimatedSprite2D
+
+@onready var sheep_path_follow: CharacterBody2D = %SheepPathFollow
+@onready var sheep_path_walk_behavior: PathWalkBehavior = %SheepPathWalkBehavior
+@onready var townie_path_walk_behavior: PathWalkBehavior = %TowniePathWalkBehavior
+
+
+func _ready() -> void:
+	sheep_sprite = sheep_path_follow.get_node("AnimatedSprite2D")
```
manuq: This is hacky. It would be better to give the sheep a "look at side" property like the townies have. Or for now, use a different approch, flip the entire sheep with:

```
sheep_path_follow.scale.x = -1 if sheep_path_follow.velocity.x < 0 else 1
```

manuq (approved):
One more sheep joke! I like it!

felixwalberg:
> One more sheep joke! I like it!

Hopefully this is not overkill! The sheep have just been easier to use than other animals/decorations

scenes/dev/components/sheep_chase.gd
felixwalberg: I might look into adding the "look at side" property. I think I had done something with inverting the scale in `champ_outro.tscn` and had to change it.

felixwalberg: `look_at_side` seems to also use this actually!

manuq: In the case of the townie, I did it that way because of the multiple AnimatedSprite2D nodes it contains. Although I'm not happy. Now that I think about it, each part (hair, head, body, legs) should have a CharacterSpriteBehavior node. That behavior takes care of the flip for you :)

manuq: > In the case of the townie, I did it that way because of the multiple AnimatedSprite2D nodes it contains. Although I'm not happy. Now that I think about it, each part (hair, head, body, legs) should have a CharacterSpriteBehavior node. That behavior takes care of the flip for you :)

Oh actually they have it. The "look at side" is used for the initialization, in a wrong way. Now I understand a bug I've seen in which townies walked looking behind.

felixwalberg: The CharacterSpriteBehavior works great and simplifies what has to be done in the dev archipelago scene! @manuq previously approved this, but I wanted to verify the changes are sufficient for merging now that I have changed `sheep.tscn` which has 8 other owners.

wjt (approved):
This is great. I found a pre-existing bug in the sheep while testing it.

PR #2707, opened 2026-08-13
tools: Add script to load+save all scenes that use some resource

wjt:
This is useful when making some change across all instances of a
script/scene:

1. Add migration code to the script that runs when the node is
   instantiated
2. Run this script
3. Remove the migration code
4. Commit the result

As commented here you can do this by running a built-in tool that does
this to *all* scenes but this can be much more disruptive.

You can also do this in the editor by using View Owners..., selecting
all, opening all the scenes, then saving each one. You can't use "save
all scenes" unless you have marked each scene as unsaved; and
`EditorInterface.mark_scene_as_unsaved()` only operates on the foreground
tab (haven't actually checked what happens if you bulk-open scenes and
call this during the loading of each one); so doing this by hand
involves hitting Ctrl+S, Ctrl+Shift+W, repeatedly.

### DIFF
--- a/tools/bulk_resave_scenes.gd
+++ b/tools/bulk_resave_scenes.gd
@@ -1,0 +1,45 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+class_name BulkResaveScenes
+extends EditorScript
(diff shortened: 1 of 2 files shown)


manuq (approved):
Yes!

PR #2712, opened 2026-08-14
Storybook: Adding language bookmarks

### LINKED ISSUE #2548: Tabs, bookmarks, or some other form of organization for the Storyquests in the book

alcole2:
Add TOC (table of contents), EN, and ES bookmarks to the storybook, to allow players to filter quests by language in addition to the full list.

The current bookmarks are placeholder buttons for now, they should be themed in the future.

It behaves like a real book, although the animation shows a single page swipe even when going to pages far away. For example: if there are 30 quests, of which 20 are EN and 10 are ES, the calculation should be:

- 2 spreads for ToC (30 < 16 * 2)
- 2 spreads for EN (20 < 16 * 2)
- 1 spread for ES (10 < 16)

Then the spread index is:

- _current_spread_index == 0 -> ToC 1st spread
- _current_spread_index == 1 -> ToC 2nd spread
- _current_spread_index == 2 -> EN 1st spread
- _current_spread_index == 3 -> EN 2nd spread
- _current_spread_index == 4 -> ES only spread
- _current_spread_index == 7 -> Quest 1
- _current_spread_index == 8 -> Quest 2
- ...
- _current_spread_index == 24 -> Quest 30

Having quests in a single language (all English or all Spanish) hides the EN and ES bookmarks and the per-language spreads.

Resolves #2548

### DIFF
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -237,7 +237,7 @@ func _switch_to_page(spread_index: int) -> void:
 	if spread_index == _current_spread_index:
 		return
 
-	var total_spreads: int = _bookmark_indexes[ContentTab.QUESTS] + quests.size()
+	var total_spreads: int = _bookmark_indexes[ContentTab.QUESTS] + quests.size() - 1
 	if spread_index < 0 or spread_index > total_spreads:
 		return
 
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -12,6 +12,13 @@ extends CanvasLayer
 ## [param quest] set to [code]null[/code].
 signal selected(quest: Quest, restart: bool)
 
+enum ContentTab { TOC, EN, ES }
+const TAB_LANGUAGES := {
+	ContentTab.TOC: "",
+	ContentTab.EN: "en",
+	ContentTab.ES: "es",
+}
+
 ## Quests to show in the storybook.
 @export var quests: Array[Quest]
 @export var quests_per_page: int = 8:
@@ -22,13 +29,12 @@ signal selected(quest: Quest, restart: bool)
 
 var quests_per_spread: int = 16
 
+var _current_tab: ContentTab = ContentTab.TOC
 var _current_spread_index: int = -1
 var _navigation_locked: bool = false
 var _current_list_page: int = 0
 
 var _filtered_quests: Array[Quest] = []
-
-var _current_tab_index: int = 0  # 0: TOC, 1: EN, 2: ES
 
 @onready var left_quest_list: VBoxContainer = %LeftQuestList
 @onready var right_quest_list: VBoxContainer = %RightQuestList
@@ -250,7 +256,7 @@ func _on_left_button_pressed() -> void:
 			return
 
 		# Flipping back from the first page returns to full table of contents
-		_switch_bookmark_tab(0, "")
+		_switch_bookmark_tab(ContentTab.TOC)
 		return
 
 	_switch_to_page(_current_spread_index - 1)
@@ -309,15 +315,15 @@ func reset_focus() -> void:
 
 
 func _on_toc_bookmark_pressed() -> void:
-	_switch_bookmark_tab(0, "")
+	_switch_bookmark_tab(ContentTab.TOC)
 
 
 func _on_en_bookmark_pressed() -> void:
-	_switch_bookmark_tab(1, "en")
+	_switch_bookmark_tab(ContentTab.EN)
 
 
 func _on_es_bookmark_pressed() -> void:
-	_switch_bookmark_tab(2, "es")
+	_switch_bookmark_tab(ContentTab.ES)
 
 
 func _show_table_of_contents() -> void:
@@ -326,9 +332,7 @@ func _show_table_of_contents() -> void:
 
 
 func _filter_quests_by_language(lang_code: String) -> void:
-	_filtered_quests = quests.filter(
-		func(quest: Quest) -> bool: return quest.language == lang_code
-	)
+	_filtered_quests = quests.filter(func(quest: Quest) -> bool: return quest.language == lang_code)
 	_reset_storybook_list_view()
 
 
@@ -339,19 +343,20 @@ func _reset_storybook_list_view() -> void:
 	_update_page_visibility()
 
 
-func _switch_bookmark_tab(target_tab_index: int, lang_code: String) -> void:
-	if _navigation_locked or target_tab_index == _current_tab_index:
+func _switch_bookmark_tab(target_tab: ContentTab) -> void:
+	if _navigation_locked or (target_tab == _current_tab and _current_spread_index == 0):
 		return
 
 	_navigation_locked = true
-	var old_tab_index: int = _current_tab_index
-	_current_tab_index = target_tab_index
+	var old_tab: ContentTab = _current_tab
+	_current_tab = target_tab
+	var lang_code: String = TAB_LANGUAGES.get(target_tab, "")
 
 	# Fade out UI
 	await _fade_out_ui()
 
 	# Flip forward if target > current, otherwise flip backward
-	if target_tab_index > old_tab_index:
+	if target_tab > old_tab:
 		animated_book.play("book_right")
 	else:
 		animated_book.play("book_left")
--- a/scenes/menus/storybook/components/storybook.gd
+++ b/scenes/menus/storybook/components/storybook.gd
@@ -78,9 +78,9 @@ func _ready() -> void:
 		else:
 			_quests_per_language[language].append(q)
 
-	toc_bookmark_button.pressed.connect(_on_toc_bookmark_pressed)
-	en_bookmark_button.pressed.connect(_on_en_bookmark_pressed)
-	es_bookmark_button.pressed.connect(_on_es_bookmark_pressed)
+	toc_bookmark_button.pressed.connect(_switch_bookmark_tab.bind(ContentTab.TOC))
+	en_bookmark_button.pressed.connect(_switch_bookmark_tab.bind(ContentTab.EN))
+	es_bookmark_button.pressed.connect(_switch_bookmark_tab.bind(ContentTab.ES))
 
 	_update_bookmark_visibility()
 	_populate_quest_lists()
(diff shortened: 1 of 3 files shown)


scenes/menus/storybook/components/storybook.gd
```diff
@@ -35,6 +39,11 @@ var _current_list_page: int = 0
 @onready var animated_book: AnimatedSprite2D = %AnimatedSprite2D
 @onready var ui_container: Control = %StoryBookContent
 
+#Bookmark button references
```
manuq:
```suggestion
# Bookmark button references
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -51,6 +60,18 @@ func _fade_in_ui() -> void:
 
 func _ready() -> void:
 	animated_book.animation_finished.connect(_on_animation_finished)
+
+	# Saving a copy of all quests
+	_all_quests = quests.duplicate()
```
manuq: This does:
```
_all_quests = quests.duplicate()
```

Then `show_table_of_contents()` does:
```
quests = _all_quests.duplicate()
```

While it works, it's a bit odd that the script changes the exported `quests` property. An exported property is supposed to be changed from the outside (from the user of this storybook).

Instead, add a `_flltered_quests` internal property. And change `_populate_quest_lists()` to use `_filtered_quests`. Consider it the variable containing the filtered view. In `show_table_of_contents()`  do:
```
_filtered_quests = quests
```

Note that there is no need for `duplicate()` at all. In the same way the `filter()` calls in the per-language handlers don't duplicate the quests. Arrays in Godot are always passed by reference.

manuq (changes requested):
This is a great start and already working! In addition to the quests variable handling request:

The page is empty for "es" in Dev Archipelago storybooks:

Image

Should a language bookmark/section appear if there are no quests for that language? The page would be empty so I don't think so.

Should language bookmarks/sections appear at all if all the quests have the same language? The content of that section would be the same as the table of contents, so I don't think so.

The bookmarks are now hardcoded to the existing StoryQuest languages. But what will happen when a StoryQuest in a different language is added? Maybe it's fine for now, as I don't expect contributions in languages that we can't review (note to self: we should document that somewhere).

Minor detail: the buttons fade in and out when the pages flip. They should stay in place, considering that they are not part of the pages content.

scenes/menus/storybook/components/storybook_test.gd
```diff
@@ -7,13 +7,23 @@ const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
...
+]
+
+var spanish_titles := [
+	"El Señor de las Agujas",
+	"El tejido oscurot",
+]
+
+var french_titles := [
+	"Le Seigneur des Aiguilles",
 ]
```
manuq, line 24: This works. But there is a better approach to not duplicate the actual titles in the arrays:

```
var english_titles := [
	"The Secret of Crochet",
	"The Needle and the Sorcerer",
	"Conan the Weaver",
	"Return to Fray's End",
	"The Little HushRoom",
]

var spanish_titles := [
	"El Señor de las Agujas",
	"El tejido oscurot",
]

var french_titles := [
	"Le Seigneur des Aiguilles",
]

var titles := english_titles + spanish_titles + french_titles
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -26,6 +26,12 @@ var _current_spread_index: int = -1
 var _navigation_locked: bool = false
 var _current_list_page: int = 0
 
+## A copy of all quests loaded on ready (before filtering)
+var _all_quests: Array[Quest] = []
```
manuq: The copy of `quests` is not needed anymore! You can remove this variable and replace all instances of it with `quests` and it should still work the same way.
```suggestion
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -51,6 +62,20 @@ func _fade_in_ui() -> void:
 
 func _ready() -> void:
 	animated_book.animation_finished.connect(_on_animation_finished)
+
+	# Saving a copy of all quests
+	_all_quests = quests
```
manuq:
```suggestion
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +301,92 @@ func _on_back_button_pressed() -> void:
...
+func _on_en_bookmark_pressed() -> void:
+	_switch_bookmark_tab(1, "en")
+
+
+func _on_es_bookmark_pressed() -> void:
+	_switch_bookmark_tab(2, "es")
+
+
+func _show_table_of_contents() -> void:
+	_filtered_quests = _all_quests
```
manuq:
```suggestion
	_filtered_quests = quests
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +301,92 @@ func _on_back_button_pressed() -> void:
...
+	_reset_storybook_list_view()
+
+
+func _filter_quests_by_language(lang_code: String) -> void:
+	_filtered_quests = _all_quests.filter(
```
manuq:
```suggestion
	_filtered_quests = quests.filter(
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +301,92 @@ func _on_back_button_pressed() -> void:
...
+		animated_book.play("book_right")
+	else:
+		animated_book.play("book_left")
+
+	# Wait for animation to finish before rebuilding list
+	await animated_book.animation_finished
+
+	# Filter quests array
+	if lang_code == "":
+		_filtered_quests = _all_quests
```
manuq:
```suggestion
		_filtered_quests = quests
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +301,92 @@ func _on_back_button_pressed() -> void:
...
+	else:
+		_filtered_quests = _all_quests.filter(
```
manuq:
```suggestion
		_filtered_quests = quests.filter(
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +301,92 @@ func _on_back_button_pressed() -> void:
...
+	_update_page_visibility()
+
+	# Fade back in
+	_fade_in_ui()
+	_navigation_locked = false
+
+
+func _update_bookmark_visibility() -> void:
+	var unique_languages := []
+	for q in _all_quests:
```
manuq:
```suggestion
	for q in quests:
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +301,92 @@ func _on_back_button_pressed() -> void:
...
+		if q.language != "" and q.language not in unique_languages:
+			unique_languages.append(q.language)
+
+	# Only show language bookmarks if there is more than 1 language
+	var show_language_tabs: bool = unique_languages.size() > 1
+
+	# Check if at least one quest exists for each language
+	var has_en: bool = _all_quests.any(func(q: Quest) -> bool: return q.language == "en")
+	var has_es: bool = _all_quests.any(func(q: Quest) -> bool: return q.language == "es")
```
manuq:
```suggestion
	var has_en: bool = quests.any(func(q: Quest) -> bool: return q.language == "en")
	var has_es: bool = quests.any(func(q: Quest) -> bool: return q.language == "es")
```

manuq (changes requested):
Almost there!

scenes/menus/storybook/components/storybook.gd
```diff
@@ -26,6 +26,10 @@ var _current_spread_index: int = -1
 var _navigation_locked: bool = false
 var _current_list_page: int = 0
 
+var _filtered_quests: Array[Quest] = []
+
+var _current_tab_index: int = 0  # 0: TOC, 1: EN, 2: ES
```
manuq: I understand that the reason you added this index is to flip the page left or right depending on the actual tab or section. And also to do nothing when the index is the same (for example, user clicks on "EN" button when the English tab is already being displayed.

Godot has enums for a better representation of this. So you could instead have:

```
enum ContentTabs { TOC, EN, ES }
var _current_tab := ContentTabs.TOC
```

scenes/menus/storybook/components/storybook.gd
```diff
@@ -271,3 +297,92 @@ func _on_back_button_pressed() -> void:
...
+func _on_toc_bookmark_pressed() -> void:
+	_switch_bookmark_tab(0, "")
+
+
+func _on_en_bookmark_pressed() -> void:
+	_switch_bookmark_tab(1, "en")
+
+
+func _on_es_bookmark_pressed() -> void:
+	_switch_bookmark_tab(2, "es")
```
manuq: The two arguments of `_switch_bookmark_tab` seem redundant. Try passing only one and then within that function obtain the other. For example, you could map the index (or enum, see above) to the String using a Dictionary.

manuq (changes requested):

scenes/menus/storybook/components/storybook_test.gd
```diff
@@ -7,13 +7,23 @@ const STORYBOOK_SCENE := preload("uid://bhm7fdjvppt8b")
 @export_range(0, 100, 1, "or_greater") var quests_amount: int = 30
 
 var titles := [
-	"Lord of the Needles",
+	"El Señor de las Agujas",
 	"The Secret of Crochet",
-	"The Dark Knit",
+	"El tejido oscurot",
```
manuq:
```suggestion
	"El tejido oscuro",
```

manuq:
@alcole2 something is not right.

- Open the book
- Switch to the next page

I would expect to see the EN content. But I see the first quest detail page.

- Click on ES
- Switch to the next page

I would expect to see the first quest detail page. But I see the first ES quest (After The Tremor).

https://github.com/user-attachments/assets/e33b647d-a191-4d63-9ceb-e5cea61fdd41

manuq:
I'm summarizing here what I talked with @alcole2 today. The current language tabs acts as filters for the book. They should act as bookmarks of a real book. So 1. `quests` should be used instead of `_filtered_quests` for the quests to display. And 2. is probably better to drop `_current_tab` in favor of using `_current_spread_index` directly for all the cases, and calculate which spread to show. And for the quests of the list of content, how to filter the quests to list inside the spread using the `quests_per_spread` amount (which is 16 by default).

Example: if there are 30 quests, of which 20 are EN and 10 are ES, the calculation should be:
- 2 spreads for ToC (30 < 16 * 2)
- 2 spreads for EN (20 < 16 * 2)
- 1 spread for ES (10 < 16)

Then the index would be:

- `_current_spread_index` == 0 -> ToC 1st spread
- `_current_spread_index` == 1 -> ToC 2nd spread
- `_current_spread_index` == 2 -> EN 1st spread
- `_current_spread_index` == 3 -> EN 2nd spread
- `_current_spread_index` == 4 -> ES only spread
- `_current_spread_index` == 7 -> Quest 1
- `_current_spread_index` == 8 -> Quest 2
- ...
- `_current_spread_index` == 24 -> Quest 30

manuq:
In the fixup commits I ended up refactoring a lot of the storybook script to make it work like https://github.com/endlessm/threadbare/pull/2712#issuecomment-5375506411

manuq:
I merged this because a learner plans to work on another Storybook task, to avoid conflicts.

PR #2714, opened 2026-08-14
Void quest: Add Void retreat transition

manuq:
In both the Void runner and closing transition.


manuq:
Stacked on top of https://github.com/endlessm/threadbare/pull/2699

scenes/quests/lore_quests/quest_002/1_void_runner/components/void_runner.gd
```diff
@@ -7,13 +7,16 @@ extends Node2D
...
+@onready var animation_player: AnimationPlayer = %AnimationPlayer
+
 
 func is_enemy_defeated() -> bool:
 	return not is_instance_valid(enemy)
 
 
 func repel_void() -> void:
 	whole_scene_camera.priority += 20
-	await void_layer.uncover_all(3.0)
```
manuq, line 17: If we go with this, should we remove `TileMapCover.uncover_all()` ??

manuq:
Rebased with main!

PR #2715, opened 2026-08-16
quest_004: Redesign horizontal door passage as vertical to use the do…

GianPierooo:
…or as intended


GianPierooo:
imagen_2026-08-16_003351862

Hi @manuq @wjt, we've just finished almost everything in our LoreQuest. We had a small request: you mentioned in the document that we should combine the humming ability with Repel, but we think it could work well as a new standalone ability within Threadbare, since we see a lot of potential in it, it has sound effects, particles, and it's dynamic enough to be reused on other maps. We're looking forward to your response :)

manuq (approved):

PR #2717, opened 2026-08-17
Reinstate LoreQuests in StoryQuest kit

### LINKED ISSUE #2710: Should we consider moving LoreQuests into the StoryQuest Kit?
There are several interesting and fun applications of game mechanics and components that sit within some of the LoreQuests that may be nice to offer learners as they work within the smaller StoryQuest Kit. 

Could we consider trying to move them? Should we? 

One concern would be size of Kit and another may be how best to organize these so they don't distract from the now simplified folders/assets learners have in their smaller kit.

Thoughts? @wjt

wjt:
Reinstate LoreQuests in StoryQuest kit

These quests contain useful examples of using the components of our
game, not all of which are demonstrated in Dev Archipelago yet.

Stop stripping the lore_quests folder from the StoryQuest kit. This
increases the size of the kit by ~180 files and ~5 MB - not a lot.

Adjust the end of the tutorial quest to point to the home scene rather
than Fray's End. (I deliberately didn't do this before because it uses a
particular spawn point in Fray's End, but this still works with the "go
home" reference. If that spawn point does not exist (as it does not in
Dev Archipelago), it will just be ignored.)

Keep Fray's End excluded from the kit and don't add an elder that offers
these quests to Dev Archipelago. They can be found in the filesystem.

Resolves https://github.com/endlessm/threadbare/issues/2710

### DIFF
--- a/scenes/quests/lore_quests/quest_000/6_grappling/tutorial_grappling.tscn
+++ b/scenes/quests/lore_quests/quest_000/6_grappling/tutorial_grappling.tscn
@@ -242,7 +242,6 @@ position = Vector2(-575, 315)
 item = SubResource("Resource_u8qfb")
 collected_dialogue = ExtResource("21_x2ecx")
 dialogue_title = &"thread_collected"
-next_scene = "uid://cufkthb25mpxy"
 spawn_point_path = NodePath("SpawnPointAfterIntro")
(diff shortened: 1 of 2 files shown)


manuq (approved):
Yes I read all the conversation and I agree.

PR #2726, opened 2026-08-17
Frank/beach scene

franko276:
Added new dialougue file and restore cinematic dialouge


rencarle:
Sorry for being very nitpicky, but could you remove the new line after "Replace this with dialogue for a cutscene." in scenes/ui_elements/cinematic/cinematic_placeholder.dialogue

PR #2727, opened 2026-08-17
Skeleton dialogue fixes and Temple room 1 layout modified

rencarle:
The skeleton now has all of its dialogue and they play at appropriate times. The temple room 1 layout was changed as discussed in Discord and the meeting. I also did a minor script fix for the lever sequence, so the debugger shouldn't be throwing red errors now that the lever objects are saved to the correct type of array.


scenes/quests/lore_quests/side_quest_001/4_temple_room_1/components/lever_sequence.gd
lynette023, line 1: Warnings on here to look at once merged

PR #2728, opened 2026-08-17
Added Sound and Book

franko276:
Add sound effect and change book sprite


rencarle:
Could you make a copy of the lever (script and its scene) with your changes and revert the original "scenes/game_elements/props/hookable_lever/components/hookable_lever.gd" to how it was? This is so we prevent messing with the game's original code.

PR #2730, opened 2026-08-18
cat, sheep: Fix meow/baa position

wjt:
Previously the AudioStreamPlayer2D was a child of a `Node` (not
`Node2D`). `Node`s do not have a position in the 2D scene, so the
resulting position of the AudioStreamPlayer2D was (0, 0). The
AudioStreamPlayer2D is set to attenuate over 2000px of distance so if
the cat/sheep was more than 2000px from the (0, 0) position of the
scene, its sound would be completely inaudible, and before that it would
be quieter. We never noticed this before because the sheep and cat
happen to be close to the origin of the scene! However it was noticeable
on the new enemy interaction island in Dev Archipelago.

Change the structure of each scene so that the AudioStreamPlayer2D node
is not a child of the PetInteraction (Node) but of the root (subclass of
Node2D).

### DIFF
--- a/scenes/game_elements/props/decoration/sheep/sheep.tscn
+++ b/scenes/game_elements/props/decoration/sheep/sheep.tscn
@@ -64,18 +64,18 @@ metadata/_custom_type_script = "uid://du8wfijr35r35"
 visible = false
 position = Vector2(5, -28)
 shape = SubResource("RectangleShape2D_tsjbs")
-debug_color = Color(0.6290859, 0.52632225, 0.15803415, 0.41960785)
+debug_color = Color(0.6, 0.545, 0, 0.42)
+
+[node name="AudioStreamPlayer2D" type="AudioStreamPlayer2D" parent="." unique_id=1930789457]
+stream = SubResource("AudioStreamRandomizer_3vki4")
+bus = &"SFX"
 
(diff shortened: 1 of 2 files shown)


wjt:
@felixwalberg this fixes the silent sheep you added in #2706

scenes/game_elements/props/decoration/sheep/sheep.tscn
```diff
@@ -64,19 +64,19 @@ metadata/_custom_type_script = "uid://du8wfijr35r35"
 visible = false
 position = Vector2(5, -28)
 shape = SubResource("RectangleShape2D_tsjbs")
-debug_color = Color(0.6290859, 0.52632225, 0.15803415, 0.41960785)
+debug_color = Color(0.6, 0.545, 0, 0.42)
```
manuq, line 67: Another missing one!

manuq (approved):
Oh good finding! The alternative would be to make PetInteraction a Node2D, but this is better, I think.

scenes/game_elements/props/decoration/sheep/sheep.tscn
felixwalberg: In my latest PR it had automatically changed to this newer color, but I didn't realize it was intentional, so I reverted it.

PR #2731, opened 2026-08-18
dev archipelago: Add interaction to the townie chasing a sheep

wjt:
The sound effect is my own “work”.


wjt:
@felixwalberg I'll leave it to you to decide whether this is a good addition 😀

felixwalberg (approved):
I think this is great!

PR #2751, opened 2026-08-19
git_describe: Simplify "v0.1.2" to "0.1.2" not "0.1.2."

### LINKED ISSUE #2711: application/config/version has trailing dot in releases
`application/config/version` can only contain dot-separated numbers for Windows reasons. So the `git describe` output `v0.2.2-31-g65a4ae129` becomes `0.2.2.31` with the fourth field representing the number of patches since the tag.

I noticed while jamming with @jgbourque that in the `v0.2.2` StoryQuest kit the version is set to `0.2.2.` (sic.) with trailing dot. I think this is wrong and it should either be `0.2.2.0` or `0.2.2`.

wjt:
On the title screen we show the full output of `git describe --tags`,
which for an arbitrary change on that is not a tag looks like
`v0.2.2-24-g14ed33fee` and for a tag looks like `v0.2.2`.

For Windows executables, the version can only be up to 4
period-separated integers. So we simplify `v0.2.2-24-g14ed33fee` to
`0.2.2.24`, i.e. the nano field is the number of patches since the tag
represented by the major.minor.micro fields, and set this as the project
version.

However this logic was incorrect for tags: `v0.2.2` was reduced to
`0.2.2.` with a trailing period. The regexp nests a named match group
inside an optional non-capturing match group: `(?:-(?<count>\\d+).*)?`.
If the outer group doesn't match, the `"count"` field is not in the
`names` field of the `RegExMatch` object, but there is still an empty
entry for it in the `strings` field.

Indexes into the `strings` field via the `names` field (a map from group
name to its index in `strings`). Explicitly check whether `count`
matched, and don't include the empty non-match as the nano field if not.

Resolves https://github.com/endlessm/threadbare/issues/2711

### DIFF
--- a/addons/threadbare_git_describe/version.gd
+++ b/addons/threadbare_git_describe/version.gd
@@ -31,7 +31,12 @@ static func simplify(full_version: String) -> String:
 	var result := regex.search(full_version)
 	if not result:
 		return "0.0.0"
-	return ".".join(result.strings.slice(1))
+
+	var base := result.strings[result.names["base"]]
+	if "count" in result.names:
+		var count := result.strings[result.names["count"]]
+		return base + "." + count
+	return base
 
 
 ## Gets the full `git describe`-style version


manuq (approved):
Looks good!

Side note: This it the kind of things that would welcome a unit test. But we don't have the setup for them. Maybe a simple EditorScript with assert() lines could be a start?

PR #2755, opened 2026-08-19
TalkBehavior: Document all exported properties

manuq:
And also do like the Cinematic node: add a setter to the title property and add a warning in the editor if such title doesn't exist within the dialogue.


manuq:
@jgbourque asked me about documentation for adding dialogue to a character. Does this help?

wjt (approved):

PR #2764, opened 2026-08-19
Updated Side Quest to have main's updates

rencarle:
Side quest now has all the updates (as of the time I had merged the changes into the branch on my copy). I also fixed the files that were throwing warnings, so those should be something to look at (although it is simply adding types, which shouldn't break anything).


lynette023 (approved):
Warnings all fixed, looks good to go!

PR #2769, opened 2026-08-20
Dev Archipelago: Add background music

wjt:
Once again, composed with strudel.cc, and dissected in Ardour.

This time I exported separate stems for the rhythm section and the
xylophone melody. There are 3 sections which connect like this via
AudioStreamInteractive's auto-advance:

    A --> B <--> C

There are two variations of B: one with the xylophone, and one without.
There are three variations of C: two different xylophone melodies, and
one without.

I would have liked to add even more mixing and matching, particularly
sometimes switching between the two C melodies midway through, but this
is OK for now.

I used Ogg Vorbis's -q0 quality level, ~64kbps, to keep the file sizes
down. My ears aren't what they once were but to my surprise it sounds
fine. Still, this adds 1.1 MB to the project...


manuq (approved):
I love it! Although I hear the sheeps repel sound effect on top of the music at about the same level from everywhere now.

> There are two variations of B: one with the xylophone, and one without.
> There are three variations of B: two different xylophone melodies, and
> one without.

I think you meant to say "three variations of C".

wjt:
Good catch!

Yes - I thought I filed a bug about the repel sound, I'll do that.

wjt:
Filed https://github.com/endlessm/threadbare/issues/2771

PR #2774, opened 2026-08-20
Revert "Dev Archipelago: Add background music (#2769)"

wjt:
This reverts commit 27d11af881a37672e41a95488909d20523be7317.

This musical style for Dev Archipelago has not been met with universal
approval!


manuq (approved):
OK. Is it possible to iterate on this music instead of reverting it?

wjt:
Could do but based on the reaction in our build review I think having this version in the game is a release blocker :) and I'm on holiday next week. Plus, another piece would almost certainly have a different collection of files, so would need the resource to have a different structure. I think it's simpler to just remove it for now.

Issue #2704, opened 2026-08-13
Prevent StoryWeaver from getting stuck behind townie in Dev Archipelago


felixwalberg:
Upon entering Dev Archipelago, the first island has a townie that greets the StoryWeaver with information about how the islands work. The townie is standing close to a tree node and rock node. If the StoryWeaver walks behind the townie, between them and the tree/rock, they get stuck! 

Update the position of at least one of the three nodes so that there is more space between the townie and the tree and rock, making it big enough for the StoryWeaver to walk both into _and_ out of the gap.

Alternatively, redecorate! Change the objects that are behind the townie or add some new ones that completely block the gap.

<img width="1466" height="756" alt="Image" src="https://github.com/user-attachments/assets/acf817fc-d61a-45f3-bbc9-3fabd90e8ae4" />

### XP Summary (total 150)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 50 |
| Game Design: Core Gameplay & Mechanics | Creativity & Design Thinking | 50 |
| Game Design: Core Gameplay & Mechanics | Technical Literacy | 50 |

manuq:
I created a minimum reproducible project and opened a bug upstream: https://github.com/godotengine/godot/issues/122708

wjt:
For clarity, to fix this specific instance of the issue: rearrange the objects in this scene so that StoryWeaver doesn't get stuck. The upstream bug @manuq reported is in search of a general solution to this "player gets stuck" family of bugs.

PR #2776, opened 2026-08-20
Player: Make sight and harm collision shapes outline only

manuq:
Too many collision shapes overlapping can be hard to understand for debugging. Changing them to alpha zero leaves them as outlines, as long as `debug/shapes/collision/draw_2d_outlines` is `true` (the default).

### DIFF
--- a/scenes/game_elements/characters/player/player.tscn
+++ b/scenes/game_elements/characters/player/player.tscn
@@ -418,7 +418,7 @@ character = NodePath("../..")
 [node name="CollisionShape2D" type="CollisionShape2D" parent="PlayerInteraction/CharacterSight" unique_id=255765935]
 position = Vector2(47, -30)
 shape = SubResource("RectangleShape2D_blfj0")
-debug_color = Color(0.600391, 0.54335, 0, 0.42)
+debug_color = Color(0.6, 0.54, 0, 0)
 
 [node name="PlayerRepel" parent="." unique_id=1324972992 instance=ExtResource("12_tnibl")]
 unique_name_in_owner = true
@@ -452,7 +452,7 @@ collision_mask = 256
 [node name="CollisionShape2D" type="CollisionShape2D" parent="PlayerHarm/HitBox" unique_id=1807630823]
 position = Vector2(0, -28.5)
 shape = SubResource("RectangleShape2D_je7p5")
-debug_color = Color(0.94902, 0.231373, 0, 0.419608)
+debug_color = Color(0.95, 0.2375, 0, 0)
 
 [node name="GotHitAnimation" type="AnimationPlayer" parent="PlayerHarm" unique_id=1067804411]
 unique_name_in_owner = true


manuq:
While debugging for https://github.com/endlessm/threadbare/issues/2704 I changed these collision shapes debugging color to look like this:

image

What do you think? Tries to be less busy visually. Up for discussion.

wjt (approved):
I think this looks way better.

I was surprised to see how you implemented this though: setting the alpha channel on the `debug_color` to `0`. I dug through the Godot code to see why this works:

```cpp
void RectangleShape2D::draw(const RID &p_to_rid, const Color &p_color) {
	RenderingServer::get_singleton()->canvas_item_add_rect(p_to_rid, Rect2(-size * 0.5, size), p_color);
	if (is_collision_outline_enabled()) {
		// Draw an outlined rectangle to make individual shapes easier to distinguish.
		Vector<Vector2> stroke_points;
		stroke_points.resize(5);
		stroke_points.write[0] = -size * 0.5;
		stroke_points.write[1] = Vector2(size.x, -size.y) * 0.5;
		stroke_points.write[2] = size * 0.5;
		stroke_points.write[3] = Vector2(-size.x, size.y) * 0.5;
		stroke_points.write[4] = -size * 0.5;

		Vector<Color> stroke_colors = { Color(p_color, 1.0) };

		RenderingServer::get_singleton()->canvas_item_add_polyline(p_to_rid, stroke_points, stroke_colors);
	}
}
```

Notice that the outline's stroke_color has its alpha channel set back to `1.0`.

```cpp
bool Shape2D::is_collision_outline_enabled() {
	return GLOBAL_GET_CACHED(bool, "debug/shapes/collision/draw_2d_outlines");
}
```

OK. So is this documented anywhere? I can't see it in any of these:

- https://docs.godotengine.org/en/stable/classes/class_collisionshape2d.html#class-collisionshape2d-property-debug-color
- https://docs.godotengine.org/en/stable/classes/class_shape2d.html#class-shape2d-method-draw
- https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-property-debug-shapes-collision-draw-2d-outlines

I don't think it's a problem to rely on this. I was just surprised that it works!

manuq:
> OK. So is this documented anywhere? I can't see it in any of these:
> 
> * https://docs.godotengine.org/en/stable/classes/class_collisionshape2d.html#class-collisionshape2d-property-debug-color
> * https://docs.godotengine.org/en/stable/classes/class_shape2d.html#class-shape2d-method-draw
> * https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-property-debug-shapes-collision-draw-2d-outlines
> 
> I don't think it's a problem to rely on this. I was just surprised that it works!

I was also surprised to find it, and happy as well because I really wanted this feature. I agree that this is not documented anywhere. I will open a pull request for godot-docs. Your pointer to the Godot source will be very helpful.

PR #2777, opened 2026-08-20
InputWalkBehavior: Check if the player got stuck

manuq:
And if so, defeat the player.

### InputWalkBehavior: Check if the character got stuck

That is, if after calling move_and_slide() it still collides with anything.
And emit a signal when this state changes.

### Player: Defeat after being stuck for 2 seconds

Add a 2 seconds timer and use it to call defeat on timeout. Start or stop the
timer depending on the input walk signal "is stuck".

Add a shaking animation to the player sprite when it becomes stuck.

And if so, set velocity to zero and go back to the previous position otherwise.

### Dev Grapple: Add example of pin positioned in water

Helps https://github.com/endlessm/threadbare/issues/2704

### DIFF
--- a/scenes/game_logic/walk_behaviors/input_walk_behavior.gd
+++ b/scenes/game_logic/walk_behaviors/input_walk_behavior.gd
@@ -21,6 +21,9 @@ var input_vector: Vector2
 var is_running: bool:
 	set = _set_is_running
 
+var _is_stuck: bool
+var _last_safe_position: Vector2
+
 
 func _set_is_running(new_is_running: bool) -> void:
 	if is_running == new_is_running:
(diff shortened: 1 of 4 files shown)


manuq:
Doesn't work :/


https://github.com/user-attachments/assets/e47914ec-d546-4115-bc83-405b292ee6ef


The "is stuck" check seems reliable. It doesn't give a false positive running in front of elevation or water, for example. But the character can still get stuck (although less than before).

Someone from my local Godot community recommended me to report this bug upstream. I can do that.

- https://github.com/godotengine/godot/issues/122708

manuq:
I have repurposed this. Now the player will become defeated after 2 seconds of getting stuck. Which is not ideal but is slightly better. A level designer could put a hookable pin on water. I added an example of this to Dev Archipelago:

https://github.com/user-attachments/assets/67bb7801-9a72-46a7-98a9-150a77dc2e63

scenes/game_logic/walk_behaviors/input_walk_behavior.gd
```diff
@@ -53,6 +67,10 @@ func _physics_process(delta: float) -> void:
 	character.velocity = character.velocity.move_toward(input_vector, step * delta)
 	character.move_and_slide()
 
+	# Check if the player got stuck (is colliding after the move_and_slide).
+	var collision := character.move_and_collide(Vector2.ZERO, true)
+	is_stuck = collision != null
```
wjt, line 72: `move_and_slide()` affects various state variables of the `CharacterBody2D` so I was curious whether `move_and_collide()` does, too. Now I understand the importance of the second argument, `test_only = true`!

wjt (approved):

wjt:
Great, this is better than being blocked forever and having to manually do the same thing.

I think the timer makes sense because you could imagine situations where it's temporary, e.g. being squidged into a corner by an NPC which then moves the other way.

PR #2782, opened 2026-08-24
Remove Y-sort from root nodes

### LINKED ISSUE #2781: Fix other lore scenes with Y Sort Enabled in root node
### Threadbare version

_No response_

### Affected scene

_No response_

### Description

Like in https://github.com/endlessm/threadbare/issues/2780 I've found more levels with this issue:

- scenes/world_map/linenville_path.tscn
- scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/stealth_sequence_combo.tscn

The problem with this that the TileMapLayer nodes are sorted too. So if we ever extend these scene upwards, above the X axis into a negative Y position, the characters and props would be drawn below the TileMapLayers.

### Checklist

manuq:
LinenVillePath: Remove Y-Sort from root node

StealthSequenceCombo: Remove Y-sort from root node

Because the player and other elements can be hidden behind the TileMapLayer
nodes if the level is extended upwards, reaching Y negative values.

The level already has Y-sort element grouped in the OnTheGround node. But the
SequencePuzzle (which has the musical rocks) is outside it. So, move
SequencePuzzle inside OnTheGround.

Fix https://github.com/endlessm/threadbare/issues/2781

### DIFF
--- a/scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/stealth_sequence_combo.tscn
+++ b/scenes/quests/lore_quests/song_sanctuary_2/1_stealth_sequence_combo/stealth_sequence_combo.tscn
@@ -51,7 +51,6 @@ type = 2
 metadata/_custom_type_script = "uid://bgmwplmj3bfls"
 
 [node name="StealthSequenceCombo" type="Node2D" unique_id=1761996320]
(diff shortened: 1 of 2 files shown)


wjt:
Oops, these are both scenes I made! Good catch.

PR #2787, opened 2026-08-25
tools: Add script to check for orphan assets

manuq:
I've been using this in StoryQuests submissions.

## StoryQuests: Remove unused assets

Using the output of tools/check_orphan_assets.gd and confirming that they are
also not referenced from scripts by filename or UID.

Except for Champ, because these assets could be actual placeholders.


manuq:
If we merge this, we should update https://github.com/endlessm/threadbare/wiki/Reviewing-a-StoryQuest-submission#checking-for-unneeded-files

tools/check_orphan_assets.gd
```diff
@@ -0,0 +1,108 @@
...
+## [code]load()[/code] calls are not recorded as dependencies. A path assembled
+## at run time cannot be detected either way.
+
+## Where to look for assets. Everything under it is checked.
+const ROOT_FOLDER := "res://scenes/quests/story_quests/stella"
+
+## Extensions considered assets, lowercase and without the dot.
+const ASSET_EXTENSIONS: PackedStringArray = [
+	"png", "jpg", "jpeg", "webp", "svg", "ogg", "wav", "mp3", "ttf", "otf", "ogv"
+]
```
wjt, line 26: I was trying to see whether this hardcoded list could be avoided. For example, `.dialogue` files are potentially-orphaned "assets" of a sort...

I tried using `ResourceLoader.list_directory()` rather than the `DirAccess` methods. By definition this only includes files that Godot recognises as resources (which includes `.gd` files but not, for example, Markdown files). It also includes files where you have an `.import` file but the corresponding file has been deleted (!!) but only because the imported texture (for example) is in the `.godot` cache.

Not sure it's an improvement.

tools/check_orphan_assets.gd
```diff
@@ -0,0 +1,108 @@
...
+	var pending := PackedStringArray([folder])
+
+	while not pending.is_empty():
+		var current := pending[-1]
+		pending.remove_at(pending.size() - 1)
+
+		for directory: String in DirAccess.get_directories_at(current):
+			# Godot's own cache holds copies of everything and would count as
+			# references to assets that are otherwise unused.
+			if directory != ".godot":
```
wjt, line 102: I think if you use the non-static methods of `DirAccess` - i.e. open the dir with `DirAccess.open()` then iterate `d.get_directories()` - then you could use the `include_hidden` property (i.e. set it to `false`) to avoid having to filter this by hand.

wjt (approved):
If this is useful, let's ship it!

I can think of some changes I might make - noted below - but this is the kind of script that we can incrementally improve as we use it.

tools/check_orphan_assets.gd
manuq: Yes this list is actually the file extensions in `.gitattributes`, excluding files not recognized by Godot (currently just `.aseprite`). Let me check with `ResourceLoader.list_directory()`!.

tools/check_orphan_assets.gd
manuq: Thanks for the hints! I will try them.

Issue #1331, opened 2025-10-13
Replace bushes with crops in stealth challenge


wjt:
The two patches of crops in  `res://scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn` are bushes from Tiny Swords. We replaced the bush-crops in Fray's End https://github.com/endlessm/threadbare/issues/677 and we should do the same here.

<img width="774" height="477" alt="Image" src="https://github.com/user-attachments/assets/2df7b3c5-a8a4-4bc4-9433-8ff8dff5a73d" />

<img width="428" height="318" alt="Image" src="https://github.com/user-attachments/assets/e0031c2a-1d04-43a5-a69f-f6549aae9da5" />

In res://scenes/game_elements/props/decoration/crops/ there are 3 prop scenes: `barley.tscn`, `carrot.tscn`, and `tomato.tscn`.

For each patch of bush-crops in `res://scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn`, pick one of the 3 prop scenes and replace the bushes with that crop. Perhaps the field in the middle of the level could be barley, while the patch at the end could be carrots?

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Problem Solving | 50 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 50 |
| Game Design: Game Feel & UX | Technical Literacy | 75 |

Leonardo-UTP:
i plan to work on this

Leonardo-UTP:
work in progress....

Leonardo-UTP:
This document contains the recorded progress on the issue so far....

https://docs.google.com/document/d/1IUveQhEFz1i5-BexRg82cyb-hNCJro4Z5lHYY_w3Jn4/edit?usp=sharing

Leonardo-UTP:
I first identified the bushes belonging to the two crop areas involved in this issue. During the initial review, I noticed that these bushes were placed directly under the Level node together with other scene elements, making it difficult to distinguish which bushes belonged to each area.

<img width="567" height="411" alt="Image" src="https://github.com/user-attachments/assets/e858aeb8-fac6-40e9-8242-809c2c001198" />

<img width="567" height="422" alt="Image" src="https://github.com/user-attachments/assets/9a3651e0-5f32-465e-b3f2-276645de3e56" />

Leonardo-UTP:
[...]

PR #2788, opened 2026-08-25
Issue 1331 replace bushes - Replace field bushes with crops

Leonardo-UTP:
Replaces the bushes in the middle and end fields with barley and carrot crops, preserving their original positions.

The crops were organized under their corresponding Node2D containers with Y Sort enabled to preserve the correct visual layering with the player.

The old bush nodes were removed after review, and the level was tested again to confirm that the crops and player layer correctly.

Issue: #1331


manuq (approved):
Good work!

Leonardo-UTP:
Thank you! I'll look for another issue to work on. C:

PR #2801, opened 2026-08-28
Add a visual clue to indicate the projectiles are about to dissapear

### LINKED ISSUE #2442: Projectiles: add clue to indicate that they are about to disappear
When projectiles disappear in games there is usually a clue. Visually: the asset could start blinking. We should add the same.

### XP Summary (total 500)
| Subdomain | Skill | XP |
| - | - | - |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |

Salonso928:
Affected scene: res://scenes/quests/template_quests/NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_projectile.tscn
Affected script: res://scenes/game_elements/props/projectile/components/projectile.gd

When a projectile has spent the 80% of its duration, a dissapear animation will start. While this animation is playing the proyectile can be interacted with the repel comand and all Area2D nodes if posible, but this will no longer reset its duration.
An AnimationPlayer was added to the projectile scene and imported using the @onready annotation in the projectile.gd

image

https://github.com/user-attachments/assets/e03976d4-2c42-40cf-b195-21256a5555cb

Fixes #2442

### DIFF
--- a/scenes/game_elements/props/projectile/components/projectile.gd
+++ b/scenes/game_elements/props/projectile/components/projectile.gd
@@ -68,7 +68,11 @@ var _trail_particles: GPUParticles2D
 @onready var visible_things: Node2D = %VisibleThings
 @onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
 @onready var trail_fx_marker: Marker2D = %TrailFXMarker
+@onready var animation_player: AnimationPlayer = %AnimationPlayer
 
+var time_passed:float = 0.0
+var time_to_dissapear:float = 0.0
+var is_dissapearing:bool = false
 ## How long the projectile lives in the scene.
 ## [br][br]
 ## This timer is restarted each time the projectile collides with anything.
@@ -111,6 +115,7 @@ func _ready() -> void:
 	duration_timer.start()
 	var impulse: Vector2 = direction * speed
 	apply_impulse(impulse)
+	time_to_dissapear = duration*0.8
 
 
 func _process(_delta: float) -> void:
@@ -121,6 +126,11 @@ func _process(_delta: float) -> void:
 		)
 		var force: Vector2 = direction_to_target * speed
 		constant_force = force
+	
+	time_passed+=_delta	
+	if time_passed>= time_to_dissapear && not is_dissapearing:
+		is_dissapearing=true
+		animation_player.play("dissapear")
 
 
 ## Add a small effect scene to the current scene in the current position.
@@ -136,7 +146,12 @@ func add_small_fx() -> void:
 
 func _on_body_entered(body: Node2D) -> void:
 	add_small_fx()
-	duration_timer.start()
+	
+	# Logic for the dissapear animation
+	# When the proyectile is dissapearing, the duration_timer will not reset
+	if not is_dissapearing:
+		duration_timer.start()
+		time_passed=0.0
 
 	# Logic for Fragile Barrel
 	# We must check for the specific subclass first because it inherits from FillingBarrel
@@ -157,7 +172,9 @@ func _on_body_entered(body: Node2D) -> void:
 ## enters the repel area.
 func got_repelled(repel_direction: Vector2) -> void:
 	add_small_fx()
-	duration_timer.start()
+	if not is_dissapearing:
+		duration_timer.start()
+		time_passed=0.0
 	var hit_vector: Vector2 = repel_direction * hit_speed
 	hit_sound.play()
 	animated_sprite_2d.speed_scale = 2
--- a/scenes/game_elements/props/projectile/components/projectile.gd
+++ b/scenes/game_elements/props/projectile/components/projectile.gd
@@ -68,11 +68,14 @@ var _trail_particles: GPUParticles2D
 @onready var visible_things: Node2D = %VisibleThings
 @onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
 @onready var trail_fx_marker: Marker2D = %TrailFXMarker
-@onready var animation_player: AnimationPlayer = %AnimationPlayer
 
-var time_passed:float = 0.0
-var time_to_dissapear:float = 0.0
-var is_dissapearing:bool = false
+#Attributes that control the blinking effect
+var time_passed: float = 0.0
+var time_to_dissapear: float = 0.0
+var is_dissapearing: bool = false
+var blink_effect: bool = true
+var blink_timer: Timer
+
 ## How long the projectile lives in the scene.
 ## [br][br]
 ## This timer is restarted each time the projectile collides with anything.
@@ -115,8 +118,19 @@ func _ready() -> void:
 	duration_timer.start()
 	var impulse: Vector2 = direction * speed
 	apply_impulse(impulse)
-	time_to_dissapear = duration*0.8
-
+	
+	##the effect will start when the projectile has spent 70% of its lifetime
+	time_to_dissapear = duration * 0.7
+	
+	blink_timer = Timer.new()
+	#each 0.25 seconds the blink effect will show
+	blink_timer.wait_time = 0.25
+	blink_timer.one_shot = false
+	blink_timer.timeout.connect(_on_blink_timer_timeout)
+	
+	add_child(blink_timer)
+	
+	
 
 func _process(_delta: float) -> void:
 	visible_things.rotation = linear_velocity.angle()
@@ -127,10 +141,11 @@ func _process(_delta: float) -> void:
 		var force: Vector2 = direction_to_target * speed
 		constant_force = force
 	
-	time_passed+=_delta	
-	if time_passed>= time_to_dissapear && not is_dissapearing:
-		is_dissapearing=true
-		animation_player.play("dissapear")
+	time_passed += _delta	
+	if time_passed >= time_to_dissapear && not is_dissapearing:
+		is_dissapearing = true
+		blink_timer.start()
+		
 
 
 ## Add a small effect scene to the current scene in the current position.
@@ -151,7 +166,7 @@ func _on_body_entered(body: Node2D) -> void:
 	# When the proyectile is dissapearing, the duration_timer will not reset
 	if not is_dissapearing:
 		duration_timer.start()
-		time_passed=0.0
+		time_passed = 0.0
 
 	# Logic for Fragile Barrel
 	# We must check for the specific subclass first because it inherits from FillingBarrel
@@ -172,9 +187,12 @@ func _on_body_entered(body: Node2D) -> void:
 ## enters the repel area.
 func got_repelled(repel_direction: Vector2) -> void:
 	add_small_fx()
+	
+	# When the proyectile is dissapearing, the duration_timer will not reset
 	if not is_dissapearing:
 		duration_timer.start()
-		time_passed=0.0
+		time_passed = 0.0
+		
 	var hit_vector: Vector2 = repel_direction * hit_speed
 	hit_sound.play()
 	animated_sprite_2d.speed_scale = 2
@@ -203,3 +221,11 @@ func _on_duration_timer_timeout() -> void:
 func remove() -> void:
 	await get_tree().create_timer(randf_range(0., 3.)).timeout
 	explode()
+
+func _on_blink_timer_timeout() -> void:
+	if blink_effect:
+		animated_sprite_2d.modulate.a = 0.2
+		blink_effect = false
+	else:
+		animated_sprite_2d.modulate.a = 0.8
+		blink_effect = true	
--- a/scenes/game_elements/props/projectile/components/projectile.gd
+++ b/scenes/game_elements/props/projectile/components/projectile.gd
@@ -65,16 +65,16 @@ extends RigidBody2D
 
 var _trail_particles: GPUParticles2D
 
-@onready var visible_things: Node2D = %VisibleThings
-@onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
-@onready var trail_fx_marker: Marker2D = %TrailFXMarker
-
 #Attributes that control the blinking effect
 var time_passed: float = 0.0
 var time_to_dissapear: float = 0.0
 var is_dissapearing: bool = false
 var blink_effect: bool = true
 var blink_timer: Timer
+
+@onready var visible_things: Node2D = %VisibleThings
+@onready var animated_sprite_2d: AnimatedSprite2D = %AnimatedSprite2D
+@onready var trail_fx_marker: Marker2D = %TrailFXMarker
 
 ## How long the projectile lives in the scene.
 ## [br][br]
@@ -228,4 +228,4 @@ func _on_blink_timer_timeout() -> void:
 		blink_effect = false
 	else:
 		animated_sprite_2d.modulate.a = 0.8
-		blink_effect = true	
+		blink_effect = true
--- a/scenes/quests/template_quests/NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_projectile.tscn
+++ b/scenes/quests/template_quests/NO_EDIT/2_NO_EDIT_combat/NO_EDIT_combat_components/NO_EDIT_projectile.tscn
@@ -10,6 +10,43 @@ bounce = 1.0
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_o6xl0"]
 size = Vector2(44, 44)
 
+[sub_resource type="Animation" id="Animation_w1dj1"]
+length = 0.001
+tracks/0/type = "value"
+tracks/0/imported = false
+tracks/0/enabled = true
+tracks/0/path = NodePath("VisibleThings/AnimatedSprite2D:modulate")
+tracks/0/interp = 1
+tracks/0/loop_wrap = true
+tracks/0/keys = {
+"times": PackedFloat32Array(0),
+"transitions": PackedFloat32Array(1),
+"update": 0,
+"values": [Color(1, 1, 1, 1)]
+}
+
+[sub_resource type="Animation" id="Animation_lrx48"]
+resource_name = "dissapear"
+loop_mode = 1
+tracks/0/type = "value"
+tracks/0/imported = false
+tracks/0/enabled = true
+tracks/0/path = NodePath("VisibleThings/AnimatedSprite2D:modulate")
+tracks/0/interp = 1
+tracks/0/loop_wrap = true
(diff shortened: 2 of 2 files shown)


scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -121,6 +126,11 @@ func _process(_delta: float) -> void:
 		)
 		var force: Vector2 = direction_to_target * speed
 		constant_force = force
+	
+	time_passed+=_delta	
+	if time_passed>= time_to_dissapear && not is_dissapearing:
+		is_dissapearing=true
+		animation_player.play("dissapear")
```
manuq: This crashes the game if you play `res://scenes/quests/lore_quests/quest_000/4_ink_combat/tutorial_ink_combat.tscn` with:

```
E 0:00:08:038   projectile.gd:71 @ @implicit_ready(): Node not found: "%AnimationPlayer" (relative to "/root/TutorialInkCombat/InkBlobProjectile").
E 0:00:12:055   Projectile._process: Cannot call method 'play' on a null value.
```

manuq (changes requested):
The work you did looks good for the `NO_EDIT_combat.tscn`, which uses the `NO_EDIT_projectile.tscn`. But not with other projectiles, like `ink_blob_projectile.tscn` or any other projectile in StoryQuests.

I have more to comment about the visual clue (which is a great start, by the way), but let's fix this first!

Salonso928:
Hi @manuq I made some changes. I removed the AnimationPlayer node, now the blink effect is only controlled by the projectile.gd script, therefore all the scenes that use this script should display the effect.
Here is a video playing ink_combat_round_4.tscn


https://github.com/user-attachments/assets/0b42bb28-1a9e-4ecc-9b18-65809be0b02e

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -111,7 +118,19 @@ func _ready() -> void:
 	duration_timer.start()
 	var impulse: Vector2 = direction * speed
 	apply_impulse(impulse)
-
+	
+	##the effect will start when the projectile has spent 70% of its lifetime
+	time_to_dissapear = duration * 0.7
+	
+	blink_timer = Timer.new()
```
manuq, line 125: Instead of using a Timer node to do the blinking, it would be better to use [a Tween](https://docs.godotengine.org/en/stable/classes/class_tween.html#class-tween).

But this leads to my main feedback, whether or not to animate the blink effect programmatically. I'll comment about it separately.

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -121,6 +140,12 @@ func _process(_delta: float) -> void:
 		)
 		var force: Vector2 = direction_to_target * speed
 		constant_force = force
+	
+	time_passed += _delta	
+	if time_passed >= time_to_dissapear && not is_dissapearing:
```
manuq, line 145: The `DurationTimer` node is already tracking the time, you could check its `time_left` property. Or even better, create a separate Timer for this, and call it StartBlinkingTimer.

The `_ready()` method already does:
```
	duration_timer.wait_time = duration
	duration_timer.start()
```

So that's the place to setup the new timer:

```
	duration_timer.wait_time = duration
	duration_timer.start()
	start_blinking_timer.wait_time = duration - blink_times * blink_duration
```

manuq (changes requested):
This looks good, although I think that the projectile should blink a predictable amount of times, let's say 3 blinks by default before dissapearing. Sorry that it wasn't part of the specification!

I understand that you switched from AnimationPlayer to make the blink programmatically because, as you saw, your first attempt changed only one of the 18 projectile scenes currently in the game.

There are pros and cons of doing it programmatically: the main benefit is that changing the script changes all 18 at the same time, great! But it leaves all the projectiles with the same look and feel. Someone wanting a different blink effect will need to duplicate the script. And artists wanting to improve the effect will be more comfortable editing it in an AnimationPlayer. Also we may want to add a sound effect in addition to the blink as audio clue.

Is fine if you want to keep doing it programmatically. I may add a followup ticket to move to AnimationPlayer. If so, check my suggestion to switch to Tween.

Alternatively, if you want to go back to AnimationPlayer you can make it optional, and add AnimationPlayer nodes to the `NO_EDIT_projectile.tscn` and the `ink_blob_projectile.tscn`, leaving the others (all StoryQuests) as before (without blink effect). For making it optional, you should not set a unique name (as I suggested before):

```
@onready var animation_player: AnimationPlayer = %AnimationPlayer
```

Because that will fail with `Node not found: "%AnimationPlayer"` in the scenes that don't have it. Instead you can export a new property:

```
@export var blink_animation_player: AnimationPlayer
```

Under `@export_group("FXs")`. And assign it in the 2 scenes mentioned above.

manuq:
The [lint and format checks](https://github.com/endlessm/threadbare/actions/runs/33285013532/job/99569245395?pr=2801) are failing too for boring reasons (new lines and spacing). Please consider installing pre-commit [as explained here](https://github.com/endlessm/threadbare/wiki/Contributing#coding-style).

Salonso928:
Hi! sorry for being late I was very busy. I followed the AnimationPlayer Node route, so I deleted the animation effect via programming. Now the projectile.gd have the following attributes, dissapearing point and blink times.
The dissapearing point is the percentage when the projectile will start its animation, and the blink times is the number of times the projectiles will blink before dissapearing. The effect is still the same as seen in the video.
image
image



I also added the animationPlayer to the ink__projectile, here is the video: 

https://github.com/user-attachments/assets/46af5e4a-3bad-47c4-9bcf-6a7a10472f74

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -122,6 +138,12 @@ func _process(_delta: float) -> void:
 		var force: Vector2 = direction_to_target * speed
 		constant_force = force
 
+	if blink_animation_player:
+		if duration_timer.time_left <= time_to_dissapear && not is_dissapearing:
+			is_dissapearing = true
+			var animation_duration = blink_animation_player.get_animation("dissapear").length
+			blink_animation_player.play("dissapear", -1, animation_duration / blink_effect_time)
```
manuq: This is inside the `_process` function so it's being checked every frame. It would be better to add a timer, similar to the existing `duration_timer`.

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -136,7 +158,11 @@ func add_small_fx() -> void:
 
 func _on_body_entered(body: Node2D) -> void:
 	add_small_fx()
-	duration_timer.start()
+
+	# Logic for the dissapear animation
+	# When the proyectile is dissapearing, the duration_timer will not reset
+	if not is_dissapearing:
+		duration_timer.start()
```
manuq: This changes the behavior. Now projectiles that have a `blink_animation_player` will behave differently and don't do what the documentation comment of `duration_timer` says:

```
## This timer is restarted each time the projectile collides with anything.
## So the life of the projectile is extended in each collision.
@onready var duration_timer: Timer = %DurationTimer
```

We could discuss this behavior separately, but that's what it is right now. If you add a timer to trigger the animation, as suggested above, you should reset it here too.

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -157,7 +183,10 @@ func _on_body_entered(body: Node2D) -> void:
 ## enters the repel area.
 func got_repelled(repel_direction: Vector2) -> void:
 	add_small_fx()
-	duration_timer.start()
+
+	# When the proyectile is dissapearing, the duration_timer will not reset
+	if not is_dissapearing:
+		duration_timer.start()
```
manuq: Same here, changes the existing behavior.

scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
manuq, line 1: I understand that you changed this level by accident, while testing. Please revert it.

manuq (changes requested):
Excellent, this works very well now. Thank you so much for taking my advice on using AnimationPlayer. See my new feedback.

Salonso928:
Hi! I made some changes. I followed your feedback, now the blink effect will reset if the projectiles collides or gets repelled (as seen in the following video) and also I use a timer to control the effect. The effect iis and works the same as before but now the code is better.


https://github.com/user-attachments/assets/bd6e6f9a-d19c-4d4a-8f66-579f77dd14fe

this also works for the blob projectile because de gd script is the same

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -63,6 +63,19 @@ extends RigidBody2D
 ## root node. When the projectile gets hit, the [member GPUParticles2D.amount_ratio] is set to 1.
 @export var trail_fx_scene: PackedScene
 
+@export var blink_animation_player: AnimationPlayer
+
+@export_group("Dissapear Clue")
+
+## The percentage of the projectile's lifespan before it starts its disappearing animation.
+@export_range(0.0, 1.0, 0.1) var dissapearing_point: float = 0.7
```
manuq: I played 2 levels that has this default, at 70% the projectile starts blinking:

- res://scenes/quests/lore_quests/quest_000/4_ink_combat/tutorial_ink_combat.tscn
- res://scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn

And because the enemies have different projectile duration, 5 seconds in first one and 10 seconds in the other, the clue is much quicker in the first one. If the projectile duration is too short, the dissapearing clue is too short too, giving the player no time to react. Would you consider changing it so that the clue duration is be the same regardless of the projectile duration? If so you can replace the `dissapearing_point` with a `blink_duration` export variable, in seconds. Basically replace this calculation:

```
var time_to_dissapear = duration * dissapearing_point
```

With:

```
var time_to_dissapear = min(0, duration - blink_duration)
```

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -136,8 +160,16 @@ func add_small_fx() -> void:
...
 	add_small_fx()
+
 	duration_timer.start()
 
+	#If the projectile collides with another object or is repelled the animation will reset
+	if blink_animation_player:
+		if is_dissapearing:
+			reset_animation_and_timer()
+		else:
+			blink_timer.start()
```
manuq: This is repeated below. Can you add a new function `_reset_timers()` and factor out this?

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -112,6 +125,17 @@ func _ready() -> void:
 	var impulse: Vector2 = direction * speed
 	apply_impulse(impulse)
 
+	if blink_animation_player:
+		var time_to_dissapear = duration * dissapearing_point
```
manuq: If you check this line in the editor you will see that it has a warning about the static type.



Image

Godot can infer it though, it just needs:

```suggestion
		var time_to_dissapear := duration * dissapearing_point
```

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -186,3 +226,15 @@ func _on_duration_timer_timeout() -> void:
 func remove() -> void:
 	await get_tree().create_timer(randf_range(0., 3.)).timeout
 	explode()
+
+
+func _on_blink_timer_timeout() -> void:
+	is_dissapearing = true
+	var animation_duration = blink_animation_player.get_animation("dissapear").length
```
manuq: This is the second one having a warning in the editor:

```suggestion
	var animation_duration := blink_animation_player.get_animation("dissapear").length
```

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -157,7 +189,15 @@ func _on_body_entered(body: Node2D) -> void:
 ## enters the repel area.
 func got_repelled(repel_direction: Vector2) -> void:
 	add_small_fx()
+
 	duration_timer.start()
+
+	#If the projectile collides with another object or is repelled the animation and timer will reset
+	if blink_animation_player:
+		if is_dissapearing:
+			reset_animation_and_timer()
```
manuq: Nice, you have factored out repeated code into `reset_animation_and_timer()`. There is more repeated code at the beginning of `got_repelled()` and `_on_body_entered()`. So what about doing more refactoring? 

```
func _on_projectile_hit() -> void:
	add_small_fx()

	# Reset timers and blink animation:

	duration_timer.start()

	if blink_animation_player:
		if is_dissapearing:
			blink_animation_player.play("RESET")
			blink_timer.start()
			is_dissapearing = false
		else:
			blink_timer.start()
```

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -63,6 +63,19 @@ extends RigidBody2D
...
+
+@export_group("Dissapear Clue")
+
+## The percentage of the projectile's lifespan before it starts its disappearing animation.
+@export_range(0.0, 1.0, 0.1) var dissapearing_point: float = 0.7
+
+## The number of times the projectile blinks before disappearing.
+@export_range(1, 10, 1) var blink_times: int = 3
+
+var is_dissapearing: bool = false
```
manuq: Now that there is a timer, `is_dissapearing` is not needed. You can check `blink_timer.is_stopped()` to know if the timer is not running (in other words the projectile is dissapearing). Or `blink_animation_player.is_playing()`.

manuq (changes requested):
Great work! Thank you for following my recommendation and using a timer, it's much better now. This is almost ready, check my new comments below.

Salonso928:
Hi, I followed all your advices, refactor some of the code, delete the is_dissapearing variable and now the time the animation will spend is fixed, the animation effect is still the same and now the minimun it will be shown is 3 seconds

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -63,6 +63,18 @@ extends RigidBody2D
 ## root node. When the projectile gets hit, the [member GPUParticles2D.amount_ratio] is set to 1.
 @export var trail_fx_scene: PackedScene
 
+@export var blink_animation_player: AnimationPlayer
```
manuq, line 66: Please add a documentation comment explaining that this AnimationPlayer needs to have an animation called "dissapear". By the way, I think that the animation can be called "blink". And even better, we can add a warning in the editor when the AnimationPlayer that was set in the Inspector doesn't have one. But I can open a separate ticket for it.

Also please move the `@export_group("Dissapear Clue")` line above this one, so this export belongs to that group. Would you agree with that?

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -63,6 +63,18 @@ extends RigidBody2D
...
+
+@export_group("Dissapear Clue")
+
+## The time in seconds to show the disappear animation.
+@export_range(0.0, 10.0, 0.5) var blink_duration: float = 3.0
```
manuq, line 71: "blink duration" reads to me as the duration of one blink. Sorry if I caused this trouble in previous reviews, but this should be named `dissapear_duration` to match the documentation comment.

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -63,6 +63,18 @@ extends RigidBody2D
...
+
+## The number of times the projectile blinks before disappearing.
+@export_range(1, 10, 1) var blink_times: int = 3
+
+var blink_effect_time: float
```
manuq, line 76: This variable is not needed, it can be calculated when needeed with duration, dissapear_duration and blink_times.

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -71,7 +83,7 @@ var _trail_particles: GPUParticles2D
 
 ## How long the projectile lives in the scene.
 ## [br][br]
-## This timer is restarted each time the projectile collides with anything.
+## This timer is restarted each time the projectile collies with anything.
```
manuq, line 86:
```suggestion
## This timer is restarted each time the projectile collides with anything.
```

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -112,6 +124,19 @@ func _ready() -> void:
 	var impulse: Vector2 = direction * speed
 	apply_impulse(impulse)
 
+	if blink_animation_player:
+		#If the blink duration is greater than the duration
+		#The minimun time the animation will spend is 3 seconds.
+		var time_to_dissapear: float = max(3.0, duration - blink_duration)
```
manuq, line 130: Where does this 3 seconds come from? I think that if the dissapear duration is greater than the duration, it should start blinking immediately. You had this right before.

scenes/game_elements/props/projectile/ink_blob_projectile.tscn
```diff
@@ -28,6 +66,8 @@ script = ExtResource("1_gkhfg")
 small_fx_scene = ExtResource("3_inmjt")
 big_fx_scene = ExtResource("4_41322")
 trail_fx_scene = ExtResource("5_3lot6")
+blink_animation_player = NodePath("AnimationPlayer")
+blink_times = 5
```
manuq, line 70: As requested before, let's reset this to the default.

scenes/quests/lore_quests/quest_001/2_ink_combat/ink_combat_round_3.tscn
manuq: This is still the case.

scenes/game_elements/props/projectile/components/projectile.gd
```diff
@@ -186,3 +209,23 @@ func _on_duration_timer_timeout() -> void:
...
+	# Reset timers and blink animation:
+	duration_timer.start()
+
+	if blink_animation_player:
+		#If the projectile collides with another object or is repelled the animation and timer will reset
+		if blink_timer.is_stopped():
+			blink_animation_player.play("RESET")
+			blink_timer.start()
+		else:
+			blink_timer.start()
```
manuq, line 231: If and else have the same content, is better to do:
```suggestion
		blink_timer.start()
```

manuq (changes requested):
Thank you so much! Since this was almost done, I commented but also did the changes myself. I have appended commits and will merge it in:  https://github.com/endlessm/threadbare/pull/2948

Issue #2248, opened 2026-05-26
tower in stealth level has no physics and does not occlude guard sight


wjt:
While testing #2232 I discovered that the tower at the end of `res://scenes/quests/lore_quests/quest_001/3_stealth_level/stealth_level.tscn` has no physics and does not occlude sight: you can walk through it, and you can't hide from the guard behind it.

<img width="602" height="579" alt="image" src="https://github.com/user-attachments/assets/423f4073-e828-4188-961a-7735fc223378" />

https://github.com/user-attachments/assets/d4d0e58a-5d2c-4849-931c-9e39cf931548

I think that's wrong.

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 150 |
| Game Design: Core Gameplay & Mechanics | Creativity & Design Thinking | 150 |
| Game Design: Core Gameplay & Mechanics | Technical Literacy | 150 |

KarmDK:
I plan to work on this

KarmDK:
https://docs.google.com/document/d/1tPtwo7bROx0Tf-sUhmKlXeNcN4VJpl7GxP4-20ZM1S0/edit?tab=t.0

KarmDK:
Hi! I added collision to the tower, now it can hide, but the light doesn't get blocked by the tower. The character can no longer pass through it.

<img width="358" height="165" alt="Image" src="https://github.com/user-attachments/assets/9279afbc-940a-499e-a6b0-fe85169da8e1" />

https://github.com/user-attachments/assets/b60deceb-c5c2-40cc-8faa-a97139f0f9b4

KarmDK:
The solution I saw would be to add a LightOccluder2D node.

https://github.com/user-attachments/assets/5bb5f136-3cff-477a-a8ca-842d425895f8

KarmDK:
When using it in the game, I see that it's possible to make the character undetectable when the enemy is below, but detectable from above, simply by adjusting the collision. It's also possible to make it undetectable in both cases. What do you think? How should I configure it?

<img width="933" height="670" alt="Image" src="https://github.com/user-attachments/assets/0c04d2df-b3de-4615-812f-c93caf89c26c" />

manuq:
@KarmDK excellent, yes as the task title suggests, this tower needs both a physic collider and a light occluder. Please open a pull request.

PR #2805, opened 2026-08-28
Fix missing collision and occluder on tower prop

KarmDK:
This pull request adds collision physics and visual occlusion to the tower in the stealth mission level. Previously, the tower only existed as a visual sprite, allowing the player to pass through it and the guards to detect them through the structure.

The tower scene has been updated with a StaticBody2D containing collision shapes, preventing player overlap and ensuring the guard's line of sight is properly blocked. A LightOccluder2D has also been added to block the guard's light through visual projection.

Resolves
https://github.com/endlessm/threadbare/issues/2248


manuq (approved):
Tested and works as a one-off. If we ever need this tower in another level we can factor it out as its own scene.

Issue #1102, opened 2025-08-20
Write dialogue for Template Elder


wjt:
The current LoreQuest Elder and StoryQuest Elder have quite similar dialogue to one another. The main difference is that the [StoryQuest Elder says](https://github.com/endlessm/threadbare/blob/67f352962b48b213c0a60c1ba3a8b68bd6815991/scenes/game_elements/characters/npcs/elder/components/story_quest_starter.dialogue#L12):

> Become the hero, and see where the path leads.

which is meant to indicate that the hero of a StoryQuest is normally not StoryWeaver, and that these stories are not part of the main plotline.

Currently the StoryQuest Elder offers the player the Template StoryQuest, but we would like to move this to a third elder. Write a new variation of this dialogue that indicates that the quests being offered are templates, intended to be explored & adapted by contributors. (Reach out to @saspiers for guidance if needed!)

The new dialogue should be added to the project in the same directory & format as the existing dialogue files. You can edit it in the Dialogue tab of the editor. (Reach out to us if you need pointers.)

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Narrative & Storytelling | Problem Solving | 50 |
| Game Design: Narrative & Storytelling | Creativity & Design Thinking | 50 |
| Game Design: Narrative & Storytelling | Technical Literacy | 75 |

edwardvilcamonje-max:
I plan to work on this

edwardvilcamonje-max:
https://docs.google.com/document/d/1Jf6FztanLAz6nBjNqU06De4-CLkml9i-w7kC_OsZuVo/edit?usp=sharing
My Planning Sheet

edwardvilcamonje-max:
Is this dialogue okay, or should I change anything?

SPDX-FileCopyrightText: The Threadbare Authors
SPDX-License-Identifier: MPL-2.0
~ start
TemplateQuest Elder: [[Hi|Hello|Greetings]]! We've been waiting for you. Our world is full of stories!
TemplateQuest Elder: We have some StoryQuest templates for you to explore. These stories are yours to adapt and make your own!
do show_storybook()
if chosen_quest == null:
% TemplateQuest Elder: Take a look at these templates. Perhaps one will inspire a story of your own!
% TemplateQuest Elder: These stories are waiting to be explored, adapted, and brought to life!
% TemplateQuest Elder: We hope you'll find something that sparks your imagination!
else:
[...]

PR #2806, opened 2026-08-29
Write dialogue for Template Elder  #1102

edwardvilcamonje-max:


manuq:
Based on your comment https://github.com/endlessm/threadbare/issues/1102#issuecomment-5459733883 and the content of this file (mentions StoryQuest Elder), it looks like we should close this one in favor of https://github.com/endlessm/threadbare/pull/2807/changes.

PR #2808, opened 2026-08-30
Stealth tutorial: Prevent guard from detecting player on spawn point

### LINKED ISSUE #2529: Tutorial stealth: First spawn point is not safe from first guard
### Threadbare version

v0.2.1-16-gdf61c5443

### Affected scene

res://scenes/quests/lore_quests/quest_000/2_stealth/tutorial_stealth.tscn

### Description

Checkpoints (or rather the SpawnPoint within) should (in my opinion) be placed in positions where the player cannot be caught. It is unfair to the player if they respawn at a position where they will quickly be detected again.

The first spawn point in the tutorial stealth level fails this test: the lower guard can see you as they turn around at the end of their path.

https://github.com/user-attachments/assets/73a90968-0264-4b8a-9776-576000321dfc

Either the position of the checkpoint or the guard's path should be changed. If the checkpoint is moved, the patch of sand should be moved/expanded as needed to still approximate the area that triggers the checkpoint. (c.f. [this discussion about checkpoints being invisible until triggered](https://github.com/endlessm/threadbare/discussions/2528) - the patch of sand is a hint to the player that there is something special about that spot.)

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Core Gameplay & Mechanics | Problem Solving | 50 |

IDynamixI985:
In the tutorial stealth level, the lower guard was previously able to detect the player immediately upon respawning at the first checkpoint/spawn point when turning around at the end of their patrol path.

To fix this and provide better pacing:
- Adjusted the initial path and level geometry near the spawn point to ensure the player remains safe upon respawning.
- Updated the path layout slightly so the movement toward the next section feels more active and engaging rather than just walking in a straight line.
- Adjusted the sand path visual cues accordingly to align with the new safe spawn area.
- Replaced the dim area's 'CollisionShape2D' with a 'CollisionPolygon2D' set to solid mode to properly adapt the cave darkness to the new layout.

Resolves https://github.com/endlessm/threadbare/issues/2529


IDynamixI985:
Hi, I've changed the path a little so it's not straight. Is this okay, or should I leave it as it was at the beginning—but keep the change to the spawn point so the guard doesn't catch the player?

image

manuq:
First off, I confirm that the spawn point is safe now!

> Hi, I've changed the path a little so it's not straight. Is this okay, or should I leave it as it was at the beginning—but keep the change to the spawn point so the guard doesn't catch the player?

I think the non-straight path works, although wasn't part of the issue. But if you do it, please adjust the shadows and foam tiles. I'm marking in red the extra shadows and foam that should be removed, and in yellow the tiles missing foam:

This is how it looked before, for comparing them:

Screenshot From 2026-08-31 13-28-05
Screenshot From 2026-08-31 13-29-13

manuq (changes requested):
Almost there!

IDynamixI985:
Hi, I've already made the changes. There's nothing else to change, right?

manuq (changes requested):
One more small thing: the interior transition to dim the lights has changed. Before it was exactly at the cave openings: 

Image

Now it is:

Image

Can you adjust it to be as before?

IDynamixI985:
I can do that, but the transition would happen when the player is in this spot. I could move the cave's exit one block further, and it would be exactly right.
image

manuq:
> I can do that, but the transition would happen when the player is in this spot. I could move the cave's exit one block further, and it would be exactly right. image

I see. How can it be fixed? Suggestions:
- Use a polygon shape instead of a rectangle.
- Or change the elevation tiles at the cave exit.
- Or redo your work, now considering this, and focusing on doing as few changes as possible to prevent guard from detecting player on spawn point.

wjt:
I like the new curved path, nice solution. Personally I would take options 1 or 2 from @manuq's suggestions :)

IDynamixI985:
I've changed the CollisionShape2D type to CollisionPolygon2D. Is that okay? Is there something else I need to do?
image

manuq (approved):
Works great! Thanks for adjusting the dim area! Please update the pull request description with this new change, so we can merge it.

IDynamixI985:
Hi, I've changed the description

manuq:
@IDynamixI985 thanks!

PR #2826, opened 2026-09-03
StoryQuest kit: Preconfigure Backstitch server

### LINKED ISSUE #2825: Preconfigure Backstitch server in SQ kit
The latest Backstitch release does not offer Ink & Switch's alpha server as an option by default: it must be manually configured by each user. @jgbourque asks us to preseed the StoryQuest kit project with a `backstitch.cfg` file that points to `alpha.backstitch.dev` so that learners don't have to configure it by hand. Later, when we self-host the server, we can change the kit.

wjt:
The latest Backstitch release does not offer
https://alpha.backstitch.dev/ by default, to encourage people not to use
this server for important data.

We want to reduce the number of manual/error-prone steps a learner must
follow to create/join a Backstitch project for their StoryQuest. Add a
backstitch.cfg file to the StoryQuest kit bundle.

As and when we have an Endless-provided server, we will switch this
preseeded file to point there.

Resolves https://github.com/endlessm/threadbare/issues/2825

### DIFF
--- a/tools/sqckck.sh
+++ b/tools/sqckck.sh
@@ -31,6 +31,14 @@ rm .gitignore.template
 git add backstitch-launcher-*
 echo "::endgroup::"
 
+echo "::group::Preconfiguring Backstitch server"
+cat >backstitch.cfg <<EOF
+[backstitch]
+available_servers = "https://alpha.backstitch.dev/"
+server_url = "https://alpha.backstitch.dev/"
+EOF
+echo "::endgroup::"
+
 echo "::group::Committing pruned project"
 git switch --detach
 git add --update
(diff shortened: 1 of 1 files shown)


wjt:
Converting to draft to test the output of the "Make StoryQuest Kit" workflow

tools/sqckck.sh
```diff
@@ -41,7 +49,7 @@ echo "tag_name=$TAG_NAME" >> "$GITHUB_OUTPUT"
 echo "::endgroup::"
 
 echo "::group::Creating zip file"
-git archive --format=zip --prefix=threadbare-storyquest/ --output="threadbare-storyquest-kit.zip" @
+git archive --add-file=backstitch.cfg --format=zip --prefix=threadbare-storyquest/ --output="threadbare-storyquest-kit.zip" @
```
wjt: Doesn't work because the --prefix is not applied to the --add-file!

wjt: Turns out it is sensitive to the order of arguments.

wjt:
@jgbourque Please test https://github.com/endlessm/threadbare/actions/runs/33758342664?pr=2826

jgbourque:
This worked for me! I had the alpha server pre selected as the choice in the plugin.

PR #2828, opened 2026-09-03
Upgrade Dialogue Manager to v4.0.3

### LINKED ISSUE #2820: Update Godot Dialogue Manager addon to v4
Like in https://github.com/endlessm/threadbare/issues/1825

We currently have v3.10.5 but the latest at the time of writing is v4.0.3.

Generally the steps to update this addon are as follows:

1. Check whether we have any local changes to the addon (`git log addons/dialogue_manager/`). If so, stop and investigate whether these are changes we want to keep. Only continue if there are no important changes.
2. In the Dialogue tab in the editor, click the "v3.9.1 available" button, click "Download update"
3. Commit the result
4. Play the game a little, going through some scenes with interesting dialogue (e.g. the musician uses branching dialogue and references other nodes in the scene) to make sure nothing is broken.
5. (If things are broken, fix them and commit the result; or give up and document what broke on this issue.)
6. Open a PR!

Any further changes to use new features should be made in separate PRs.

<details>
<summary>XP Summary (total 375)</summary>

| Subdomain | Skill | XP |
| - | - | - |
</details>

manuq:
This is a major upgrade with breaking changes so it can't be done through the
editor interface going to the Dialogue tab of the addon. Instead, download the
ZIP file of latest release, v4.0.3.

This commit only replaces the addons/dialogue_manager/ folder with the contents
of the release ZIP file, and adds the file upon restarting Godot.

### Update dialogue balloon

None of this is needed, this is only to keep the differences with the example
balloon in Dialogue Manager as small as possible.

- Export the dialogue resource.

- Reset next and skip actions to their defaults in the script, and change them
  in the .tscn file instead.

- Rename NextButton to ProgressButton, to make it more similar to the Progress
  (a non-clickable polygon) in the example balloon.

- Add an empty AudioStreamPlayer to the scene, just to satisfy the script which
  shows how to use it for voice over.

- Include _on_dialogue_label_spoke within the signals region.

- Move player_name to the locals Dictionary, which is precisely for that.

### Project: Update dialogue manager settings

The setting to translate (or not) the template files automatically has been
renamed.

### Cinematic, TalkBehavior: Adapt to Dialogue Manager changes

The titles have been renamed to "cues".

Resolve https://github.com/endlessm/threadbare/issues/2820


manuq:
Ready for review. I did a playthrough and everything looks fine now.

I looked at `DialogueActionable2D` and it certainly is very similar to our `TalkBehavior` + `InteractArea`. The problem is that we have `InteractArea` inheriting `Area2D` already, and `TalkBehavior` being the way to trigger dialogue with that area, but `InteractArea` can be used for other interactions (switching a lever, playing music with rocks, etc). So is not a drop-in replacement. Although it is a good reference! I would borrow the node icon :)

project.godot
```diff
@@ -48,14 +48,14 @@ shapes/paths/geometry_color=Color(0.72156864, 0.003921569, 0.9098039, 0.7764706)
 
 [dialogue_manager]
 
+editor/translations/UPDATE_TRANSLATION_TEMPLATES_AUTOMATICALLY=false
```
wjt, line 51: Bizarre choice upstream to use ALL_CAPS here!

scenes/game_logic/talk_behavior.gd
wjt, line 21: I think it's the right decision to not search-and-replace everywhere in our code & documentation that uses `title` but we should probably mention that Dialogue Manager now calls these cues in the docstring?

scenes/ui_elements/dialogue/components/balloon.gd
```diff
@@ -7,25 +7,34 @@ extends CanvasLayer
...
-var resource: DialogueResource
-
 ## Temporary game states
 var temporary_game_states: Array = []:
 	set(new_value):
 		temporary_game_states = new_value
 		for state: Variant in new_value:
 			if state is Player:
-				_player_name = (state as Player).player_name
+				locals["player_name"] = (state as Player).player_name
```
wjt, line 37: Yep, this is an improvement. But - is it actually necessary? I thought `temporary_game_states` was part of the list of things that is checked to resolve variables.

And there is also a pre-ëxisting bug here - `locals["player_name"]` should be cleared before this loop. (I think in practice it doesn't matter because the states aren't changed once the balloon is instantiated, and the balloon is not reused.)

wjt (approved):
LGTM. Thanks for taking the time to resync the balloon scene & code. I have been meaning to do that because the new upstream bits allow instantiating the balloon directly - which would let us replace Cinematic with a balloon instance + a SceneLink triggered by the dialogue finishing.

scenes/game_logic/talk_behavior.gd
manuq: True. What do you think to change the documentation comments for now, and open a issue to do the full search-and-replace at one point, after current programs end?

scenes/ui_elements/dialogue/components/balloon.gd
manuq: Sorry I got convinced by the comment in locals: "A dictionary to store any ephemeral variables". I can revert it!

wjt: Needs testing! I think I added this originally...

scenes/game_logic/talk_behavior.gd
wjt: Sounds good!

scenes/ui_elements/dialogue/components/balloon.gd
manuq: I remember testing this in my full playthrough when upgrading Dialogue Manager to v4, but I will check again!

PR #2833, opened 2026-09-03
Fix scene link at end of tutorial

### LINKED ISSUE #2834: Tutorial doesn't link to home anymore
### Threadbare version

main

### Affected scene

scenes/quests/lore_quests/quest_000/6_grappling/tutorial_grappling.tscn

### Description

I'm doing a full playthrough for https://github.com/endlessm/threadbare/pull/2828. I wasn't teleported from the tutorial to Fray's End. It looks like we forgot to set the target to be "Home Scene" in the imagination thread (which is a SceneLink).

### Checklist

manuq:
The imagination thread is a SceneLink which pointed to a specific scene (Fray's End) before. Since 5d704eda we removed the specific scene but the target enum wasn't set to "Home Scene". Since e2d8a726 the target enum was introduced.

Fix https://github.com/endlessm/threadbare/issues/2834


wjt (approved):
Oops, thanks for catching this!

PR #2835, opened 2026-09-03
menu_ui_player: Simplify connecting to control signals

wjt:
menu_ui_player: Simplify connecting to control signals

`Object.is_connected(StringName, Callable)` and
`Signal.is_connected(Callable)` are equivalent but I prefer the non-string
form. I also prefer the Python-style `not` rather than the C-style `!`.

The copy-pasted code was hiding an error: for `TabBar` `tab_clicked`,
the check was wrongly for `_on_slider_value_changed`, while the Callable
that is connected in that case is `_on_tab_clicked`.

Fix this by extracting a function to do the "connect once" behaviour, so
it's not possible to make this mistake.

### DIFF
--- a/scenes/globals/uisfx_player/menu_ui_player.gd
+++ b/scenes/globals/uisfx_player/menu_ui_player.gd
@@ -20,17 +20,19 @@ func _enter_tree() -> void:
 
 
 func _on_node_added(node: Node) -> void:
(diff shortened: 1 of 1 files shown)


manuq (approved):
Oh I now realize that I skipped this script entirely in my "review".

Issue #2463, opened 2026-06-24
Make collectible threads more visually enticing


wjt:
On a recent build review @saspiers pointed out that collectible threads just sit on the ground with no idle animation, while other items (notably ability powerups and coins) have idle animations and in the case of the powerup has a visual effect behind it clearly suggesting "pick me up, pick me up!"

https://github.com/user-attachments/assets/f378b656-5cdc-4f10-87d1-9f5985d2f71f

https://github.com/user-attachments/assets/6a41de02-6a71-4fa7-8347-46232edb2090

But threads are just as important as powerups, and more important than coins!

We could add the same "collect me!" effect behind threads, and/or add an idle animation, and/or do something else to draw more attention to them.

### XP Summary (total 375)
| Subdomain | Skill | XP |
| - | - | - |
| Art: Animation & Effects | Problem Solving | 75 |
| Art: Animation & Effects | Creativity & Design Thinking | 150 |
| Art: Animation & Effects | Technical Literacy | 150 |

wjt:
(I feel honour-bound to say that in several hours of watching at least a dozen playtesters at Serious Play, I never saw anyone fail to realise they needed to collect a thread.)

melalicialorenamoyanocastaneda-png:
I plan to work on this

melalicialorenamoyanocastaneda-png:
Here is my planning sheet https://docs.google.com/document/d/18JE26dQ9wN1N0mwG-UlAE1FMzJjhPn5VvCQm1EfEi6o/edit?tab=t.0

melalicialorenamoyanocastaneda-png:
<img width="1600" height="1200" alt="Image" src="https://github.com/user-attachments/assets/d44bd048-d31a-4574-93d4-babff1e0348d" />

I tried to do a bocet first in my book with the threads in different textures, because I wanted to try give them their own texture

melalicialorenamoyanocastaneda-png:
<img width="1600" height="1200" alt="Image" src="https://github.com/user-attachments/assets/6914cbe8-bfd9-46be-9188-93097c6ca625" />

and after that I tried to pass it on digital with ibis paint but... when I was doing the pixelart I saw that each color in the original images they already had names for each color, so i decided to keep the same colors for the threads because if they have that name it must be for something

melalicialorenamoyanocastaneda-png:
[...]

PR #2839, opened 2026-09-04
Make collectible threads more visually enticing  #2463

melalicialorenamoyanocastaneda-png:
Here are the archives that I said how they are colocated in the carpets "NEW" and "aseprite" and how I tried to implement the animation in the godot but I failed miserably

### DIFF
--- a/scenes/globals/uisfx_player/menu_ui_player.gd
+++ b/scenes/globals/uisfx_player/menu_ui_player.gd
@@ -20,17 +20,19 @@ func _enter_tree() -> void:
 
 
 func _on_node_added(node: Node) -> void:
-	if node is CheckButton and !node.is_connected("toggled", _on_toggle_pressed):
-		node.toggled.connect(_on_toggle_pressed)
+	if node is CheckButton:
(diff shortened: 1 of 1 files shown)


manuq (changes requested):
@melalicialorenamoyanocastaneda-png thanks! But please don't move any files. We use Git and it remembers what's old and what's new. There is no problem if you can't do the integration yourself, you already did the artwork. Please if you can, provide just the spritesheets (one frame next to each other), not one PNG per frame.

melalicialorenamoyanocastaneda-png:
Hi I think there's a problem, because, sorry if I moved a file, I just wanted to add the carpets to organize every part of the three threads, because they had many parts and I did'nt know if erase the old draws so I separated in the carpets New and Old and also colocated the assetsprites in another carpet there, but also I had a question, you can't see the spritesheets? because the spritesheets are there too, let me show you


https://github.com/user-attachments/assets/ab844e72-4cf3-43bf-9463-a6b258117185

Also as I say if you put the animation up and down in the godot then all the spritesheets and images are without that animation, or I can just put the spritesheets and assetprites without the other images?

PR #2845, opened 2026-09-04
Dev Archipelago: Translate Basics islands to Spanish

manuq:
This translates:
- The dialogue of the Basics Elder to engage with the quests.
- The content in the StoryBook that this elder presents.
- The signposts of the "Basics" Dev islands / quests.

### ThreadbareEditor addon: Add translation parser for quests

So the .tres file is recognized by the translation system, and title and
description of Quests can be extracted to the catalog.

### Update the list of translatable files:

Add the following new files:
- All three `.tscn` files from Dev Basics.
- Basics Elder dialogue
- Dev Basics quests (two quest.tres files)

Regenerate the template POT file and update Spanish translation.

Resolves https://github.com/endlessm/threadbare/issues/2844

PR #2848, opened 2026-09-07
Update to Dialogue Manager 4.1.0

wjt:
Update to Dialogue Manager 4.1.0

This was done with the in-engine tool.

https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v4.1.0
shows a handful of fixes compared to 4.0.3 plus a couple of things that
might be interesting to us:
https://github.com/nathanhoad/godot_dialogue_manager/pull/1300 says that
“If using gettext, then you probably want to disable [this new option]”.
I have not made that change in this commit.

### DIFF
--- a/addons/dialogue_manager/DialogueManager.cs
+++ b/addons/dialogue_manager/DialogueManager.cs
@@ -993,6 +993,7 @@ namespace DialogueManagerRuntime
             InlineMutations = (Array<Godot.Collections.Array>)data.Get("inline_mutations");
             Time = (string)data.Get("time");
             Tags = (Array<string>)data.Get("tags");
+            ExtraGameStates = (Array<Variant>)data.Get("extra_game_states");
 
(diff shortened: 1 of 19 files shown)


manuq (approved):
Already a minor update, nice! And it doesn't have changes for the example balloon, good.

manuq:
> https://github.com/nathanhoad/godot_dialogue_manager/releases/tag/v4.1.0 shows a handful of fixes compared to 4.0.3 plus a couple of things that might be interesting to us: [nathanhoad/godot_dialogue_manager#1300](https://github.com/nathanhoad/godot_dialogue_manager/pull/1300) says that “If using gettext, then you probably want to disable [this new option]”. I have not made that change in this commit.

Good to know! I will rebase https://github.com/endlessm/threadbare/pull/2845 and check. The default of that new setting is "true" so enabled. So yes I'll confirm and disable it.

Issue #2842, opened 2026-09-04
Request to update the StoryQuest template creation flow


jgbourque:
**Current issues** 

1. The StoryQuest template creation tool currently asks for a StoryQuest name before teams may have chosen one. But the tool is useful from the start of their SQ design as it establishes folders and organization for learner teams.

2. Separately, the generated template suggests a specific example structure with an intro, several minigames, and an outro. This can make the template feel more prescribed than intended. The game has evolved to offer more variety and choice for learners than this particular set up from prior versions.

**Suggested changes**

- [x] Change StoryQuest name to something lighter, such as Working title or team name, so teams can create their project early without needing a final title.
    - https://github.com/endlessm/threadbare/pull/2850
- [ ] Have the template creation tool set a fresh, generic playable-area scene with basic TileMap layers, the Player, and Camera2D as the "main scene" instead of one of the older template scenes to reinforce the idea of constructing scenes from scratch.
    - #2905 
- [ ] Move the existing intro/minigame/outro scenes into a Reference Scenes subfolder so they read as examples and reusable pieces, not the expected StoryQuest structure.

Thoughts on this evolution of this tool? @manuq @wjt

wjt:
This is three (or two depending on how you count), separate requests. :)

The first one - tweaking the tool's UI - is an easy & uncontroversial change.

The second and third - to what extent does it matter that playing NO_EDIT would not match what you get when you run the tool then play your own quest?

Should we instead e.g. remove the sequence puzzle example (this tends to be the least interesting scene in submissions) and replace it with a simple "walk from A to B" scene with the correct structure. Then make this the initial scene of the quest when we run the tool?

wjt:
The “Open guide with instructions.” link points to:

- https://github.com/endlessm/threadbare/discussions/599

@jgbourque do you plan to update/move that guide? What should the link point to instead, if anything?

wjt:
I'm working on the first bit!

manuq:
> * Change StoryQuest name to something lighter, such as Working title or team name, so teams can create their project early without needing a final title.
> *

Yes I think is good to rename the "Title" label to say "Working Title". And even to merge the files with the working title as prefix. It would be too much to ask for renaming all the files before submitting the StoryQuest upstream. I would recommend that the working title is something sufficiently unique. Not "group 1". For example, if the team is considering a SQ about a dragon that throws snowballs instead of fire, call it "Dragon Snow" even if the final title is "How to Train Your Dragon".

[...]

PR #2850, opened 2026-09-07
Make StoryQuest bootstrap dialog less final-sounding

wjt:
Change "Title" to "Working Title" and add a note that the title and
description can be changed later.

Improve the presentation in various ways.

Helps https://github.com/endlessm/threadbare/issues/2842

### DIFF
--- a/addons/storyquest_bootstrap/new_storyquest_dialog.gd
+++ b/addons/storyquest_bootstrap/new_storyquest_dialog.gd
@@ -6,6 +6,8 @@ extends Window
 signal create_storyquest(title: String, description: String, filename: String)
 signal cancel
 
+const HELP_URL := "https://github.com/endlessm/threadbare/discussions/599"
+
 @export var storyquests_path: String
 @export var validate_title: Callable
 @export var validate_filename: Callable
@@ -20,13 +22,16 @@ var _valid := false
 @onready var panel: Panel = %Panel
 @onready var title_edit: LineEdit = %TitleEdit
 @onready var folder_edit: LineEdit = %FolderEdit
-@onready var full_path_label: Label = %FullPathLabel
+@onready var full_path_control: Control = %FullPath
 @onready var errors_label: RichTextLabel = %ErrorsLabel
 @onready var description_edit: TextEdit = %DescriptionEdit
+@onready var guide_button: Button = %GuideButton
 @onready var progress_bar: ProgressBar = %ProgressBar
 
 
 func _ready() -> void:
+	guide_button.icon = EditorInterface.get_editor_theme().get_icon("ExternalLink", "EditorIcons")
+
 	var style := get_theme_stylebox("PanelForeground", "EditorStyles")
 	panel.add_theme_stylebox_override("panel", style)
 	title_edit.grab_focus()
@@ -37,6 +42,16 @@ func _ready() -> void:
 	_invalid_char_regex = RegEx.new()
 	var error := _invalid_char_regex.compile("\\W+", true)
 	assert(error == OK, error_string(error))
+
+	full_path_control.text = storyquests_path
+
+
+func _notification(what: int) -> void:
+	match what:
+		NOTIFICATION_EDITOR_PRE_SAVE:
+			guide_button.icon = null
+			panel.remove_theme_stylebox_override("panel")
+			errors_label.remove_theme_color_override("default_color")
 
 
 func _input(event: InputEvent) -> void:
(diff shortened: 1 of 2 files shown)


wjt:
image

addons/storyquest_bootstrap/new_storyquest_dialog.tscn
```diff
@@ -102,6 +91,13 @@ placeholder_text = "This quest is about..."
 wrap_mode = 1
 tab_input_mode = false
 
+[node name="TipsLabel" type="Label" parent="PanelContainer/VBoxContainer" unique_id=1470208228]
+custom_minimum_size = Vector2(400, 0)
+layout_mode = 2
+text = "You can change your quest's title and description later!"
```
manuq, line 97: 👍

addons/storyquest_bootstrap/new_storyquest_dialog.tscn
```diff
@@ -34,64 +35,52 @@ theme_override_constants/margin_bottom = 10
...
 layout_mode = 2
-text = "Folder Name:"
+text = "Folder Name"
 
 [node name="FolderEdit" type="LineEdit" parent="PanelContainer/VBoxContainer/GridContainer" unique_id=146935302]
 unique_name_in_owner = true
 layout_mode = 2
 size_flags_horizontal = 3
-placeholder_text = "dragonhouse"
+placeholder_text = "project_green"
```
manuq, line 60: 👍

addons/storyquest_bootstrap/new_storyquest_dialog.tscn
```diff
@@ -34,64 +35,52 @@ theme_override_constants/margin_bottom = 10
...
-uri = "https://github.com/endlessm/threadbare/discussions/599"
-
 [node name="GridContainer" type="GridContainer" parent="PanelContainer/VBoxContainer" unique_id=1146920114]
 layout_mode = 2
 columns = 2
 
 [node name="TitleLabel" type="Label" parent="PanelContainer/VBoxContainer/GridContainer" unique_id=2129724723]
 layout_mode = 2
-text = "Title:"
+text = "Working Title"
```
manuq, line 44: 👍

addons/storyquest_bootstrap/new_storyquest_dialog.gd
```diff
@@ -38,6 +43,16 @@ func _ready() -> void:
...
 
+	full_path_control.text = storyquests_path
+
+
+func _notification(what: int) -> void:
+	match what:
+		NOTIFICATION_EDITOR_PRE_SAVE:
+			guide_button.icon = null
+			panel.remove_theme_stylebox_override("panel")
+			errors_label.remove_theme_color_override("default_color")
```
manuq, line 54: Interesting. Can you explain why is this needed? Is this because we have a custom theme set in the project settings? So now creating UI for the editor itself becomes more difficult?

```
[gui]

theme/custom="uid://cvitou84ni7qe"
```

addons/storyquest_bootstrap/new_storyquest_dialog.gd
```diff
@@ -119,3 +134,7 @@ func _revalidate() -> void:
 	_valid = errors.size() == 0
 	errors_label.visible = not _valid
 	create_button.disabled = not _valid
+
+
+func _on_guide_button_pressed() -> void:
+	OS.shell_open(HELP_URL)
```
manuq, line 140: Interesting! Can you explain why `OS.shell_open(HELP_URL)` triggered from a regular `Button` is preferred over using a `LinkButton`?

manuq (approved):
All good! I'm only curious about some changes.

addons/storyquest_bootstrap/new_storyquest_dialog.gd
wjt: That point about the custom theme in project settings is true! But actually that's not why this is needed. Strictly speaking it's *not* needed but without it, when I save this scene, whatever styles/icons happen to come from *my* editor theme get saved into the .tscn file, including the full SVG icon. This is unnecessary - they're replaced at runtime - and also is liable to cause spurious diffs in future (e.g. if you use the light theme and I use the dark theme, our editors will fight over what's saved into the file).

But actually your point about the custom theme is interesting... maybe we can assign the editor's theme to this control to make it easier to edit despite our global theme! I'll try that.

addons/storyquest_bootstrap/new_storyquest_dialog.gd
wjt: I thought the link-button (with underlined text) looked bad relative to the rest of the dialog box. Elsewhere in Godot external links look like this:

image

so I thought I'd match that.

wjt: ...and LinkButton doesn't support setting an icon.

PR #2857, opened 2026-09-08
Apply low-pass filter to music bus when paused

wjt:


wjt:
Doesn't work on the web. I don't know why not because I did the same thing on https://wjt.itch.io/bgcolor-2025 and it does work on the web. But it's not worth thinking about this any more... it was just an idea!

manuq:
Oh too bad, I really like this idea! I heard it on Hades. One special case is when adjusting the music volume, the music goes back to normal.

PR #2858, opened 2026-09-08
storyquest_bootstrap: Show editor theme when editing dialog

wjt:
Thanks Manuel for this great idea: we can set the editor's theme on the
dialog window when editing it, rather than it being displayed with the
game's theme (which is the global default for the project).

Clear this when saving the scene. Reapply these theme properties
post-save. Explain in comments why we do this.


wjt:
image

Much better than:

image

manuq (approved):
Much better! By the way it was your idea, I only made a question :)

PR #2863, opened 2026-09-09
Make collectible threads more visually enticing

melalicialorenamoyanocastaneda-png:
#2463

I added the spritesheets of the tails of the threads and the lightning of the threads too in a carpet called "threads_animation" there are also the aseprites solicited, and in the carpet "new_threads" where are the new threads, they don't change a lot, just a little bit the texture, doing the red thread more fat for example and the green one more thinner.


melalicialorenamoyanocastaneda-png:
I solucionated the problems in the previous request, and now there are the images and spritesheets and less archives changed, also there was a problem with my repository but it's already fixed, sorry for the incoveniences.

manuq:
@melalicialorenamoyanocastaneda-png thanks! I'll take it from here, and open a separate ticket for the implementation. I will not merge the thread PNGs as new files, they are almost identical:

https://github.com/user-attachments/assets/8b5e7364-3e38-4a49-b3aa-d7040845d364

The only thing that should be kept, I think, is separating the shadow into a different PNG.

Also the highlight animation is the same  for all colors, so we don't need 3 separate spritesheets.

manuq:
I had to open https://github.com/endlessm/threadbare/pull/2875 because of Git LFS I can't push to the remote.

Issue #2771, opened 2026-08-20
Dev Archipelago sheep's repel sound is audible from too far away


wjt:
### Threadbare version

v0.2.2-29-g682357b48

### Affected scene

res://scenes/dev/dev_archipelago.tscn

### Description

In Dev Archipelago there are a pair of sheep playing tennis:

<img width="1330" height="807" alt="Image" src="https://github.com/user-attachments/assets/4361c944-a758-4f7c-8707-3abe69499fc3" />

They each have a `PlayerRepel` component. This component plays a sound effect whenever the repel action fires.

The problem is that the sound effect can be heard from a long way away. For example, you can hear it from here:

<img width="1330" height="807" alt="Image" src="https://github.com/user-attachments/assets/e41c0c37-121b-4a64-b1bb-c68ea911e8cb" />

and here:

<img width="1330" height="807" alt="Image" src="https://github.com/user-attachments/assets/cfba4a2b-62df-4082-aeb1-06b7ec285b6e" />

This feels too far away for the sound to be audible.

<details>
<summary>XP Summary (total 500)</summary>

| Subdomain | Skill | XP |
| - | - | - |
| Art: Music & Sound | Problem Solving | 150 |
| Art: Music & Sound | Creativity & Design Thinking | 75 |
| Art: Music & Sound | Technical Literacy | 75 |
| Game Design: Game Feel & UX | Problem Solving | 75 |
| Game Design: Game Feel & UX | Creativity & Design Thinking | 50 |
| Game Design: Game Feel & UX | Technical Literacy | 75 |
</details>

wjt:
The AudioStreamPlayer2D has the default max distance (2000px) and attenuation curve (1.0, i.e. linear). In this scene we do not have an AudioStreamListener2D so attenuation & stereo panning is relative to the middle of the viewport.

So you might think: OK, reduce the distance to 1280px (say). This means that you can still hear the sound effect even when the sheep are offscreen by up to half a horizontal viewport, but beyond that they are silent. The problem with this is that this attenuation also applies to the player's repel component: if the player is on the left-hand side of the screen, that's 640px from the centre, so the sound effect would be half-way down the attenuation curve. (This is already the case of course - if you repel near the edge of the ink levels the sound is slightly quieter. It's just not very noticeable because the max distance is 2000px so you're never more than about 1/3 away from the centre of the screen.)

[...]

PR #2867, opened 2026-09-09
PlayerRepel: Use AudioStreamPlayer2D.play() not audio animation track

wjt:
PlayerRepel: Use AudioStreamPlayer2D.play() not audio animation track

This works around a bug/limitation of the web build where spatial
attenuation does not work when using an audio animation track.

Without this change, in Dev Archipelago, the sheep's repel sound is
audible wherever you are in the map, at full volume. With this change,
it is attenuated in the same way as on desktop.

The consequence of this change is that, on the web, the player's own
repel sound is also now panned and attenuated when the player is not at
the centre of the viewport (i.e. when they move near the camera's limits
so the camera can't follow them). I think this is bad! But it's a
pre-existing problem on desktop: this change just makes it consistent
between platforms.

https://github.com/endlessm/threadbare/issues/2771#issuecomment-5598904910

### DIFF
--- a/scenes/game_elements/characters/player/components/player_repel.tscn
+++ b/scenes/game_elements/characters/player/components/player_repel.tscn
@@ -136,21 +136,20 @@ tracks/4/keys = {
 "update": 0,
 "values": [Color(1, 1, 1, 0), Color(1, 1, 1, 0.456), Color(1, 1, 1, 0)]
 }
-tracks/5/type = "audio"
+tracks/5/type = "method"
 tracks/5/imported = false
 tracks/5/enabled = true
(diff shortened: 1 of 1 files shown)


manuq:
Nice workaround! From the islands at the north (grapple elder) and at the east (enemy elder) I can hear it attenuated. From the island at the south east (level design elder) I don't hear it without headphones.

> The consequence of this change is that, on the web, the player's own repel sound is also now panned and attenuated when the player is not at the centre of the viewport (i.e. when they move near the camera's limits so the camera can't follow them). I think this is bad! But it's a pre-existing problem on desktop: this change just makes it consistent between platforms.

I trust you on this!

manuq (approved):

PR #2869, opened 2026-09-09
Dev Archipelago: Reduce the ping-pong sheep's frequency

wjt:
Previously they each repelled on a 1-second timer, with one offset by
0.5s.

Double these timeouts this so the ball is repelled once per second, not
twice. This makes the sound less annoying.


wjt:
This is orthogonal to the other changes related to #2771 but I think it's better to be less frequent. WDYT?

manuq (approved):
Better!

PR #2873, opened 2026-09-10
Add pre-alpha label to main menu

### LINKED ISSUE #959: Obviously flag the pre-release state of the game on the main title
As we start to talk about and link out to the game more, we should make it more clear in the game itself what its status is—I would hate for someone to start the game with higher expectations, and then not understand that it's an early work in progress.

My initial thought was a big obvious banner across/below the Threadbare logo that says `PRE-ALPHA` or something, but @wjt also pointed out that we have the specific version number in the bottom-right of the main title; perhaps near there makes more sense.

### XP Summary (total 175)
| Subdomain | Skill | XP |
| - | - | - |
| Game Design: Game Feel & UX | Technical Literacy | 50 |

ChiguireDigital:
Adds a `PRE ALPHA` label to the main menu next to the current game version.

- Added the `PRE ALPHA` label before the current version number.
- Kept the existing version label, font, styling, and positioning.

Resolves #959

### DIFF
--- a/addons/threadbare_git_describe/version_label.gd
+++ b/addons/threadbare_git_describe/version_label.gd
@@ -8,4 +8,4 @@ extends Label
 
 
 func _ready() -> void:
-	text = preload("./version.gd").get_full_version()
+	text = "PRE ALPHA · " + preload("./version.gd").get_full_version()


ChiguireDigital:
image

Hi! Here is the result so far. I think it looks good.

What do you think?

wjt (approved):
Interesting! I was expecting an additional `Label` in `res://scenes/menus/title/components/main_menu.tscn`. The consequence of adding it here is that it also applies to the other scene that uses `version_label.gd`:

image

Namely the pause screen:

Screenshot From 2026-09-10 14-54-59

I'm not sure if you noticed this. I hadn't thought of it before reviewing this, but actually I think it is good to also show the version number here.

The downside of doing it like this, in code, is that it is harder to localise the "PRE ALPHA" text. I only speak English so for me it's fine ;-) and I will leave it to a Spanish speaker to let me know whether different text should be shown if the game is in Spanish. If so we can change the approach.

manuq:
> The downside of doing it like this, in code, is that it is harder to localise the "PRE ALPHA" text. I only speak English so for me it's fine ;-) and I will leave it to a Spanish speaker to let me know whether different text should be shown if the game is in Spanish. If so we can change the approach.

We should start a style guide for translations, I discussed a while ago with @Stregoica777. From what I remember (I'm not expert) "pre alpha" would be acceptable as slang that's easy to understand, so we probably don't want to translate it. Another example could be "spawnear" as acceptable for "spawning".

manuq:
That is for Spanish though!

PR #2877, opened 2026-09-10
Checkpoint: Add tile change persistence

### LINKED ISSUE #1208: Support checkpoints in expanding void levels

NataliaOspinal:
Whenever an enemy changes part of the tilemap as the VoidSpreadingEnemy does, if the player reaches a checkpoint and dies, the scene resets to its original state instead of keeping the changes that we'd like to keep in previous traveled zones for storytelling or possible gameplay mechanics.

To fix this, I created a new Checkpoint child class (I wasn't sure I could edit the original, but if that's the idea I could just add a checkbox for persistent enemies and the respective exports to incorporate the code to the original) that effectively tracks with both _process and _physics_process for accuracy of tile changes, checking the respective paths and its enemy states.

Note: I've included (as part of the code) a way to track the changes of a possible void-spreading guard, since that was the idea shown in the screenshots of the issue!

Resolves: https://github.com/endlessm/threadbare/issues/1208

### DIFF
--- a/scenes/game_elements/props/checkpoint_void/components/checkpoint_void.gd
+++ b/scenes/game_elements/props/checkpoint_void/components/checkpoint_void.gd
@@ -3,22 +3,21 @@
 @tool
 extends Checkpoint
 class_name CheckpointVoid
+## A checkpoint that saves and restores the state of specific enemies and the void layer.
+##
+## This checkpoint extends the base Checkpoint functionality to track specific enemies 
+## (like guards or void-spreading enemies) and prevents them, along with consumed tiles, 
+## from resetting when the scene is reloaded.
 
-# This checkpoint variation will allow some enemies to bypass the scene 
-# reset after re-spawning in a checkpoint, keeping their tilemap changes
-# and last positions
-
-# Static variables
 static var saved_enemy_states: Dictionary = {}
 static var saved_consumed_tiles: Array[Vector2i] = []
 static var pending_consumed_tiles: Array[Vector2i] = []
 static var _tracker_instance: Node = null
 
-# Variables you need to assign in Inspector
-## Assign the enemies you want to bypass the reset scene
+## Specific enemies that should retain their position and state across scene reloads.
 @export var persistent_enemies: Array[CharacterBody2D]
 
-## Assign the modified layer
+## The tilemap layer that is being consumed by the void.
 @export var shared_void_layer: Node2D
 
 const _NEIGHBORS := [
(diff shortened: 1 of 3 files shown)


scenes/game_elements/props/checkpoint_void/components/checkpoint_void.gd
```diff
@@ -0,0 +1,199 @@
...
+class_name CheckpointVoid
+## A checkpoint that saves and restores the state of specific enemies and the void layer.
+##
+## This checkpoint extends the base Checkpoint functionality to track specific enemies 
+## (like guards or void-spreading enemies) and prevents them, along with consumed tiles, 
+## from resetting when the scene is reloaded.
+
+# Store everything including position, paths, and consumed tiles in a single dictionary 
+# bound strictly to each enemy to prevent synchronization issues.
+static var saved_enemy_states: Dictionary = {}
```
wjt: As commented on https://github.com/endlessm/threadbare/issues/1208#issuecomment-5634192552 I don't think static variables are an appropriate approach.

wjt (changes requested):

manuq:
@NataliaOspinal I see you have added new changes. Let us know when it's ready for review again.

NataliaOspinal:
Hello! The code is ready for review, I gave notice on the Issue itself a few days ago, but just in case I'm giving notice here too! @manuq

scenes/game_elements/characters/enemies/guard/components/guard.gd
manuq, line 1: The comment in the ticket "Imagine that the template enemies are void-spreading enemies which chase you, but can't cross the bridges" was hypothetical. Sorry for the confusion. Please revert any changes to `guard.gd`.

manuq (changes requested):
This is looking good, but I wasn't able to test it! Please provide a working example. You could add one in `scenes/dev/level_design/`. Check for instance, the `scenes/dev/level_design/doors_and_levers/`.

scenes/game_elements/characters/enemies/guard/components/guard.gd
NataliaOspinal: Okay got it! I created a testing scene inside scenes/dev/level_design/void_persistence and reverted the changes in the guard.gd script! Let me know if more changes are needed!

PR #2882, opened 2026-09-11
Dev Archipelago: Improve camera in Dev Grapple and add long needles example

manuq:
### Dev Grapple: Add example of long, long needles path

As a test that this can be done, only with the corresponding camera.

### Dev Grapple: Add PhantomCamera

The camera in this demo had no smoothing. All cameras in lore levels have
position smoothing activated. But for grapple scenes is better to use a
PhantomCamera. Add one and set its limits to the water tilemap layer.
Explanation will come in next commit.


manuq:
To test it, play [scenes/dev/grapple/grapple_gym_1/grapple_gym_1.tscn](https://play.threadbare.game/branches/endlessm/dev-grapple-camera/#dev/grapple/grapple_gym_1/grapple_gym_1) and at the end, use this needle to go back to the beginning:

Screenshot From 2026-09-11 14-02-12

wjt (approved):
Image

😂 😂 😂 😂

PR #2883, opened 2026-09-11
UI: Expand credits screen vertically

### LINKED ISSUE #2383: Credits screen does not expand vertically
Image

The control hierarchy should be set up so that the scroll expands to fill the full height of the screen, if the screen is taller than 16:9 (the base resolution's aspect ratio).

The screenshot above is an extreme aspect ratio, but for example resizing the window from 1280×720 to 1280×800 should (I think) cause the scroll to expand vertically into the additional 80 pixels. If perhaps above some limit its vertical size is clamped, it might look better to vertically centre it. 

### XP Summary (total 200)
| Subdomain | Skill | XP |
| - | - | - |

Zer0LoL:
The credits screen layout has been updated so that the content expands vertically to fill additional screen height on aspect ratios taller than 16:9 (such as 16:10 or 5:4).

- Wrapped the credits panel in a CenterContainer and MarginContainer hierarchy inside credits.tscn.

- Configured size limits so that on extreme vertical aspect ratios, the panel stops expanding and remains vertically centered, matching the behavior of the options menu.

Resolves https://github.com/endlessm/threadbare/issues/2383

### DIFF
--- a/scenes/menus/title/components/credits.tscn
+++ b/scenes/menus/title/components/credits.tscn
@@ -372,32 +372,35 @@ texture = ExtResource("8_1fvwr")
 expand_mode = 1
 stretch_mode = 5
 
-[node name="PanelContainer" type="PanelContainer" parent="." unique_id=586483173]
+[node name="CenterContainer" type="CenterContainer" parent="Control" unique_id=361753202]
 layout_mode = 1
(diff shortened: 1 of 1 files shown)


manuq (approved):
Excellent!

PR #2887, opened 2026-09-14
Add challenge scenes

### LINKED ISSUE #2871: Add Justin's new challenge scenes
@jgbourque has some scenes in `xQVHhCi8Nbr2Y58qc4EwYzEVzLX` that should be imported into the game. The request is to add them to the StoryQuest Kit, which implies they need to live in Dev Archipelago.

wjt:
These are deliberately unfinished scenes to accompany some guides.

Place them alongside the NO_EDIT template since they are deliberately
minimally themed (and use the template player rather than StoryWeaver).

Resolves https://github.com/endlessm/threadbare/issues/2871


manuq (commented):
This is the first time in Dev Archipelago that players would see a template character instead of the StoryWeaver, and scenes that by themselves have no playability at all. I understand that they would depend on companion guides, as explained in the quest description.

In [Dev Archipelago MVP](https://github.com/endlessm/threadbare/milestone/8) we described:
> The player has ultimate freedom to play and experiment with different testing levels, gyms, and simple ("toy") examples with the hopes of sparking inspiration for unique StoryQuests and other contributions to the Threadbare game.

So I wonder if the archipelago is suitable for this. Could these be templates instead? The first siblings of the NO_EDIT template? Maybe a brand new "Challenges" template?

Otherwise, what are ways to make these scenes playable before being edited, while still open-ended?

wjt:
Putting them into dev archipelago was my idea. We could instead just put them into `scenes/challenges` and not have them reachable in-game, only in the editor. I'm fine either way!

Edit: I reread your comment. Interesting idea, they could be in `scenes/quests/template_quests`!

wjt:
I revised this to drop the playground elder, and put them alongside NO_EDiT.

@jgbourque wdyt?

manuq (approved):
Looks good to me! Waiting for @jgbourque to confirm.

PR #2888, opened 2026-09-14
Void spreading enemy: Don't crash if void_layer or path are unset

wjt:
Previously if either of these properties were unset, the game would crash.

Both these properties are required for proper functionality but it's
kinder to show a warning in the editor and then have the enemy do the
best it can at runtime. Doing this will also help to make a challenge
scene for learners where they have to correctly configure the enemy
before improving its path.


scenes/game_elements/characters/enemies/void_spreading_enemy/components/void_spreading_enemy.gd
```diff
@@ -97,11 +112,30 @@ func _set_state(new_state: State) -> void:
...
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	var warnings: PackedStringArray
+
+	if not void_layer:
+		warnings.append("Void Layer not set. The enemy cannot cover the world or consume props")
+
+	if not idle_patrol_path:
+		warnings.append("Idle Patrol Path not set. The enemy will be invisible when idle")
```
manuq, line 125: Yeah, it will be invisible as a consequence of being in the same position / not moving.

manuq (approved):
Great!

PR #2891, opened 2026-09-14
Sign post: Wrap long lines

### LINKED ISSUE #2843: Sign posts: Wrap lines correctly
Currently the signpost text have newlines for cutting long sentences: 

Image

But this is not ideal. If we want to translate this text, the wrapping should handled by the Control node.

manuq:
Give the label a maximum horizontal size of 300, and keep -1 for the maximum
vertical size. Also change autowrap mode to "Words", and reset vertical and
alignment to the default (top).

Give the LabelContainer (the first Control node) a custom minimum size of 320 x
32. This is in the X axis, the max size of the label plus the horizontal margins
(10 on each side). And on the Y axis, the size of a line of text plus the
vertical margins (5 on each side). And to allow smaller sizes (for example a
signpost saying "Go!"), set the horizontal container sizing of the
PanelContainer child to Shrink Center.

### Dev Basics: Remove line breaks from text in signposts

As a test of how this looks, before changing all of it in batch.

Update Spanish translation.

Fix https://github.com/endlessm/threadbare/issues/2843


scenes/quests/lore_quests/quest_002/2_grappling_hook/grappling_hook_pins.tscn
```diff
@@ -278,8 +278,7 @@ position = Vector2(-79, 26)
 [node name="Sign" parent="OnTheGround" unique_id=911109431 instance=ExtResource("12_50o53")]
 position = Vector2(2929, 1233)
 direction = 1
-text = "Go!
-"
+text = "Go!"
```
manuq, line 281: Fixes this bug which can be seen before and after the changes in `sign.tscn`.

Image

manuq:
### Screenshots of a scene that has been updated (line breaks removed):

Before:

image

After:

image

### Screenshots of a scene that hasn't been updated:

Before:

image


After:

image

wjt (approved):
Nice improvement!

If the message is really long then the label ends up obscuring the signpost. One way to solve this would be to make the label grow upwards. I think this could be done by adjusting the anchor/pivot position of the controls, though I tried and failed to make it work. However the real solution is “Don't do that”. For longer text, the sign could be made interactible and the text could be shown as dialogue.

manuq:
> Nice improvement!
> 
> If the message is really long then the label ends up obscuring the signpost. One way to solve this would be to make the label grow upwards. I think this could be done by adjusting the anchor/pivot position of the controls, though I tried and failed to make it work. However the real solution is “Don't do that”. For longer text, the sign could be made interactible and the text could be shown as dialogue.

Good point! We can revisit the anchoring. I was even considering adding an InteractArea plus a TalkBehavior to signposts and use the dialogue bubble for displaying the text. Maybe we can consider further changes in the future, for now I wanted to unblock the Spanish translation of Dev Archipelago.

manuq:
@wjt I took your suggestion in https://github.com/endlessm/threadbare/pull/2892

PR #2896, opened 2026-09-15
Dev Archipelago: Add Spanish translation

manuq:
### Tools: Add tool to better maintain the list of translatable files

Because maintaining this from the Project Settings Localization tab is tedious
and error-prone.

### custom_hookable_objects_test.tscn: Remove leftover signpost

It was left with a FIXME label outside the playable area. This is also to remove
the FIXME line from being translatable.

### grapple_gym_2.tscn: Mark pixel labels as not translatable

These are just numbers, in principle they shouldn't be translatable.

### Project: Update list of translatable files

Using the new tool.

This keeps all existing files, adds all Dev Archipelago, and also adds
scenes/globals/pause/quest_abandoned.dialogue which I think is needed anyways
for Dev Archipelago (when player abandons the demo).


manuq:
This is work in progress. For now I worked on an editor script to help maintain the list of translatable files, similar to [what we had in Block Coding](https://github.com/endlessm/godot-block-coding/blob/006e31ebe7951fe4230021f54fd2201cee29d966/addons/block_code/translation/utils.gd#L132). Because maintaining this from the Project Settings Localization tab is tedious and error-prone.

tools/update_translatable_files.gd
```diff
@@ -0,0 +1,128 @@
...
+			elif extension == "tres" and _is_quest(path):
+				files.append(path)
+
+	files.sort()
+	return files
+
+
+## Returns whether [param path] is a [Quest] resource.
+static func _is_quest(path: String) -> bool:
+	return ResourceLoader.load(path, "", ResourceLoader.CACHE_MODE_IGNORE) is Quest
```
wjt, line 128: I'm curious whether `ResourceLoader.exists(path, "Quest")` would also work.

tools/update_translatable_files.gd
```diff
@@ -0,0 +1,128 @@
+# SPDX-FileCopyrightText: The Threadbare Authors
+# SPDX-License-Identifier: MPL-2.0
+@tool
+extends EditorScript
```
wjt, line 4: If you add `class_name UpdateTranslatableFiles` then it can be run from the Ctrl+Shift+P command palette.

wjt (approved):
Looks great! Writing all those translations must have been a lot of work.

tools/update_translatable_files.gd
manuq: That's smart! I'll add it.

tools/update_translatable_files.gd
manuq: I tried and it added:

```
The POT generation list now has 62 file(s).
  + res://scenes/ui_elements/input_hints/components/drop_shadow_material.tres
```

Which is not a `Quest` resource, it's a `ShaderMaterial` resource. Strange!

wjt: I guess that means it doesn't work :)

PR #2897, opened 2026-09-15
RandomFrameSpriteBehavior: re-randomize after SpriteFrames changes

### LINKED ISSUE #2890: AnimatedDecoration + RandomFrameSpriteBehavior don't play nicely together
In https://github.com/endlessm/threadbare/pull/2882 I added an area filled with capycorns:

https://github.com/user-attachments/assets/8ecfd2b4-7b2a-4852-b93a-b2be8642a9ec

And as you can see, all of them play the idle animation in sync. Instead, the ones I instantiated one by one don't have that issue (the small groups in the video above).

I suspect that AnimatedDecoration is the problem. It's sole purpose seems to be to proxy the SpriteFrames property to the inner AnimatedSprite2D. I checked areas filled with trees (which don't have AnimatedDecoration script attached) and they don't have this problem.

<details>
<summary>XP Summary (total 300)</summary>

| Subdomain | Skill | XP |
| - | - | - |
</details>

wjt:
This is necessary because the structure we often have (e.g. in capycorn)
is:

- AnimatedDecoration
  - AnimatedSprite2D
    - RandomFrameSpriteBehavior

_ready() is called from leaf to root, so
`RandomFrameSpriteBehavior._ready()` is called before
`AnimatedDecoration._ready()`. But if `AnimatedDecoration.sprite_frames` is
set, it is only applied to `AnimatedSprite2D` in
`AnimatedDecoration._ready()` - which is after
`RandomFrameSpriteBehavior._ready()` has run and randomized the frame.
Setting `AnimatedSprite2D.sprite_frames` resets the frame & progress to
zero.

Handle AnimatedSprite2D.sprite_frames_changed signal and re-randomize.
Do the same after the scene is saved so that saving the scene doesn't
cause everything to synchronise in the editor.

Resolves https://github.com/endlessm/threadbare/issues/2890


scenes/game_logic/sprite_behaviors/random_frame_sprite_behavior.gd
```diff
@@ -9,6 +9,11 @@ extends BaseSpriteBehavior
 
 
 func _ready() -> void:
+	randomize_frame()
+	sprite.sprite_frames_changed.connect(randomize_frame)
```
manuq, line 13: Ohh `sprite_frames_changed` exists!

manuq (approved):
🥇

scenes/game_logic/sprite_behaviors/random_frame_sprite_behavior.gd
wjt: Yes, I was happy to find that!

PR #2901, opened 2026-09-16
Make cat walk around

### LINKED ISSUE #1290: Make the cat move
Currently the cat in Fray's End does not move around.

Once we have a walking animation for it (#1288), we can make it move around the map. My suggestion is that it could mostly sit still, but occasionally walk around erratically. Here's one way this could be done:

1. Add an ErraticWalkBehavior to the cat. Set its process mode to DISABLED by default.
2. Add a Timer to the cat. Autostart the timer but make it one-shot.
3. When the timer fires, set the ErraticWalkBehavior to process.
4. When ErraticWalkBehavior emits its `direction_changed` signal, set its process mode back to DISABLED, and start the timer again (perhaps with a randomised interval).
5. Similarly, when the player interacts with the cat, disable ErraticWalkBehavior. Start the timer when the interaction ends.

### XP Summary (total 450)
| Subdomain | Skill | XP |
| - | - | - |

IDynamixI985:
Makes the cat walk around *Fray's End*.

* **Map Boundaries (`frays_end.tscn`):** Added a `StaticBody2D` (`CatLimit`) using layer 2 (`npcs`) at the map exits, positioned so they block the cat from escaping without obstructing the player's path.
* **Cat Nodes (`cat.tscn`):** Added `%ErraticWalkBehavior` (with speeds and travel distance set) and a `%Timer` (2s, one-shot, autostart) to control walk and idle cycles.
* **Signal Connections:** Handled state transitions using editor signals:
  * `%Timer.timeout` enables `%ErraticWalkBehavior`.
  * `%ErraticWalkBehavior.direction_changed` disables walk behavior, resets `velocity` to zero (triggers idle animation), and restarts `%Timer`.
  * `%InteractArea.interaction_started` stops movement and the timer while being petted; `interaction_ended` restarts `%Timer`.

Resolves: https://github.com/endlessm/threadbare/issues/1290


IDynamixI985:
Hi! I completed all requirements for the task, I think, but the cat sprite currently only faces right while walking. What's the preferred way to flip the sprite here?

wjt:
> Hi! I completed all requirements for the task, I think, but the cat sprite currently only faces right while walking. What's the preferred way to flip the sprite here?

scenes/game_logic/sprite_behaviors/character_sprite_behavior.gd

> Map Boundaries (frays_end.tscn): Added a StaticBody2D (CatLimit) with a CollisionPolygon2D in Segments mode on layer 11. This confines the cat to its area without blocking the player.

I don't think we should use up a whole new layer for this.

If we're happy to let the cat roam the whole map, but not let it escape, we can place some invisible colliders that block NPCs (the cat is an NPC) at the boundaries of the map.

If you want to confine the cat to a smaller area you could use a NavigationRegion and pathfind within that. But personally I think "whole map, block exits" would be fine for an erratic walk.

IDynamixI985:
Hi, thanks for the feedback! The changes have been made and the PR description is updated. Everything is all set now.

scenes/world_map/frays_end.tscn
```diff
@@ -101,6 +101,15 @@ _data = {
...
 point_count = 4
 
+[sub_resource type="RectangleShape2D" id="RectangleShape2D_gm6tl"]
+size = Vector2(28, 320)
+
+[sub_resource type="RectangleShape2D" id="RectangleShape2D_6fau3"]
+size = Vector2(33, 195)
+
+[sub_resource type="RectangleShape2D" id="RectangleShape2D_u5bk5"]
+size = Vector2(334.125, 49)
```
wjt: I think these need to be a bit further out, outside the camera bounds (the yellow box). Because the player can also collide with them, these mean that the player no longer walks off screen when you move out of fray's end:

Image

https://github.com/user-attachments/assets/3616c40e-510c-4afe-8e50-829e9a6716c5

Try moving the barrier outside the camera limits.

I also think they should be in the "walls" layer not the "npcs" layer.

scenes/world_map/frays_end.tscn
```diff
@@ -838,6 +847,22 @@ position = Vector2(969.4194, 2045.982)
 scale = Vector2(1.1060888, 1.0290669)
 polygon = PackedVector2Array(-877.343, -208.91, -817.673, -167.124, -758.004, -191.418, -697.43, -150.605, -638.664, -170.04, -579.899, -209.882, -460.559, -201.136, -390.04, -228.345, -324.042, -206.966, -255.332, -160.322, -168.539, -205.023, -99.8287, -193.362, -2.18738, -175.87, 55.6742, -183.644, 109.919, -223.486, 171.397, -226.401, 217.506, -200.164, 320.572, -186.559, 404.652, -186.559, 505.005, -189.475, 517.662, -237.091, 589.455, -238.273, 670.946, -241.251, 812.201, -217.423, 889.163, -140.978, 974.222, 0.017334, -874.675, -0.992798)
 
+[node name="CatLimit" type="StaticBody2D" parent="OnTheGround" unique_id=1094201463]
```
wjt, line 850: There's an exit route you didn't consider: the bridge at the bottom right:

Image

scenes/game_elements/characters/npcs/cat/cat.tscn
```diff
@@ -65,3 +177,36 @@ fact_counter = NodePath("../FactCounter")
...
+autostart = true
+
+[connection signal="interaction_ended" from="InteractArea" to="Timer" method="start"]
+[connection signal="interaction_started" from="InteractArea" to="Timer" method="stop" unbinds=2]
+[connection signal="interaction_started" from="InteractArea" to="." method="set_velocity" unbinds=2 binds= [Vector2(0, 0)]]
+[connection signal="interaction_started" from="InteractArea" to="ErraticWalkBehavior" method="set_process_mode" unbinds=2 binds= [4]]
+[connection signal="direction_changed" from="ErraticWalkBehavior" to="Timer" method="start"]
+[connection signal="direction_changed" from="ErraticWalkBehavior" to="." method="set_velocity" binds= [Vector2(0, 0)]]
+[connection signal="direction_changed" from="ErraticWalkBehavior" to="ErraticWalkBehavior" method="set_process_mode" binds= [4]]
+[connection signal="timeout" from="Timer" to="ErraticWalkBehavior" method="set_process_mode" binds= [0]]
```
wjt, line 212: Wow, I didn't expect all this to be done in the scene editor! Very clever.

(I think it might be clearer to set this all up in code, with a `cat.gd` script attached to the root node, but don't feel you have to change it.)

scenes/game_elements/characters/npcs/cat/cat.tscn
wjt, line 1: The cat can walk on water! You need to change his collision_mask to include `non_walkable_floor`.

Image

wjt (changes requested):
Great fun! Just a few little changes.

I think we put him in his current position when he was a static asset with no interactive behaviours, so he was a bit out of the way:

Image

Now that he moves around, he tends to get stuck in this corner because he's tucked away between lots of objects he can collide with. Maybe that's OK – he's hanging around his home, and only occasionally goes exploring!

IDynamixI985:
Changes are all done! Now the cat can roam around the beach without taking an accidental swim.
all_done

KarmDK:
Hi! I've been testing it out, and one thing I've noticed is that I can't cross West End, I don't think that should be happening.

KarmDK:
> Ahora que se mueve, tiende a quedarse atrapado en esta esquina porque está escondido entre muchos objetos con los que puede chocar. Quizás esté bien – ¡está rondando por su casa y sólo ocasionalmente sale a explorar!

The cat tends to hang around the blue zone a lot. I suppose, as wjt said, this is because there are many objects it bumps into. Perhaps you could place the cat in the orange zone, where it could move more easily without bumping into things so much or block access to the blue zone so it doesn't stay there for long.
Captura de pantalla 2026-09-20 050036

KarmDK:
Something similar happens here, the cat enters this area and then stays there because it struggles to get out. I’d say you could simply make it so the cat doesn't cross the bridge and therefore doesn't reach that area.

imagen_2026-09-20_053544673

IDynamixI985:
I set the cat boundaries to Layer 6 (interactable). This keeps the cat contained without blocking the player on the bridge or exits, and it fits well since the cat is an interactable NPC. I also moved the cat to a more open area so it has more room to wander.

manuq:
This looks great!


https://github.com/user-attachments/assets/4f802822-0caf-40ac-9c49-07fc035c02d2

manuq:
> I set the cat boundaries to Layer 6 (interactable). This keeps the cat contained without blocking the player on the bridge or exits, and it fits well since the cat is an interactable NPC. I also moved the cat to a more open area so it has more room to wander.

I see this in the cat:
image

The interactable layer is for something very different. It is used exclusively for any area that the player can interact with. The cat scene has one for the "pet the cat" interaction:

image

But the cat (CharacterBody2D) shouldn't be added to that area, and shouldn't collide with it.

scenes/game_elements/characters/npcs/cat/cat.tscn
```diff
@@ -65,3 +177,36 @@ fact_counter = NodePath("../FactCounter")
 [node name="FactCounter" type="Node" parent="." unique_id=539544383]
 script = ExtResource("6_0bwy3")
 prefix = "cats_petted"
+
+[node name="AnimationPlayer" type="AnimationPlayer" parent="." unique_id=1636558305]
```
manuq: I think that this works well just with the CharacterSpriteBehavior set to play animations. Please:
- Remove AnimationPlayer and CharacterAnimationPlayerBehavior nodes
- Reset play animations in CharacterSpriteBehavior to the default (on)

manuq (changes requested):

manuq:
Actually I was able to do the requested changes myself. Thank you @IDynamixI985 !

manuq:
I forgot to also change the CatLimit collision layers. In https://github.com/endlessm/threadbare/pull/2953 I ended up using a new collision layer for it, named "npcs_limit".

@KarmDK thanks for your review! It was really useful.

Issue #1918, opened 2026-02-14
Add grappling sound effects


wjt:
Currently grappling is silent. We should identify what sound effects are needed and add them:

- Throwing thread
- Thread making contact with a pin, needle, or other hookable item (e.g. button)
- Sound of StoryWeaver whooshing through the air
- Sound of failed throw
- Sound of StoryWeaver collecting a button (by pulling it in, by grappling through it, or even by walking through it)

### XP Summary (total 475)
| Subdomain | Skill | XP |
| - | - | - |
| Art: Music & Sound | Problem Solving | 75 |
| Art: Music & Sound | Creativity & Design Thinking | 150 |
| Art: Music & Sound | Technical Literacy | 75 |
| Engineering: Gameplay & Systems Programming   | Problem Solving | 50 |
| Engineering: Gameplay & Systems Programming   | Creativity & Design Thinking | 50 |
| Engineering: Gameplay & Systems Programming   | Technical Literacy | 75 |

wjt:
This could be done in smaller pieces, not necessarily all sounds added by the same person.

For each one, we need to decide what it should sound like, find or make a good sound with an appropriate license, and then do the technical work to play the sound at the right time.

Codex-1010:
I plan to work on this

Codex-1010:
I added a sound effect that plays when the player fires the grappling hook cable; the "Whip woosh" sound by ErikCruzDev (CC0) was used, under the license specified in a ".license" file, following the project's REUSE convention.

PR #2902, opened 2026-09-16
Player: Add sound effect for throwing the grapple string

Codex-1010:
Adds a sound effect that plays when the player throws the grappling hook string, using the "string_thrown" signal already emitted in "_new_hook_string()". The sound "Whip woosh" by ErikCruzDev (CC0) was used, licensed via a ".license" file following the project's REUSE convention. This addresses part of #1918 — specifically the "throwing thread" sound.


scenes/game_elements/characters/player/components/player_hook.gd
```diff
@@ -100,6 +100,9 @@ var hook_string: Line2D
 ## A [PhantomCamera2D] at the tip of the string.
 @onready var phantom_camera_2d: PhantomCamera2D = %PhantomCamera2D
 
+## Plays a sound effect when the string is thrown.
+@onready var throw_audio_player: AudioStreamPlayer2D = $ThrowAudioPlayer
```
manuq, line 104: Please give a unique name to the new node, and reference it here with `%`. See the other `@onready` variables in this same file. This is documented in https://github.com/endlessm/threadbare/wiki/Contributing#coding-style

manuq (changes requested):
I'm not an expert but I think this sounds very well! Check my request on unique names. Also your pull request description seems outdated:

> Adds a sound effect that plays when the player throws the grappling hook string, using the "string_thrown" signal already emitted in "_new_hook_string()". The sound "Whip woosh" by ErikCruzDev (CC0) was used, licensed via a ".license" file following the project's REUSE convention. This addresses part of https://github.com/endlessm/threadbare/issues/1918 — specifically the "throwing thread" sound.

This PR is playing the sound directly in `_new_hook_string()`. This means that it will also play when making contact with the first hookable item (pin, needle, or any other). I don't think it's bad, I just mention it because reading the ticket https://github.com/endlessm/threadbare/issues/1918 that case could have a differrent sound effect.

Probably you intended to use the "string_thrown" signal and add the sound to the player scene. Another option would be to add it to the AnimationPlayer `throw_string` animation. I think what you have is good for now, just update the description to match the implementation.

Also please add the following as the last line in the description:

`Helps https://github.com/endlessm/threadbare/issues/1918`

We use "Helps" when the PR doesn't fully adress the issue. Thanks!

scenes/game_elements/characters/player/components/player_hook.gd
Codex-1010: Done, thanks for the pointer, updated all 5 audio player nodes to use unique names

manuq:
@Codex-1010 I did a number of fixes myself because we would like to add this into the next release. Please continue iterating on the sound effects if you want, in a separate pull request. I found that the "fail" sound wasn't transmiting failure, and the opposite was happening with the "hooked" sound, it was a bit too sad. So I played with the volume and pitch. All of the sound effects were playing at the default 0 db recorded sound. Also:
- The bus was not correct, it was in the master bus. It should be in SFX.
- I removed the collected sound, because that seems more proper to be added into the item, when collected.
- I moved the AudioStreamPlayer2D nodes to the PlayerHook scene, and updated their names to match the code.
- I also changed when to play them in the script, and always play the throw effect combined with the others.
- Also I updated the name of the license files. They should have the audio extension too. For example `grapple_fail.ogg.license` (not `grapple_fail.license`)

manuq:
@Codex-1010 these two license files are empty:
- assets/third_party/sounds/grapple/grapple_collect.wav.license
- assets/third_party/sounds/grapple/grapple_hit.wav.license

Sorry but I will have to leave this out of the next release if you don't provide them. Don't worry though, there will be another release soon!

Codex-1010:
@manuq  Sorry about that! Just fixed both license files grapple_hit.wav.license, grapple_collect.wav.license, they now have the proper CC0 attribution. Thanks a lot for all the fixes and for taking the time to polish this for the release, really appreciate the feedback on the sound

manuq:
@Codex-1010 thank you!

PR #2918, opened 2026-09-17
Void runner: Fix level broken after void retreating

### LINKED ISSUE #2917: Void retreat transition broke the Void runner
### Threadbare version

v0.2.2-110-g3c16aee32

### Affected scene

res://scenes/quests/lore_quests/quest_002/1_void_runner/void_runner.tscn

### Description

After the transition the void tilemaplayer is still active, while invisible, so it isn't possible to go back:

Image

### Checklist

manuq:
In 851ef55e the fade-out of the Void retreating was replaced by a dissolve shader. That was done before by uncover_all(). But that function was also responsible for clearing all tiles, and fading-in all the consumed props.

Fix this regression by calling uncover_all() again, and leaving the fade-out of the Void as optional. Call it without the await so both transitions (props fade-in and Void dissolve) happen in parallel.

Fix https://github.com/endlessm/threadbare/issues/2917


scenes/game_elements/characters/enemies/void_spreading_enemy/components/tile_map_cover.gd
wjt:
```suggestion
	if animate_modulate:
		self.modulate.a = 1.0
```

 I'm not sure we need to keep this optional behaviour anyway - so maybe it's fine as-is.

wjt (approved):

PR #2920, opened 2026-09-17
Ignore media keys when handling pause input.

justin-calleja:
Keyboard volume knobs can emit volume and media key events that were opening the pause menu; filter those keys before toggling pause.


scenes/globals/pause/pause_overlay.gd
```diff
@@ -16,18 +16,38 @@ var home_scene: String
...
 
 func _ready() -> void:
 	home_scene = ThreadbareProjectSettings.get_setting(ThreadbareProjectSettings.HOME_SCENE)
 	visible = false
 
 
 func _unhandled_input(event: InputEvent) -> void:
+	if _is_ignored_media_key(event):
+		return
 	if event.is_action_pressed(&"pause"):
```
wjt, line 38: I don't believe this is necessary because none of those keys are bound to the `pause` action in `project.godot`.

justin-calleja: Unfortunately, I can't upload a video of this as it's over 10MB. Technically, I only need these:

```gd
	Key.KEY_VOLUMEMUTE,
	Key.KEY_VOLUMEDOWN,
	Key.KEY_VOLUMEUP,
```

Basically, my keyboard has a knob on it which can be scrolled to volume down / up or pressed to mute.

For some reason, Godot is mapping them to "pause":

Image

I tried to grab a screenshot of it in action (note: must be outside dialogue):

Image

manuq: @justin-calleja that's interesting indeed. Can you debug which of the 4 events is doing that? Could be "Pause (Physical)"?

manuq: I did a quick check in Godot issues, I couldn't find your exact issue, although this one is similar: https://github.com/godotengine/godot/issues/69978 . Maybe you can open a bug upstream, after confirming?

justin-calleja: Hi @manuq , it's `event.is_action_pressed(&"pause")` (line 26 in the screenshot is `true` and then the `event.keycode` is either `4194382` when increasing volume, `4194380` when decreasing volume or `4194381` when muting (didn't include this one in screenshot above).

```gdscript
	if event.is_action_pressed(&"pause"):
		print("event.keycode: ", event.keycode, \
			" Key.KEY_VOLUMEUP: ", Key.KEY_VOLUMEUP, \
			" Key.KEY_VOLUMEDOWN: ", Key.KEY_VOLUMEDOWN, \
			" Key.KEY_VOLUMEMUTE: ", Key.KEY_VOLUMEMUTE)
```

prints e.g.

```
event.keycode: 4194382 Key.KEY_VOLUMEUP: 4194382 Key.KEY_VOLUMEDOWN: 4194380 Key.KEY_VOLUMEMUTE: 4194381
event.keycode: 4194380 Key.KEY_VOLUMEUP: 4194382 Key.KEY_VOLUMEDOWN: 4194380 Key.KEY_VOLUMEMUTE: 4194381
event.keycode: 4194380 Key.KEY_VOLUMEUP: 4194382 Key.KEY_VOLUMEDOWN: 4194380 Key.KEY_VOLUMEMUTE: 4194381
event.keycode: 4194381 Key.KEY_VOLUMEUP: 4194382 Key.KEY_VOLUMEDOWN: 4194380 Key.KEY_VOLUMEMUTE: 4194381
```

So yes Godot is picking up those volume events as "pause" for some reason. I saw your other message while typing this - yes at some point I will try reporting this upstream.

Re this, the other ones:

```
	Key.KEY_MEDIAPLAY,
	Key.KEY_MEDIASTOP,
	Key.KEY_MEDIAPREVIOUS,
	Key.KEY_MEDIANEXT,
```

were defensive - as none of them should pause either but I cannot reproduce with my keyboard... Aula F75 if it means anything. I turned off my software keyboard mapper (kanata) and still the same.

I can remove those extra key maps - note it will take a while for me to see further replies here (at least 12 hrs).

wjt: Can you print the whole event, not just `event.keycode`?

justin-calleja: Hi @wjt . Yes if i just `print(event)` it prints:

```
InputEventKey: keycode=4194382 (VolumeUp), mods=none, physical=false, location=unspecified, pressed=true, echo=false
InputEventKey: keycode=4194380 (VolumeDown), mods=none, physical=false, location=unspecified, pressed=true, echo=false
InputEventKey: keycode=4194381 (VolumeMute), mods=none, physical=false, location=unspecified, pressed=true, echo=false
```

Breaking on print gives this in the inspector:

image

So it seems like the "Keycode" and the "Key label" match but the "Physical Keycode" differs for some reason.

justin-calleja: If I remove "Pause (Physical)" entry under "pause" action in project settings; then I can remove the code addition in this PR i.e. the volume up / down / mute via keyboard knob does not count as "pause" action. That gives this (removal) change in `project.godot`:

> -, Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":16,"window_id":0,"alt_pressed":false,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":4194313,"key_label":0,"unicode":0,"location":0,"echo":false,"script":null)

The `4194313` physical keycode matches the `4194313` physical keycode i get on all 3 volume up / down / mute. i.e. those events vary in keycode+key label from their physical keycode which is always `4194313`.

wjt: Interesting! Is this on Windows?

I think this is probably a Godot bug but I wonder whether you can try changing the binding of the Pause key from "physical" to "keycode":

image

On the basis that, unlike "WASD" where we care about the physical location on the keyboard rather than what is printed on the key (see #2185), in the case of Pause we really want the logical "Pause" key. (The reason the `P` binding is physical even though it's a mnemonic for "Pause" is: you could imagine we might want to bind an action to the key that on US keyboards is labelled `R`, because it's easy to reach from the `WASD` key cluster. However on my keyboard, that physical key enters the character `P`.)

wjt: I don't want us to remove the Pause binding entirely because the key on my keyboard that says "Pause" on it should surely pause the game!

justin-calleja: Ah fair enough - so that's what maps to it 😅  was wondering what "Pause (Physical)" meant.

Of course - a button to pause should pause.

Re your question - when i change this:

image

Then it works as expected.

I'm on Windows yes.

I don't have a dedicated pause button but i can set kanata to trigger it and it's working as expected with physical or keycode setting.

I don't understand the full implications but it seems like switching from physical to keycode would solve this ... and hopefully introduce no issues.

btw - I set a MediaPlayPause (i.e. a button to toggle pause and play) and that doesn't trigger on either physical or keycode - which is good... not sure if i was expecting it to trigger every 2nd press or something.

wjt: Great, I do have a physical pause button so if you can update this PR to change the binding rather than filtering out specific keys, I am happy to test on Linux and in a web browser!

justin-calleja: Ok - will do - but later in the day.

Thank you

justin-calleja:
Closing in favor of: https://github.com/endlessm/threadbare/pull/2927

Thank you @wjt for the solution.

PR #2927, opened 2026-09-18
Input Map: Change pause event to keycode

justin-calleja:
The current physical pause event leads Godot to treat volume up, down, mute from some keyboards as pause events in:

event.is_action_pressed(&"pause")

Changing to keycode avoids this issue and keeps actual physical pause buttons (in keyboards that have it) working as before.

Replaces https://github.com/endlessm/threadbare/pull/2920


manuq (approved):
I don't have a keyboard with the Pause key to test, but the change looks good.

manuq:
@justin-calleja thanks for going to the bottom of the problem!

